PR #24, opened 2023-03-27
Item ability improvements

appable0:
**Item abilities:**
- Add item cooldown for Crimson Isles items (Ragnarock Axe, Fire Freeze Staff, Sword of Bad Health, Staff of the Volcano)
- Change all existing items to sound detection for all items where that's feasible (some do not have a sound or have inconsistent sounds)
- Fix issue with abilities not being detected item clicked not on a block (breaking gyrokinetic wand ability)

**Mask timer:**
- Update bonzo mask timer due to Hypixel change that no longer allows stacking fragged and unfragged bonzo mask
- Add spirit mask timer

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/ItemClickData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/ItemClickData.kt
@@ -29,7 +29,7 @@ class ItemClickData {
             BlockClickEvent(ClickType.LEFT_CLICK, position, itemInHand).postAndCatch()
         }
         if (packet is C0APacketAnimation) {
-            ItemClickEvent(itemInHand)
+            ItemClickEvent(itemInHand).postAndCatch()
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/at/hannibal2/skyhanni/data/ItemClickData.kt
```diff
@@ -21,11 +23,14 @@ class ItemClickData {
             val position = packet.position.toLorenzVec()
             BlockClickEvent(ClickType.RIGHT_CLICK, position, packet.stack).postAndCatch()
         }
+        val itemInHand = Minecraft.getMinecraft().thePlayer.heldItem
```
hannibal002, line 26: Any reason why you moved the variable outside the if block?
This could be problematic when we get other packets from the server while the player is null. (Yes, this rarely happens)

src/main/java/at/hannibal2/skyhanni/events/ItemClickEvent.kt
```diff
@@ -0,0 +1,8 @@
+package at.hannibal2.skyhanni.events
+
+import at.hannibal2.skyhanni.data.ClickType
+import at.hannibal2.skyhanni.utils.BlockUtils.getBlockStateAt
+import at.hannibal2.skyhanni.utils.LorenzVec
+import net.minecraft.item.ItemStack
+
+class ItemClickEvent(val itemInHand: ItemStack?) : LorenzEvent()
```
hannibal002, line 8: Naming can be confusing since BlockClickEvent already handles left or right-clicking.

hannibal002 (approved):

PR #26, opened 2023-03-31
Add AutoUpdater and license info gui

lineargraph:


hannibal002 (changes requested):
See discord private message

PR #29, opened 2023-03-31
Fix visitor request pattern when amount is not displayed

appable0:
Minor change to patternInFront, which previously extracted "&nbsp;Polished Pumpkin" as itemName and then failed to find an internal name.
Screenshot 2023-03-31 at 3 03 30 PM


appable0:
Closing as already fixed

PR #33, opened 2023-04-05
Added Dyes and Skins to the price calc (also removed funny prinLn's

jani270:
image
image

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
@@ -135,7 +135,7 @@ object SkyBlockItemModifierUtils {
         return null
     }
 
-    fun ItemStack.getDye(): String? {
+    fun ItemStack.getArmorDye(): String? {
         for (tags in tagCompound.keySet) {
             if (tags != "ExtraAttributes") continue
             val extraAttributes = tagCompound.getCompoundTag(tags)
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
```diff
@@ -121,6 +121,34 @@ object SkyBlockItemModifierUtils {
...
+            if (tags != "ExtraAttributes") continue
+            val extraAttributes = tagCompound.getCompoundTag(tags)
+            for (attributes in extraAttributes.keySet) {
+                if (attributes == "skin") {
+                    return extraAttributes.getString(attributes)
+                }
+            }
+        }
+
+        return null
```
lineargraph:
```suggestion
        return tagCompound?.getCompoundTag("ExtraAttributes")?.getString("skin")?.takeUnless { it.isBlank() }
```
This should be fine, since `getTagCompound` returns an empty default tag if that key is not present (similar with `getString` returning a blank string by default)

src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
```diff
@@ -121,6 +121,34 @@ object SkyBlockItemModifierUtils {
...
+            if (tags != "ExtraAttributes") continue
+            val extraAttributes = tagCompound.getCompoundTag(tags)
+            for (attributes in extraAttributes.keySet) {
+                if (attributes == "dye_item") {
+                    return extraAttributes.getString(attributes)
+                }
+            }
+        }
+
+        return null
```
lineargraph: Same here:
```suggestion
        return tagCompound?.getCompoundTag("ExtraAttributes")?.getString("dye_item")?.takeUnless { it.isBlank() }
```

lineargraph (changes requested):

PR #39, opened 2023-04-12
Relocate libautoupdate

lineargraph:


src/main/java/at/hannibal2/skyhanni/config/features/About.java
```diff
@@ -11,7 +11,7 @@ public class About {
 
     @ConfigOption(name = "Current Version", desc = "This is the SkyHanni version you are running currently")
     @ConfigVersionDisplay
-    public Void currentVersion = null;
+    public transient Void currentVersion = null;
```
hannibal002, line 14: What does transient do in this context?

src/main/java/at/hannibal2/skyhanni/config/features/About.java
```diff
@@ -26,6 +26,7 @@ public class About {
 
     @ConfigOption(name = "Used Software", desc = "Information about used software and licenses")
     @Accordion
+    @Expose
```
hannibal002, line 29: Do we really need to save this into the config? we could refetch it from the jar on every startup anyway

src/main/java/at/hannibal2/skyhanni/config/features/About.java
lineargraph: explicitly not serialized (supresses the warning for missing serialization)

src/main/java/at/hannibal2/skyhanni/config/features/About.java
lineargraph: its just gonna save an empty object, and this is there to stop the warning as well

PR #47, opened 2023-04-17
fix tps display getting getting too large

PaulProjects:
the tps display looks really bad when it gets too large
and the dark gray for tps under 12 is not really readable, so I changed it to dark red


hannibal002:
The “Not playable” text was meant as an Easteregg-like thing, but of course, it's more compact and definitely more readable this way.

hannibal002:
image
What happened here? :D

PaulProjects:
ahhhhh... good question
why does github even detect that as change, da heck

hannibal002:
image
IntelliJ thinks you have added one empty line 🤷

hannibal002:
@PaulProjects and your changes are already live in the newest beta.
I could not find you on discord. Wanna send me a pm?

PR #52, opened 2023-04-20
Rename HypixelTIck.kt to HypixelTick.kt

AfkUserMC:


AfkUserMC:
i figured it would break some stuff but github wouldn't stfu

PR #54, opened 2023-04-22
Fixed level 0 skill names being parsed as levels

Sunderw3k:
Level 0 skills don't have a roman level, so the skill name was being parsed. I changed the code to use CollectionAPI (probably needs a better name) to check if the skill level is 0


hannibal002:
Oh wow, this was a small but annoying bug. Thanks for contributing! :)

PR #55, opened 2023-04-23
Hotkeys dont look like fun

hannibal002:
I have tried to rewrite the hotkey logic. Instead of changing the MC hotkey binding, this approach modifies the isPressed and isKeyDown methods via mixin.
Looks promising, but does not work perfect still.

Sometimes it even kicks you from Hypixel for spamming left click. It is unclear to me why exactly ignoreCheck does not work there.


hannibal002:
Closing this since Nea did #62 already

PR #59, opened 2023-04-25
Option to stay within the display limit for Crop Time/Money

thesefer:


src/main/java/at/hannibal2/skyhanni/features/garden/farming/CropMoneyDisplay.kt
```diff
@@ -126,7 +126,7 @@ class CropMoneyDisplay {
             number++
             val cropName = cropNames[internalName]!!
             val isCurrent = cropName == GardenAPI.cropInHand
-            if (number > config.moneyPerHourShowOnlyBest && !isCurrent) continue
+            if (number > config.moneyPerHourShowOnlyBest && (!config.moneyPerHourShowCurrent && isCurrent || !isCurrent)) continue
```
lineargraph: I'm pretty sure this equivalent (and also less confusing with operator precedence):

```suggestion
            if (number > config.moneyPerHourShowOnlyBest && (!config.moneyPerHourShowCurrent || !isCurrent)) continue
```

src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenBestCropTime.kt
```diff
@@ -62,7 +62,7 @@ class GardenBestCropTime {
             val duration = TimeUtils.formatDuration(millis, maxUnits = 2)
             val isCurrent = crop == currentCrop
             number++
-            if (number > config.cropMilestoneShowOnlyBest && !isCurrent) continue
+            if (number > config.cropMilestoneShowOnlyBest && (!config.cropMilestoneShowCurrent && isCurrent || !isCurrent)) continue
```
lineargraph:
```suggestion
            if (number > config.cropMilestoneShowOnlyBest && (!config.cropMilestoneShowCurrent || !isCurrent)) continue
```

lineargraph (changes requested):

src/main/java/at/hannibal2/skyhanni/config/features/Garden.java
```diff
@@ -275,6 +275,12 @@ public class Garden {
     @ConfigAccordionId(id = 7)
     public int cropMilestoneShowOnlyBest = 10;
 
+    @Expose
+    @ConfigOption(name = "Expand top list by current Crop", desc = "Add active crop to the end of the list by expanding the set limit")
```
hannibal002: A title that long is not really readable with the current render logic. Please consider using a shorter title.
grafik

src/main/java/at/hannibal2/skyhanni/config/features/Garden.java
```diff
@@ -625,6 +631,12 @@ public static class CustomSpeed {
     @ConfigAccordionId(id = 13)
     public int moneyPerHourShowOnlyBest = 5;
 
+    @Expose
+    @ConfigOption(name = "Expand top list by current Crop", desc = "Add active crop to the end of the list by expanding the set limit")
```
hannibal002: Furthermore, this wording in the description sounds to me like with this toggle set to false, you don't see the current crop at all, or with it enabled, the current crop will always stay at the end of the list. Maybe find a better explanation to avoid confusing users and causing more support cases.

src/main/java/at/hannibal2/skyhanni/features/garden/farming/CropMoneyDisplay.kt
thesefer: you are right, forgot to delete it because I did it only for me first and didn't care. thanks

src/main/java/at/hannibal2/skyhanni/config/features/Garden.java
thesefer: How about **"If enabled the current farmed crop will be displayed at the end of the list regardless of it's ranking and the set top limit, else it is only visible if its ranking matches the limit"** will this make it clear? What do you think should the character limit be for the title? **"Expand top list"** could fit, too.

hannibal002: "Always show current crop" would work even better ig

PR #61, opened 2023-04-26
Added decimals and time before reset to blocks/second

CalMWolfs:
Made a slider for the precision of blocks broken a second (default 2 dp.)
Made blocks broken a second only reset after not breaking a block for x seconds (default 10)

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCropMilestoneDisplay.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCropMilestoneDisplay.kt
@@ -154,12 +154,14 @@ class GardenCropMilestoneDisplay {
     private var currentCrop: CropType? = null
     private var blocksBroken = 0
     private var lastBlocksBroken = 0
+    private var secondsStopped = 0
 
     private fun resetSpeed() {
         currentSpeed = 0
         averageBlocksPerSecond = 0.0
         blocksBroken = 0
         blocksSpeedList.clear()
+        secondsStopped = 0
     }
 
     init {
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCropMilestoneDisplay.kt
```diff
@@ -182,21 +183,27 @@ class GardenCropMilestoneDisplay {
...
                 blocksSpeedList.removeFirst()
+                blocksSpeedList.add(blocksBroken)
             }
+
             averageBlocksPerSecond = blocksSpeedList.average()
+            secondsStopped = 0
 
+        } else {
+            secondsStopped++
+            if (secondsStopped >= 5) {
```
hannibal002, line 200: You might want to use `config.blocksReset` instead of `5` here

src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCropMilestoneDisplay.kt
```diff
@@ -182,21 +183,27 @@ class GardenCropMilestoneDisplay {
...
         this.blocksBroken = 0
 
-        // If the difference in speed between the current and the previous bps value is too high, disregard this value
-        if (abs(lastBlocksBroken - blocksBroken) > 4) {
-            if (blocksSpeedList.isNotEmpty()) {
-                blocksSpeedList.removeLast()
-            }
-        } else if (blocksBroken >= 2) {
+        // skipping seconds when not farming momentarily e.g. typing in chat hopefully doesn't skip the time it takes to change between rows
+        if (blocksBroken >= config.blocksReset) {
```
hannibal002, line 187: You should replace `config.blocksReset` with `2` again.

hannibal002 (changes requested):
We maybe don't need the `blocksReset` value at all, since the calculation stops when the user pauses farming anyway.
I like the approach of removing the first data-second altogether if the second data-second arrives. But you removed my 2 checks. Now the last second of farming data will be saved, even if it's not a full second of farming, corrupting the data once the user starts farming again. Additionally, I liked the old logic of resetting the whole data as soon as the last 2 seconds have a too big of a difference in blocks broken. This impacts the change in farming when testing different variant ants of farm design. e.g. breaking only 1 pumpkin in a 3 blocks farm should give you 1/3 × 20 BPS, when you start breaking 2/3 of the blocks the BPS should jump up instantly instead of moving up very slowly.

PR #68, opened 2023-04-27
Update DevData.java

AfkUserMC:


hannibal002:
You can delete the setting and the related feature logic altogether. That was from a time before I know spark does exist

PR #72, opened 2023-04-28
6th visitor changes

CalMWolfs:
Making the 6th visitor display persist after restarting the game or rejoining the garden (if you are in a coop and you join and there is 5 visitors but not a 6th ready it will still show 6 as ready)
Added a visual pop up when the 6th visitor should be waiting
Added config option for a ding to play when this is the case (default off) this also will ding when joining the garden and 6 visitors are predicted to be waiting)
Made visitor timer not rely on tabList updates and instead update once a second

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorTimer.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorTimer.kt
@@ -39,7 +39,7 @@ class GardenVisitorTimer {
             updateVisitorDisplay()
         }
     }
-// check money/hr display try reset command made
+
     private fun updateVisitorDisplay() {
         if (!isEnabled()) return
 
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorTimer.kt
```diff
@@ -64,13 +73,20 @@ class GardenVisitorTimer {
         }
 
         if (queueFull) {
-            if (sixthVisitorArrivalTime != 0L && visitorJustArrived) {
+            if (visitorJustArrived && visitorsAmount - lastVisitors == 1) {
```
CalMWolfs, line 76: stops the 6th visitor counter being set to 12 minutes on entering the garden for the first time since joining hypixel

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorTimer.kt
```diff
@@ -64,13 +73,20 @@ class GardenVisitorTimer {
...
                 updateSixthVisitorArrivalTime()
                 visitorJustArrived = false
+                sixthVisitorReady = false
             }
             millis = sixthVisitorArrivalTime - System.currentTimeMillis()
-            if (isSixthVisitorEnabled() && sixthVisitorArrivalTime != 0L && millis < 0) {
+            SkyHanniMod.feature.hidden.nextSixthVisitorArrival = System.currentTimeMillis() + millis + (5 - visitorsAmount) * visitorInterval
```
CalMWolfs, line 82: If possible could only save when exiting the garden

PR #73, opened 2023-04-28
Yaw and pitch module

thesefer:
Guess it could be done better. "Simple" yaw and pitch, nothing too fancy for now because I do not know whether you would accept it (referring to our DM conversation).


src/main/java/at/hannibal2/skyhanni/features/garden/GardenYawAndPitch.kt
```diff
@@ -0,0 +1,35 @@
...
+import at.hannibal2.skyhanni.SkyHanniMod
+import at.hannibal2.skyhanni.events.GuiRenderEvent
+import at.hannibal2.skyhanni.utils.LorenzUtils.round
+import at.hannibal2.skyhanni.utils.RenderUtils.renderStrings
+import net.minecraft.client.Minecraft
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+class GardenYawAndPitch {
+    private val config get() = SkyHanniMod.feature.garden
+    private var yawandpitchDisplay = listOf<String>()
```
hannibal002: Why using a field at all if you never read it?

src/main/java/at/hannibal2/skyhanni/config/features/Garden.java
```diff
@@ -1030,4 +1030,12 @@ public static class CustomSpeed {
 
     @Expose
     public Position cropSpeedMeterPos = new Position(278, -236, false, true);
+
+    @Expose
+    @ConfigOption(name = "Yaw / Pitch", desc = "Show Yaw and Pitch")
```
hannibal002: Please add a point at the end of the description.

hannibal002 (changes requested):
See above

src/main/java/at/hannibal2/skyhanni/config/features/Garden.java
skupfer: Second time I forgot about the dot. Added and will push it after resolving the other review.

thesefer: done

jani270:
I don't really see a reason to have this in SkyHanni, as it is already present in NEU.

hannibal002 (approved):

thesefer:
@jani270 Oh, you are right, didn't see because I never used it. But aren't there always overlapping features?

hannibal002:
> I don't really see a reason to have this in SkyHanni, as it is already present in NEU.

NEU only has 2 digit precision. This was a problem for multiple users already.

jani270:
> > I don't really see a reason to have this in SkyHanni, as it is already present in NEU.
> 
> NEU only has 2 digit precision. This was a problem for multiple users already.

Then the better solution to make the neu feature better?

jani270:
> @jani270 Oh, you are right, didn't see because I never used it. But aren't there always overlapping features?

Well, there are always some, but i think it would be better not to do that.

PR #76, opened 2023-04-28
Sixth visitor changes

CalMWolfs:
Made visitor timer not rely on tab list updates and instead update once a second
6th visitor timer persists on restarting the game and joining/leaving the garden
warning for when the 6th visitor is predicted to be ready
sound for when this occurs that is toggleable (default on)


hannibal002:
When having 5 visitors waiting and the 6th visitor is ready too, accepting/refusing a visitor will not change the tab list.
The timer should jump from 6 visitors to 5 visitors + 12 min time.

PR #79, opened 2023-04-29
Contest fortune changes

CalMWolfs:
Made the needed farming fortune reflect the user's setting of "Show as drop multiplier"
Added setting to toggle cactus mode (sets blocks/sec to 17) - rough value but much more accurate than ~20
Added slider so that users can put in a rough estimate of their blocks/second (19.8 default value which is probably still higher than most users get)


hannibal002:
Maybe we should rethink the -100 calculation thingy

image

PR #87, opened 2023-05-01
Added predicted socre to Jacob's menu

CalMWolfs:
Added a line saying predicted score based on the user's stats.
can help them if they are curious or improved their farm/setup since last contest or didn't participate for the full contest


hannibal002:
You forgot to account for the baseDrops factor. I have fixed the formula.

hannibal002 (changes requested):
I find the text confusing. In the gold/silver/bronze line we talk about `crops` and in the new line at the bottom we talk about `score`. 
image

PR #93, opened 2023-05-06
Added a few commands for debugging and testing

CalMWolfs:
Added the ability to copy the tablist and scoreboard to the clipboard
Added a testmessage command to test colours and/or how mods respond to chat events

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/test/command/TestChatCommand.kt
+++ b/src/main/java/at/hannibal2/skyhanni/test/command/TestChatCommand.kt
@@ -1,6 +1,9 @@
 package at.hannibal2.skyhanni.test.command
 
 import at.hannibal2.skyhanni.utils.LorenzUtils
+import net.minecraft.util.ChatComponentText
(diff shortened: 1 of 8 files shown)


src/main/java/at/hannibal2/skyhanni/test/command/TestChatCommand.kt
```diff
@@ -0,0 +1,19 @@
...
+            LorenzUtils.chat("§c[SkyHanni] Specify a chat message to test")
+            return
+        }
+        val resultList = mutableListOf<String>()
+        for (arg in args) {
+            resultList.add(arg)
+        }
+        val string = resultList.joinToString(" ")
+        LorenzUtils.chat("§a[SkyHanni] testing message: §7$string")
+        LorenzUtils.chat(string.replace("&", "§"))
```
lineargraph: I would recommend you use a packet to send that message, otherwise other mods have a decent chance of not receiving that event

hannibal002: Implemented :)

PR #94, opened 2023-05-07
Copy nearby particles to clipboard

CalMWolfs:
Added the ability to copy nearby particles
Also removed empty lines returned from /shcopytablist 
Fun fact players names are on a second hidden tablist (the old one without the extra info)


hannibal002:
Instead of this
```particle type: ENCHANTMENT_TABLE
particle location: LorenzVec{x=-361.2354431152344, y=81.0, z=-442.0961608886719}
distance from player: 1.4523382526524677
particle offset: LorenzVec{x=0.20000000298023224, y=1.0, z=0.20000000298023224}
is long distance: true
particle count: 12
particle speed: -1.15
particle arguments: [I@5a66f44e
```

it now looks like this
```particle type: FLAME
particle location: LorenzVec{x=-484.89, y=138.82, z=-1016.92}
distance from player: 7.18
particle offset: LorenzVec{x=0.0, y=0.0, z=0.0}
is long distance: true
particle count: 1
particle speed: 0.0
particle arguments: []
```

PR #99, opened 2023-05-10
Fix for calendar reading going above 100 percent

### LINKED ISSUE #98: Calendar reading progress goes above 100%
The overlay in the calendar screen that tracks how many contests have been viewed goes above 100%.

Contrabass26:
Attempts to fix #98 - not tested yet for the start of a new Skyblock year

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
@@ -114,24 +114,26 @@ class GardenNextJacobContest {
             }
         }
 
-        for (item in items) {
-            val lore = item.getLore()
-            if (!lore.any { it.contains("§6§eJacob's Farming Contest") }) continue
+        if (contests.size < maxContestsPerYear) {
+            for (item in items) {
+                val lore = item.getLore()
+                if (!lore.any { it.contains("§6§eJacob's Farming Contest") }) continue
 
-            val name = item.name ?: continue
-            val matcherDay = patternDay.matcher(name)
-            if (!matcherDay.matches()) continue
+                val name = item.name ?: continue
+                val matcherDay = patternDay.matcher(name)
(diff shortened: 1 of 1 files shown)


hannibal002:
Your fix will ignore loading new contests if they are already at 100%
The >100% display is only a visual bug while inside the calendar.
We should try to find the root cause instead of fixing the visuals. Since atm, loading more contests is necessary for the next year. Otherwise, the next year's contests would not load at all.

Contrabass26:
I considered that, but don't the lines above check for a new year and clear the old contests? So the contests map will never need to be overwritten, only added to. That was my thought process, but obviously you know your own code better than I do 😄 . I'll try and look for the deeper cause as well.

hannibal002:
> That was my thought process, but obviously you know your own code better than I do

Please don't assume I remember my almost 2-month-old code haha

You looked into the logic for creating this PR, you are probably right.
Maybe we can just add this check and it's fine. But I don't want to risk errors that cause the inventory data to lot load at all when right now it is just an annoying bug while loading, not really “game breaking”

If you can find the root cause instead, this would be perfect, thanks!

Contrabass26:
Probably the best thing is to wait until a new year and see what happens - like you say, it's not game breaking so can wait. I'll test it in a couple of days.

Contrabass26:
@hannibal002 Just tested it for the skyblock year that just started - it's all looking fine

hannibal002:
> @hannibal002 Just tested it for the skyblock year that just started - it's all looking fine

And the first contests after couple days loaded correctly?

Contrabass26:
It went from 0% all the way up to 100% as I moved through the year, and it's still working now (unless all the contests are the same by coincidence!)

hannibal002:
Marked as draft – further testing required

Contrabass26:
@hannibal002 I have been regularly "testing" (using) this feature for several months now without any issues... But if you want to test it personally that's perfectly understandable.

hannibal002 (approved):

PR #103, opened 2023-05-11
Visitor drop counter

CalMWolfs:
Visitor Drop Counter
**Excuse the repeating lines of code, I don't know if they can be compressed**
Counts up all the drops that you get from visitors
- Count each rarity of visitor **Accepted**
- Count copper and farming exp
   - The xp count is not affected by wisdom, just the raw xp it says in chat
- Count coins "spent", Bazaar prices so market manipulation will show
- Count rare drops

Settings:
- Setting to show number or drop first
- Setting to show as the icon instead of the name
- Setting to show only on the barn plot

Made show as drop multiplier default on
Added support for numbers that have 1 decimal into format number

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/Garden.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/Garden.java
@@ -175,6 +175,69 @@ public class Garden {
     public boolean visitorHideChat = true;
 
     @Expose
+    @ConfigOption(name = "Visitor Drops", desc = "")
+    @ConfigEditorAccordion(id = 23)
+    public boolean visitorDrops = false;
+
+    @Expose
+    @ConfigOption(name = "Visitor Drop Summary", desc = "Tallies up statistic about visitors and the rewards you have received from them." +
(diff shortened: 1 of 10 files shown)


src/main/java/at/hannibal2/skyhanni/config/features/Garden.java
```diff
@@ -997,7 +1060,7 @@ public static class RecentTeleportPads {
     )
     @ConfigEditorBoolean
     @ConfigAccordionId(id = 22)
-    public boolean farmingFortuneDropMultiplier = false;
+    public boolean farmingFortuneDropMultiplier = true;
```
CalMWolfs, line 1065: This is optional but I feel it’s the metric that more people use

src/main/java/at/hannibal2/skyhanni/utils/NumberUtil.kt
CalMWolfs, line 1: I guess you could also set this up for more than 1 decimal point by counting the amount of digits after the point and then dividing accordingly but I’m not sure where uses more than 1 decimal point of precision

PR #160, opened 2023-05-18
Fortune guide  v1

### LINKED ISSUE #123: Cheapest Farming Fortune Gain

CalMWolfs:
/ff to open the menu

- Mostly done, It should work well enough to show the breakdown of their farming fortune, It needs use by a larger amount of people to find bugs and suggestions
Daedalus axe won't work
If the pet levels up or they gain a new visitor milestone they will have to reopen that menu to update it

Will continue working on it, fixing bugs as they are found, adding new features and making the upgrade guide


src/main/java/at/hannibal2/skyhanni/features/garden/ToolTooltipTweaks.kt
CalMWolfs, line 1: This file could go back to how it was as I don't use the function anymore, but it still could be used by something else later on, up to you

src/main/java/at/hannibal2/skyhanni/config/Storage.java
```diff
@@ -17,6 +18,10 @@ public class Storage {
     @Expose
     public String apiKey = "";
 
+//todo could not work out how to access the player specific part :/
+    @Expose
+    public Integer gardenCommunityUpgrade = 0;
```
CalMWolfs: could not work out how to use the player specific storage for this

PR #178, opened 2023-05-26
Garden Plot Icon

superhize:
- Change editing mode by clicking on the wooden axe in the bottom right corner (not sure about keeping the item there, was thinking of using the "The Barn" slot instead or maybe a clickable text out of the inventory)
- Click an item in your inventory then click any item in the Configure Plots menu to change it to that item (keep lore, name, etc)

~~There is a bug if using patcher, the item name is not saved and display the name of the item you clicked in your inventory instead.
Tried to disable all patcher feature but it's still doing it~~ 
Found a workaround, thanks @CalMWolfs


src/main/java/at/hannibal2/skyhanni/features/garden/inventory/GardenPlotIcon.kt
```diff
@@ -0,0 +1,136 @@
...
+            if (plotList.isNotEmpty() && plotList.contains(event.slotNumber)) {
+                val old = originalStack[event.slotNumber]?: return
+                val new = event.slotNumber.getItem()?: return
+                val name = old.displayName
+                val lore = old.getLore().toTypedArray()
+                if (lastClickedSlotId == event.slotNumber) {
+                    lastClickedSlotId = -1
+                    return
+                }
+                val replaceStack = Utils.editItemStackInfo(new, name, true, *lore)
```
superhize: There, i have no idea why patcher override the new name with the name of the stack you clicked in your inventory

hannibal002 (changes requested):
Cool feature! It works correctly for me.

Please make these changes before merge:

- You should be able to just save the internal name instead of the full item stack in the config. Build the item stack with the `NEUItems.getItemStack(internalName)` function on inventory open once.
- Please cache the items for use in `ReplaceItemEvent`. The event gets fired every frame! We should avoid complex logic like the building of the ItemStack in this event. My FPS went to 90 in the Configure Plots inventory, compared to 200 FPS in other inventories.
- Please add the `inSkyBlock` check in every event-method before checking other stuff (comparing booleans is less performance than string comparisons)
- The methods `Int.setItem` and `Int.getItem` can be inlined/removed. Why aren't we using the companion object `plotList` here? You could change the whole class a `object` to access the companion objects from everywhere in that file
- Can we explain/describe the feature and the current mode in the item lore of the wooden axe "edit mode"? This would help users understand the feature.

If something is unclear, or you need help implementing the changes, just ask for help.

hannibal002:
- Fixed NPEs when opening the inventory or clicking on an item without loaded profile data
- Renamed `showItem` to `inInventory`
- Created and used method `isEnabled`
- Fixed Crash if `copyStack` is null (clicking in edit mode 1 on a plot icon without selecting a replacement item beforehand)

hannibal002 (approved):

PR #188, opened 2023-05-29
v2 for fortune guide

CalMWolfs:
Added bazaar items to recommended upgrades

got the stuff in place for knowing when an item is outdated


src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
```diff
@@ -47,7 +47,7 @@ object SkyBlockItemModifierUtils {
 
     fun ItemStack.getPetItem(): String? {
         val data = cachedData
-        if (data.heldItem == "") {
+        if (data.heldItem == "" || data.heldItem == null) {
```
hannibal002: null assumes no pet item loaded. we would check each time this way, making the cache unnecessary

src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
```diff
@@ -135,6 +135,23 @@ object ItemUtils {
             .getTagList("textures", Constants.NBT.TAG_COMPOUND).getCompoundTagAt(0).getString("Value")
     }
 
+    fun ItemStack.getItemRarity(): Int {
```
hannibal002, line 138: Maybe make an enum out of the rarity for the future

src/main/java/at/hannibal2/skyhanni/data/ChatManager.kt
```diff
@@ -85,18 +85,15 @@ class ChatManager {
 
         val last = lines.last()
         if (last.startsWith("§f§lCOMMON")) return true
```
hannibal002, line 87: Here we could use the rarity enum

src/main/java/at/hannibal2/skyhanni/config/Storage.java
```diff
@@ -171,6 +171,12 @@ public static class PlotIcon {
 
             public static class Fortune {
 
+                @Expose
+                public Map<FarmingItems, Boolean> outdatedItems = new HashMap<>();
+
+                @Expose
+                public int uniqueVisitors = 0;
```
hannibal002: Maybe store this outside of Fortune, directly in Garden. Will be useful for standalone gui element maybe?

hannibal002 (changes requested):

hannibal002 (approved):

PR #191, opened 2023-05-31
Added price in bronze/silver Trophy Fishing Sack

superhize:


superhize:
#192 modify the gemstone regex, may want to merge it before this one

PR #214, opened 2023-06-04
fix

### LINKED ISSUE #213: Bugs with fortune guide
int limit ff on pets - fixed
garden lvl not being detected? - cant reproduce

CalMWolfs:
No error when wearing 1/4 farming armor
No int limit ff on pets

Cannot reproduce pet item not working at all

tried resetting config, setting garden exp to weird amounts but always worked fine. Also tried numbers instead of roman numerals. Need more data from people with issues


hannibal002:
This will close #213

PR #216, opened 2023-06-04
Updated Sack-In-A-Sack display

superhize:
Ability to replace how it is displayed in lore


superhize:
i think replaceLore should be ON by default, idk

PR #234, opened 2023-06-13
Add more details to "Profile" and add "Stacking" to Rich Presence

NetheriteMiner:
Stacking is dynamic-only, requires holding an item with a stacking enchantment.


hannibal002 (changes requested):
See Discord PM

hannibal002 (approved):

PR #236, opened 2023-06-17
Made visitor drop statistics on by defualt

CalMWolfs:
I think a lot of people don't know about this
It also only shows on barn plot by default so it doesn't clutter their GUI otherwise. Feel free to reject but this was my thoughts


hannibal002 (approved):
agreed!

PR #239, opened 2023-06-18
Maybe people will read

CalMWolfs:
Made it more clear where NEU is
Not sure if it would be good to not include the NEU GitHub for now as that doesn't actually give them the alpha download


hannibal002 (approved):
Yes maybe

PR #265, opened 2023-06-26
mirror verse jump and run

hannibal002:
Started working on the mirror verse jump and run


hannibal002:
Todo missing:
Option to disable the rainbow effect / allow one custom colour instead

PR #275, opened 2023-06-28
Trying to fix ghost counter

superhize:
- It now work as it (no other mods is needed and conflicts should ever happen again)
- It work even if the action bar show percent +xxx (xx%) instead of +xxx (xxx/xxx)
- Added command /shresetghostcounter to reset everything (not the bestiary data)
- Added time format in ETA formatting (can show days, hours, minutes and seconds remaining)

Also fixed few things that were not working as intended (i forgot what)

Big thanks to @ksipli on discord for helping me debugging everythings


hannibal002:
In the future, we should find a way to split the 700+ lines of GhostCounter.kt into multiple smaller files

hannibal002 (approved):

PR #277, opened 2023-06-29
Update Rich Presence to support The Rift

NetheriteMiner:
There are actually a lot more miscellaneous changes:

- Separated ActionBarStatsData regex (technically slower, but supports caching now)
- Fixed names unnecessarily getting cut off when visiting an island in Location
- Added a detection for ambiguous names so the image can actually be specific (e.g. Community Center exists in both the hub and crimson isle, image will now show the correct image)
- Added several images to locations, so less of them are "skyblock-logo"
- Optimized detection of the name of the held item in stacking (also fixes the bug where holding an item that isn't in the neu repo caused the rich presence to stop updating)


NetheriteMiner:
Some older commits are in this pull request and I don't know why
Also, it's a draft because it just needs testing, but it should theoretically be working

PR #283, opened 2023-07-01
changed the blocking of visitors

CalMWolfs:
If you click it twice the villager will be denied
Also stopped you from being kicked from the inventory


hannibal002:
You could click accidentally the refuse button twice, since you think it didn't register the first time (only to see in chat that you lost a grass)
I prefer you get kicked from the inventory and have to open the visitor again to bypass the deny prevention.
How about saving the `blocked once` value inside the visitor object to accomplish this? Maybe also save the time of the first click so that the user has to refuse the visitor twice in under 10 seconds?

hannibal002:
Also, we should make `FLOWERING_BOUQUET` default disabled in `config.visitorRewardWarning.drops`

jani270:
Might just wanna do the same as neu does if we block smth?

hannibal002:
Showing a popup with a confirm button? It might be a better idea overall.

CalMWolfs:
Closing due to HiZe's better solution and me not having time to fix this up

PR #292, opened 2023-07-04
Living Metal Suit Progress

superhize:
https://www.youtube.com/watch?v=tj0wpWKF0n4


hannibal002:
Reworked some code, among other things:
- Making progressMap var, using a buildMap to change it directly.
- Calling update() only if progressMap has changed since the last check.
- Change the value of progressMap from int (0-100) to double (0-1).
- Using collection util methods like average, count and filterNotNull in drawDisplay()

PR #302, opened 2023-07-07
Vampire slayer features

superhize:
someone asked

- Highlight your own boss (color can be changed)
- Change color when the boss is below 20% (can use steak) (can change color)
- Highlight others players boss only if you hit them (can be toggled)
- Highlight Co-Op members boss (you need to write their name in the config)
- Ability to see boss through blocs (not sure if keeping that)
- Highlight Blood Ichor (option to render a beacon beam too)
- Highlight Killer Spring
- TwinClaws warning (work with co-op members highlight, and others people boss if you hit them)


Fix3dll:
@superhize Is it possible to highlight Blood Ichor as beacon beam?

src/main/java/at/hannibal2/skyhanni/features/rift/VampireSlayerFeatures.kt
```diff
@@ -0,0 +1,246 @@
...
+                val vec = stand.position.toLorenzVec()
+                val distance = start.distance(vec)
+                val isIchor = stand.hasSkullTexture(bloodIchorTexture)
+                if (isIchor || stand.hasSkullTexture(killerSpringTexture)) {
+                    val color = (if (isIchor) config.bloodIchor.color else config.killerSpring.color)
+                        .toChromaColor().withAlpha(config.withAlpha)
+                    RenderLivingEntityHelper.setEntityColor(
+                        stand,
+                        color
+                    ) { distance <= 15 }
```
hannibal002: This will not work as intended. The value inside the field distance will be hard copied into the condition. the value is always true or false, does not check on every frame as intended. you could move the distance check out of the setEntityColor() call and use isEnabled() as the condition instead

hannibal002 (approved):

PR #310, opened 2023-07-12
Update both dicer rng drop messages

StarZebra:


CalMWolfs:
there are 3 tiers of dicers and each one drops different amounts maybe you could add regex that detects all of them and only matches the starts and ends of strings exactly

StarZebra:
oh right I forgot about that, well I can't do regex yet so I will close this pr.

PR #321, opened 2023-07-17
Feature: Snakes highlighter for rift cave

hannibal002:
## What
After 1.5 years, ive stated working on this again.
and im happy with the results. 

## Changelog New Features
+ Added Rift Snake Highlighter. - hannibal2
    * Highlights moving snakes in the Living Cave in the Rift.
    * Uses colors to indicate different snake states.
    * Highlights head or tail based on whether the held item calms or breaks the snake.


src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
```diff
@@ -0,0 +1,313 @@
...
+import net.minecraft.block.Block
+import net.minecraft.client.Minecraft
+import net.minecraft.init.Blocks
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.seconds
+
+@SkyHanniModule
+object LivingCaveSnakeFeatures {
+    private val config get() = RiftApi.config.area.livingCave.livingCaveLivingMetalConfig
+    private var snakes = Mutex(mutableListOf<Snake>())
```
Thunderblade73: Kotlin has a native implementation for mutex behaviour. ```synchronized```
https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/synchronized.html
https://medium.com/@appdevinsights/how-do-you-handle-concurrency-in-kotlin-using-synchronized-blocks-3e209251d47d

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
```diff
@@ -0,0 +1,313 @@
...
+                } else head
+                event.renderBlock(location)
+            }
+            for (block in blocks) {
+                if (block == head && lastAddTime.passedSince() < 200.milliseconds) {
+                    continue
+                }
+                for ((x, y) in edges) {
+                    event.draw3DLine(block + x, block + y, state.color.toColor(), 2, true)
+                }
```
Thunderblade73: Pls use a line drawer here to save redundante calls.

src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt
```diff
@@ -130,4 +131,70 @@ object LocationUtils {
...
+                val unitDirection = direction * (1.0 / magnitude)
+                vertex + unitDirection * delta
+            } else {
+                vertex // If the vertex is at the center, no shift
+            }
+        }
+
+        // Shift all vertices outward
+        val shiftedVertices = originalVertices.map { shiftOutward(it) }
+
```
Thunderblade73: This code looks a lot like the ```AxisAlignedBB.expand()``` code.

Thunderblade73 (changes requested):
Gave the Pr just a quick look and found some stuff that should be changed. Did not read the  full code. I focused on the render part first and foremost.

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
lineargraph: While synchronizing on an object is technically possible, it is recommended against by kotlin itself. Additionally the mutex type makes it impossible to access something without acknowleding the lock, by wrapping the object in type level knowledge of the associated lock.

Thunderblade73: > While synchronizing on an object is technically possible, it is recommended against by kotlin itself. Additionally the mutex type makes it impossible to access something without acknowleding the lock, by wrapping the object in type level knowledge of the associated lock.

I just don't think that a custom class is the recommended way to to do it. And I just remembered that you used synchronized block in the RepoPattern, so I just assumed that is the preferred way to do it in kotlin. When you are fine with then I'm also fine with it.

lineargraph: The custom class here is maybe not neccessarily the best way of doing it, (especially with seperate read and write locks, or with threads that wait on each other using condvars), but i specifically wanted something that encodes the need to synchronize on the object in the type system, and this is what i found. Custom lock implementations also allow for more things (such as a checked access method using `isHeldByCurrentThread`).

Thunderblade73: I just looked at the code more closely and noticed that a mutex is here the wrong approach to begin with. Since all 3 lists are simply queue where one side fills it and the other drains it. So a ```ConcurrentLinkedQueue``` would work better here imo.

hannibal002: Using a ConcurrentLinkedQueue feels wrong for me (assuming I understand it correctly) since we have three separate actions, fill, read, drain. And without merging read and drain in one locked state, the fill action could maybe happen during read. That would leave unprocessed data to be drained.

hannibal002: If we could iterate and drain at the same time this would be ideal. But I don't want to create three helper lists every tick

src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt
hannibal002: Yeah I  already thought about that existing logic while writing this here. sadly I didn't found a neat way of parsing a location to a axis aligned aabb object so I used a different, more primitive solution. But I think the math should be put into a utils again

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
Thunderblade73: Either way I'm currently porting the code to the ConcurrentLinkedQueue.

> If we could iterate and drain at the same time this would be ideal. But I don't want to create three helper lists every tick

```CollectionUtils.drainForEach``` exists

> Using a ConcurrentLinkedQueue feels wrong for me (assuming I understand it correctly) since we have three separate actions, fill, read, drain. And without merging read and drain in one locked state, the fill action could maybe happen during read. That would leave unprocessed data to be drained.

Drain and fill work at the same time
Read and fill work at the same time
Just read and drain could have strange behaviour

File: src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt (hunk 2 above)
Comment (hannibal002) [reply src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt]: What does the line drawer save? The GL state changes between each line?

Comment (Thunderblade73) [reply src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt]: All of this(expect the draws.invoke)
image

File: src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt (hunk 1 above)
Comment (hannibal002) [reply src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt]: But this is exactly what we do here? First we fill in async block packet event, then we read and drain in tick event. So this doesn't help us then?

Comment (Thunderblade73) [reply src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt]: In tick we just drain. Drain is a destructive read so you don't have to read again since you already did read it during the drain.

File: src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt (hunk 2 above)
Comment (hannibal002) [reply src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt]: At a glance this doesn't look like it supports our see behind blocks parameter, will test out later

File: src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt (hunk 1 above)
Comment (hannibal002) [reply src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt]: So we could read and drain at the same time? I'm confused, you just said
Just read and drain could have strange behaviour

Comment (hannibal002) [reply src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt]: Also can you explain to me the difference between this concurrent list approach and the mutex approach? The internal value still needs to be copied over or the write process be halted, so how is this better

Comment (Thunderblade73) [reply src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt]: > So we could read and drain at the same time? I'm confused, you just said Just read and drain could have strange behaviour

Yes, if you read in on thread but at the same time drain it in another. But that does not happen in your code.

Comment (Thunderblade73) [reply src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt]: > Also can you explain to me the difference between this concurrent list approach and the mutex approach? The internal value still needs to be copied over or the write process be halted, so how is this better

No. Mutex does need to stop. The ConcurrentLinkedQueue is designed so it can work continuously.
Sure there might some small artifacts in the render but I would argue those are better than frame hiccups.

File: src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt (hunk 3 above)
Comment (hannibal002) [reply src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt]: I looked again into this, and while `location.boundingToOffset(1.0, 1.0, 1.0).expand(0.0001, 0.0001, 0.0001)` exists, i dont find a way to render this `AxisAlignedBB` by only drawing the corner lines. @Thunderblade73

Comment (Thunderblade73) [reply src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt]: yeah, I know that is not implemented. Simply use the point of the Corners of the AxisAlignedBB as you used your points. Otherwise feel free to implement such an method, since it will be needed in the future anyway.

Comment (hannibal002) [reply src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt]: done

File: src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt [hunk 4]
```diff
@@ -0,0 +1,297 @@
...
+object LivingCaveSnakeFeatures {
+    private val config get() = RiftApi.config.area.livingCave.snakeHelper
+    private val snakes = mutableListOf<Snake>()
+    private val edges = LocationUtils.generateCubeEdges(0.005)
+
+    private val originalBlocks = mutableMapOf<LorenzVec, Block>()
+
+    private var selectedSnake: Snake? = null
+
+    private val directions = setOf(
```
ItsEmpa: this could probably be moved to the companion object of lorenzvec

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
```diff
@@ -0,0 +1,297 @@
...
+        LorenzVec(0, 0, -1),
+    )
+
+    // TODO maybe move this in repo
+    private val pickaxes = setOf(
+        "SELF_RECURSIVE_PICKAXE",
+        "ANTI_SENTIENT_PICKAXE",
+        "EON_PICKAXE",
+        "CHRONO_PICKAXE",
+    ).map { it.toInternalName() }
```
ItsEmpa:
```suggestion
    ).toInternalNames()
```

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
```diff
@@ -0,0 +1,297 @@
...
+            snake.blocks.add(0, location)
+            snake.lastAddTime = SimpleTimeMark.now()
+            snake.invalidHeadSince = null
+        }
+    }
+
+    @HandleEvent
+    fun onItemInHandChange(event: ItemInHandChangeEvent) {
+        currentRole = when (event.newItem) {
+            "FROZEN_WATER_PUNGI".toInternalName() -> Role.CALM
```
ItsEmpa: make this a property in the class to not call .toInternalName constantly

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
```diff
@@ -0,0 +1,297 @@
...
+        if (!isEnabled()) return
+
+        lastClickedBlock = event.position
+
+        val snake = getClosest() ?: return
+        if (event.position !in snake.blocks) return
+
+        selectedSnake = snake
+        if (event.clickType == ClickType.RIGHT_CLICK) {
+            if (InventoryUtils.itemInHandId == "FROZEN_WATER_PUNGI".toInternalName())
```
ItsEmpa: same with this internal name

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    private fun getClosest(): Snake? =
+        snakes.minByOrNull { it.blocks.minOfOrNull { block -> block.distanceSqToPlayer() } ?: Double.MAX_VALUE }
+
+    private fun LorenzVec.isNotTouchingAir(): Boolean = directions.none { plus(it).getBlockAt() == Blocks.air }
+
+    private fun isEnabled() = RiftApi.inRift() && (RiftApi.inLivingCave() || RiftApi.inLivingStillness()) && config.highlight
+
+    private class Snake(
```
ItsEmpa: make this a data class

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
```diff
@@ -0,0 +1,297 @@
...
+            found.firstOrNull()
+        }
+    }
+
+    private fun getClosest(): Snake? =
+        snakes.minByOrNull { it.blocks.minOfOrNull { block -> block.distanceSqToPlayer() } ?: Double.MAX_VALUE }
+
+    private fun LorenzVec.isNotTouchingAir(): Boolean = directions.none { plus(it).getBlockAt() == Blocks.air }
+
+    private fun isEnabled() = RiftApi.inRift() && (RiftApi.inLivingCave() || RiftApi.inLivingStillness()) && config.highlight
```
ItsEmpa: remove the rift check from isEnabled and add rift island check in the handle event annotations

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
```diff
@@ -0,0 +1,297 @@
...
+                if (block == head && lastAddTime.passedSince() < 200.milliseconds) {
+                    continue
+                }
+                for ((x, y) in edges) {
+                    event.draw3DLine(block + x, block + y, state.color.toColor(), 2, true)
+                }
+            }
+        }
+
+        fun SkyHanniRenderWorldEvent.renderBlock(location: LorenzVec) {
```
ItsEmpa: can be private

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
```diff
@@ -0,0 +1,297 @@
...
+            val isSelected = this@Snake == selectedSnake
```
ItsEmpa: a function for isSelectedSnake() could be useful seeing how you are checking this in a couple of places

ItsEmpa (changes requested):
some things i noticed

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
hannibal002: done

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
hannibal002: will do

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
hannibal002: alr done

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
hannibal002: what for? do we use compare/tostring?

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
hannibal002: done

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
hannibal002: done

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
ItsEmpa: imagen

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
hannibal002: done

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
hannibal002: done

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
hannibal002: For a data class, the generated `equals` method compares property values, so two different instances with identical properties are considered equal. For a non-data class (without an overridden `equals`), the default implementation checks reference equality, meaning only the same instance is equal to itself.

we dont want to compare the properties, just the reference. so this is not needed here

ItsEmpa: fair enough

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveSnakeFeatures.kt
hannibal002: we now use line drawer

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveLivingMetalHelper.kt
```diff
@@ -63,14 +67,10 @@ object LivingCaveLivingMetalHelper {
     fun onRenderWorld(event: SkyHanniRenderWorldEvent) {
         if (!isEnabled()) return
         val (a, b) = pair ?: return
-        if (System.currentTimeMillis() > startTime + 5_000) return
+        if (animationStartTime.passedSince() > 4.seconds) return
```
Thunderblade73, line 70: Is this intended to be reduced from 5 to 4 seconds?

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
```diff
@@ -1121,6 +1126,52 @@ object RenderUtils {
         }
     }
 
+    fun SkyHanniRenderWorldEvent.drawEdges(location: LorenzVec, color: Color, lineWidth: Int, depth: Boolean) {
+        LineDrawer.draw3D(partialTicks) {
+            for ((p1, p2) in location.edges) {
+                draw3DLine(p1, p2, color, lineWidth, depth)
+            }
+        }
```
Thunderblade73: Add this to the LineDrawer, so that multiple of those can be drawn, You can keep the wrapper but it should be possible to draw 2+ of those things inside on LineDrawer so the "logic" part should be moved to a methode inside LineDrawer.

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
```diff
@@ -1121,6 +1126,52 @@ object RenderUtils {
...
+    }
+
+    fun SkyHanniRenderWorldEvent.drawEdges(axisAlignedBB: AxisAlignedBB, color: Color, lineWidth: Int, depth: Boolean) {
+        // TODO add cache. maybe on the caller site, since we cant add a lazy member in AxisAlignedBB
+        LineDrawer.draw3D(partialTicks) {
+            for ((p1, p2) in axisAlignedBB.calculateEdges()) {
+                draw3DLine(p1, p2, color, lineWidth, depth)
+            }
+        }
```
Thunderblade73: Same as above

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
```diff
@@ -1121,6 +1126,52 @@ object RenderUtils {
...
+            topLeftBack to topRightBack,
+            topRightBack to topRightFront,
+            topRightFront to topLeftFront,
+            // Vertical edges
+            bottomLeftFront to topLeftFront,
+            bottomLeftBack to topLeftBack,
+            bottomRightBack to topRightBack,
+            bottomRightFront to topRightFront,
+        )
+    }
```
Thunderblade73: Move that to locationUtils or LorencVec. Since that is not really a render method just a helper for AxisAlignedBB.

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/snake/LivingCaveSnake.kt
```diff
@@ -0,0 +1,124 @@
...
+    var state: State = State.SPAWNING,
+    var lastCalmTime: SimpleTimeMark = SimpleTimeMark.farPast(),
+    var lastHitTime: SimpleTimeMark = SimpleTimeMark.farPast(),
+    var invalidHeadSince: SimpleTimeMark? = null,
+    private var lastBrokenBlock: LorenzVec? = null,
+) {
+    val head get() = blocks.first()
+    private val tail get() = blocks.last()
+
+    fun invalidSize(): Boolean = blocks.isEmpty() || blocks.zipWithNext().any { (a, b) ->
```
Thunderblade73: It checks invalide in general not dependent on the size. So a better name would be preferred.

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/snake/LivingCaveSnake.kt
```diff
@@ -0,0 +1,124 @@
...
+                LocationUtils.interpolateOverTime(lastAddTime, 200.milliseconds, blocks[1], head)
+            } else head
+            event.renderBlock(location)
+        }
+        for (block in blocks) {
+            if (block == head && lastAddTime.passedSince() < 200.milliseconds) {
+                continue
+            }
+            event.drawEdges(block, state.color.toColor(), 2, true)
+        }
```
Thunderblade73: Line Drawer! The Line Drawer should wrap around the for.

Thunderblade73 (changes requested):
Just some cleanup is needed. Nothing major.

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveLivingMetalHelper.kt
hannibal002: yes.
while testing snake, i accidentally clicked the ore block that activated the metal helper, and it took forever for the highlight to dissapear. 
while actively farming those blocks you dont need that long of a highlight imo.

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
hannibal002: wdym? this is alr in line drawer

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
hannibal002: i dont see a need for the calculate edges logic outside of a render context.

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/snake/LivingCaveSnake.kt
hannibal002: invalidSizeOrInvalidDistanceBetweenBlocks? that better?

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
Thunderblade73: > wdym? this is alr in line drawer

It uses a LineDrawer but it the method is not inside a LineDrawer.

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/snake/LivingCaveSnake.kt
Thunderblade73: invalidShape or something smaller than invalidSizeOrInvalidDistanceBetweenBlocks

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/snake/LivingCaveSnake.kt
hannibal002: thougth about this, but idk.
this would require us to move the internal line draw logic outside the utils, and idk if i like this

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
Thunderblade73: > i dont see a need for the calculate edges logic outside of a render context.

This could be used for collision detection or whatever. (Not a realistic use but possible). I just don't want as much as possible outside of RenderUtils that does not need to be there. Since that file is way to big.

src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/snake/LivingCaveSnake.kt
Thunderblade73: > thougth about this, but idk. this would require us to move the internal line draw logic outside the utils, and idk if i like this

This is no problem if you apply the other change I suggested.

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
Thunderblade73: Something like that:
```kotlin
LineDrawer.drawEdges(....)
```
(But probably not as an extension, just make it a normal method)

hannibal002: the function draw3DLine we call inside the loop is inside line drawer. idk what you mean
image

Thunderblade73: Yes but not the whole functions. Eg: The loop

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
hannibal002: now i get it. will test if my impl works

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
hannibal002: fair

Thunderblade73 (approved):

PR #323, opened 2023-07-18
Added shadow fury to ItemAbilityooldown

superhize:


src/main/java/at/hannibal2/skyhanni/features/itemabilities/abilitycooldown/ItemAbilityCooldown.kt
```diff
@@ -42,6 +42,9 @@ class ItemAbilityCooldown {
             if (event.pitch == 0.61904764f && event.volume == 1f) {
                 ItemAbility.GYROKINETIC_WAND_LEFT.sound()
             }
+            if (event.pitch == 1f && event.volume == 1f){
+                ItemAbility.SHADOW_FURY.sound()
```
hannibal002, line 50: Maybe check here if the player has a shadow fury in the hand (since the sound could be caused by different mechanics)

hannibal002 (changes requested):

hannibal002 (approved):

PR #324, opened 2023-07-18
Use New Elite API

ptlthg:
- Updates the old `elitebot.dev/api` references to instead use the new `api.elitebot.dev` endpoints 
- Makes the upcoming players for the ETA display into a list, which allows the next player to instantly be displayed when passing someone on the leaderboard
- Changes the crop farming weight values to the updated ones
- Removes rounding from the middle of the weight calculation, as it's no longer done that way


ptlthg:
While it would be ideal for this to be included in 0.19, I'm not entirely confident I didn't mess anything up. 

In the meantime I've added temporary api routes that will keep supporting old SkyHanni users regardless so there's no rush with this.

hannibal002:
I have tested it, its working fine. Will add it in the next beta in couple minutes

hannibal002 (approved):

PR #325, opened 2023-07-19
Eman slayer Features

dragon99z:
I've added some highlighting for the beacons and skulls + a warning then a beacon is thrown


hannibal002 (approved):
Looking good and worked for me

PR #335, opened 2023-07-25
Feat: Sync Jacob Contests

ptlthg:
Adds config settings to fetch and share the upcoming Jacob's Contests for the year to save users from all needing to open their calendar every week.

To-Do
- [x] Add clickable message to submit contests for the default setting if the player has loaded them manually. (Don't really need to check if the server has them already, not that big of a deal)
- [X] Test everything more, ensure that toggling everything off works still


ptlthg:
This also happens to fix #98

ptlthg:
_Should_ be working and ready for review now. (Also note the 'master' branch I merged from my repo is equal to 'beta' here)

hannibal002 (approved):

PR #339, opened 2023-07-27
Feature: AH estimated item value comparison

hannibal002:
## What
Makes it easier to find good/bad items in ah compared to their estimated item value.
<details>
<summary>Images</summary>

image
image

</details>

## Changelog New Features
+ Added AH Show Price Comparison. - hannibal2
    * Highlight auctions based on the difference between their estimated value and the value they are listed for.
    * Options to change the colours


CalMWolfs:
I have fixed this up so now it is usable.
Also does not feel very op or too powerful

CalMWolfs (approved):
your changes look good, just changed the pattern slightly and used pattern groups

PR #342, opened 2023-07-28
Watchdog hider improvement

appable0:
Attempts to hide watchdog messages by matching the watchdog message to a multi-line template including the empty lines around it. 

This is intended to resolve the issue where the Watchdog hider leaves two extra blank lines in chat. The empty line hider is a current work-around, but messes with other chat elements.

I am not sure why it doesn't properly hide the empty lines currently, as they seem to be properly filtered in logs.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
@@ -12,7 +12,7 @@ class ChatFilter {
     @SubscribeEvent
     fun onChatMessage(event: LorenzChatEvent) {
         if (!LorenzUtils.onHypixel) return
-
+        if (event.blockedReason != "") return
         val blockReason = block(event.message)
         if (blockReason != "") {
             event.blockedReason = blockReason
@@ -27,9 +27,7 @@ class ChatFilter {
         isGuildExp(message) && config.guildExp -> "guild_exp"
         friendJoin(message) && config.friendJoinLeft -> "friend_join"
         killCombo(message) && config.killCombo -> "kill_combo"
-        watchdogAnnouncement(message) && config.watchDog -> "watchdog"
         profileJoin(message) && config.profileJoin -> "profile_join"
-
         bazaarAndAHMiniMessages(message) && config.others -> "bz_ah_minis"
         slayer(message) && config.others -> "slayer"
         slayerDrop(message) && config.others -> "slayer_drop"
@@ -40,7 +38,6 @@ class ChatFilter {
         winterIsland(message) && config.others -> "winter_island"
         uselessWarning(message) && config.others -> "useless_warning"
         annoyingSpam(message) && config.others -> "annoying_spam"
-
         isWinterGift(message) && config.winterGift -> "winter_gift"
         isPowderMining(message) && config.powderMining -> "powder_mining"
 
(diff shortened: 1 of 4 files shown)


appable0:
This actually worked, I just had a ChatTriggers prototype that didn't work and was re-adding the blank lines. Oops!

hannibal002:
I am not sure that I understand the logic of the changed code properly.
It looks to me like it only blocks the same messages as before?

appable0:
The Watchdog message looks like this:
Screenshot 2023-07-29 at 2 44 44 PM
The current Watchdog hider correctly filters the Watchdog lines, but it leaves behind two empty lines.
Screenshot 2023-07-29 at 2 48 22 PM
The best current solution is to use the empty lines hider, but that solution takes away many other empty lines in other contexts.
Screenshot 2023-07-29 at 2 51 47 PM
Screenshot 2023-
07-29 at 2 50 28 PM
Ideally, the Watchdog hider should hide the complete Watchdog message, including the empty lines. This is difficult because it's not possible to tell whether the first empty line received will be associated with Watchdog.

PatternHider solves this by immediately hiding lines and storing a buffer of hidden lines. If it encounters a line that doesn't match the multi-line pattern it expected, it can simply revert using its buffer (this happens either if a chat message doesn't match or if it times out in two client ticks).

src/main/java/at/hannibal2/skyhanni/features/chat/PatternHiderManager.kt
```diff
@@ -0,0 +1,51 @@
...
+                Minecraft.getMinecraft().thePlayer.addChatMessage(counter.chat)
+            }
+        }
+    }
+
+    companion object {
+        val watchdogPattern = listOf(
+            Regex("§f"),
+            Regex("§4\\[WATCHDOG ANNOUNCEMENT]"),
+            Regex("§fWatchdog has banned §r§c§l(.*)§r§f players in the last 7 days."),
```
hannibal002, line 43: The brackets for `.*`in this  regex can be removed

src/main/java/at/hannibal2/skyhanni/features/chat/PatternHiderManager.kt
```diff
@@ -0,0 +1,51 @@
...
+import at.hannibal2.skyhanni.utils.LorenzUtils
+import net.minecraft.client.Minecraft
+import net.minecraft.util.IChatComponent
+
+class PatternHiderManager(skyHanniMod: SkyHanniMod) {
+
+    private val referenceCounter = mutableListOf<Counter>()
+
+    init {
+        val watchdogHider = PatternHider(this, "watchdog", watchdogPattern) {
```
hannibal002, line 13: We could make pattern Hider a data class and keep the logic methods and the forge events in the manager class

src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
```diff
@@ -27,9 +27,7 @@ class ChatFilter {
         isGuildExp(message) && config.guildExp -> "guild_exp"
         friendJoin(message) && config.friendJoinLeft -> "friend_join"
         killCombo(message) && config.killCombo -> "kill_combo"
-        watchdogAnnouncement(message) && config.watchDog -> "watchdog"
         profileJoin(message) && config.profileJoin -> "profile_join"
```
hannibal002, line 30: The empty lines you removed here were intentional, to mark the legacy `config.others` code that should be moved into separate config options anyway

hannibal002 (changes requested):
All other changes are looking good to me.

PR #350, opened 2023-08-01
Dicer changes

CalMWolfs:
- Added to $/hr
- Made rare drops into a regex

Notes:
- Due to the way the current $/hr thing is I don't have the time to add it to /shshowcropmoneycalculation yet and I'm not really sure how to
- Maybe it is good to put the dicer drop regex in the repo instead
- Can prob use same logic to add to the contest estimates and whatever

requires https://github.com/hannibal002/SkyHanni-REPO/pull/19


CalMWolfs:
Seemed to be within 0.5% of what the expected value of pumpkin should be and melon looked accurate as well

hannibal002 (approved):

PR #351, opened 2023-08-01
Performance improvements for the damage indicator

lineargraph:
I dont know why, but this feature was pulling about 5% of my frametimes because it did not disable itself while turned off (also because iterating over all entities instead of just keeping a list of relevant boss entities using entity world add and entity world remove events is kinda weird, but i didnt want to bother fixing that for a feature i dont use).

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/slayer/enderman/EndermanSlayerHideParticles.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/slayer/enderman/EndermanSlayerHideParticles.kt
@@ -14,7 +14,7 @@ import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
 
 class EndermanSlayerHideParticles {
 
-    private var endermanLocations = listOf<LorenzVec>()
+    private var endermanLocations = sequenceOf<LorenzVec>()
 
     @SubscribeEvent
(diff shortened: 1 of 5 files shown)


lineargraph:
Yeah also i made it so that there isnt a new allocation of the entire map for each entity for each tick (not ideal, ideally we would still filter over which entities we iterate, but this will do for now)

src/main/java/at/hannibal2/skyhanni/features/slayer/enderman/EndermanSlayerHideParticles.kt
```diff
@@ -14,7 +14,7 @@ import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
 
 class EndermanSlayerHideParticles {
 
-    private var endermanLocations = listOf<LorenzVec>()
```
lineargraph, line 17: are you sure this works? sequences can only be consumed once

hannibal002 (approved):

src/main/java/at/hannibal2/skyhanni/features/slayer/enderman/EndermanSlayerHideParticles.kt
hannibal002: No i am not and it didnt work at all

PR #352, opened 2023-08-01
Ghost bestiary update

superhize:
Fixed fews thing too


src/main/java/at/hannibal2/skyhanni/features/misc/ghostcounter/GhostCounter.kt
```diff
@@ -74,8 +76,15 @@ object GhostCounter {
     private var killETA = ""
     private var currentSkill = ""
     private var currentSkillLevel = -1
+    private val configUpdateVersion = 1
 
-    var bestiaryUpdate = false
+    init {
```
hannibal002: This `init` block gets called on startup, way before `profileSpecific` gets initialized. The field `hidden` will always be null here.

hannibal002 (changes requested):

hannibal002 (approved):

PR #353, opened 2023-08-02
Ender Node Tracker

ericpretzel:
Tracks items and profit obtained from mining ender nodes and killing normal endermen.
Sample image 
image

Mite gel calculation is super scuffed. I don't think there's a good way to do it though. The drops go directly to your (dragon/mining/combat) sacks, so the on-pickup event is not triggered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,27 @@
 # SkyHanni - Change Log
 
-## Version 0.19 (2023-08-01)
+## Version 0.20 (Unreleased)
+
+### New Features
++ Replacing command `/warp is` with `/is`.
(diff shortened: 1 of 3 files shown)


CalMWolfs:
Haven't tested it yet but code seems alright to me. Just maybe you want to change the only in end setting. I don't see any reason for it to be on outside of the end but maybe this could be turned into a setting so it only shows in that main part of the end and not the dragons nest, the enderman slayer area, the bruiser hideout or above the end island

Also in future you should make a new branch instead of merging from beta

hannibal002 (approved):
Thank you for this feature <3

hannibal002:
@ericpretzel message me on discord for the contributor role :)

PR #357, opened 2023-08-03
Ender node tracker

ericpretzel:
Implements ender node tracker. Same as #353, but I fixed the branch this time :) I think

Sample image again:
image

Also uses a somewhat better mite gel calculation than last time, and only shows the tracker in the specific area where end nodes are.


hannibal002 (approved):
I will approve this pr as is since the feature is almost perfect imo.

Two things for the future or maybe even for a follow-up PR:
use S0DPacketCollectItem as seen in https://github.com/hannibal002/SkyHanni/blob/cb0f62d90fce450ea5b663feb2df36b5998d959e/src/main/java/at/hannibal2/skyhanni/features/slayer/SlayerItemProfitTracker.kt#L125 This should be more accurate as your current approximation with random.

Please storage fields like totalNodesMined, totalEndermiteNests or lootCount directly from the config instead of loading/saving via the events.

PR #367, opened 2023-08-06
Basic test for dungeon items

lineargraph:
Red X is not caused by this PR, but by an already invalid behaviour in `beta` that i just wrote a test for.


hannibal002 (approved):
"basic" = adding 20 new files in 3 languages and 6 packages overall. sure nea, surely basic

PR #369, opened 2023-08-08
Fix typo

buthed010203:
I noticed this and it bothered me so I fixed it :)


hannibal002 (approved):
Thanks for finding and fixing this!

PR #375, opened 2023-08-10
Using NEUInternalName in MinionCraftHelper

hannibal002:
Using NEUInternalName in MinionCraftHelper.
This still has problems that it broke some stuff in MinionCraftHelper


walkerselby:
This PR has been replaced by #619

PR #378, opened 2023-08-10
Better Clipboard

hannibal002:
Added delay and only showing error message on the 3rd error in a row

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/ClipboardUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/ClipboardUtils.kt
@@ -1,0 +1,83 @@
+package at.hannibal2.skyhanni.utils
+
+import at.hannibal2.skyhanni.SkyHanniMod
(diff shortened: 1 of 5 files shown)


src/main/java/at/hannibal2/skyhanni/utils/ClipboardUtils.kt
```diff
@@ -0,0 +1,83 @@
...
+        return result
+    }
+
+    private fun runDelayed(duration: Duration, runnable: () -> Unit) {
+        Timer().schedule(object : TimerTask() {
+            override fun run() {
+                runnable()
+            }
+        }, duration.inWholeMilliseconds)
+    }
```
lineargraph, line 35: There is not really a point to use suspend if youre just gonna use threads anyway, no? Shouldnt this just be https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/delay.html

PR #383, opened 2023-08-12
Add Translator

NetheriteMiner:
There is some slight buggy behavior when translating from English but only when translating multiple sentences. I will fix this eventually, but it should not be a big problem in the meantime.


NetheriteMiner:
Fixed translating from Other -> English always failing.
Fixed the extra characters when translating multiple sentences in English -> Other.

hannibal002 (changes requested):
The current handling of click event will break other click events on player messages from other mods, e.g., neu's open PV feature.
How about we only replace the message part with the translator click and keep the PV open click action on the player name part?

hannibal002:
image
This bug still happens with neu alpha 19. 
Tested outside dev environment

src/main/java/at/hannibal2/skyhanni/features/chat/Translator.kt
```diff
@@ -0,0 +1,217 @@
...
+            if (!SkyHanniMod.feature.chat.translator) return
+            var message = ""
+            for (i in args) {
+                message = "$message$i "
+            }
+
+            val translation = getTranslationToEnglish(message)
+
+            if (translation == "Unable to translate!") LorenzUtils.chat("§c[SkyHanni] Unable to translate message :(")
+            else LorenzUtils.chat("§6[SkyHanni] §eFound translation: $translation")
```
jani270: pls fix color code

jani270 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/chat/Translator.kt
Rohan-Sangani: which color code needs changing? the gold §6?

jani270: correct

hannibal002 (approved):
causes neu problems in dev env, but production is working fine so it doesnt matter

PR #387, opened 2023-08-15
Add more z-indexing

lineargraph:
This should also fix the itemstacks peeking through chat or the scoreboard.


hannibal002 (approved):
bad stack size, will look into that later someday maybe (or never)

lineargraph:
> bad stack size, will look into that later someday maybe (or never)

its pretty easy just out of scope for me

PR #389, opened 2023-08-15
Potion display bug fix and timer class

appable0:
- Fixed duration format code issue due to value being in milliseconds rather than seconds
- Refactor to a timer class (future plan: save potion effects in ProfileStorage since API support is gone)


hannibal002:
> Fixed duration format code issue due to value being in milliseconds rather than seconds

How can I reproduce this error?

hannibal002 (approved):
looks fine to me

PR #400, opened 2023-08-19
Add default option selector

lineargraph:
Users will be prompted the first time they open SkyHanni to choose which features they want to enable or disable. Later on people can still use the command to turn off feature toggles in an entire category using /shdefaultoptions.

Future improvements: Prompt *just* for new feature toggles after an update.

Implementation note: I gave up with giving things the @FeatureToggle annotation at some point, so a few config categories are still missing it. @FeatureToggle shall be added to all boolean config options that toggle an entire feature (but not a suboption of an already existing @FeatureToggle)


hannibal002:
I'm sure this is not intentional?

image


image

lineargraph:
Code 55% faster with GitHub Copilot!

hannibal002 (approved):

PR #401, opened 2023-08-22
Garden Plot border feature

VixidDev:
Adds Garden plot borders with the use of F3 + G, similar to how chunk borders are rendered in later mc versions.

Also I removed the line `GlStateManager.disableDepth()` from `RenderUtils.kt` since it was actually making the depth parameter obsolete since this line always disabled the depth test regardless.


hannibal002 (approved):
works very well

PR #413, opened 2023-08-27
feat: show alert when reaching max superpairs clicks

ericpretzel:
plays beep sound and sends a message in chat when player reaches maximum clicks gained for superpairs minigames.


hannibal002 (approved):
I like this feature!
image

PR #422, opened 2023-08-29
Feature: Trevor tracker and other small stuff

CalMWolfs:
Trevor changes:
- Hotkey can also accept the quest
- More detection for when to show the solution
- Removed old hp values  - got all thx fycron
- Show the mob name when displaying the waypoint and message on the screen
- Tracker for Trevor
        - Shows quests done, breakdown of their rarity, animals killed vs animals that kill themself
        -  Pelts per hour

Other Change:
- Fixed up ice/hr calculation
- Fixed naming of frozen treasure fixed rate timer


NetheriteMiner:
Surprisingly, I don't think this conflicts with #416

hannibal002:
image
lets see :D

CalMWolfs:
Well these conflicts weren't directly related but fixed now

src/main/java/at/hannibal2/skyhanni/features/misc/FrozenTreasureTracker.kt
```diff
@@ -25,13 +25,13 @@ class FrozenTreasureTracker {
...
     private var lastEstimatedIce = 0L
-    private val icePerMin = mutableListOf<Long>()
+    private val icePerSecond = mutableListOf<Long>()
     private var icePerHour = 0
     private var stoppedChecks = 0
     private var compactPattern = "COMPACT! You found an Enchanted Ice!".toPattern()
 
     init {
-        fixedRateTimer(name = "skyhanni-dungeon-milestone-display", period = 1000) {
+        fixedRateTimer(name = "skyhanni-frozen-treasure-tracker", period = 1000) {
```
lineargraph, line 35: A fixed rate timer can cause a ConcurrentModificationException crash if you swap worlds right as the timer ticks.

src/main/java/at/hannibal2/skyhanni/features/misc/FrozenTreasureTracker.kt
```diff
@@ -41,31 +41,33 @@ class FrozenTreasureTracker {
...
-                icePerMin.clear()
+                icePerSecond.clear()
                 icePerHour = 0
             }
-            return
+            while (stoppedChecks > 0) {
+                stoppedChecks -= 1
+                icePerSecond.add(0)
+            }
+            icePerSecond.add(difference)
```
lineargraph, line 69: Maybe consider capping off the length of this list?

lineargraph (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/FrozenTreasureTracker.kt
CalMWolfs: My original idea was to use the LorenzTickEvent.repeatSecond(1) but this means that if the game lags and isn't at 20tps then the pelts/hr plus other parts of the tracker stop working. Right now it just has the try, catch but if you know a better way to do this lmk

lineargraph: The better way would be to have exclusivity on the data. Only one thread can modify or access that objects content. You could achieve that by using `icePerSecond = mutableListOf()` inside of the onworldchange method

lineargraph:
Yes, that is what i meant (there is still an RNG crash right now btw,  with a TOCTOU in `while (icePerSecond.size > 1200) icePerSecond.removeAt(0)`)

hannibal002 (approved):

PR #425, opened 2023-08-29
Fix translator missing a character at the end

NetheriteMiner:


hannibal002:
3 minutes too slow

NetheriteMiner:
That's ok, next beta. That 3 minutes was spent double checking that nothing else had broken

PR #427, opened 2023-08-30
Feature: Armor drops to $/hr

CalMWolfs:
Added money drops to $/hr


src/main/java/at/hannibal2/skyhanni/utils/InventoryUtils.kt
```diff
@@ -40,7 +43,14 @@ object InventoryUtils {
     fun countItemsInLowerInventory(predicate: (ItemStack) -> Boolean) =
         getItemsInOwnInventory().filter { predicate(it) }.sumOf { it.stackSize }
 
-    fun getArmor(): Array<ItemStack?> = Minecraft.getMinecraft().thePlayer.inventory.armorInventory
+    fun getArmor(dataAgeLimit: Int): Array<ItemStack?> {
```
lineargraph: Please do not pass `dataAgeLimit` as an `Int` use a `Duration`  instead.

src/main/java/at/hannibal2/skyhanni/utils/InventoryUtils.kt
```diff
@@ -40,7 +43,14 @@ object InventoryUtils {
...
+        if (lastArmorCheck.passedSince() > dataAgeLimit.milliseconds) {
+            wornArmorNames = Minecraft.getMinecraft().thePlayer.inventory.armorInventory
+            lastArmorCheck = SimpleTimeMark.now()
+        }
```
lineargraph: This does not look to me like proper delay on the armor. I can only guess the functionality of this method (which makes it bad to begin with and deserves to be commented on), but I will assume this is meant to counteract the fact that the server takes about a second to calculate whether or not you are wearing an armor set. In this case the delay might be offset from when you actually switch armor, because the timer is not aligned with any armor switch, just with the last method call.

src/main/java/at/hannibal2/skyhanni/features/garden/farming/FarmingArmorDrops.kt
```diff
@@ -88,9 +93,50 @@ class FarmingArmorDrops {
...
+
+    @SubscribeEvent
+    fun onRepoReload(event: RepositoryReloadEvent) {
+        try {
+            armorDropJson = event.getConstant("ArmorDrops") ?: error("ArmorDrops not found in repo")
+
+        } catch (e: Exception) {
+            e.printStackTrace()
+            LorenzUtils.error("error in RepositoryReloadEvent")
+        }
```
lineargraph: Please do not save the JSON file, but instead immediately parse it into a narrow value space. (See https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)

src/main/java/at/hannibal2/skyhanni/features/garden/farming/FarmingArmorDrops.kt
```diff
@@ -88,9 +93,50 @@ class FarmingArmorDrops {
...
+    }
+
+    companion object {
+        var armorDropJson = JsonObject()
+        private var currentArmorDropChance = 0.0
+        private var lastCalculationTime = SimpleTimeMark.farPast()
```
lineargraph: Is there a reason to put this in a companion object, instead of turning the entire class into an object? This just seems like a pointless separation. Alternatively you could use a dedicated API class.

src/main/java/at/hannibal2/skyhanni/features/garden/farming/FarmingArmorDrops.kt
```diff
@@ -88,9 +93,50 @@ class FarmingArmorDrops {
...
+                    .mapNotNull { it?.getInternalName_old() }
+                    .count { it.contains(armorName) || it.contains("FERMENTO") }
+
+
+                val dropRates = specialDropObject["chance"].asJsonArray
+                var dropRate = 0.0
+                if (pieceCount > 0 && dropRates.size() >= pieceCount) {
+                    dropRate = dropRates[pieceCount - 1].asDouble
+                }
+                currentArmorDropChance = (dropRate * 60 * 60.0) / 100
```
lineargraph: There are absolutely no checks on this JSON handling. Please move the json parsing into the event (as per my last review comment), and also consider using GSONs object mapping feature, instead of manually parsing it.

lineargraph (changes requested):

hannibal002 (approved):

PR #428, opened 2023-08-30
Following Line

hannibal002:


hannibal002:
causes big lags: 300 fps -> 50 fps for me

hannibal002:
Lags are gone

PR #430, opened 2023-08-31
Fix + Changes: Ender node tracker tweaks

ericpretzel:
- Now directly modifies variables in profile storage as recommended
- Fixed new issue that made the profit tracker use npc/george price instead of lowest bin for some items
- Accurately tracks mite gel now
- set ProfileStorageData.loaded to true before broadcasting ConfigLoadEvent (pretty sure this doesnt break anything). I did this so the display would update properly when first joining


hannibal002:
> set ProfileStorageData.loaded to true before broadcasting ConfigLoadEvent (pretty sure this doesnt break anything).

I have the bad feeling there was a good reason to have the value change after the event call. But I can't remember anymore, and I'm not uncertain.
Finger crossed, this will work fine.

hannibal002 (approved):

PR #437, opened 2023-09-02
Feature: Crimson Isle Fetch Quest Helper

NetheriteMiner:
This suggestion has been marked as "planned" for 28 days, so I thought it would be a good idea


hannibal002:
Please work with QuestLoader

hannibal002:
I have marked this pr as version 0.21, meaning I will merge it after 0.20 got released

lineargraph (changes requested):
I feel as though this is not fully rule complicit (or at least is violating the spirit of the rules in some way). Maybe add a button?

hannibal002 (changes requested):
I agree with nea here. It should not run the command on its own (SBE auto-refill ch pass is a cheat in my mind as well)
maybe just show a chat message "click here to grad x items from your sacks"?
Moreover, please move this logic into the sacks API, so we can use it for many other cases as well.

hannibal002:
I mark this PR as draft (see comments above)

hannibal002 (changes requested):
I found too many bigger problems to fix them myself right now: 
the fetch quest text appears even before the quest is accepted.
It doesn't check if you have the needed items already in the inventory.
It allows for multiple clicks. (It should only do one click per day)

hannibal002:
One solution to all those problems: don't work with the inventory at all. Just detect when the player accepts this quest and then shows a chat message, the user can click on.
This is more work for the user, but feels less "op" and therefore better imo

hannibal002 (approved):

PR #442, opened 2023-09-06
Anniversary Event Active Player Ticket Timer

lineargraph:


src/main/java/at/hannibal2/skyhanni/config/commands/Commands.kt
```diff
@@ -247,6 +249,10 @@ object Commands {
             "shtestmessage",
             "Sends a custom chat message client side in the chat"
         ) { TestChatCommand.command(it) }
+        // TODO: remove for commit
```
hannibal002, line 252: You want to delete that before i merge?

hannibal002 (approved):
i will remove the debug command

PR #443, opened 2023-09-06
Fix: minion pickup not removing the nametag

brandonwamboldt:
It's possible i'm missing something, but I believe the only problem was that the InventoryCloseEvent fires **after** the chat message so the logic never got triggered. I couldn't think of any reason why it needed to be done that way, so I moved code to delete the minion from the profile storage to the chat event and couldn't produce any errors during my testing.


hannibal002:
This is so simple, this could actually work.

hannibal002 (approved):

PR #441, opened 2023-09-06
Fix + Feature: Fixed inconsistencies with arachne brood highlighting and add countdown for arachne spawn

brandonwamboldt:
Sorry to bombard you with PRs today. Just fixing a few things that i've noticed 😄 

I did quite a few tests with the level 300 and 500 variants of arachne and had several runic mobs, everything highlighted correctly though.

I also added a spawn timer for after crystals are placed. It can be 1-2s off the actual spawn but i still think its useful. If you want, I can remove it though since its not entirely accurate. For the level 500 version, the timer is a few seconds after spawn, ending when she drops down (she is invincible before that). Saves some clicking when you're spawning, dont need to start swinging your sword right away

PR #444, opened 2023-09-07
Fix: 3k hp spiders being falsely highlighted as Arachne's Keepers

realooffyy:


hannibal002:
Is this different from the fix in https://github.com/hannibal002/SkyHanni/pull/441?

realooffyy:
Don't think so, this one's fixing the fact that runic Voracious Spiders with 3,000 hp are getting highlighted as Arachne's Keepers by detecting Cave Spiders instead of Spiders
edit: yeah it's different oops

hannibal002 (approved):

PR #452, opened 2023-09-08
Fix: Fix one wiltered berberis farm plot not working

### LINKED ISSUE #395: wilted berberis bug
image
Helper does not work in some locations

brandonwamboldt:
If player is in the West Village, check if they are in the bounding box of one of the farm plots that isn't in the "Dreadfarm" area.

java_2023-09-08_10-32-49

Resolves https://github.com/hannibal002/SkyHanni/issues/395


hannibal002:
Please move the logic behind the area location, so that the rift api class has nothing to change. The result should simply fake the dreadfarm location when inside the West village and inside those coordinates.

The same kind of fix is aos needed for locations in the voidglooom area and the Sven cave, would be good to have those all in once place

PR #458, opened 2023-09-09
fix crash with item rarity

CalMWolfs:
- Converting all fails to common rarity may not be the best solution but it is better than the user crashing
- We should try as much as possible to not just crash the user when something is wrong and instead use LorenzUtils.error and then return some value that we can deal with
- This will not fix the initial problem that caused the item to have no rarity and I haven't looked into that

[bug report of the initial issue](https://discord.com/channels/997079228510117908/1149879403061461086)


CalMWolfs:
closing for now

PR #462, opened 2023-09-09
Fix: Rare sea creature detection 

brandonwamboldt:
Use nametag detection for rare sea creatures to ensure no false positives (Water Hydras and T4 Revs both have 1.5m health and inherit from EntityZombie). Also added runic support to EntityLivingBase.hasMaxHealth as well as corrupted runic mobs.


hannibal002:
Runic support this way will cause many false positives I'm certain

brandonwamboldt:
Updated the PR so that it doesn't trigger in dungeons, and to properly set the boss flag where applicable. Looked through the other use cases, shouldn't cause any false positives

hannibal002 (approved):

PR #463, opened 2023-09-09
Feature: Highlight party members

brandonwamboldt:
Never lose a teammate again. ONLY works for parties to avoid griefing

java_2023-09-09_14-29-03


lineargraph (approved):

src/main/java/at/hannibal2/skyhanni/features/misc/PartyMemberOutlines.kt
```diff
@@ -0,0 +1,70 @@
...
+            partyMembers.addAll(usernames)
+        } else if (message.matches(partyKickMessage)) {
+            partyMembers.remove(partyKickMessage.find(message)!!.groupValues[1])
+        } else if (message.matches(partyLeftMessage)) {
+            partyMembers.remove(partyLeftMessage.find(message)!!.groupValues[1])
+        } else if (message.matches(partyListMessage)) {
+            val usernames = partyListMembers.findAll(message).map { it.groupValues[1] }
+            partyMembers = usernames.toMutableSet()
+        }
+    }
```
CalMWolfs: Can we work with PartyAPI for this?

CalMWolfs (commented):
Looking good i think best to use existing party list features, though if there is any problems with the regex there feel free to fix but i think its fine. Also maybe an option to change outline colour (either colour selector or drop down list of all lorenz colours)

src/main/java/at/hannibal2/skyhanni/features/misc/PartyMemberOutlines.kt
brandonwamboldt: Didn't see PartyAPI back when i built this so i will definitely clean this up and use that!

brandonwamboldt: Color is configurable and now using PartyAPI

java_2023-10-03_17-16-40
java_2023-10-03_17-16-48

CalMWolfs (approved):

hannibal002 (approved):

PR #464, opened 2023-09-10
Feature: gui scaling

lineargraph:


hannibal002 (approved):
Upgrading from a config without the scale element causes the factor field to be 0, not 1. this will make all guis very small. will fix this in a commit after this pr

PR #467, opened 2023-09-11
Feature: Player symbols in chat

CalMWolfs:
Adds chat symbols such as ironman/bingo/nether faction like SBA had/has
- Having an emblem doesn't break detection of these symbols
- Ability to change where it displays (left of name, right of name, hide altogether)
- Works without having sb levels in chat (Hypixel's emblems don't)
- Also remove ClientChatReceivedEvent from everywhere that didn't need it
- And most importantly doesn't break chat components from Hypixel or other mods


src/main/java/at/hannibal2/skyhanni/data/ChatManager.kt
```diff
@@ -83,7 +83,7 @@ object ChatManager {
 
         if (message.startsWith("§f{\"server\":\"")) return
         val key = IdentityCharacteristics(original)
-        val chatEvent = LorenzChatEvent(message, original)
+        val chatEvent = LorenzChatEvent(message, original, event.type)
```
lineargraph: Not needed at all. Check line 79:
```kt
if (event.type.toInt() == 2) return
```

src/main/java/at/hannibal2/skyhanni/features/misc/PlayerChatSymbols.kt
```diff
@@ -0,0 +1,98 @@
...
+
+            val split = usernameWithSymbols.split("$username ")
+            var emblemText = if (split.size > 1) split[1] else ""
+            emblemText = StringUtils.removeResets(emblemText)
+
+            if (emblemText != "") {
+                event.chatComponent = StringUtils.replaceFirstChatText(event.chatComponent, "$emblemText ", "")
+
+                StringUtils.modifyFirstChatComponent(event.chatComponent) { component ->
+                    if (component is ChatComponentText && component.text.contains(username)) {
```
lineargraph: Use `getUnformattedTextForChat()` instead of `text`. Also avoids having to cast to ChatComponentText

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -152,4 +156,33 @@ object StringUtils {
...
+        }
+        return false
+    }
+
+    // replaces without breaking any click or hover events (unless that whole text is removed)
+    fun replaceFirstChatText(chatComponent: IChatComponent, toReplace: String, replacement: String): IChatComponent {
+        modifyFirstChatComponent(chatComponent) { component ->
+            if (component is ChatComponentText && component.text.contains(toReplace)) {
+
+                component.text = component.text.replace(toReplace, replacement)
```
lineargraph: Use `getUnformattedTextForChat()` here too

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -152,4 +156,33 @@ object StringUtils {
...
+                true
+            } else {
+                false
```
lineargraph: You can remove the if and instead directly use a `return@modifyFirstChatComponent /* your if condition here 
*/`

lineargraph (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -44,7 +44,7 @@ object StringUtils {
     /**
      * From https://stackoverflow.com/questions/10711494/get-values-in-treemap-whose-string-keys-start-with-a-pattern
      */
-    fun <T> subMapWithKeysThatAreSuffixes(prefix: String, map: NavigableMap<String?, T>): NavigableMap<String?, T>? {
+    fun <T> subMapWithKeysThatAreSuffixes(prefix: String, map: NavigableMap<String?, T>): Map<String?, T>? {
```
lineargraph: Any reason for this change? NavigableMap is an interface that provides additional functionality and guarantees over Map, so this feels like a bit of a waste.

hannibal002 (approved):

PR #11, opened 2023-02-17
Amazing Yaw Snapping

lineargraph:

PR #470, opened 2023-09-11
Feature: Add /shmouselock command to lock mouse rotation for farming

brandonwamboldt:
Works be setting mouse sensitivity to 0 under the hood


lineargraph:
I do not feel comfortable reviewing this, after https://github.com/hannibal002/SkyHanni/pull/11 got declined.

brandonwamboldt:
Understandable but i think this should be different since you can just manually adjust sensitivity to achieve the same effect

lineargraph:
Same with my PR, tbh. And still way less cheaty than the wallhacks that some other people have felt comfortable adding recently. To clarify: i dont feel uncomfortable because i think you or my PR is cheating, i feel uncomfortable because my original PR got rejected as a cheat, and this mod has just accepted cheats nowadays.

hannibal002 (approved):

PR #471, opened 2023-09-11
Added clickable fake garden level up message

j10a1n15:
Makes this fake garden level up message clickable, to open /gardenlevels, like all the other level up messages 
image

dont ask, feels more natural like this


hannibal002 (approved):
I did not know this command even existed!

PR #474, opened 2023-09-12
Fix: Translator changing colors and hiding messages

NetheriteMiner:
- Uses the existing chat style instead of creating a new one so that messages don't become white
- Added detection for if a message is by a human instead of just using a regex for if it contains a colon (not perfect because hypixel dislikes me)

Please test this. The chat components that my dev env received were different than the ones that my prod env received, so I expect there are more situations that I need to account for.


NetheriteMiner:
Fixed in #482 in a better way

PR #477, opened 2023-09-13
Fix: Improve performance for tooltips and item ability cooldown feature (fixes item cooldown flickering)

brandonwamboldt:
Reduce calls to NEU's getInternalName as it's a very costly function that we're calling quite often (replaced one occurrence of it being called for each line in item lore). Optimize item ability cooldown to reduce the CPU overhead as well as fix the item cooldown text "flickering" every other frame.

Profiling with Spark indicated a roughly 30% CPU usage reduction for normal walking around with items in my hotbar, and significant improvements to tooltips (specifically the RNG meter tooltips which were the slowest).

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/ItemDisplayOverlayFeatures.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/ItemDisplayOverlayFeatures.kt
@@ -148,7 +148,7 @@ class ItemDisplayOverlayFeatures {
         }
 
         if (SkyHanniMod.feature.inventory.itemNumberAsStackSize.contains(12)) {
-            if (item.getInternalName_old() == "LARVA_HOOK") {
+            if (itemName.contains("Larva Hook")) {
                 for (line in item.getLore()) {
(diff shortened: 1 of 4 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/ItemDisplayOverlayFeatures.kt
```diff
@@ -148,7 +148,7 @@ class ItemDisplayOverlayFeatures {
         }
 
         if (SkyHanniMod.feature.inventory.itemNumberAsStackSize.contains(12)) {
-            if (item.getInternalName_old() == "LARVA_HOOK") {
+            if (itemName.contains("Larva Hook")) {
```
hannibal002, line 151: I would prefer to use the internal name for checks like this in the future (I know internal name check is slower than item name check, but hypixel tends to rename the item names sometimes and internal names dont change) also this whole method should get an cache so this doenst call each frame

hannibal002 (approved):
It feels like immediately after right click spamming an ability, the item background still flickers a bit. but feels definetly better now. thanks for the improvements!

PR #478, opened 2023-09-14
Feature: Require confirmation on certain quick craft items

brandonwamboldt:
Optional feature to require confirmation to quick craft items (via ctrl+clicking the quick craft slot). Items that do NOT require confirmation are configurable in the repo as we may wish to expand the list in the future. For now, any item that goes in sacks can be quick crafted (e.g. enchanted blocks), as well as a few carpentry items, but the vast majority of items require quick crafting. My logic is that this feature is for people who normally only use quick crafting to compact items, craft enchanted blocks, etc, and to prevent accidentally crafting a 2nd pair of melon boots 😭 

See corresponding REPO change here: https://github.com/hannibal002/SkyHanni-REPO/pull/35

java_2023-09-13_23-14-35


lineargraph (changes requested):
There definitely needs to be a way to configure this as a user. Either via command, orrr if you are able to via a GUI. I would be able to help with a GUI, but it shouldn't be too hard. A very similar GUI you could look at is https://github.com/NotEnoughUpdates/NotEnoughUpdates/blob/master/src/main/kotlin/io/github/moulberry/notenoughupdates/miscgui/DynamicLightItemsEditor.kt

brandonwamboldt:
My main concern is that there are both too many items to expect a user to whitelist or blacklist, so shipping a base set of items we think wouldn't be accidental crafts would make the feature the most useful. But a command to add additional items to the whitelist makes sense and would be easy to add so i'll look at that

hannibal002 (approved):

PR #480, opened 2023-09-14
Fix: Compensate for mage ability cooldown reduction while in dungeons

brandonwamboldt:
I wasn't able to test with all of the weapons, but the ones I could test with worked as expected in dungeons.


hannibal002:
`Gyrokinetic Wand right click` ability cooldown no longer works inside dungeon.

hannibal002 (approved):

PR #482, opened 2023-09-16
Backend: Chat Component and LorenzChatEvent Stuff

CalMWolfs:
Adds functions for replacing parts of a chat component. Also turns ClientChatReceivedEvent into LorenzChatEvent where possible and adds the event type into LorenzChatEvent

also maybe fix more translator bugs by adding a from player function


hannibal002:
Do we really need the event type byte in the translator check?

src/main/java/at/hannibal2/skyhanni/events/LorenzChatEvent.kt
```diff
@@ -2,5 +2,9 @@ package at.hannibal2.skyhanni.events
 
 import net.minecraft.util.IChatComponent
 
-class LorenzChatEvent(val message: String, var chatComponent: IChatComponent, var blockedReason: String = "") :
+class LorenzChatEvent(
+    var message: String,
+    var chatComponent: IChatComponent,
+    var blockedReason: String = ""
+) :
     LorenzEvent()
```
lineargraph: Maybe move the LorenzEvent() one line up then?

lineargraph (approved):

hannibal002 (approved):

PR #489, opened 2023-09-18
 Backend: more efficient implementation of String.removeColor() method (StringUtils.kt:)

brainage04:
Tiny optimization, will reduce the number of loop iterations in `String.removeColor()` by `n`, where `n` is the number of formatting characters (`§`) in the string being formatted.

Resubmitted with more appropriate branch name.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
@@ -19,20 +19,17 @@ object StringUtils {
     }
 
     fun String.removeColor(): String {
-//        return replace("(?i)\\u00A7.", "")
-
         val builder = StringBuilder()
-        var skipNext = false
-        for (c in this.toCharArray()) {
-            if (c == '§') {
-                skipNext = true
+
+        var counter = 0
+        while (counter < this.length) {
+            if (this[counter] == '§') {
+                counter += 2
                 continue
+            } else {
+                builder.append(this[counter])
+                counter++
             }
-            if (skipNext) {
-                skipNext = false
-                continue
-            }
-            builder.append(c)
         }
 
         return builder.toString()


hannibal002:
I have tested this optimization: ([Test code](https://gist.github.com/hannibal002/44086924bc0c236c06e1e4f1421b56ec))



After 350k runs with real life data:
```
total time: '89.416600ms' vs '58.966100ms' vs '325.373200ms' vs '419.861600ms'
testLog.log("total time: '$totalOld' vs '$totalNew' vs '$totalRegexCached' vs '$totalRegexUncached'")
```

image
Interesting results indeed.

hannibal002:
https://gist.github.com/hannibal002/fe5f04991841b0f2e00cfaa0b101241b
I changed the logging process in this example and added the number of runs:
```
numberOfRuns: 346963 = '73.690300ms' vs '101.772400ms' vs '281.423500ms' vs '420.686900ms'
```

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -19,20 +19,17 @@ object StringUtils {
...
-                continue
-            }
-            if (skipNext) {
-                skipNext = false
+
+        var counter = 0
+        while (counter < this.length) {
+            if (this[counter] == '§') {
+                counter += 2
                 continue
```
lineargraph, line 28: I think that continue can be removed

lineargraph (approved):

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -19,20 +19,17 @@ object StringUtils {
     }
 
     fun String.removeColor(): String {
-//        return replace("(?i)\\u00A7.", "")
-
         val builder = StringBuilder()
```
lineargraph, line 22: Also: we can initialize this StringBuilder with a capacity
```suggestion
        val builder = StringBuilder(this.length)
```

hannibal002 (approved):

PR #492, opened 2023-09-21
Feature: Modify Visual Words

CalMWolfs:
Just like SBE modify visual words
- Can modify words anywhere
- Option to work on all of Hypixel (not other servers or singleplayer)
- Can quickly enable/disable phrases
- Can change the order of the phrases
- See preview of what the text will look like
- Slightly changed other things in /ff (to stop code duplication)
- Something else i think

Also while in the menu no visual words will be modified. Future plans include making a way to mark text as unmodifyable


Obsidianninja11:
Visual words:
It would be nice to be able to highlight and move the cursor in the config.
Pasting doesn't work in the config.
You can add a blank word to replace (having blank for the replacer is fine, but having a replaced word be blank is useless?) maybe grey out back button when there's nothing?
Color codes don't work as far as I can tell (I tried with &). You can't get § without pasting.
Why are there double lines when you add/edit words?

CalMWolfs:
> It would be nice to be able to highlight and move the cursor in the config. 

Yeah, but this is a lot of effort that I do not have time for. I've made it so you can control/shift + backspace to get to the start. I may add this later when I have time

> Pasting doesn't work in the config. 

Added in newest commit

> You can add a blank word to replace (having blank for the replacer is fine, but having a replaced word be blank is useless?) maybe grey out back button when there's nothing? 

Yeah, but I don't want to lock them into not being able to do anything. And besides if the string to replace is blank it is skipped anyways.  Now I've also added a thing where if the phrase and replacement are both empty they get removed

> Color codes don't work as far as I can tell (I tried with &). You can't get § without pasting. 

You need to use && instead of just & I've added a thing to explain this

> Why are there double lines when you add/edit words?

One is the preview of the raw text (with the && plus the formatting character afterwards) I've added a thing to explain this

ThatGuyWhomAsked:
So I think since its already been added to the 0.21 milestone we should merge this into the beta 22

CalMWolfs:
> So I think since its already been added to the 0.21 milestone we should merge this into the beta 22

0.21 refers to mod version 0.21 not beta 0.21. So a few more betas, then version 0.20 fully releases then once 0.21 betas start this will be added
 Also, this prob needs more testing.

hannibal002:
> So I think since its already been added to the 0.21 milestone we should merge this into the beta 22

The 0.21 milestone means this pr will only get added after the 0.20 full release is out.
For this, all 0.20 milestone PRs need to be merged first.

ThatGuyWhomAsked:
> > So I think since its already been added to the 0.21 milestone we should merge this into the beta 22
> 
> The 0.21 milestone means this pr will only get added after the 0.20 full release is out. For this, all 0.20 milestone PRs need to be merged first.

Ok, thanks for explaining

Vahvl:
could this be made to work with neu's per letter chroma thingie? (&&z (z not Z) by default in neu)

CalMWolfs:
> could this be made to work with neu's per letter chroma thingie? (&&z (z not Z) by default in neu)

It could but any of that will be done in a different pr

Also i think 2-3 of you were using this instead of the actual versions so I've updated this fork to be up to date

j10a1n15:
The command to open the gui isnt working anymore (Unknown command. Type "/help" for help. ('shwords') (3)), only works now using the button in the config

hannibal002:
> Slightly changed other things in /ff
> Something else i think

Why merge entirely different stuff in the same pr :(

Vahvl:
the ultimate pr

CalMWolfs:
> The command to open the gui isnt working anymore (Unknown command. Type "/help" for help. ('shwords') (3)), only works now using the button in the config

weird i think github just deleted it when i updated the pr. anyways fixed now

Vahvl:
hi guys merge beta 3 please

j10a1n15:
Bug: If you try to replace the last char in a scoreboard line, it doesnt work.
So for example: trying to replace `www.hypixel.net` doenst work, but `www.hypixel.ne`
Same with ` Late Autumn 24th` doenst work, ` Late Autumn 24t` does

hannibal002:
> Bug: If you try to replace the last char in a scoreboard line, it doesnt work. So for example: trying to replace `www.hypixel.net` doenst work, but `www.hypixel.ne` Same with ` Late Autumn 24th` doenst work, ` Late Autumn 24t` does

Technically, the sidebar (scoreboard to the right) is rendered as 2 separate strings next to each other. We would need to hard-code this edge case.

CalMWolfs:
> Bug: If you try to replace the last char in a scoreboard line, it doesnt work.
> 
> So for example: trying to replace `www.hypixel.net` doenst work, but `www.hypixel.ne`
> 
> Same with ` Late Autumn 24th` doenst work, ` Late Autumn 24t` does

I do not plan on fixing this as checking each string for one of these 15 characters then removing it seems like a waste of resources unless anyone feels like it would be fine to do or has a more efficient idea in mind

j10a1n15:
Hello please update to beta 4 thank you

hannibal002 (approved):

PR #497, opened 2023-09-23
Fix: Fragged spirit mask cooldown doesn't work

brandonwamboldt:
Ha just kidding, can't let you fix it first


hannibal002 (approved):
Fastest one liner in the next beta

PR #498, opened 2023-09-23
Additon: Add /gd as supplement to /warp garden

RayDeeUx:


Absterge:
Not against this being added per se, just that it's already possible in both Skytils and DSM. Is this really needed?

hannibal002:
I would either implement the full logic of https://discord.com/channels/997079228510117908/1149744503234764821 directly, or nothing at all. Currently, this is just Skytils with fewer options.

hannibal002:
As cal found out, it is the feature "optimised font renderer" from patcher

src/main/java/at/hannibal2/skyhanni/features/commands/WarpGdCommand.kt
```diff
@@ -0,0 +1,22 @@
...
+
+class WarpGdCommand {
+
+    @SubscribeEvent
+    fun onSendPacket(event: PacketEvent.SendEvent) {
+        if (!LorenzUtils.inSkyBlock) return
+        if (!SkyHanniMod.feature.commands.supplementWarpGarden) return
+
+        val packet = event.packet
+        if (packet is C01PacketChatMessage && packet.message.lowercase() == "/gd") {
```
hannibal002, line 17: We should start using a "player send message to server" event

hannibal002:
If this pr prepared for additional garden warp features (e.g. with key bind that only works inside the garden or hub) then I would merge it.

But just as a command alias, this is useless compared to Skytils' feature.

PR #508, opened 2023-09-26
Fix: Item with no UUID do not show cooldown correctly

brandonwamboldt:
Fix an issue where items with no UUID won't correctly show their cooldown, and where some items may incorrectly show a cooldown when there isnt one.

If no UUID or ID is present on the item, don't check it for cooldowns.


hannibal002 (changes requested):
Have you forgotten to commit changes in SkyBlockItemModifierUtils.kt ?
image

hannibal002:
Found the method in #509, nevermind!

hannibal002 (approved):

PR #519, opened 2023-09-29
Feature: Make normal clicks to shift clicks in equipment inventory

Thunderblade73:


src/main/java/at/hannibal2/skyhanni/config/features/InventoryConfig.java
```diff
@@ -415,5 +415,10 @@ public static class TiaRelayConfig {
     @FeatureToggle
     public boolean highlightAuctions = true;
 
+    @Expose
+    @ConfigOption(name = "Shift Click For Equipment", desc = "Makes normal clicks to shift clicks in equipment inventory")
```
hannibal002: Please remove the word "for" in the option name, the text gets too small :/ 
image

hannibal002 (changes requested):
Even though the actual new code is just 3 lines, please move it inside a new class. The `HideNotClickableItems` class is already 500+ lines big and needs heavier reworks, and adding new stuff that has nothing to do with it just makes this process more confusing later.
The logic itself works perfectly for me.


Also, I don't know if we really should have it default enabled. While it's only useful for everyone and I don't see any downsides at all, it could be a bit problematic when someone argues "we change packets per default" so technically it's a bit more nuanced. In the end, your call if you want to keep it default enabled, I'm fine with it.

Thunderblade73:
Made it a self contained Feature and the default is now false (better safe than sorry)

hannibal002 (approved):

PR #524, opened 2023-09-30
Fix: Detect arachne 'quickspawn' vs regular spawn

brandonwamboldt:
This cost me so much precious enchanted string to test 😭 😭 😭 

In all my tests, quickspawning only ever had particles once, most of the time it had zero. Regular spawning always had 60+, so these margins should be fairly safe. Same with the offset from the chat message. I did about 15 tests with a mix of regular vs quickspawns to confirm. Tests that consumed enchanted string I needed for rev viscera 😭


brandonwamboldt:
Hmm the very next test i did has issues so im gonna do more tests

brandonwamboldt:
Adjusted the timer to capture particles to be more forgiving of server lag

PR #525, opened 2023-09-30
Feature: Add spawn timer support for Arachne quickspawns

brandonwamboldt:
This cost me so much precious enchanted string to test 😭 😭 😭 

In all my tests, quickspawning only ever had particles once, most of the time it had zero. Regular spawning always had 60+, so these margins should be fairly safe. Same with the offset from the chat message. I did about 15 tests with a mix of regular vs quickspawns to confirm. Tests that consumed enchanted string I needed for rev viscera 😭


hannibal002:
How do I test those quick spawns? 
Do you need the enchanted strings back? :D

brandonwamboldt:
For quick spawn, wait til the boss spawns the second round of minis, then go back to the altar and spam click the middle with your crystal as quickly as you can. This is the only approach that worked for me, but it was very consistent if you click quickly. Needs other people around to kill the boss though. Doesnt matter if the boss is a normal one or T2, just matters that you spam click your crystal as it dies.

hannibal002 (approved):

PR #526, opened 2023-09-30
Wrong/broken Changelog: Kill Detection

Thunderblade73:
This Pr will add the possibility  to recognize if the player has Kill-credit at a Mob.
Limitations: If the Player is not under the top 5 damage Dealer the Feature will break

Currently working: melee Hit detection and the recognition when a mob in the Hit-List dies

There is a lot of work ahead 😔


hannibal002:
old

PR #528, opened 2023-10-02
Feature: Highlight Jerries (Jerrypoclaypse)

RayDeeUx:
will add desc to config in a moment

highlights jerries. simple as that


hannibal002 (approved):
Highlight Jerries during the Jerrypoclaypse.

PR #530, opened 2023-10-03
Feature: Show alerts when the player catches a legendary sea creature

brandonwamboldt:
See repo update: https://github.com/hannibal002/SkyHanni-REPO/pull/40/files


src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureFeatures.kt
```diff
@@ -39,8 +43,24 @@ class SeaCreatureFeatures {
...
+    }
+
+    @SubscribeEvent
+    fun onSeaCreatureFish(event: SeaCreatureFishEvent) {
+        if (!LorenzUtils.inSkyBlock) return
+        if (!config.alertOwnCatches) return
+
+        if (event.seaCreature.getRarity() >= LorenzRarity.LEGENDARY) {
+            TitleUtils.sendTitle("§6LEGENDARY CATCH!", 3.seconds, 3.6)
+            if (config.playSound) SoundUtils.createSound("random.orb", 1.0f).playSound()
```
CalMWolfs: this should be `SoundUtils.playBeepSound()` and so should the one above

CalMWolfs (changes requested):
Also see my coment on the repo pr. Looks good though
One other thing you could consider is making the repo file into one of the json classes (see the jsonobjects folder) - I have explained this horribly so if you have a question just ask in contributor chat as hanni or I can prob explain better there but this also isn't necessary for you to change

hannibal002 (dismissed):
Thank you for adding the rarity of sea creatures to the repo and using this data in SkyHanni.
Some problems I have encountered, though:
It looks to me like you disregarded the custom rare tag we were using in the repo prior.
Before your pr: All sea creatures have the chat color of their type and are additionally bold if they are "rare".
Now with this pr: Nothing is bold anymore, their "type of origin" is totally ignored in chat messages, and it only uses their rarity color. This change is definitely not intended or wanted.

A feature that "highlights/Warns legendary sea creatures" should not trigger on the real rarity of the sea creature, but if the sea creature, itself is special/rare. For example currently, mobs like Thunder, Lord Jawbus, Carrot King and Deep Sea Protectors aren't legendary and yet are marked as rare. (I know that the current logic would still catch the mythic Lord Jawbus, but for that case the feature name "highlight legendary sea creatures" is not a correct match either.

Therefore, I suggest you still keep using the existing "rare" option instead of checking the rarity. We can still store the rarity in the field, but we should not use it for this feature toggle (we might decide to change the highlight color depending on the real rarity of a sea creature, making Lord Jawbus pink and Thunder purple, instead of all yellow/orange.

This was all about design, now some technical stuff:
The variable `toString` in the class `SeaCreature` uses the chat color variable, only to override it with the result of the rarity method. This is unlogical to me.
You could store the `LorenzRarity` data type directly in the field instead of parsing the string to the rarity on each access.
(Also maybe changing the loading of the sea creature repo file via java objects entirely (most likely out of scope for this pr)

brandonwamboldt:
Looks like i misunderstood some of the existing functionality, I'll get those things fixed right up

brandonwamboldt:
@hannibal002 I believe i have addressed all of your feedback

hannibal002 (approved):

PR #539, opened 2023-10-08
Fix double rendering of item tooltips with chat triggers

lineargraph:
The root cause here is
https://github.com/ChatTriggers/ChatTriggers/blob/master/src/main/kotlin/com/chattriggers/ctjs/launch/plugin/guiContainer.kt not separating their code into its own method, so the original injection point would find their popMatrix attribute as well as the vanilla one. I changed this to now use a different injection point which is not being used by chattriggers. The real fix would be to just move the chattriggers code into a hook method.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/mixins/transformers/gui/MixinGuiContainer.java
+++ b/src/main/java/at/hannibal2/skyhanni/mixins/transformers/gui/MixinGuiContainer.java
@@ -54,8 +54,9 @@ public abstract class MixinGuiContainer extends GuiScreen {
     @Inject(method = "drawScreen",
             at = @At(
                     value = "INVOKE",
-                    target = "Lnet/minecraft/client/renderer/GlStateManager;popMatrix()V",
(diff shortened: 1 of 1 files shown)


RayDeeUx:
will you PR to chattriggers too, or...

lineargraph:
> will you PR to chattriggers too, or...

doesn't look like they accept many PRs

hannibal002:
This is literally the oldest known SkyHanni bug: https://discord.com/channels/997079228510117908/1014624630541144144/1068119940818796545

hannibal002 (approved):

lineargraph:
> This is literally the oldest known SkyHanni bug: [discord.com/channels/997079228510117908/1014624630541144144/1068119940818796545](https://discord.com/channels/997079228510117908/1014624630541144144/1068119940818796545)

He, fun, i just assumed this was fixed by now until alea told me it wasn't this morning

RayDeeUx:
common chattriggers L

PR #540, opened 2023-10-09
Feature: Lobby halloween waypoints, per request of Tobbbb

RayDeeUx:
tried to keep the helper waypoint names as short as possible


src/main/java/at/hannibal2/skyhanni/features/misc/halloweenlobbywaypoints/BasketWaypoints.kt
```diff
@@ -0,0 +1,61 @@
...
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+class BasketWaypoints {
+    private val config get() = SkyHanniMod.feature.misc.halloweenBasket
+    private var waypoint: LorenzVec? = null
+    private var waypointName: String? = null
+
+    @SubscribeEvent
+    fun onRenderWorld(event: RenderWorldLastEvent) {
+        if (LorenzUtils.inSkyBlock || !LorenzUtils.onHypixel) return
```
hannibal002: How about checking the sidebar the same way we do for SkyBlock, just for the lobbies? This logic feels like a slippery slope. Without the checks below, this call would be valid everywhere on Hypixel except SkyBlock.

src/main/java/at/hannibal2/skyhanni/features/misc/halloweenlobbywaypoints/Basket.kt
```diff
@@ -0,0 +1,34 @@
+package at.hannibal2.skyhanni.features.misc.halloweenlobbywaypoints
+
+import at.hannibal2.skyhanni.utils.LorenzVec
+
+enum class Basket(
```
hannibal002, line 5: Generally, I would prefer the repo instead of enums like this. You could define the basketName with a function instead of using a static variable instead.
Since it's a one-time event, we can keep it outside repo if you prefer that.

src/main/java/at/hannibal2/skyhanni/features/misc/halloweenlobbywaypoints/BasketWaypoints.kt
```diff
@@ -0,0 +1,61 @@
...
+
+        if (LorenzUtils.skyBlockArea == "?") return
+
+        waypoint?.let {
+            event.drawWaypointFilled(it, LorenzColor.GOLD.toColor())
+            event.drawDynamicText(it, "§6" + waypointName!!, 1.5)
+        }
+    }
+
+    private fun chechScoreboardHalloweenSpecific(): Boolean {
```
hannibal002, line 58: Please don't make this "heavy function" (compared to a boolean access) in a render call. Instead, use a tick event every second or so and store a variable "isHalloweenScoreboard"

src/main/java/at/hannibal2/skyhanni/features/misc/halloweenlobbywaypoints/BasketWaypoints.kt
```diff
@@ -0,0 +1,61 @@
...
+        return (
+            ( ScoreboardData.sidebarLinesFormatted.any {
+                it.contains("Hypixel Level")
+            } ) && 
+            ( ScoreboardData.sidebarLinesFormatted.any {
+                it.contains("Halloween ")
+            } ) && 
+            ( ScoreboardData.sidebarLinesFormatted.any {
+                it.contains("Baskets ")
```
hannibal002: You can use `list.any { siA && isB && isC}` instead of 3 `any` blocks

hannibal002 (changes requested):
I have not tested the feature yet, but the overall code looks fine.

src/main/java/at/hannibal2/skyhanni/features/misc/halloweenlobbywaypoints/BasketWaypoints.kt
hannibal002: For the sake of consistency, I thought it would be cool to have a method like `LorenzUtils.inLobby()`. Since this is the first lobby feature, this would help write more lobby only features in the future.

src/main/java/at/hannibal2/skyhanni/features/misc/halloweenlobbywaypoints/Basket.kt
hannibal002: Fine.

src/main/java/at/hannibal2/skyhanni/features/misc/halloweenlobbywaypoints/BasketWaypoints.kt
RayDeeUx: something something "i walk so others can run" 😅

src/main/java/at/hannibal2/skyhanni/features/misc/halloweenlobbywaypoints/BasketWaypoints.kt
RayDeeUx: will do in a moment eating brunch rn

RayDeeUx: @hannibal002 
tried it, failed. it seems that this only works if we're checking bits and pieces of one specific line in the scoreboard

src/main/java/at/hannibal2/skyhanni/features/misc/halloweenlobbywaypoints/BasketWaypoints.kt
RayDeeUx: ok u have fun with that, the helper method i wrote is specifically for halloween also the most consistent general purpose solution might be locraw

hannibal002: > also the most consistent general purpose solution might be locraw

Why request additional data when Hypixel already provides us with the sidebar data :(

src/main/java/at/hannibal2/skyhanni/features/misc/halloweenlobbywaypoints/BasketWaypoints.kt
hannibal002: Yeah, my bad, this was wrong from me. My approach does obviously not work if those strings aren't all in the same line.

src/main/java/at/hannibal2/skyhanni/features/misc/halloweenlobbywaypoints/BasketWaypoints.kt
RayDeeUx: > > also the most consistent general purpose solution might be locraw
> 
> Why request additional data when Hypixel already provides us with the sidebar data :(

do i need to remind you of the `skiblock` incident

hannibal002 (approved):

PR #543, opened 2023-10-09
Added smoldering to estimated item value

jani270:


hannibal002:
While I agree with you here, how about making a dynamic check "if tier 1 book price is above x" in the future?
What would x be in your opinion?

PR #559, opened 2023-10-11
Fix: Enable translator and fix modifying links

NetheriteMiner:
I haven't found any other reported translator bugs, so it should be good to enable.


hannibal002:
I'm waiting with this PR until no more chat icon errors get reported. I don't want to have the guessing game what feature is breaking what part of the chat :D

hannibal002 (approved):

PR #555, opened 2023-10-10
Internal Changes: add support for compacted numbers parsing

Helfull:
This PR does add support for item amounts with compacted numbers

e.g.

`10k` `1m`

This also adds support for another amount prefix from the garden visitor rewards
e.g.
```
 §8+§319k §7Farming XP
 §8+§215 §7Garden Experience
 §8+§c21 Copper
 §8+§b10 Bits
```

Additionally i pulled log of items that did get pushed through this by visiting anita, the composter, and the garden plot configuration

```
§5Basket of Seeds
§5Hoe of Greatest Tilling
§6Gold medal
§9Delicate V
§9Nether Wart Pouch
§aJacob's Ticket
§aJacob's Ticket §8x20
§aJacob's Ticket §8x30
§aJacob's Ticket §8x32
§cBronze medal
§fSilver medal
§fSilver medal §8x2
  §81x §9Enchanted Sugar Cane
```

Tested on the places i could find of the usage from it

I am not 100% sure if i got all the places it gets used, if i missed something, let me know, i ll go ahead and add those test cases.

PR #561, opened 2023-10-12
Internal Changes: Refactor VisitorAPI to event based interaction

Helfull:
With this PR the `VisitorAPI` or `GardenVisitorFeature` is moved to an event based system.

There is one central "Listener" for the base events from MC/Forge and spawns the required Visitor events which then will be handled by the required features.

I had to also change the `RenderUtils` as i didnt want to expose the `RenderWorldLastEvent` through the `VisitorRenderEvent` and since it was not used anyway it was safe to be changed, this can be seen in this commit https://github.com/hannibal002/SkyHanni/pull/561/commits/b453b934b56efc214c874ae6d2f82f50d058564d

This PR should be merged/checked after this pr https://github.com/hannibal002/SkyHanni/pull/555


hannibal002:
I don't like having two independent changes in the same pr :/

image

Helfull:
> I don't like having two independent changes in the same pr :/
> 
> image

As i describe, it was this, or expose the `RenderWorldLastEvent`

I can pull those changes out into its own PR, which would then be blocking this one

hannibal002:
I don't see the harm in doing so.
See https://github.com/hannibal002/SkyHanni/blob/39adca7678bb675e6917b35c1de699c6a0bd55dc/src/main/java/at/hannibal2/skyhanni/features/bazaar/BazaarApi.kt#L62

hannibal002:
Or a third approach would be to not expose the render event directly but provide the drawString method inside the visitor event that then calls the render event internally.

hannibal002:
Any way, I don't like the idea of being able to run render calls outside their intended scope. This is a part of bug prevention framework design in my mind.

Helfull:
i will expose the event, over the third option, which i was thinking about as well, but it felt wrong.

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/VisitorTooltipParser.kt
Helfull, line 1: This one was supposed to be used in the `readTooltip` method to make it easier to read.

But i threw most of it away, after i started working on the event based refactor. I will add those changes back in the next change set after the events are done.

Helfull:
Converting to draft, as the last changes on beta broke a lot here with conflicts. Will add them

Helfull:
I dont know why these tests fail now, need to check that later

Helfull:
Okay, the cause for this is the definition of `fun SkyBlockTime.formatted(): String` this seems not to be resolvable with mockk

Helfull:
@hannibal002 maybe we should stop declaring extensions inside other classes, if they dont need any internals of that class to work.

This would remove the size of those classes, in additions we could create a `utils.extensions` package which contains all of those extensions which are created, this way it is easier to identify if a function is a member or an extension we created.

hannibal002 (approved):

PR #567, opened 2023-10-13
Fix: Chat symbols

CalMWolfs:
should fix the issues with the chat symbol.
Chat symbol no longer works on some if not all private messages (hypixel splits up the rank prefix on the +)

this works in dev env but not actual game now? (soopy stat beside name bug)


hannibal002 (changes requested):
Idk what i did wrong while testing but the feature was no longer working at all for me.

hannibal002 (approved):

PR #575, opened 2023-10-14
HOTFIX: single test file hotfix

RayDeeUx:
bring back both tests under one file to prevent double block registration


RayDeeUx:
helfull fixed on their end, closing

PR #579, opened 2023-10-15
Fixed highlight showcase items not working in some areas

Obsidianninja11:


src/main/java/at/hannibal2/skyhanni/features/misc/items/GlowingDroppedItems.kt
```diff
@@ -48,16 +55,18 @@ class GlowingDroppedItems {
         return rarity?.color?.toColor()?.rgb
     }
 
-    private fun isShowcaseArea() =
+    private fun isShowcaseArea() = RecalculatingValue(1.seconds) {
```
hannibal002: This should be a member object, not a function. Moreover, the name is no longer optional. See https://github.com/hannibal002/SkyHanni/blob/24c8ce90e3925a397975b617eedf796ae52208ae/src/main/java/at/hannibal2/skyhanni/features/misc/TimeFeatures.kt#L26 as a comparison.

hannibal002: Also, can we first check the enums and only the strings after they don't have a match? Also, could create a list of the islands, but this is not needed.

hannibal002 (changes requested):

hannibal002 (approved):

PR #583, opened 2023-10-16
Backend: Disable Diana when not Diana

CalMWolfs:
Stops some stuff from happening when its not Diana which should slightly increase performance
Adds force Diana option in case mayor detection fails

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/event/diana/GriffinBurrowParticleFinder.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/event/diana/GriffinBurrowParticleFinder.kt
@@ -1,7 +1,6 @@
 package at.hannibal2.skyhanni.features.event.diana
 
 import at.hannibal2.skyhanni.SkyHanniMod
-import at.hannibal2.skyhanni.data.IslandType
 import at.hannibal2.skyhanni.events.BlockClickEvent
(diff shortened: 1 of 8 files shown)


src/main/java/at/hannibal2/skyhanni/features/event/diana/GriffinBurrowParticleFinder.kt
```diff
@@ -121,21 +116,17 @@ class GriffinBurrowParticleFinder {
...
     @SubscribeEvent
     fun onBlockClick(event: BlockClickEvent) {
-        if (!LorenzUtils.inSkyBlock) return
+        if (!DianaAPI.featuresEnabled()) return
         if (!config.burrowsSoopyGuess) return
-        if (LorenzUtils.skyBlockIsland != IslandType.HUB) return
 
         val pos = event.position
-        if (event.itemInHand?.isSpade != true || pos.getBlockAt() !== Blocks.grass) return
+        if (!DianaAPI.hasSpadeInHand() || pos.getBlockAt() !== Blocks.grass) return
```
hannibal002: No, please don't do this. this can cause problems.

hannibal002 (changes requested):
Please revert the spade checks. Everything else looks fine.

hannibal002 (approved):

PR #591, opened 2023-10-16
Update: Option to disable chest value in dungeons

superhize:
Someone asked


walkerselby:
Maybe I am weird, but I feel like the option should be to enable it, rather than an option to disable it.

hannibal002 (approved):

PR #595, opened 2023-10-17
Internal: Ban imports

lineargraph:
By royal decree of Sir Hannibal the second, it is now forbidden to import some packages from others. The provided config is just an example, but should be easy enough to understand and adapt.

### DIFF
--- a/.github/workflows/ban-imports.yaml
+++ b/.github/workflows/ban-imports.yaml
@@ -1,0 +1,23 @@
+name: Check banned imports
+on:
+  - pull_request
+  - push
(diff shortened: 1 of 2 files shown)


hannibal002 (changes requested):
the check failed again (even after merging beta into banimports)

lineargraph:
Yes, the check is currently failing because of an example import ban that i put in. obviously we wouldnt want to keep that one, it is purely for demonstration.

CalMWolfs (approved):

hannibal002 (approved):

PR #597, opened 2023-10-17
Internal Change: Replace void with unit

walkerselby:
Void is not equivalent to Java void but java.lang.Void.


lineargraph:
In java when using generics you still use java.lang.Void (or a nullable Void, since non null Void is a bottom type, not a unit type). But yes, since Kotlin has it's own unit type that can be used as a generic, Unit is more appropriate.

PR #606, opened 2023-10-19
Internal Change: Chat filter cleanup

walkerselby:
Refactor some if to whens
Change consecutive spaces to regex values for improved readability
Extracted others to their own function to prevent multiple checks
Flip check orders in block
Config check is faster than message checks. Should have a minimal speed improvement as messages are now only compared if the config value is enabled.


walkerselby:
Most recent revision: 

Rewrote a lot. Removed separate functions, changed how system was accessed. 

This should also make it easier to migrate to the REPO eventually.

hannibal002 (approved):

PR #612, opened 2023-10-22
Change: Custom Keybinds for Harp Helper

Thunderblade73:


src/main/java/at/hannibal2/skyhanni/features/inventory/HarpFeatures.kt
```diff
@@ -39,12 +55,12 @@ class HarpFeatures {
...
         val chest = event.gui as? GuiChest ?: return
 
-        for (key in keys) {
+        for ((index, key) in keys.withIndex()) {
             if (key.isKeyHeld()) {
                 if (lastClick.passedSince() > 200.milliseconds) {
                     Minecraft.getMinecraft().playerController.windowClick(
                         chest.inventorySlots.windowId,
-                        35 + key,
+                        37 + index,
```
Rohan-Sangani, line 63: Why 37 and not 35?

src/main/java/at/hannibal2/skyhanni/config/features/InventoryConfig.java
```diff
@@ -311,6 +313,42 @@ public static class HarpConfig {
...
+                @ConfigOption(name = "Key 5", desc = "Key for the fifth Node")
+                @ConfigEditorKeybind(defaultKey = Keyboard.KEY_5)
+                public int key5 = Keyboard.KEY_5;
+                @Expose
+                @ConfigOption(name = "Key 6", desc = "Key for the sixth Node")
+                @ConfigEditorKeybind(defaultKey = Keyboard.KEY_6)
+                public int key6 = Keyboard.KEY_6;
+                @Expose
+                @ConfigOption(name = "Key 7", desc = "Key for the seventh Node")
+                @ConfigEditorKeybind(defaultKey = Keyboard.KEY_1)
```
Rohan-Sangani: Default should be 7, not 1

src/main/java/at/hannibal2/skyhanni/features/inventory/HarpFeatures.kt
Thunderblade73: Because you had it with Key1, that has value 1 and Index start with 0. So in total +2

Thunderblade73: Wait I think this should only be +1 not +2. But it somehow worked. Need to test this tomorrow

Thunderblade73: O found why it is +2. Because Key1 starts with 2 and not 1
image

hannibal002 (approved):

PR #614, opened 2023-10-23
Internal Change: Composter overlay NEUInternalName

walkerselby:
Update to use NEUInternalName rather than String
Some code cleanup


walkerselby:
@Alea1337 your review has been requested

hannibal002 (changes requested):
I think we can set `composterCurrentFuelItem` and `composterCurrentOrganicMatterItem` in the storage to be of type `NEUInternalName`

hannibal002 (approved):

PR #621, opened 2023-10-24
Add ping for incoming visitors, and accept hotkey

lineargraph:
This patch adds a ping that plays every second for the last 10 seconds before a visitor spawns, and also adds a hotkey to accept a visitor.


lineargraph:
(this is obviously for spooky thingies) ephemeral dingsibumsi

hannibal002 (approved):

PR #639, opened 2023-10-28
Feature: Minion Xp Calculator

Thunderblade73:
Shows the xp of an item in its Tooltip and the total in the collect all item as Tooltip + works with Storage


Thunderblade73:
Please provide some feedback on the formatting:

image
image
image

src/main/java/at/hannibal2/skyhanni/features/minion/MinionXp.kt
```diff
@@ -0,0 +1,404 @@
...
+    companion object {
+        private val xpItemMap: MutableMap<PrimitiveItemStack, String> = mutableMapOf()
+        private val collectItemXpList: MutableList<String> = mutableListOf()
+        fun onMinionClose() {
+            xpItemMap.clear()
+            collectItemXpList.clear()
+        }
+    }
+
+    private val xpInfoMap: HashMap<NEUInternalName, XpInfo> = hashMapOf(
```
hannibal002: Please move this into the repo

walkerselby:
This is all my personal preference. 

For all: 
I don't think the colon is necessary, you can remove it. 
I would capitalize XP to match Hypixel capitalizing XP. 
I would change the delimator for decimal values to be a period rather than a comma, as that is more common for the American format and matches Hypixel's number formatting. 
I would prettyprint this the numbers with commas separating 123,456,789

For the first image, I would put the text lower. I would skip a line under rare and then put it there. 

Second image looks fine to me. 

Third image, I would add an empty line between collect to get and the No Minion Storage data. I would also preprend that as "Error: No Minion Storage Data". 


Also, maybe just me, I feel like there is a lot going on with colors. But not exactly sure what I would prefer at the moment. Only think I think in general, though, would be to change that error message to light red, to match the rest of skyhanni, and also change the fix suggestion to light yellow, also to match the rest of skyhanni.

hannibal002:
> I would change the delimator for decimal values to be a period rather than a comma, as that is more common for the American format and matches Hypixel's number formatting.

Just use Number.addSeparators()

hannibal002 (approved):

PR #656, opened 2023-11-01
Feature: Only warn for certain crops for Jacobs contest

Obsidianninja11:


src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
```diff
@@ -315,6 +315,7 @@ object GardenNextJacobContest {
     private fun warn(timeInMillis: Long, crops: List<CropType>) {
         if (!config.warn) return
         if (config.warnTime <= timeInMillis / 1000) return
+        if (!warnForCrop()) return
```
Apo-th, line 364: Why are you checking for negative here? Why don't you just return the opposite in warnForCrop() ?

Obsidianninja11: That would be kind of confusing
also this sort of thing is used for isenabled functions

hannibal002:
ja, lass mal machen

hannibal002 (approved):

PR #660, opened 2023-11-01
Feature: The Great Spook Display Utilities and Notif Sound

RayDeeUx:
- Primal Fear Timer
- Fear Stat
- IRL Time Left
- (+ notif sound by thunderblade73)


RayDeeUx:
yes all enabled by default they just have to `/sh gui` and move everything around

jani270:
> yes all enabled by default they just have to `/sh gui` and move everything around

Bad.

walkerselby:
> yes all enabled by default they just have to `/sh gui` and move everything around

Can you set them to be disabled by default?

hannibal002 (approved):

PR #673, opened 2023-11-02
add expired pumpkin to farming fortune

CalMWolfs:
no chat message or whatever because im not gonna play enough to get the stuff for it and it'll be in a beta for only 3 days until the event is over


puff:
I think you will still be able to use them after the event is over (since they are in bazaar) so maybe this should be added?
SkyHanni's chat log shows this when I gave all pumpkins to Carrolyn:
`[allowed]  §r§a§lPUMPKINS EXPORTATION COMPLETE!`

walkerselby:
Good point, especially as they are on the bazaar

hannibal002 (approved):

PR #675, opened 2023-11-02
added toggle to hide autopet messages

CalMWolfs:
Stopped auto-hiding the autopet messages as this confused people who toggled the feature without realising that it hides the messages
Also moved the pets position config into the pets config


hannibal002 (approved):
Technically hideAutopet will also hide "you summoned/despawned chat messages, but this should not matter.

PR #684, opened 2023-11-04
Fixed scoreboard date number suffix missing

RayDeeUx:
also inspired me to add a second toggle to include the date ordinal because why not.

proof it actually happened before:

{i know it's a screenshot of a `.txt` file but you gotta believe me}


src/main/java/at/hannibal2/skyhanni/config/features/GUIConfig.java
```diff
@@ -133,6 +133,15 @@ public static class InGameDateConfig {
         @FeatureToggle
         public boolean includeSunMoon = true;
 
+        @Expose
+        @ConfigOption(
+            name = "Show Date Ordinal",
+            desc = "Show the date's ordinal suffix. Ex: (1st <-> 1, 22nd <-> 22, 23rd <-> 3, 24th <-> 24, etc.)"
+        )
+        @ConfigEditorBoolean
+        @FeatureToggle
```
hannibal002: This is no standalone feature, hence no feature toggle.

hannibal002 (changes requested):
see above

src/main/java/at/hannibal2/skyhanni/config/features/GUIConfig.java
RayDeeUx: > This is no standalone feature, hence no feature toggle.
pardon, pls elaborate

hannibal002 (approved):

PR #468, opened 2023-09-11
Feature: Item Stack Size

RayDeeUx:
Added:
- Abbreviations of Skyblock player stats
- Skyblock level
- Skill average
- collection % unlocked display
- collection maxed out % display
- recipes unlocked %
- bank tier
- dojo grades
- highest crafted minion tier
- wardrobe slot #s
- progress to next minion slot
- SO MUCH MORE

PR #694, opened 2023-11-05
Feature: Item Stack Size (Restores #468)

RayDeeUx:
turns out renaming branches has consequences.

anyways, todo:

~~convert config lists to enums~~
~~move pattern declarations to individual values~~
hanni wants me to split each category into an individual PR gg gaming
some caching stuff hanni wanted to do


RayDeeUx:
good news: all the configs are enum draggable lists.
bad news: we (ie someone more experienced than me) might need to write a configmigrator for the original set of inventory stack size numbers.

hannibal002:
Is there a chance we cannot use `removeColor` and `contains` and use regex and equals instead? There might be unexpected side effects from doing this not as precise as possible.

RayDeeUx:
# CLOSED. CHECK PRS #759, #760, #761, #762, #763, #764, #765, #766, #767, #768.

PR #712, opened 2023-11-16
Backend: Mob Detection

Thunderblade73:
# What
Introduces a system that can detect and identify every mob in Skyblock. It is a sub-PR from #526

The system wraps each entity (assuming valid) in a Mob class and gives it a name which is the same as the nametag of the mob (if it has one) and is formatted for better use. 
Example: "§8[§7Lv90§8] §cKada Knight§r §a800k§f/§a800k§c❤§r" the name would then be "Kada Knight"
There are two ways to interact with the system:
* By Event
  There a the Spawn and Despawn Event that give back the mob. If you know that you only need a mob from a specific group use the sub event instead. The groups are declared with the Mob.Type Enum. Most of the time the SkyblockMob one will be used.
  At the point of the event, the mob has every data that could be needed.
  Note: when an spawn happens a despawn is guaranteed (even when changing islands)
* By set
  The sets hold all mobs of their type, The sets also hold a entityList which gives every entity associated with the group back.
  When there is a need to find the association of an entity to it's mob the MobData.entityToMob map should be used,
  
  The Sets are:
  * MobData.currentMobs  (every Mob regardless of Type)
  * MobData.players
  * MobData.displayNPCs
  * MobData.skyblockMobs
  * MobData.summoningMobs
 
MobData.entityToMob
# Why
We had some sort of mob detection with the MobFinder and in MobHighlight, but there approach was find a specific Mob and do some stuff. The mob detection turns every entity into a mob and abstracts it into a (somewhat) easy-to-understand structure. With the old approach, you would have to write a specific detection for each mob needed, but with the mob detection, you only need to know the name of the mobs needed for the feature and can get every instance of it (by event or by set). This reduces unexpected behavior.
# How
* The simple versions:
  * For spawn:
  Detect entity spawn -> wait until in range (22.8 Blocks) -> get armorstand -> extract name and format it -> call event
  * For despawn:
  Detect entity despawn -> find the matching mob for the entity -> call event
* The long version:

For detection of spawn/despawn the Forge events could have been used, but since we try to avoid Forge stuff, I made it my way. I used a simple algorithm with sets to detect it. I save the entities from the previous tick and compare them with the current entities. 
Note: It could also use the packets directly but since Skyblock sends packets in a weird order this task is quite difficult to do.
``` Kotlin
(currentEntityLiving - previousEntityLiving) //Spawn
(previousEntityLiving - currentEntityLiving) //Despawn
```

Then the entity gets categorized into three main groups (RealPlayer, DisplayNPC and SkyblockMob) by filter methods that can be applied to the entity alone without the need for the armorstand to be loaded. Depending on what group the entity lies, the further steps will be different
* RealPlayer
Wait one loop (because I don't care) and then it is done at this point, and the event is called. Definitely the easiest.
* DisplayNPC
Wait one loop -> check if in range (22.8 Blocks) (not correct for all DisplayNPCs but good enough) -> if not in range, loop again, but if in range, the armorstand is grabbed and the name is saved -> then the event is called
* SkyblockMob (the main thing)
Does the wait and range check as the DisplayNPC (but the range is 99.998% correct for it). After that, the crazy stuff happens.
Note: any exception returns the mob directly and skips further filtering.
First check if the mob is listed as noArmorStandMobs exception
Get the next Mob (ID + 1) because Skyblock Spawns it mobs in order. For example, for basic mobs, the armorstands directly spawned after the base entity (server-side). Therefore the relation of the two things is there ID where the difference is only 1.
Check for the main exceptions (every mob that doesn't follow the rules).
After that, looks if the next mob is already present in the maps, and if so, it "late stacks" itself to the associated mob.
Then the stacking happens. It takes every next mob (+2, +3, …) that is not a null, a valid armorstand, or an already present entity in the maps. 
After exiting this looping, that stacked mob get parsed to the stackedMobsException which handles wrong stacks. Then it will check the last entity that was found. If it was null, the base entity will be put back into the loop cycle. If it was an already present entity, it does a "late stack" for the whole list. If the armorstand was valid, the regex fitting begins.
First, it tries to fit the summoningRegex, then the SlayerNameFilter, then the bossMobNameFilter, and then mobNameFilter.
If each of those cases, a specfic SkyblockMob will be created. and for mobNameFilter, a SkyblockBasicMob or DungeonMob will be created (depending on if in a dungeon or not).

Then the event is called.

Since I used the term "late stack" earlier, I will explain it now, what that means. Because Skyblock does have some mobs that are made up out of multiple entities, especially in newer islands. Therefore, I implemented a way to stack up mobs. In the normal case, it is quite easy just get every follow-up entity until the armorstand is hit. But since entities don't have to be all loaded currently, there can be some entities missed from the stack (but only with smaller ID as the first entity). Therefore, I made the "late stack" mechanism that adds entities to an already present mob,

For the despawn only thing that is special is that when the player disconnects from Hypixel the mobs won't be cleared correctly that is why once the player joins back in all mobs will be cleared.

Import to know is that the entities inside the mob can change at any time. Because of "Late Stack" and because Hypixel spawns some mobs multiple times which breaks the reference of the entity inside the Minecraft Code.

# Notes
This is only the system; there are no changes yet to any other code. I want to separate the system from the integration because this is already a big PR, and more code in random places wouldn't make it easier to review. Also, please feel free to migrate stuff to this system
Also, every kind of feedback is welcome (now I can makes changes quite easily, but later on, it will be painful to do)

+ Added Mob Detection - Thunderblade73


Thunderblade73:
Current Class Structure (for the result of the Filtering and the event input)
UMI2
The Colors denotes:
Cyan : the 4 different event Types (the 4th is just an EntityPlayer)
Red : Only internal logic
Grey : an outside class

Thunderblade73:
Finally done with the rewrite. I would love to get some feedback (not only from hannibal)
The description is updated to match the changes (only a few parts changed)
Changes:
It is now Enum based instead of Class based. It works in Dungeon 100% instead of 50% of the time. 
And only a handful of mobs are missing in the exception logic (2 bats are missing and a fix. And testing of F6, F7 & M7 is still missing + Kuudra (I always forget it exists))

walkerselby:
I didn't go fully through everything but went through a fair bit. Can't wait to see this being used. Also, idea: in data, feel free to make a package for mob

src/main/java/at/hannibal2/skyhanni/data/mob/MobFactories.kt
```diff
@@ -0,0 +1,133 @@
...
+        Mob(baseEntity = baseEntity, mobType = Mob.Type.Special, armorStand = armorStand, name = name)
+
+    private fun String.removeCorruptedSuffix(case: Boolean) = if (case) this.dropLast(1) else this
+    fun dojo(baseEntity: EntityLivingBase, armorStand: EntityArmorStand): Mob? =
+        MobFilter.dojoFilter.matchMatcher(armorStand.cleanName()) {
+            Mob(
+                baseEntity = baseEntity,
+                mobType = Mob.Type.Special,
+                armorStand = armorStand,
+                name = if (this.group(1).isNotEmpty()) "Points: " + this.group(1) else this.group(2)
```
hannibal002: Use named regex labels

hannibal002 (approved):
worked fine for me

src/main/java/at/hannibal2/skyhanni/data/mob/MobFactories.kt
Thunderblade73: give me the names and I add them, since the pattern is to general to give names.

Thunderblade73: done, but not tested

hannibal002 (approved):
Works still fine and the changes since last approve are also fine. will merge

PR #737, opened 2023-11-28
Add unique gift highlighter

lineargraph:
Highlights players that have not been gifted before. Makes use of the envelope icon if wanted.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
+++ b/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
@@ -1,6 +1,7 @@
 package at.hannibal2.skyhanni
 
 import at.hannibal2.skyhanni.api.CollectionAPI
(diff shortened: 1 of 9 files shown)


src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
```diff
@@ -432,6 +435,7 @@ class SkyHanniMod {
         loadModule(DungeonCleanEnd())
         loadModule(DungeonBossMessages())
         loadModule(DungeonBossHideDamageSplash())
+        loadModule(SkyhanniPlayersWhenTheySeeUnsuckedCock)
```
lunaynx: 💀

lineargraph: Oh right i forgot about that

hannibal002 (approved):

PR #739, opened 2023-11-29
Internal Change: Cleanup CaptureFarmingGear

walkerselby:


walkerselby:
This PR is now stale, being 4 months old. Will revisit and re-implement when I have time.

PR #740, opened 2023-11-29
Feature Change: Unique Gift Highlighter exluded Ironman

Thunderblade73:
If you are not on ironman other players that are on ironman won't be highlighted. Also did the same logic for Bingoprofiles.


hannibal002 (changes requested):
Please replace your `TabListUpdateEvent` approach with reading the name above the player object.

hannibal002 (approved):

PR #746, opened 2023-11-30
Backend: DelayedRun sync

Thunderblade73:
Found a better sync bug fix.


src/main/java/at/hannibal2/skyhanni/utils/DelayedRun.kt
```diff
@@ -1,34 +1,24 @@
...
 
-import at.hannibal2.skyhanni.utils.LorenzUtils.editCopy
+import java.util.concurrent.LinkedBlockingQueue
 import kotlin.time.Duration
 
-// TODO find better sync bug fix than creating a new map for each use
 object DelayedRun {
-    var map = mapOf<() -> Any, SimpleTimeMark>()
+    val map = mutableListOf<Pair<() -> Any, SimpleTimeMark>>()
+    private val inMap = LinkedBlockingQueue<Pair<() -> Any, SimpleTimeMark>>()
```
lineargraph: You should not be using a `LinkedBlockingQueue` here. We don't want any blocking behaviour. Use  `ConcurrentLinkedQueue` instead.

src/main/java/at/hannibal2/skyhanni/utils/DelayedRun.kt
```diff
@@ -1,34 +1,24 @@
 package at.hannibal2.skyhanni.utils
 
-import at.hannibal2.skyhanni.utils.LorenzUtils.editCopy
+import java.util.concurrent.LinkedBlockingQueue
 import kotlin.time.Duration
 
-// TODO find better sync bug fix than creating a new map for each use
 object DelayedRun {
-    var map = mapOf<() -> Any, SimpleTimeMark>()
+    val map = mutableListOf<Pair<() -> Any, SimpleTimeMark>>()
```
lineargraph: The naming of both of these variables could use some improvement: Maybe `tasks` and `delayedTasks`.

src/main/java/at/hannibal2/skyhanni/utils/DelayedRun.kt
```diff
@@ -1,34 +1,24 @@
...
-            entries.removeIf { (runnable, time) ->
-                val inPast = time.isInPast()
-                if (inPast) {
-                    runnable()
-                }
-                inPast
-            }
+        inMap.drainTo(map)
+        map.removeIf { (runnable, time) ->
+            time.isInPast().also { if (it) runnable() }
```
lineargraph: I don't think the `also` one line style here serves to improve readability.

lineargraph (changes requested):

lineargraph (approved):

lineargraph (approved):
Just had a thought about using a TreeMap instead :upside_down_face:, but no, this looks great.

src/main/java/at/hannibal2/skyhanni/utils/DelayedRun.kt
```diff
@@ -1,34 +1,29 @@
...
-                }
-                inPast
+        tasks.removeIf { (runnable, time) ->
+            val inPast = time.isInPast()
+            if (inPast) {
+                runnable()
             }
+            inPast
         }
+        futureTasks.drainTo(tasks)
```
hannibal002, line 28: I think drainTo should happen before removeIf

src/main/java/at/hannibal2/skyhanni/utils/DelayedRun.kt
```diff
@@ -1,34 +1,29 @@
...
+    private val futureTasks = ConcurrentLinkedQueue<Pair<() -> Any, SimpleTimeMark>>()
 
     fun runDelayed(duration: Duration, run: () -> Unit) {
-        map = map.editCopy {
-            this[run] = SimpleTimeMark.now() + duration
-        }
+        futureTasks.add(Pair(run, SimpleTimeMark.now() + duration))
     }
 
     fun runNextTick(run: () -> Unit) {
```
hannibal002, line 16: The current logic feels like it is "run in 2 ticks from now", right?

hannibal002 (changes requested):
The logic looks fancy, thanks <3
Not sure if im missing something but one small thing here looks misleading or wrong to me

Thunderblade73:
Since both suggestion are for one and the same issue:
It's 1 tick must happend but up to 1 extra tick could have happend.
I wanted to have this behavior so it is guaranteed that one tick has parsed, without that we would always execute the runNextTick in the subTick before the next Tick which is misleading. 
And also some stupid stuff could happen without that:
Say you run a runNextTick in a LorenzTickEvent. Then it could be possible that the code executes immediately after that event Listener finishes since the next Listener could be the checkRuns one and therefore no time has parsed for it.

hannibal002:
The idea behind "run in next tick" for me was not a "provide at least 50ms delay" but rather "run in the next tick in the main thread".
But your logic is better.
Now the function executes in between 50 and 100 ms, instead of 1 and 50 ms. this feels more consistent.
And the extra 50ms will hurt no one.

Now only one question remains to me: should we rename the function runNextTick to something else?

Thunderblade73:
The name is fine. Also documented the behavior.

hannibal002 (approved):

PR #755, opened 2023-12-02
Cleanup ItemStack Size

walkerselby:


walkerselby:
This PR is now stale. Will revisit and re-implement when I have time.

PR #792, opened 2023-12-11
Fix typo with Not Clickable Items in Composter

Absterge:
Changed "Only sell" to "Only insert" for items in the Composter. I was torn between using "insert" or "add" (Hypixel uses both in the Composter) so feel free to change.


hannibal002 (approved):

hannibal002:
Insert is totally fine.

PR #796, opened 2023-12-13
Fix: Support Badlion sendcoords format and only show inquisitor waypoints in the hub

brandonwamboldt:
Seen a lot of false reports from Skyhanni users this week when ppl share coords in Crystal Hollows. Also adding support for Badlion coords of the format `X: 123.45 Y: 456.78 Z: 100.00`


hannibal002:
Is there a regex character to use instead of the parameter CASE_INSENSITIVE?
We want to migrate all regex checks to repo eventually and there we can't use additional parameters as well

lunaynx:
`(?i)` probably, assuming that works in Kotlin.

brandonwamboldt:
Hmm i can try (?i) but would we want the regex case sensitive? I can also just change it to [xX] etc

brandonwamboldt:
@hannibal002 just gonna use [xX] for simplicity

hannibal002 (approved):

PR #812, opened 2023-12-17
Fix next visitor time while farming

lunaynx:
The speed factor should be exactly 3, trying to determine it dynamically causes inaccuracies due to lag / rounding.


hannibal002 (commented):
This will cause inconsistency issues if someone breaks less than 20 blocks per second, doesn't it?

lunaynx:
Yeah technically it will make the second timer go slower, but nobody will really farm less than 20 BPS mid lane... well, except for cactus. I suppose we could use BPS for it, but it should use momentary BPS or last 5 seconds or something, as the whole session BPS that is displayed can be too low after short pauses.

hannibal002 (approved):
It is not perfect, but its more consistent than the current logic.

PR #811, opened 2023-12-17
Backend: ItemCategory + better LorenzeRarity

Thunderblade73:
Categorizes every item using the rarity lore line. Also a better way of getting the item rarity
Rarity Lore Line Example (Rarity: Rare, Category: Sword):
image

PR #816, opened 2023-12-17
Feature: Reforge helper

Thunderblade73:
## Dependencies
- #822
- #821
- #811
-  #746 
- #846 
- #885 
- #890 
- #1191
- #888 
# What
A helper for random Reforges similar to the one sba has but better.
Blocks the Reforge Button when the desired reforge is on the item.
Shows all possible Reforges, with there stats and the difference it would make by chancing to them
and has support for reforge stones in Hex
# Preview
image

## Changelog New Features
+ Reforge Helper. - Thunderblade73
    * Adds reforge locking similar to sba
    * Works with Hex
    * Gives a list of all possible reforge for the item, with stats
    * The reforge list can be sorted for a specific stat
+ Added SkyblockStats - Thunderblade73
+ Added ReforgeAPI - Thunderblade73


Thunderblade73:
Dependencies are:
- [ ] #829
- [ ] #822
- [ ] #821
- [ ] #811
- [ ] #746

hannibal002:
Does this only work in hex, or also in reforge anvil?

Thunderblade73:
it works in hex, at the blacksmith, at the dungeon reforge thing and in the reforge anvil in privat island if on random reforge. 
In the reforge anvil itself it doesn't work, because I do not see any reason why it should work there, since it is only used to combine stuff (and you need the stuff before)

Thunderblade73:
continued in #1437

PR #821, opened 2023-12-20
Improvement + Fix: HoverTips improvements

Thunderblade73:
## Dependencies
- #885
## What
Change the hoverTips to work with renderable and removed unnecessary code

Also added rules how the hovertip should handle out of screen behavior, so that it can't go off screen anymore.

## Changelog Improvements
+ Single line Hover messages are now closer to the cursor - Thunderblade73

## Changelog Fixes
+ Hover messages can't go off-screen anymore - Thunderblade73

+ Renderable.hoverTips now supports renderables as tips and content - Thunderblade73


CalMWolfs (approved):
Works fine for me and code cleanup looks good. Something that would be cool but is also probably out of the scope of this pr is moving the tooltip if it is rendering off screen e.g. if it goes off right side of screen shift it left a bit. Or move up/down if at top or bottom of screen. But this is optional and should be a part of a different pr

Thunderblade73:
The left/right clamping I will do in this Pr. The up down is as Minecraft default down, but can try to add an option to clamp to top. Also would make them scrollable sometime but that is out of scop of the PR.

hannibal002 (approved):

PR #826, opened 2023-12-22
Internal: Refactor Classes to Appropriate Packages

walkerselby:
Cleaned up mostly the data package


walkerselby:
This PR is now stale. Will revisit and re-implement when I have time.

PR #832, opened 2023-12-25
Add ability to get unique visitors served without Green Thumb

lunaynx:
This won't automatically update when you serve a new unique visitor, but I think it didn't do that with Green Thumb either unless you manually opened `/eq`, so this is a good start to be able to at least suggest Green Thumb to people who don't have it at all yet.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/fortuneguide/CaptureFarmingGear.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/fortuneguide/CaptureFarmingGear.kt
@@ -37,6 +37,9 @@ class CaptureFarmingGear {
 
     private val cakePattern = "(?:Big )?Yum! You (?:gain|refresh) [+]5☘ Farming Fortune for 48 hours!".toPattern()
 
+    private val tierPattern = "§7Progress to Tier (?<nextTier>\\d+):.*".toPattern()
+    private val tierProgressPattern = ".* §e(?<having>.*)§6/(?<total>.*)".toPattern()
+
     companion object {
         private val strengthPattern = " Strength: §r§c❁(?<strength>.*)".toPattern()
         private val farmingSets = arrayListOf(
@@ -89,6 +92,16 @@ class CaptureFarmingGear {
             val storage = GardenAPI.storage?.fortune ?: return
             storage.pumpkinFortune = !storage.pumpkinFortune
             LorenzUtils.chat("Toggled expired pumpkin fortune to: ${storage.pumpkinFortune}")
+        }
+
+        private fun getUniqueVisitorsForTier(tier: Int, progress: Int): Int {
+            return when {
+                tier == 0 -> 0
+                tier == 1 -> 1
+                tier == 2 -> 5
+                tier >= 3 -> 10 * (tier - 2)
+                else -> throw IllegalStateException("Unexpected unique visitors tier: $tier")
+            } + progress
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/fortuneguide/CaptureFarmingGear.kt
```diff
@@ -37,6 +37,9 @@ class CaptureFarmingGear {
 
     private val cakePattern = "(?:Big )?Yum! You (?:gain|refresh) [+]5☘ Farming Fortune for 48 hours!".toPattern()
 
+    private val tierPattern = "§7Progress to Tier (?<nextTier>\\d+):.*".toPattern()
+    private val tierProgressPattern = ".* §e(?<having>.*)§6/(?<total>.*)".toPattern()
```
hannibal002: Please use repo patterns

src/main/java/at/hannibal2/skyhanni/features/garden/fortuneguide/CaptureFarmingGear.kt
```diff
@@ -90,6 +93,16 @@ class CaptureFarmingGear {
             storage.pumpkinFortune = !storage.pumpkinFortune
             LorenzUtils.chat("Toggled expired pumpkin fortune to: ${storage.pumpkinFortune}")
         }
+
+        private fun getUniqueVisitorsForTier(tier: Int, progress: Int): Int {
```
hannibal002: I think its more clear if we remove progress from this function and add it at the function call instead. The function name is misleading otherwise

src/main/java/at/hannibal2/skyhanni/features/garden/fortuneguide/CaptureFarmingGear.kt
```diff
@@ -214,6 +227,25 @@ class CaptureFarmingGear {
                 storage.anitaUpgrade = level
             }
         }
+        if (event.inventoryName.contains("Visitor Milestones")) {
```
hannibal002, line 231: Please use early reuturn over if

src/main/java/at/hannibal2/skyhanni/features/garden/fortuneguide/CaptureFarmingGear.kt
```diff
@@ -214,6 +227,25 @@ class CaptureFarmingGear {
...
+            for ((_, item) in event.inventoryItems) {
+                if (item.displayName == "§aUnique Visitors Served") {
```
hannibal002: Please use early continue over if

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/fortuneguide/CaptureFarmingGear.kt
lunaynx: Why? There are a bunch of other similar `if`s in this function, it isn't the only one. It would look weird to only early return for the last one.

hannibal002: Same goes for all other if checks that have a longer block inside. Multiple additions control flows just increase the intendation size unnecessary, in my mind

lunaynx: You can't exactly do that when there are multiple inventory names being tested. If you early return from one, it won't check the others.

hannibal002: I looked over the whole function and you are right, one single early return feels wrong here, especially since it's not "early". Another idea I just had was to replace the chained ifs with one select block and extract each case into a function. Would increase the class size but split up a bigger function into multiple smaller logical steps.

hannibal002 (approved):

PR #833, opened 2023-12-25
Only show vacuum bag on your own vacuums

lunaynx:
This prevents the vacuum bag from showing as item stack size in SkyMart and the Auction House, as it can be confusing.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/ItemDisplayOverlayFeatures.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/ItemDisplayOverlayFeatures.kt
@@ -204,6 +204,12 @@ object ItemDisplayOverlayFeatures {
 
         if (VACUUM_GARDEN.isSelected() && item.getInternalNameOrNull() in PestAPI.vacuumVariants) {
             for (line in lore) {
+                if (line.contains("Click to trade!") || line.contains("Starting bid:") || line.contains("Buy it now:")) {
+                    return ""
+                }
+            }
+
+            for (line in lore) {
                 gardenVacuumPatterm.matchMatcher(line) {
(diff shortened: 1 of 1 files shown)


RayDeeUx:
- auction house check is not robust enough; vaccums with more than (and not including) 0 bids would fail this check
- hanni wants regex checks instead of `contains()` string operations now iirc

RayDeeUx:
what i'd suggest for the auction house check is through a chestname regex pattern instead

as for bz check, yeah there's no easy way around it except through lore line looping so just revise it to use regexpattern

lunaynx:
I didn't mention bazaar, it's (obviously) not there if it's on the auction house.

SkyMart could use an inventory title check too I suppose. I'll double check AH once Derpy is gone, since I only tried accessing it through the museum today.

RayDeeUx:
silly me, pretend i meant player trading!

hannibal002:
Isn't it smarter to use an allowed list instead of a denied list for checking where to show this? Only in own inventory, storage and chests on island, right?
Who knows how many other edge cases we have to filter out otherwise, just think of the show block on other players islands or neu PV or auction house display in hub.

lunaynx:
PV is actually not a problem since the lore shows the viewed player's own vacuum bag, but yeah I guess, there could be many other edge cases.

j10a1n15:
I feel like this could be done with a listOf(blockedLines) and then
```kotlin
val blockedLines = listOf("Click to trade!",...)

if (blockedLines.any { it in lore }) return ""
```

hannibal002 (changes requested):
Please dont use "early return" but rather a new function that has a good descriptive name and then use `if(!function()){gardenVacuumPatterm..`

hannibal002 (approved):

PR #838, opened 2023-12-26
Enough Updates

hannibal002:
This is a try to see what will happen if we would remove NEU dependency.
Note that we copy a lot of code from NEU without credits in this PR, they require credits!


hannibal002:
Note that every feature that uses item data like internal name or price is broken, and we need to comment out some other features entirely. Both workarounds are not ideal for a working non neu version of SkyHanni

PR #840, opened 2023-12-27
Improvement: Add delay after tool swap before warning about wrong speed

lunaynx:
## What
Sometimes people don't use Rancher's Boots (or just don't have enough base speed without the tool), and tools themselves often give speed, but it takes time for the tab list to update the speed after swapping.

## Changelog Improvements
+ Added a delay after tool swap before warning about wrong farming speed. - Alexia Luna


lunaynx:
Don't merge yet, looks like with my latest change it actually doesn't work, because you can't draw stuff outside the main thread.

lunaynx:
Should be good to go now.

src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
```diff
@@ -85,7 +88,9 @@ class GardenOptimalSpeed {
         if (GardenAPI.hideExtraGuis()) return
 
         val text = "Optimal Speed: §f$optimalSpeed"
-        if (optimalSpeed != currentSpeed) {
+        if (lastToolSwitch.passedSince() < 1.5.seconds) {
```
Thunderblade73:
```suggestion
        val colorCode = if (lastToolSwitch.passedSince() < 1.5.seconds) "7" else if (optimalSpeed != currentSpeed) "c" else "a"
```

src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
```diff
@@ -85,7 +88,9 @@ class GardenOptimalSpeed {
...
+            if (config.showOnHUD) config.pos.renderString("§7$text", posLabel = "Garden Optimal Speed")
```
Thunderblade73:
```suggestion
            config.pos.renderString("§$colorCode$text", posLabel = "Garden Optimal Speed")
```

hannibal002:
How can I reproduce the bug this pr is trying to fix?

lunaynx:
For example, farm Melon/Pumpkin with Fermento Boots where the tool gives you the little extra push to 400. Or Cactus with Cactus Knife and Rancher's Boots set to 464, where the tool uncaps your speed from 400 to 464.

hannibal002 (approved):

PR #842, opened 2023-12-27
Added options to only show vermin tracker with vacuum on inventory an…

ItsEmpa:
…d only in west village


src/main/java/at/hannibal2/skyhanni/features/rift/area/westvillage/VerminTracker.kt
```diff
@@ -125,6 +126,11 @@ object VerminTracker {
     @SubscribeEvent
     fun onRenderOverlay(event: GuiRenderEvent) {
         if (!isEnabled()) return
+        if (!config.showOutsideWestVillage &&
+            !LorenzUtils.skyBlockArea.let { it == "Infested House" || it == "West Village" }) return
+        if (!config.showWithoutVacuum &&
+            !InventoryUtils.getItemsInOwnInventory().any { it.getInternalName() == "TURBOMAX_VACUUM".asInternalName() }) return
```
hannibal002: This code will try to create the neu internal name object with the name turbomax vacuum each frame for each item in the invenotry. please use a cached instance instead.

hannibal002 (changes requested):

hannibal002 (approved):

PR #845, opened 2023-12-27
Mark carrot/pumpkin fortune as completed when giving to Carrolyn after already done

lunaynx:
The meaning of these messages is not obvious at a glance, but they only appear when trying to give more exportable carrots/expired pumpkin to Carrolyn beyond 3,000.

Screenshot of the feature in action


CalMWolfs (commented):
I'm not sure if the user needs a message telling them that it has been marked as completed, it should just update it in the config imo. If you do really think they need it maybe make it only send when it was previously false

lunaynx:
I could just remove the marked as completed part from the message (at least if it wasn't already), but the rest of the message is useful IMO as the NPC's message here is really confusing.

lunaynx:
Actually, since the normal exchange doesn't trigger a SkyHanni message either I guess I can just remove the "marked as completed" part entirely.

CalMWolfs (approved):

hannibal002 (approved):

PR #851, opened 2023-12-28
Backend: Add Rounded Rectangles

VixidDev:
Adds support for rounded rectangles if contributors wish to use them for hud elements. Usage: `RenderUtils.drawRoundRect(x, y, width, height, color, radius, smoothness)`


Thunderblade73:
The shader breaks if a GlStateManager.translate or GlStateManager.scale is applied before.

hannibal002 (changes requested):
Cant get it to show me anything on screen. I'm calling the logic in a world render event.

hannibal002 (approved):

PR #855, opened 2023-12-29
Backend: Added TrackSoundsCommand

Thunderblade73:
Tracks the incoming sounds for a specified amount of time and copies them into the clipboard

+ Added Track Sounds Command - Thunderblade73
+ Removed unused Sound Test - martimavocado


src/main/java/at/hannibal2/skyhanni/test/command/TrackSoundsCommand.kt
```diff
@@ -0,0 +1,50 @@
...
+import java.util.concurrent.atomic.AtomicBoolean
+import kotlin.time.Duration.Companion.seconds
+
+object TrackSoundsCommand {
+
+    var cutOfTime: SimpleTimeMark = SimpleTimeMark.farPast()
+
+    val sounds = mutableListOf<PlaySoundEvent>()
+
+    var enable = AtomicBoolean(true)
```
lineargraph: There is no point using an `AtomicBoolean` here. An `AtomicBoolean` used only with `get` and `set` provides no benefit over the primitive. Either use some form of locking, or a CAS mechanism. Both of those are overkill (the worst case scenario is that you need to redo your command). Actually all of that is not needed since `DelayedRun` runs on the main thread.

src/main/java/at/hannibal2/skyhanni/test/command/TrackSoundsCommand.kt
```diff
@@ -0,0 +1,50 @@
...
+            LorenzUtils.chat("$counter sounds copied into the clipboard!")
+            enable.set(true)
+        }
+    }
+
+    @SubscribeEvent
+    fun onSoundEvent(event: PlaySoundEvent) {
+        if (cutOfTime.isInPast()) return
+        event.distanceToPlayer // Need to call to initialize Lazy
+        sounds.add(event)
```
lineargraph: `PlaySoundEvent` is run on another thread, so access to the list here needs to be synchronized. Either implement proper locking, or use a COW list.

src/main/java/at/hannibal2/skyhanni/test/command/TrackSoundsCommand.kt
```diff
@@ -0,0 +1,50 @@
...
+        val duration = if (args.size == 1) {
+            args[0].toInt().seconds
+        } else {
+            5.0.seconds
+        }
+        LorenzUtils.chat("Now started tracking sounds for ${duration.inWholeSeconds} Seconds")
+        cutOfTime = SimpleTimeMark.future(duration)
+        DelayedRun.runDelayed(duration + 0.1.seconds) {
+            val string = sounds.joinToString("\n") { it.toString() }
+            val counter = sounds.size
```
lineargraph: Since the `sounds` are modified from off thread, the counter and string may be out of sync (and iterating over the sounds might cause a crash). In practice this won't happen because of the listener stopping earlier, but these kinds of complex invariants (especially when uncommented) aren't great.

src/main/java/at/hannibal2/skyhanni/test/command/TrackSoundsCommand.kt
```diff
@@ -0,0 +1,50 @@
...
+    val sounds = mutableListOf<PlaySoundEvent>()
+
+    var enable = AtomicBoolean(true)
+
+    fun command(args: Array<String>) {
+        if (!enable.get()) {
+            LorenzUtils.chat("Still tracking sounds, wait for the other tracking to complete before starting a new one")
+            return
+        }
+        enable.set(false)
```
lineargraph: `enable` is a pretty bad name for this, i think. Is enable the command being enabled or the listening being enabled? Maybe use something like `currentlyRecording`.

src/main/java/at/hannibal2/skyhanni/test/command/TrackSoundsCommand.kt
```diff
@@ -0,0 +1,50 @@
...
+        sounds.clear()
+        val duration = if (args.size == 1) {
+            args[0].toInt().seconds
+        } else {
+            5.0.seconds
+        }
```
lineargraph:
```suggestion
        val duration = args.firstOrNull()?.toInt()?.seconds ?: 5.0.seconds
```

lineargraph (changes requested):

hannibal002:
This or https://github.com/hannibal002/SkyHanni/pull/1013, only one of them will get merged. Have you already decided which one what? The other can be partially merged into the one, I think.

src/main/java/at/hannibal2/skyhanni/config/features/dev/DebugConfig.java
```diff
@@ -111,4 +112,7 @@ public class DebugConfig {
     @ConfigOption(name = "SkyBlock Area", desc = "Show your current area in SkyBlock while f3 is open.")
     @ConfigEditorBoolean
     public boolean currentAreaDebug = true;
+
+    @Expose
+    public Position trackSoundLog = new Position(0, 0);
```
hannibal002: rename to `trackSoundPosition` for consistency

src/main/java/at/hannibal2/skyhanni/events/PlaySoundEvent.kt
```diff
@@ -9,4 +10,7 @@ class PlaySoundEvent(val soundName: String, val location: LorenzVec, val pitch:
     LorenzEvent() {
 
     val distanceToPlayer by lazy { location.distanceToPlayer() }
+    override fun toString(): String {
+        return "PlaySoundEvent(soundName='$soundName', pitch=$pitch, volume=$volume, location=$location, distanceToPlayer=${distanceToPlayer.round(1)})"
```
hannibal002: also round location to 1 "nachkommastelle"

src/main/java/at/hannibal2/skyhanni/utils/SimpleTimeMark.kt
```diff
@@ -40,6 +40,10 @@ value class SimpleTimeMark(private val millis: Long) : Comparable<SimpleTimeMark
         fun now() = SimpleTimeMark(System.currentTimeMillis())
         fun farPast() = SimpleTimeMark(0)
 
+        fun farFuture() = SimpleTimeMark(Long.MAX_VALUE)
+
+        fun future(duration: Duration) = SimpleTimeMark.now().plus(duration)
```
hannibal002: isntead of this, create extension function Duration.inTheFuture() or somehting maybe?

src/main/java/at/hannibal2/skyhanni/utils/SimpleTimeMark.kt
```diff
@@ -41,6 +41,8 @@ value class SimpleTimeMark(private val millis: Long) : Comparable<SimpleTimeMark
         fun farPast() = SimpleTimeMark(0)
         fun farFuture() = SimpleTimeMark(Long.MAX_VALUE)
 
+        fun future(duration: Duration) = SimpleTimeMark.now().plus(duration)
```
hannibal002: isntead of this, create extension function Duration.inTheFuture() or somehting maybe?

src/main/java/at/hannibal2/skyhanni/test/command/TrackSoundsCommand.kt
```diff
@@ -0,0 +1,87 @@
...
+        val duration: Duration?
+        if (args.firstOrNull()?.isInt() != true) {
+            duration = null
+            ChatUtils.chat("Now started tracking sounds until manually ended")
+        } else {
+            duration = args.firstOrNull()?.toInt()?.seconds
+            ChatUtils.chat("Now started tracking sounds for ${duration?.inWholeSeconds} Seconds")
+        }
+        startTime = SimpleTimeMark.now()
+        cutOfTime = duration?.let { SimpleTimeMark.future(it) } ?: SimpleTimeMark.farFuture()
```
hannibal002: Less code duplication, more "Kotlin like" approach
```suggestion
        cutOfTime = args.firstOrNull()?.toInt()?.seconds?.let {
            ChatUtils.chat("Now started tracking sounds for ${it.inWholeSeconds} Seconds")
            SimpleTimeMark.future(it)
        } ?: run {
            ChatUtils.chat("Now started tracking sounds until manually ended")
            SimpleTimeMark.farFuture()
        }
        startTime = SimpleTimeMark.now()
```

src/main/java/at/hannibal2/skyhanni/test/command/TrackSoundsCommand.kt
```diff
@@ -0,0 +1,87 @@
...
+        if (args.firstOrNull() == "end") {
+            if (!isRecording) {
+                ChatUtils.userError("Nothing to end")
+            } else {
+                cutOfTime = SimpleTimeMark.now()
+            }
+            return
+        }
+        if (isRecording) {
+            ChatUtils.userError("Still tracking sounds, wait for the other tracking to complete before starting a new one or type in \"end\" to end it prematurely")
```
hannibal002:
```suggestion
            ChatUtils.userError(
                "Still tracking sounds, wait for the other tracking to complete before starting a new one, " +
                    "or type §e/shtracksoundsend end §cto end it prematurely"
            )
```

src/main/java/at/hannibal2/skyhanni/test/command/TrackSoundsCommand.kt
```diff
@@ -0,0 +1,87 @@
...
+        if (event.soundName == "game.player.hurt" && event.pitch == 0f && event.volume == 0f) return // remove random useless sound
+        if (event.soundName == "") return // sound with empty name aren't useful
+        event.distanceToPlayer // Need to call to initialize Lazy
+        sounds.addFirst(startTime.passedSince() to event)
+    }
+
+    @SubscribeEvent
+    fun onRender(event: GuiRenderEvent.GuiOverlayRenderEvent) {
+        if (cutOfTime.isInPast()) return
+        val list = sounds.takeWhile { startTime.passedSince() - it.first < 3.0.seconds }
```
hannibal002: can we make this display building in tick event and not every frame?

src/main/java/at/hannibal2/skyhanni/test/command/TrackSoundsCommand.kt
```diff
@@ -0,0 +1,87 @@
...
+        sounds.clear()
+        isRecording = false
+    }
+
+    @SubscribeEvent
+    fun onSoundEvent(event: PlaySoundEvent) {
+        if (cutOfTime.isInPast()) return
+        if (event.soundName == "game.player.hurt" && event.pitch == 0f && event.volume == 0f) return // remove random useless sound
+        if (event.soundName == "") return // sound with empty name aren't useful
+        event.distanceToPlayer // Need to call to initialize Lazy
```
hannibal002, line 83: `PlaySoundEvent.toString()` should also trigger the lazy implementaion `distanceToPlayer`. I don't see the reason for this line.

hannibal002 (changes requested):
Works perfectly fine! Just some code cleanup ideas.

src/main/java/at/hannibal2/skyhanni/test/command/TrackSoundsCommand.kt
Thunderblade73: since the distance might be not correct, because of subtick behavior. And the toString() is only called later on. So the distance might have changed, slightly.

src/main/java/at/hannibal2/skyhanni/utils/SimpleTimeMark.kt
```diff
@@ -41,6 +41,8 @@ value class SimpleTimeMark(private val millis: Long) : Comparable<SimpleTimeMark
         fun farPast() = SimpleTimeMark(0)
         fun farFuture() = SimpleTimeMark(Long.MAX_VALUE)
 
+        fun Duration.fromNow() = SimpleTimeMark.now().plus(this)
```
hannibal002:
```suggestion
        fun Duration.fromNow() = now() + this
```

hannibal002 (approved):

PR #860, opened 2023-12-29
Feature: Block pickaxe ability  on private island

Thunderblade73:
## Dependencies
 - #811 
# What
Blocks the mining ability on the private island. Main use is to prevent people to make cheese out of there island
## Discord: 
https://discord.com/channels/997079228510117908/1187705114874617947


.github/workflows/build.yml
lineargraph, line 1: Are these format changes relevant to this PR? (And with which reason are those format changes being applied?)
The stacktrace is also already available inside of the test report.

src/main/java/at/hannibal2/skyhanni/data/ItemClickData.kt
```diff
@@ -23,18 +23,18 @@ class ItemClickData {
...
-                ItemClickEvent(InventoryUtils.getItemInHand(), ClickType.RIGHT_CLICK).postAndCatch()
+                event.isCanceled = ItemClickEvent(InventoryUtils.getItemInHand(), ClickType.RIGHT_CLICK).postAndCatch()
             }
         }
         if (packet is C07PacketPlayerDigging && packet.status == C07PacketPlayerDigging.Action.START_DESTROY_BLOCK) {
             val position = packet.position.toLorenzVec()
-            BlockClickEvent(ClickType.LEFT_CLICK, position, InventoryUtils.getItemInHand()).postAndCatch()
-            ItemClickEvent(InventoryUtils.getItemInHand(), ClickType.RIGHT_CLICK).postAndCatch()
+            event.isCanceled = BlockClickEvent(ClickType.LEFT_CLICK, position, InventoryUtils.getItemInHand()).postAndCatch() ||
+                ItemClickEvent(InventoryUtils.getItemInHand(), ClickType.RIGHT_CLICK).postAndCatch()
```
lineargraph: Consider cancelling the second event instead of using `||` which completely shortcircuits the second event:
```suggestion
            val blockClickCancelled = BlockClickEvent(ClickType.LEFT_CLICK, position, InventoryUtils.getItemInHand()).postAndCatch()
            event.isCanceled = ItemClickEvent(InventoryUtils.getItemInHand(), ClickType.RIGHT_CLICK).also { it.setCancelled(blockClickCancelled) }.postAndCatch()
```

This way `@SubscribeEvent(receiveCancelled = true)` will still get called.

src/main/java/at/hannibal2/skyhanni/data/RegexData.kt
lineargraph, line 1: I think regexes should be local to the module that uses them. Maybe create an API instead to cover these use cases for multiple files instead of directly exposing the regex.

src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
```diff
@@ -205,29 +207,59 @@ object ItemUtils {
...
+                            "Could not read category for item $name",
+                            "Failed to read category from item rarity via item lore",
+                            "internal name" to getInternalName(),
+                            "item name" to name,
+                            "inventory name" to InventoryUtils.openInventoryName(),
+                            "pattern result" to itemCategory,
+                            "lore" to getLore(),
+                        )
+                    }
+                    return LorenzRarity.valueOf(rarity) to null
```
lineargraph: This line will cause a crash if the regex changes later on to include more rarities. Either be wary of that by returning `null` for a rarity, or replace the `LorenzRarity` from an enum to a value class with an internal / private constructor.

.github/workflows/build.yml
Thunderblade73: The format change, idk. Intellji did them. And the stacktrace thing is from the depended PR from this comment: https://discord.com/channels/997079228510117908/1094190239532208228/1185953806555553882

src/main/java/at/hannibal2/skyhanni/data/RegexData.kt
Thunderblade73: Ehm, it was just a fix for the repoPattern and Test problem, where you couldn't (or didn't wanted) to give me an answer.

src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
Thunderblade73: thx, forgot it for the rarity

src/main/java/at/hannibal2/skyhanni/features/misc/PrivateIslandNoPickaxeAbility.kt
```diff
@@ -0,0 +1,31 @@
...
+import at.hannibal2.skyhanni.utils.LorenzUtils.isInIsland
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+class PrivateIslandNoPickaxeAbility {
+
+    val config get() = SkyHanniMod.feature.mining
+
+    @SubscribeEvent
+    fun onClick(event: WorldClickEvent) {
+        if (!LorenzUtils.inSkyBlock || !IslandType.PRIVATE_ISLAND.isInIsland()) return
```
hannibal002: isInIsland already checks for inSkyBlock. the additional check can be removed

src/main/java/at/hannibal2/skyhanni/utils/LorenzUtils.kt
```diff
@@ -141,7 +141,7 @@ object LorenzUtils {
 
     fun getSBMonthByName(month: String): Int {
         var monthNr = 0
-        for (i in 1..12) {
```
hannibal002: Kotlin does not want empty spaces before or after the .. operator. See the examples without space here: https://kotlinlang.org/docs/ranges.html

src/main/java/at/hannibal2/skyhanni/events/BlockClickEvent.kt
```diff
@@ -4,7 +4,9 @@ import at.hannibal2.skyhanni.data.ClickType
 import at.hannibal2.skyhanni.utils.BlockUtils.getBlockStateAt
 import at.hannibal2.skyhanni.utils.LorenzVec
 import net.minecraft.item.ItemStack
+import net.minecraftforge.fml.common.eventhandler.Cancelable
 
+@Cancelable
 class BlockClickEvent(val clickType: ClickType, val position: LorenzVec, val itemInHand: ItemStack?) : LorenzEvent() {
```
hannibal002: This does not extend world click event. is this intended?

hannibal002 (changes requested):
right clicking on the island correctly blocks, but right clicking on blocks get sent to the server.

src/main/java/at/hannibal2/skyhanni/events/BlockClickEvent.kt
Thunderblade73: that is not intended. Happend during merge conflict solving. (Also the bug with the block clicking)

hannibal002 (approved):

PR #862, opened 2023-12-30
Wrong/broken Changelog: Add Emoji Replacement

martimavocado:
- Edits messages sent to the server to simulate hypixel's emojis
image
- Edits messages recieved to add colors to those emojis
image


src/main/java/at/hannibal2/skyhanni/features/chat/EmojiColorer.kt
```diff
@@ -0,0 +1,132 @@
...
+        "§r"
+    )
+
+    @SubscribeEvent
+    fun onChatReceive(event: LorenzChatEvent) {
+        if (config.colorEmoji && EmojiReplacer.inArray(event.message.removeColor(), emojis)) {
+            val newMessage = replace(event, emojis)
+            event.blockedReason = "Emoji"
+            LorenzUtils.chat(newMessage, false)
+        }
```
lineargraph, line 82: Please modify the IChatComponent instead of replacing it outright. This breaks hover events for the profile viewer or emblems (or even soopy [hand]). You will need to walk through all siblings of the component (and siblings thereof) manually and replace the text contents of just those components.

src/main/java/at/hannibal2/skyhanni/features/chat/EmojiColorer.kt
```diff
@@ -0,0 +1,132 @@
...
+    private fun findFirstIndex(message: String, lastIndex: Int): Int {
+        val subString = message.substring(0, lastIndex)
+        var maxIndex = 0
+
+        resettingCodes.forEach { code ->
+            val index = subString.lastIndexOf(code)
+            if (index > maxIndex) maxIndex = index
+        }
+        return maxIndex
+    }
```
lineargraph, line 131: This whole color logic feels off/overly complicated to me:
```suggestion
    /** Find a string that fully restores the current color state when prepended to another string */
    private fun findColorState(string: String): String {
        val lastColorIndex = resettingCodes.maxOf { string.lastIndexOf(it) }
        val lastColor = if (lastColorIndex < 0) "§r" else string.substring(lastColorIndex, lastColorIndex + 1)
        val restOfText = string.substring(it + 1)
        val modifiers = formattingCodes.filter { it in restOfText }
        return lastColor + modifiers.joinToString("")
    }
```

Keep in mind this requires `"§r"` to be deleted from `formattingCodes`, and instead be added to `resettingCodes`

src/main/java/at/hannibal2/skyhanni/features/chat/EmojiColorer.kt
```diff
@@ -0,0 +1,132 @@
...
+    @SubscribeEvent
+    fun onChatReceive(event: LorenzChatEvent) {
+        if (config.colorEmoji && EmojiReplacer.inArray(event.message.removeColor(), emojis)) {
+            val newMessage = replace(event, emojis)
+            event.blockedReason = "Emoji"
+            LorenzUtils.chat(newMessage, false)
+        }
+    }
+
+    private fun replace (event: LorenzChatEvent, array: Array<Pair<String, String>>): String {
```
lineargraph, line 85: This function should operate only on strings (inside of a ChatComponentText leaf).

src/main/java/at/hannibal2/skyhanni/features/chat/EmojiColorer.kt
```diff
@@ -0,0 +1,132 @@
...
+        if (EmojiReplacer.inArray(event.message.removeColor(), array)) {
```
lineargraph, line 86: There is not really any point to checking for the presence of emojis first. We still need to iterate over everything, so we might as well do all the work.

src/main/java/at/hannibal2/skyhanni/features/chat/EmojiColorer.kt
```diff
@@ -0,0 +1,132 @@
...
+            var oldMessage = event.message
+            array.forEach { (search, replace) ->
+                if (oldMessage.contains(search)) {
+                    val oldColor = findColor(oldMessage, search)
+                    oldMessage = oldMessage.replace(search, "$replace$oldColor")
+                }
+            }
+            return oldMessage
```
lineargraph, line 94:
```suggestion
            var oldMessage = event.message
            for ((search, replace) in array) {
                val parts = oldMessage.split(search)
                if (parts.size == 1) continue
                var oldMessage = parts[0]
                for (nextPart in parts) {
                    oldMessage += replace + findColor(oldMessage) + nextPart
                }
            }
            return oldMessage
```

lineargraph (changes requested):
Originally i wanted to look at the entire PR, but i got hung up on the emoji colorer and now i am sad.

hannibal002:
months old

PR #867, opened 2024-01-01
Fix: TiaHelper chatmessage

Thunderblade73:
Maybe fixed the chatmessages for the Tia Helper (couldn't test)
And migrated the messages to repo Pattern

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
+++ b/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
@@ -210,7 +210,6 @@ import at.hannibal2.skyhanni.features.inventory.SackDisplay
 import at.hannibal2.skyhanni.features.inventory.ShiftClickEquipment
 import at.hannibal2.skyhanni.features.inventory.SkyBlockLevelGuideHelper
 import at.hannibal2.skyhanni.features.inventory.StatsTuning
-import at.hannibal2.skyhanni.features.inventory.tiarelay.Relay
 import at.hannibal2.skyhanni.features.inventory.tiarelay.TiaRelayHelper
 import at.hannibal2.skyhanni.features.inventory.tiarelay.TiaRelayWaypoints
 import at.hannibal2.skyhanni.features.itemabilities.ChickenHeadTimer
@@ -534,7 +533,6 @@ class SkyHanniMod {
         loadModule(OdgerWaypoint())
         loadModule(TiaRelayHelper())
         loadModule(TiaRelayWaypoints())
-        Relay.entries.forEach { loadModule(it) }
         loadModule(BasketWaypoints())
         loadModule(BingoCardDisplay())
         loadModule(BingoNextStepHelper())
--- a/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
+++ b/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
@@ -210,6 +210,7 @@ import at.hannibal2.skyhanni.features.inventory.SackDisplay
 import at.hannibal2.skyhanni.features.inventory.ShiftClickEquipment
 import at.hannibal2.skyhanni.features.inventory.SkyBlockLevelGuideHelper
 import at.hannibal2.skyhanni.features.inventory.StatsTuning
+import at.hannibal2.skyhanni.features.inventory.tiarelay.Relay
 import at.hannibal2.skyhanni.features.inventory.tiarelay.TiaRelayHelper
 import at.hannibal2.skyhanni.features.inventory.tiarelay.TiaRelayWaypoints
 import at.hannibal2.skyhanni.features.itemabilities.ChickenHeadTimer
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/tiarelay/Relay.kt
```diff
@@ -2,47 +2,53 @@ package at.hannibal2.skyhanni.features.inventory.tiarelay
...
+        "§e[NPC] §dTia the Fairy§f: §b✆ §f§r§fThe next relay is on top of a statue in the dungeon hub."
     ),
     RELAY_9(
         "9th Relay", LorenzVec(-19.0, 88.5, -91.0), IslandType.HUB,
-        "§e[NPC] §dTia the Fairy§f: §b✆ §f§rThe next relay is on top of the Auction House."
-    ),
+        "§e[NPC] §dTia the Fairy§f: §b✆ §f§r§fThe next relay is on top of the Auction House."
+    );
+
+    private val repoGroup = RepoPattern.group("relay.chat")
```
hannibal002: You need to use `by` instead of `=`

hannibal002: You can put the field in a companion block instead of loading one instance for every enum instance

src/main/java/at/hannibal2/skyhanni/features/inventory/tiarelay/Relay.kt
```diff
@@ -2,47 +2,53 @@ package at.hannibal2.skyhanni.features.inventory.tiarelay
...
+    val chatPattern by repoGroup.pattern(relayName.takeWhile { it != ' ' }, chatMessage)
```
hannibal002: maybe make this `by lazy`

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
```diff
@@ -533,6 +534,7 @@ class SkyHanniMod {
         loadModule(OdgerWaypoint())
         loadModule(TiaRelayHelper())
         loadModule(TiaRelayWaypoints())
+        Relay.entries.forEach { loadModule(it) }
```
hannibal002: regex pattern is no event, you dont need to register it here.

hannibal002 (changes requested):
I  cant test it in game too. Best we test it in beta soon.

src/main/java/at/hannibal2/skyhanni/features/inventory/tiarelay/Relay.kt
lineargraph: I don't think a lazy works here.

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
Thunderblade73: But repoPatterns need to be initialized in the preinit phase to work. And I do not know how to get it working without loadModule

src/main/java/at/hannibal2/skyhanni/features/inventory/tiarelay/Relay.kt
hannibal002: oh have not noticed the by in by block then. maybe it is fine then as is?

lineargraph: Btw, did you mean a `by lazy` for the regex, or a `by lazy` for the repopattern? The repopattern needs to be eagerly initalized during preinit, so that the regex extraction works.

src/main/java/at/hannibal2/skyhanni/features/inventory/tiarelay/Relay.kt
Thunderblade73: > You need to use `by` instead of `=`

nope, it's a group that's why it needs a `=` not a `by`

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
lineargraph: You don't need to load the regexes as modules, you just need to make sure that the Relay class is initialized during preinit. You can do that by just accessing the Relay class during either static or instance init of `TiaRelayWaypoints`.

src/main/java/at/hannibal2/skyhanni/features/inventory/tiarelay/TiaRelayWaypoints.kt
```diff
@@ -17,19 +17,21 @@ class TiaRelayWaypoints {
     private var waypointName: String? = null
     private var island = IslandType.NONE
 
+    init {
+        Relay.entries.forEach { it.chatPattern }
+    }
```
Thunderblade73, line 22: @nea89o like this?

hannibal002 (approved):

PR #868, opened 2024-01-01
Improve profile switch detection

lunaynx:
Hypixel doesn't always send the "Switching to profile [name]..." message, and many modules were relying solely on that.

Now we just use the tab list and messages that always appear instead.


src/main/java/at/hannibal2/skyhanni/data/ProfileStorageData.kt
```diff
@@ -67,34 +33,28 @@ object ProfileStorageData {
...
     @SubscribeEvent
     fun onTabListUpdate(event: TabListUpdateEvent) {
-        if (profileSpecific != null) return
-        val playerSpecific = playerSpecific ?: return
-        val sackPlayers = sackPlayers ?: return
+        if (!LorenzUtils.inSkyBlock) return
+
         for (line in event.tabList) {
-            val pattern = "§e§lProfile: §r§a(?<name>.*)".toPattern()
+            val pattern = "§e§lProfile: §r§a.*".toPattern()
```
lunaynx: Note to self: Update this to handle the Rift version as well.

lunaynx:
Any chance for this to be merged for 0.23 beta soon considering 0.23 is getting delayed due to future tab list changes and Bingo started again?

It will need some adjustments for the tab list change as well, but I can't exactly do that right now because they removed that from alpha, so I'll do it later once it's back.

src/main/java/at/hannibal2/skyhanni/data/HypixelData.kt
```diff
@@ -28,6 +29,7 @@ import kotlin.concurrent.thread
 class HypixelData {
     private val group = RepoPattern.group("data.hypixeldata")
     private val tabListProfilePattern by group.pattern("tablistprofile", "§e§lProfile: §r§a(?<profile>.*)")
+    private val tabListProfilePatternRift by group.pattern("tablistprofilerift", "§d§lProfile: §r§a(?<profile>.*)")
```
CalMWolfs:
```suggestion
    private val tabListProfilePattern by group.pattern("tablistprofile", "§.§lProfile: §r§a(?<profile>.*)")
```
This could be one pattern instead of two

src/main/java/at/hannibal2/skyhanni/data/HypixelData.kt
```diff
@@ -112,6 +115,27 @@ class HypixelData {
...
+                return
+            }
+            tabListProfilePatternRift.matchMatcher(line) {
+                // Hypixel shows the profile name reversed while in the Rift
+                val newProfile = group("profile").lowercase().reversed()
+                if (profileName == newProfile) return
+                profileName = newProfile
+                ProfileJoinEvent(newProfile).postAndCatch()
+                return
+            }
```
CalMWolfs:
```suggestion
```
delete this (see above) also if it wasnt deleted it should have a in rift check then if in rift you reverse the output

src/main/java/at/hannibal2/skyhanni/data/ProfileStorageData.kt
```diff
@@ -67,34 +33,28 @@ object ProfileStorageData {
...
     @SubscribeEvent
     fun onTabListUpdate(event: TabListUpdateEvent) {
-        if (profileSpecific != null) return
-        val playerSpecific = playerSpecific ?: return
-        val sackPlayers = sackPlayers ?: return
+        if (!LorenzUtils.inSkyBlock) return
+
         for (line in event.tabList) {
-            val pattern = "§e§lProfile: §r§a(?<name>.*)".toPattern()
+            val pattern = "§[ed]§lProfile: §r§a.*".toPattern()
```
CalMWolfs: dont like 
a) that this pattern is being made for each line in the tablist
b) that it is the exact same pattern as in the HypixelData file, maybe move the pattern to UtilsPattern and reference both from there

src/main/java/at/hannibal2/skyhanni/data/ProfileStorageData.kt
```diff
@@ -22,42 +21,9 @@ object ProfileStorageData {
     var loaded = false
     private var noTabListTime = -1L
```
CalMWolfs: i know this isnt you but can you convert this file to use SimpleTimeMark

CalMWolfs (changes requested):
A few things to change, also afaik once you are in a lobby your profile name cant change until you are in a differnt lobby. Maybe once profile name is found once in a lobby a boolean is set to true until world switch and then we dont check tablist or anything else when the profile name is found.

src/main/java/at/hannibal2/skyhanni/data/ProfileStorageData.kt
lunaynx: Yeah, I tried to reuse the RepoPattern from the other file but it caused a weird completely unrelated test failure with visitors. I will try with UtilsPattern.

CalMWolfs: > Yeah, I tried to reuse the RepoPattern from the other file but it caused a weird completely unrelated test failure with visitors. I will try with UtilsPattern.

I'm 99% sure i fixed test failures since then

hannibal002 (changes requested):
Resolve suggestions from Cal.

lunaynx:
Addressed comments I think. I decided to just give the tab list pattern a new name under utils, because the new pattern could cause trouble with old SkyHanni versions that don't reverse the name while in the Rift.

hannibal002 (changes requested):
The merge conflicts with Repo Pattern PR that just got merged are a bit too complicated to easily resolve them. Please you do it.

hannibal002 (approved):

PR #874, opened 2024-01-03
feat: Party kick with reason

lineargraph:
Adds `/party kick someone Kick reason here` which informs people why you are kicking them via party chat message before doing so.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/commands/PartyCommands.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/commands/PartyCommands.kt
@@ -4,6 +4,7 @@ import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.config.ConfigUpdaterMigrator
 import at.hannibal2.skyhanni.data.FriendAPI
 import at.hannibal2.skyhanni.data.PartyAPI
+import at.hannibal2.skyhanni.events.MessageSendToServerEvent
 import at.hannibal2.skyhanni.utils.LorenzUtils
 import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
 
@@ -26,6 +27,8 @@ object PartyCommands {
         if (!config.shortCommands) return
         if (PartyAPI.partyMembers.isEmpty()) return
         if (args.isEmpty()) return
+        if (args.size > 1 && config.partyKickReason)
+            LorenzUtils.sendCommandToServer("pc Kicking ${args[0]}: ${args.drop(1).joinToString(" ").trim()}")
         LorenzUtils.sendCommandToServer("party kick ${args[0]}")
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/at/hannibal2/skyhanni/utils/LorenzUtils.kt
```diff
@@ -364,18 +369,28 @@ object LorenzUtils {
...
+    @SubscribeEvent
+    fun sendQueuedChatMessages(ignored: LorenzTickEvent) {
+        val thePlayer = Minecraft.getMinecraft().thePlayer
+        if (thePlayer == null) {
+            sendQueue.clear()
+            return
+        }
+        if (lastMessageSent.passedSince() > 300.milliseconds) {
+            lastMessageSent = SimpleTimeMark.now()
+            thePlayer.sendChatMessage(sendQueue.poll() ?: return)
```
CalMWolfs: i think this should also check if there is something in the queue so that when a command is sent it is not waiting for up to 300ms

src/main/java/at/hannibal2/skyhanni/utils/LorenzUtils.kt
```diff
@@ -364,18 +369,28 @@ object LorenzUtils {
         return this
     }
 
-    private var lastMessageSent = 0L
+    private var lastMessageSent = SimpleTimeMark.farPast()
+    private val sendQueue: Queue<String> = LinkedList()
+
+    @SubscribeEvent
+    fun sendQueuedChatMessages(ignored: LorenzTickEvent) {
```
CalMWolfs: while the event is ignored all of the existing code still calls the argument event

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/LorenzUtils.kt
lineargraph: Is that a guideline into which thought was put, or is that simply the convention that grew out of complacency?

CalMWolfs: Convention that doesn't have to be followed. Main review was for the other bit

CalMWolfs (approved):

CalMWolfs (changes requested):
doesnt work with `/p kick <player> <reason>`

CalMWolfs (approved):

src/main/java/at/hannibal2/skyhanni/features/commands/PartyCommands.kt
```diff
@@ -46,6 +49,24 @@ object PartyCommands {
         LorenzUtils.sendCommandToServer("party promote ${args[0]}")
     }
 
+    @SubscribeEvent
+    fun onSendCommand(event: MessageSendToServerEvent) {
+        if (!config.partyKickReason) {
+            return
+        }
+        if (!event.message.startsWith("/party kick ") && !event.message.startsWith("/p kick ")) {
```
hannibal002: Please compare case-insensitive.

src/main/java/at/hannibal2/skyhanni/utils/LorenzUtils.kt
hannibal002: please rename ignored to event for consistency.

src/main/java/at/hannibal2/skyhanni/utils/LorenzUtils.kt
```diff
@@ -364,18 +369,28 @@ object LorenzUtils {
...
+        val thePlayer = Minecraft.getMinecraft().thePlayer
```
hannibal002: the player -> player (consistency

hannibal002 (changes requested):
Looks okay, but please keep the consistency high

hannibal002 (approved):
works fine in all combinations to me.
the next feature would be to send a `!p kick name` chat message if the player has no kick permissions, and other party members can click on this chat message to execute the kick command

PR #823, opened 2023-12-20
Backend: Add Dark Auction as IslandType and fix IslandType detection for dungeons

j10a1n15:
This adds the Dark Auction as a new IslandType and fixes the IslandType detection for dungeons.

PR #877, opened 2024-01-04
Fix: Fix IslandType detection for The Rift

ItsEmpa:
Broke in #823


j10a1n15:
Actually while youre at it, knowing Hyixel, can you change the pattern to `(§.)*(Area|Dungeon): (§.)*(?<island>.*)`

hannibal002 (approved):

PR #882, opened 2024-01-06
Change: "Dungeon Filter" Dropdown instead of Toggle

byWambo:
Changed "Dungeon Filter" toggle to a dropdown with 5 options "All", "All w/o buffs", "All w/o puzzles", "Only start/end" and "Off". Also changed description of the setting

I don't really know what other combinations people would use

grafik


byWambo:
Discussed with hannibal2 that it would be converted to a list and not a dropdown

PR #883, opened 2024-01-06
Fix: Firesale Pattern update

Thunderblade73:
fixes the new "for a limited time" line


hannibal002 (changes requested):
I'm not sure if I'm doing something wrong, but it doesn't work for me: https://regex101.com/r/orhDp3/1

hannibal002 (approved):

PR #886, opened 2024-01-07
Backend: Renderable scrollables

Thunderblade73:
## Dependencies
- #885 
# What
Adds Scrolllist and Scrolltable. Those make it possible to have scrolling content in any GUI.

+ Added scroll able logic for Renderables. - Thunderblade73
    * Added elements: scrollList and scrollTable.


hannibal002:
I have tested this pr with https://github.com/hannibal002/SkyHanni/pull/1334


image
image

Does not work. Am i using it correctly?

hannibal002 (changes requested):
see above

Thunderblade73:
Fixed the bug (and improved the list a bit). 
I have never used the list version, only used the table version therefore it was a bit bugged. 
The issue was that I used a maxOf instead of a sumOf 😅 and that I never checked the end bound.

hannibal002 (approved):
works in the other pr

PR #887, opened 2024-01-07
Backend: Renderable progressBar

Thunderblade73:
## Dependencies
- #885 
# What
simply the progressBar we used in /ff extracted to a renderable


src/main/java/at/hannibal2/skyhanni/utils/NumberUtil.kt
```diff
@@ -206,4 +206,8 @@ object NumberUtil {
     val Int.milion get() = this * 1_000_000.0
     private val Int.bilion get() = this * 1_000_000_000.0
     val Double.milion get() = (this * 1_000_000.0).toLong()
+
+    fun Number.percentOf(maxValue: Number) = maxValue.toDouble().takeIf { it != 0.0 }?.let { max ->
```
lineargraph: Misnomer: percentage is part out of a 100. Maybe `fractionOf` or `progressTowards`?

src/main/java/at/hannibal2/skyhanni/utils/ColorUtils.kt
```diff
@@ -8,4 +10,10 @@ object ColorUtils {
...
 
     fun getAlpha(colour: Int) = colour shr 24 and 0xFF
-}
\ No newline at end of file
+
+    fun blend(start: Color, end: Color, percent: Double) = Color(
+        (start.getRed() * (1 - percent) + end.getRed() * percent).toInt(),
+        (start.getGreen() * (1 - percent) + end.getGreen() * percent).toInt(),
+        (start.getBlue() * (1 - percent) + end.getBlue() * percent).toInt()
+    )
```
lineargraph: Not sure if RGB is the best color space for blending. It is probably fine, but i think the function name should indicate that the blending is done in RGB.

lineargraph (changes requested):

hannibal002 (approved):

PR #898, opened 2024-01-09
Check if sneaking for optimal speed in garden

lunaynx:
This is useful for people who use Fermento Boots with toggle sneak for wheat, carrot, potato and nether wart to achieve the desired speed, because the only way to get below 100 speed without Rancher's Boots is sneaking or a slowness potion.

I round down the speed here because generally when farming you rather want to overshoot your speed a bit than undershoot to ensure 20 BPS.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
@@ -20,7 +20,12 @@ import kotlin.time.Duration.Companion.seconds
 class GardenOptimalSpeed {
     private val config get() = GardenAPI.config.optimalSpeeds
     private val configCustomSpeed get() = config.customSpeed
-    private var currentSpeed = 100
+    private var _currentSpeed = 100
+    private var currentSpeed: Int
+        get() = (_currentSpeed * (if (Minecraft.getMinecraft().thePlayer.isSneaking) 0.3 else 1.0)).toInt()
+        set(value) {
+            _currentSpeed = value
+        }
     private var optimalSpeed = -1
     private val currentSpeedPattern = " Speed: §r§f✦(?<speed>.*)".toPattern()
     private var lastWarnTime = 0L
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
@@ -89,8 +89,9 @@ class GardenOptimalSpeed {
 
         if (GardenAPI.hideExtraGuis()) return
 
-        val text = "Optimal Speed: §f$optimalSpeed"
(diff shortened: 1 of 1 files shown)


hannibal002 (changes requested):
While I like this addition, I think we also need to show the current speed next to the wrong speed (if speed is wrong) to make the sneak impact more user friendly.

lunaynx:
Yeah I will do that. I think we should also check to see if the player has been sneaking for some time already, otherwise it can trigger unnecessary alerts when they just sneak for a moment while standing still.

lunaynx:
Done both.

src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
```diff
@@ -15,18 +16,36 @@ import net.minecraft.client.gui.inventory.GuiEditSign
...
+        }
     private var optimalSpeed = -1
     private val currentSpeedPattern = " Speed: §r§f✦(?<speed>.*)".toPattern()
     private var lastWarnTime = 0L
     private var cropInHand: CropType? = null
     private var rancherOverlayList: List<List<Any?>> = emptyList()
 
+    @SubscribeEvent
+    fun onTick(event: LorenzTickEvent) {
+        if (sneaking) {
```
hannibal002, line 42: Maybe save the time stamp of the last non sneaking moment instead. (Mutable vs immutable state)

src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
```diff
@@ -84,9 +103,14 @@ class GardenOptimalSpeed {
...
 
-        val text = "Optimal Speed: §f$optimalSpeed"
+        var text = "Optimal Speed: §f$optimalSpeed"
         if (optimalSpeed != currentSpeed) {
+            text += " (§eCurrent: §f$currentSpeed"
+            if (sneaking) text += " §7[Sneaking]"
+            text += "§f)"
+
             if (config.showOnHUD) config.pos.renderString("§c$text", posLabel = "Garden Optimal Speed")
+            if (sneaking && !sneakingPersistent) return
```
hannibal002, line 113: This logic means "no warning when sneaking for more than 5 seconds" and this cant be right. some loigcal error maybe?

hannibal002 (changes requested):
not sure

src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
lunaynx: It means no warning if sneaking but not for long enough (5 seconds).

hannibal002: ohh nvm

hannibal002 (approved):

PR #905, opened 2024-01-11
Warn to open visitor inventory again after partially serving Spaceman

lunaynx:
Also make sure the visitor shopping list is always updated when opening the GUI.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorFeatures.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorFeatures.kt
@@ -460,7 +460,7 @@ class GardenVisitorFeatures {
         }
 
         partialAcceptedPattern.matchMatcher(event.message) {
-            LorenzUtils.chat("Talk to the visitor again to update the number of items needed!")
+            if (config.shoppingList.display) LorenzUtils.chat("Talk to the visitor again to update the number of items needed!")
         }
     }
 
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorFeatures.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorFeatures.kt
@@ -460,7 +460,7 @@ class GardenVisitorFeatures {
         }
 
         partialAcceptedPattern.matchMatcher(event.message) {
-            LorenzUtils.chat("Open visitor inventory again to update the number of items needed!")
+            LorenzUtils.chat("Talk to the visitor again to update the number of items needed!")
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorFeatures.kt
```diff
@@ -457,6 +458,10 @@ class GardenVisitorFeatures {
         if (GardenAPI.inGarden() && config.hideChat && hideVisitorMessage(event.message)) {
             event.blockedReason = "garden_visitor_message"
         }
+
+        partialAcceptedPattern.matchMatcher(event.message) {
+            LorenzUtils.chat("Open visitor inventory again to update the number of items needed!")
```
hannibal002: Maybe use the phrase "talk to the spaceman again" instead

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorFeatures.kt
lunaynx: How about "talk to the visitor again"? I think it's better not to hardcode it in case Hypixel adds another Spaceman-like visitor in the future.

CalMWolfs (approved):
i think this message is fine

lineargraph (approved):

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorFeatures.kt
```diff
@@ -457,6 +458,10 @@ class GardenVisitorFeatures {
         if (GardenAPI.inGarden() && config.hideChat && hideVisitorMessage(event.message)) {
             event.blockedReason = "garden_visitor_message"
         }
+
+        partialAcceptedPattern.matchMatcher(event.message) {
```
hannibal002, line 462: Only show the chat message when the skyhanni shopping list is enabled.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorFeatures.kt
lunaynx: Done

hannibal002: We also don't need to check the pattern if the setting is disabled.

hannibal002 (changes requested):
See comment above

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorFeatures.kt
hannibal002: Well not relevant enough ig

hannibal002 (approved):

PR #909, opened 2024-01-11
Wrong/broken Changelog: Toast Notifications

byWambo:
Cool toast notificiations for all platforms

Done:
- [ ] Windows
- [ ] MacOS
- [ ] Linux

e.g.
grafik


src/main/java/at/hannibal2/skyhanni/utils/RegistryUtil.kt
```diff
@@ -0,0 +1,56 @@
+package at.hannibal2.skyhanni.utils
+
+object RegistryUtil {
```
lineargraph, line 3: Probably not super high prio, but shelling out is usually pretty expensive. Maybe we can apply some sort of caching layer to the functions in here? Depends on how often these functions are called.

src/main/java/at/hannibal2/skyhanni/utils/RegistryUtil.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    private const val PREFIX: String = "REG"
+    private const val SUCCESS_CODE: Int = 0
+    private const val FAILURE_CODE: Int = 1
+
+    fun addKey(runtime: Runtime, location: RegistryPaths, key: String) : Boolean {
```
lineargraph, line 9: There isn't really a point to pass the `Runtime` around. There will only ever be one of those, so we might as well save us the parameter for all those methods.

src/main/java/at/hannibal2/skyhanni/utils/RegistryUtil.kt
```diff
@@ -0,0 +1,56 @@
...
+        val proc: Process = runtime.exec("$PREFIX ADD ${location.path}\\$key")
```
lineargraph, line 10: Maybe we can skip some shlexing here and use the `Runtime#exec(String[])` variant instead.

src/main/java/at/hannibal2/skyhanni/utils/RegistryUtil.kt
```diff
@@ -0,0 +1,56 @@
+package at.hannibal2.skyhanni.utils
+
+object RegistryUtil {
+
+    private const val PREFIX: String = "REG"
```
lineargraph, line 5: I don't think `PREFIX` is the correct name here.

PR #911, opened 2024-01-11
Feature: Open past contests in https {colon} {slash} {slash} elitebot {dot} dev

RayDeeUx:
impls this:

also adds two commands:
\- `/shelitebotdevcontest <month number>m <day number>d <year number>y` (but the parameters can be in any order as long as they all have the right suffix)
\- `/shopencontest [case-sensitive month name] [day] Year <year>` (parameters in `[]`s are optional)


RayDeeUx:
bruh, the commit number 😭

hannibal002:
I can't tell if this PR is ment to be serious but I can't explain to you while being honest that there is a way to fix the stuff that is wrong with this PR so I just close it.

PR #850, opened 2023-12-28
Feature: Pet Item Display

Thunderblade73:
# What
Adds icons for pets for there equipped pet item.
Implemented for Exp Share and Tier Boost implemented. More items can be easily added later on.
## Preview
image image

PR #920, opened 2024-01-13
Feature: Tier Boosted Pet Display

Thunderblade73:
image

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/misc/pets/PetConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/misc/pets/PetConfig.java
@@ -28,4 +28,10 @@ public class PetConfig {
     @ConfigEditorBoolean
     @FeatureToggle
     public boolean hideAutopet = false;
+
+    @Expose
+    @ConfigOption(name = "Show Exp Share", desc = "Shows an §5⚘ §7by a pet when an exp share is equipped")
+    @ConfigEditorBoolean
+    @FeatureToggle
+    public boolean expShare = false;
 }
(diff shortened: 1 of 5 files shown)


src/main/java/at/hannibal2/skyhanni/config/features/misc/pets/PetConfig.java
```diff
@@ -28,4 +28,10 @@ public class PetConfig {
     @ConfigEditorBoolean
     @FeatureToggle
     public boolean hideAutopet = false;
+
+    @Expose
+    @ConfigOption(name = "Show Tier Boost", desc = "Shows an §c● §7by a pet when an tier boost core is equipped")
```
CalMWolfs: Maybe it's just me but "an" sounds funny I think both could just be "a". Also missing the fullstop at the end of the description

src/main/java/at/hannibal2/skyhanni/config/features/misc/pets/PetConfig.java
```diff
@@ -28,4 +28,10 @@ public class PetConfig {
...
+    @ConfigEditorBoolean
+    @FeatureToggle
```
CalMWolfs: I don't feel this feature qualifies for this annotation but up to you

CalMWolfs (changes requested):
Pr looks fine apart from those two things. Which prob also applies to the other pr of yours that is very similar but I won't rereview that. Though I do have a suggestion, instead of having two features like this with similar logic what if you made a drop-down list for the user to select which pet items they want to display a symbol for. This would make adding new ones much easier and stop some code duplication but this is up to you

Thunderblade73:
merged with #850

PR #922, opened 2024-01-13
test scala

hannibal002:
test scala. see #595


hannibal002:
it failed. and this is good.

PR #927, opened 2024-01-15
Fix Frozen Treasure Tracker showing incorrect Compact Procs number

lunaynx:


CalMWolfs (approved):
Good catch

hannibal002 (approved):
so good fix

PR #928, opened 2024-01-15
Fix daily city project reminder triggering for released projects

lunaynx:


CalMWolfs (approved):
lgtm

src/main/java/at/hannibal2/skyhanni/features/fame/CityProjectFeatures.kt
```diff
@@ -35,8 +37,15 @@ class CityProjectFeatures {
     private var inInventory = false
     private var lastReminderSend = 0L
 
-    // TODO USE SH-REPO
-    private val contributeAgainPattern = "§7Contribute again: §e(?<time>.*)".toPattern()
+    private val patternGroup = RepoPattern.group("fame.projects")
```
hannibal002, line 40: Please use the name event.cityproject or similar. while fame is technically not wrong, this name is not used in the config or class or packet. no need to start using different names with the pattern now.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/fame/CityProjectFeatures.kt
CalMWolfs: me when: `src/main/java/at/hannibal2/skyhanni/features/fame/CityProjectFeatures.kt`

hannibal002: Well, this is fair. Otherwise, it's more an event like thing (only temporarily open) then a fame thing imo. Maybe event.fame.cityproject?

lunaynx: I copied the group name from @CalMWolfs's PR, but sure, we can change it.

CalMWolfs: Yea I thought it was fine but you can change it and then I'll update my pr to reflect this

hannibal002: yeah it is fine as is

hannibal002 (approved):

PR #929, opened 2024-01-15
Feature: Hide SC Tracker when wearing Trophy Hunter armor

martimavocado:
//dependency: https://github.com/hannibal002/SkyHanni-REPO/pull/96
moved to repo pattern

https://discord.com/channels/997079228510117908/1196324052269277284


src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
```diff
@@ -151,10 +157,25 @@ object SeaCreatureTracker {
...
 
     private fun isEnabled() = LorenzUtils.inSkyBlock &&
-        FishingAPI.lastActiveFishingTime.passedSince() < 10.minutes && config.enabled
+        FishingAPI.lastActiveFishingTime.passedSince() < 10.minutes && config.enabled && !isTrophyFishing()
+
+    private fun isTrophyFishing(): Boolean {
+        val boots = InventoryUtils.getBoots()?.getInternalName().toString().replace("internalName:", "")
+        val leggings = InventoryUtils.getLeggings()?.getInternalName().toString().replace("internalName:", "")
+        val chestplate = InventoryUtils.getChestplate()?.getInternalName().toString().replace("internalName:", "")
+        val helmet = InventoryUtils.getHelmet()?.getInternalName().toString().replace("internalName:", "")
```
CalMWolfs: Use .asString instead to remove this weirdness

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
martimavocado: makes sense something like this would exist

src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
```diff
@@ -27,6 +32,7 @@ object SeaCreatureTracker {
 
     private val tracker = SkyHanniTracker("Sea Creature Tracker", { Data() }, { it.fishing.seaCreatureTracker })
     { drawDisplay(it) }
+    private val armorNames = mutableListOf<String>()
```
CalMWolfs: This could also maybe be trophyArmorNames

CalMWolfs (changes requested):
This other little change as well. Also if it is possible you probably don't need to check the armor every tick or whatever and when checking as soon as one piece of armor is trophy armor that's probably good enough. 

On another note I'd feel a repo pattern for the armor names would be more effective than looping through a list of armor names

martimavocado:
yeah that does sound better, we don't have a armor change event do we

CalMWolfs:
> yeah that does sound better, we don't have a armor change event do we

Don't think so but you could make one. Also best that if you do this it's in its own pr

src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
lineargraph: Prefer immutable data structures wherever possible. You can also use a `Set` here instead, for faster lookups.

```suggestion
    private var armorNames = setOf<NEUInternalName>()
```

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/ItemsJson.java
```diff
@@ -11,4 +11,7 @@ public class ItemsJson {
 
     @Expose
     public Map<String, Integer> crimson_tiers;
+
+    @Expose
+    public List<String> trophy_armors;
```
lineargraph: You can use the `NEUInternalName` class here, to indicate these Strings are item ids.

```suggestion
    public List<NEUInternalName> trophy_armors;
```

src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
```diff
@@ -151,10 +157,25 @@ object SeaCreatureTracker {
         tracker.renderDisplay(config.position)
     }
 
+    @SubscribeEvent
+    fun onRepoReload(event: RepositoryReloadEvent) {
+        val data = event.getConstant<ItemsJson>("Items")
+        armorNames.clear()
+        armorNames.addAll(data.trophy_armors)
```
lineargraph: As per my previous suggestion:
```suggestion
        armorNames = data.trophy_armors.toSet()
```

src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
```diff
@@ -151,10 +157,25 @@ object SeaCreatureTracker {
...
     private fun isEnabled() = LorenzUtils.inSkyBlock &&
-        FishingAPI.lastActiveFishingTime.passedSince() < 10.minutes && config.enabled
+        FishingAPI.lastActiveFishingTime.passedSince() < 10.minutes && config.enabled && !isTrophyFishing()
+
+    private fun isTrophyFishing(): Boolean {
+        val boots = InventoryUtils.getBoots()?.getInternalName()?.asString()
+        val leggings = InventoryUtils.getLeggings()?.getInternalName()?.asString()
+        val chestplate = InventoryUtils.getChestplate()?.getInternalName()?.asString()
+        val helmet = InventoryUtils.getHelmet()?.getInternalName()?.asString()
+        return armorNames.contains(helmet) && armorNames.contains(chestplate) && armorNames.contains(leggings) && armorNames.contains(boots)
```
lineargraph:
```suggestion
        return armorNames.containsAll(InventoryUtils.getArmor().map { it?.getInternalName() })
```

lineargraph (changes requested):

src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
lineargraph: *should*

src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
```diff
@@ -151,10 +157,20 @@ object SeaCreatureTracker {
...
     fun resetCommand(args: Array<String>) {
         tracker.resetCommand(args, "shresetseacreaturetracker")
     }
 
     private fun isEnabled() = LorenzUtils.inSkyBlock &&
-        FishingAPI.lastActiveFishingTime.passedSince() < 10.minutes && config.enabled
+        FishingAPI.lastActiveFishingTime.passedSince() < 10.minutes && config.enabled && !isTrophyFishing()
+
+    private fun isTrophyFishing(): Boolean {
+        return trophyArmorNames.containsAll(InventoryUtils.getArmor().map { it?.getInternalName() })
```
hannibal002: Isnt simply one regex of `(bronze|silver|gold|diamond)_hunter_(helmet|chestplate|leggings|boots)` better?

hannibal002 (changes requested):
I dont want to break the repo work done here, but if the only goal is to check if the piece is a hunter armor part, why not simply use one regex instead?

src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
CalMWolfs: I like this much better and this can also be in the repo through a repo pattern

lineargraph:
> yeah that does sound better, we don't have a armor change event do we

Also we do, i think.
Correction: we have the EntityEquipmentUpdateEvent for *other* mobs, not the local player.

src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
jani270: Even simpler would be to just check the lore? 
image
If it says Peace Treaty (4/4)

Thunderblade73: this value bugs out some times so I wouldn't use it. Also lore checking is expensive

martimavocado: i don't think a pattern is super needed since hypixel doesn't usually change item ids, but ill do this

src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
```diff
@@ -156,5 +164,12 @@ object SeaCreatureTracker {
     }
 
     private fun isEnabled() = LorenzUtils.inSkyBlock &&
-        FishingAPI.lastActiveFishingTime.passedSince() < 10.minutes && config.enabled
+        FishingAPI.lastActiveFishingTime.passedSince() < 10.minutes && config.enabled && !isTrophyFishing()
+
+    private fun isTrophyFishing(): Boolean {
+        val armorInternalNames = InventoryUtils.getArmor().map { it?.getInternalName()?.asString() ?: return false }
```
hannibal002: The reutn false part can go: matches() returns false when the parameter is empty.

hannibal002: Also you can remove `.map {` and move it in `.all {`

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/ItemsJson.java
```diff
@@ -11,4 +12,7 @@ public class ItemsJson {
 
     @Expose
     public Map<String, Integer> crimson_tiers;
+
+    @Expose
+    public List<NEUInternalName> trophy_fishing_armors;
```
hannibal002: This can go now

src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
```diff
@@ -156,5 +164,12 @@ object SeaCreatureTracker {
     }
 
     private fun isEnabled() = LorenzUtils.inSkyBlock &&
-        FishingAPI.lastActiveFishingTime.passedSince() < 10.minutes && config.enabled
+        FishingAPI.lastActiveFishingTime.passedSince() < 10.minutes && config.enabled && !isTrophyFishing()
```
hannibal002: Do not check the player armor in every frame. Use a cached boolean in here, that you update e.g every second in a tick event or something.

hannibal002 (changes requested):
Works fine!
Some lines can go and the armor check logic should be cached.

hannibal002 (approved):

PR #881, opened 2024-01-06
Feature: `/shcopychestname` (+ customizable keybind in config)

RayDeeUx:
# CLOSED. SEE #932
needed this when i wanted a regex for finding the chestname for player trading menu. ended up improvising and using NEU's basis of `.startsWith("You  ")` instead but i'm adding this anyway to avoid having to improvise to that extent again.

PR #932, opened 2024-01-17
Feature: Player Inventory/Chest Debug

RayDeeUx:
## What

(partially supersedes #881)

## Changelog New Features
+ Added additional inventory debug commands. - Erymanthus


src/main/java/at/hannibal2/skyhanni/test/TestCopyChestData.kt
```diff
@@ -0,0 +1,80 @@
+@file:Suppress("SENSELESS_COMPARISON") // intellij does a little trolling and assumes all ItemStacks will never be null. this suppresses such warnings.
```
lineargraph: Always apply suppressions as locally as possible. I also haven't had that issue with intellij before. Maybe check that warning once again.

src/main/java/at/hannibal2/skyhanni/test/TestCopyChestData.kt
```diff
@@ -0,0 +1,80 @@
...
+        } else if (config.copyPlayerInventory.isKeyHeld()) {
+            copySlotsData(inventory = if (config.includeArmor) mc.thePlayer.inventory.mainInventory + mc.thePlayer.inventory.armorInventory else mc.thePlayer.inventory.mainInventory)
+            return
+        } else if (InventoryUtils.openInventoryName().isNotEmpty()) {
+            OSUtils.copyToClipboard(InventoryUtils.openInventoryName())
+            LorenzUtils.chat("Chest name copied to clipboard.")
+            return
+        }
+    }
+    private fun copySlotsData(chest: List<Slot> = emptyList(), inventory: Array<ItemStack> = emptyArray()) {
```
lineargraph: I don't like this method having two disjoint behaviours depending on the arguments. Maybe split this method up into more methods (like one for the header, one for just listing item stacks, and one each for listing the inventory and one for the chests)

nea89o (changes requested):

src/main/java/at/hannibal2/skyhanni/test/TestCopyChestData.kt
RayDeeUx: >Maybe check that warning once again.

already did, and that's how i get hard crashes due to `NPE`s (yep, even skipped skyhanni's usual "click to copy error" chat message)

RayDeeUx: proof: 

i could just do `?.` calls, but i added a config option to include data on null slots. may as well make use of the `== null` check at that point imo

lineargraph: The issue in this case is that you use an `Array<ItemStack>` as type, when really it should be a `Array<ItemStack?>`. The lint works correctly, your types are just wrong. Suppressing that lint here just means that you will not get proper type checking if you access the stack in another place. We should not allow implicit nulls anywhere.

RayDeeUx: 👍

src/main/java/at/hannibal2/skyhanni/test/TestCopyChestData.kt
RayDeeUx: split into two funcs, one test run and then commit

src/main/java/at/hannibal2/skyhanni/test/TestCopyChestData.kt
```diff
@@ -0,0 +1,80 @@
...
+                if ((stack.displayName.isNotEmpty() && stack.displayName.isNotBlank()) || config.includeUnnamedItems) {
+                    copyList.add("slot index: '${slot.slotIndex}'")
+                    copyList.addAll(stack.getStackInfo())
+                }
+            }
+        } else if (inventory.isNotEmpty()) {
+            copyList.addAll(listOf<String>("your inventory is below.", "", "your hotbar:", ""))
+            for ((i, stack) in inventory.withIndex()) {
+                if (i - 1 == 8 && stack != null) copyList.addAll(listOf("note: hotbar ends here", "the rest of your inventory:", "", ""))
+                if (i - 1 == 35 && config.includeArmor && stack != null) copyList.addAll(listOf("note: the rest of your inventory data ends here", "your armor:", "", ""))
```
lineargraph: Please use meaningful named constants instead:
```kt
val HOTBAR_SIZE = 9
val INVENTORY_SIZE = 27
```



```suggestion
                if (i == HOTBAR_SIZE) copyList.addAll(listOf("note: hotbar ends here", "the rest of your inventory:", "", ""))
                if (i == HOTBAR_SIZE + INVENTORY_SIZE && config.includeArmor) copyList.addAll(listOf("note: the rest of your inventory data ends here", "your armor:", "", ""))
```

I also removed the `stack != null` checks, since these would otherwise omit the note if the first slot of the respective inventory section is empty (irrespective of later slots being filled)

src/main/java/at/hannibal2/skyhanni/test/TestCopyChestData.kt
lineargraph:
```suggestion
    private fun copySlotsData(chest: List<Slot> = emptyList(), inventory: Array<ItemStack?> = emptyArray()) {
```

The player inventory contains nullable stacks.

nea89o (changes requested):

src/main/java/at/hannibal2/skyhanni/test/TestCopyChestData.kt
RayDeeUx: done

src/main/java/at/hannibal2/skyhanni/test/TestCopyChestData.kt
```diff
@@ -0,0 +1,81 @@
...
+    private fun copyInventoryData(inventory: Array<ItemStack?>) {
+        val copyList = mutableListOf<String>("relevant config:", "includeNullSlots: ${config.includeNullSlots}", "includeUnnamedItems: ${config.includeUnnamedItems}", "includeArmor: ${config.includeArmor}", "", "your inventory is below.", "", "your hotbar:", "")
+        for ((i, stack) in inventory.withIndex()) {
+            if (i == HOTBAR_SIZE) copyList.addAll(listOf("note: hotbar ends here", "the rest of your inventory:", "", ""))
+            if (i == HOTBAR_SIZE + INVENTORY_SIZE && config.includeArmor) copyList.addAll(listOf("note: the rest of your inventory data ends here", "your armor:", "", ""))
+            if (stack == null) {
+                if (config.includeNullSlots) copyList.addAll(listOf("(there is nothing inside slot $i; it is null)", "", ""))
+                continue
+            }
+            if ((stack.displayName.isNotEmpty() && stack.displayName.isNotBlank()) || config.includeUnnamedItems) {
```
lineargraph:
```suggestion
            if (stack.displayName.isNotBlank() || config.includeUnnamedItems) {
```

nea89o (commented):
This is still very confusing code to me, but i cant find anything else to critique.

src/main/java/at/hannibal2/skyhanni/config/features/dev/DebugConfig.java
```diff
@@ -101,4 +91,44 @@ public class DebugConfig {
...
+    @ConfigEditorKeybind(defaultKey = Keyboard.KEY_NONE)
+    public int copyItemData = Keyboard.KEY_NONE;
+
+    @Expose
+    @ConfigOption(name = "Copy Compressed Item Data", desc = "Copies compressed item NBT data on key press in a GUI to clipboard.")
+    @ConfigEditorKeybind(defaultKey = Keyboard.KEY_NONE)
+    public int copyItemDataCompressed = Keyboard.KEY_NONE;
+
+    @Expose
+    @ConfigOption(name = "Copy Chest Name", desc = "Copies the chest name on key press in a GUI to clipboard.")
```
hannibal002: Best create a new config class for all menu/inventory stuff here

hannibal002 (commented):
no real review yet, just a comment

src/main/java/at/hannibal2/skyhanni/test/TestCopyChestData.kt
```diff
@@ -0,0 +1,82 @@
...
+import at.hannibal2.skyhanni.utils.OSUtils
+import net.minecraft.client.Minecraft
+import net.minecraft.client.gui.inventory.GuiChest
+import net.minecraft.client.gui.inventory.GuiContainer
+import net.minecraft.inventory.Slot
+import net.minecraft.item.ItemStack
+import net.minecraftforge.client.event.GuiScreenEvent
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+object TestCopyChestData {
```
Thunderblade73, line 18: Remove the Test from the name. It is redundant since it is in the test package.

src/main/java/at/hannibal2/skyhanni/test/TestCopyChestData.kt
```diff
@@ -0,0 +1,82 @@
...
+    private val config get() = SkyHanniMod.feature.dev.debug.copyInventory
+    private val mc = Minecraft.getMinecraft()
+    private const val HOTBAR_SIZE = 9
+    private const val INVENTORY_SIZE = 27
+    private const val DEBUG_SUFFIX: String = "data copied into the clipboard! §lMake sure to save it into a .txt file; these tend to get quite long."
+    @SubscribeEvent
+    fun onKeybind(event: GuiScreenEvent.KeyboardInputEvent.Post) {
```
Thunderblade73, line 25: make this into a single when. Only thing that should not be that when is the Gui Container check  but that as the first thing.

src/main/java/at/hannibal2/skyhanni/test/TestCopyChestData.kt
```diff
@@ -0,0 +1,82 @@
...
+        } else if (config.copyPlayerInventory.isKeyHeld()) {
+            copyInventoryData(inventory = if (config.includeArmor) mc.thePlayer.inventory.mainInventory + mc.thePlayer.inventory.armorInventory else mc.thePlayer.inventory.mainInventory)
+            return
+        } else if (InventoryUtils.openInventoryName().isNotEmpty()) {
+            OSUtils.copyToClipboard(InventoryUtils.openInventoryName())
+            ChatUtils.chat("Chest name copied to clipboard.")
+            return
+        }
+    }
+    private fun copyInventoryData(inventory: Array<ItemStack?>) {
```
Thunderblade73, line 40: this function is so overcomplicated. My idea split the array into 3 sub list. Make a sepreate function that handles the item printing. Also make it contsistent. Eg. Slot Number : Item Info. This is mainly the issue with the null display.

src/main/java/at/hannibal2/skyhanni/test/TestCopyChestData.kt
```diff
@@ -0,0 +1,82 @@
...
+                copyList.add("slot index: '$i'")
+                copyList.addAll(stack.getStackInfo())
+            }
+        }
+        OSUtils.copyToClipboard(copyList.joinToString("\n"))
+        ChatUtils.chat("Inventory $DEBUG_SUFFIX")
+    }
+    private fun copyChestData(chest: List<Slot>) {
+        val copyList = mutableListOf<String>("relevant config:", "includeNullSlots: ${config.includeNullSlots}", "includeUnnamedItems: ${config.includeUnnamedItems}", "includeArmor: ${config.includeArmor}", "", "chest name: '${InventoryUtils.openInventoryName()}'", "")
+        copyList.addAll(listOf<String>())
```
Thunderblade73, line 59: wtf is this. This so cursed. Also does nothing. Since adding a empty list to an list does nothing.

src/main/java/at/hannibal2/skyhanni/test/TestCopyChestData.kt
```diff
@@ -0,0 +1,82 @@
...
+            }
+            if (stack.displayName.isNotBlank() || config.includeUnnamedItems) {
+                copyList.add("slot index: '$i'")
+                copyList.addAll(stack.getStackInfo())
+            }
+        }
+        OSUtils.copyToClipboard(copyList.joinToString("\n"))
+        ChatUtils.chat("Inventory $DEBUG_SUFFIX")
+    }
+    private fun copyChestData(chest: List<Slot>) {
```
Thunderblade73, line 57: use the item print function here as well, see comment on copyInventoryData.

src/main/java/at/hannibal2/skyhanni/test/TestCopyChestData.kt
```diff
@@ -0,0 +1,82 @@
...
+            if (stack in mc.thePlayer.inventory.mainInventory) break
+            if ((stack.displayName.isNotEmpty() && stack.displayName.isNotBlank()) || config.includeUnnamedItems) {
+                copyList.add("slot index: '${slot.slotIndex}'")
+                copyList.addAll(stack.getStackInfo())
+            }
+        }
+        OSUtils.copyToClipboard(copyList.joinToString("\n"))
+        ChatUtils.chat("Chest $DEBUG_SUFFIX")
+    }
+    private fun ItemStack.getStackInfo(): List<String> {
```
Thunderblade73, line 75: make the format more json like, eg. brackets. Not necessarily but makes it more readable in my opinion. Also does not need to be a full json.

src/main/java/at/hannibal2/skyhanni/test/command/CopyItemCommand.kt
```diff
@@ -36,6 +36,14 @@ object CopyItemCommand {
...
 
     fun copyItemToClipboard(itemStack: ItemStack) {
+        val resultList = grabItemData(itemStack)
+
+        val string = resultList.joinToString("\n")
+        OSUtils.copyToClipboard(string)
+        ChatUtils.chat("Item info copied into the clipboard!")
+    }
+
+    fun grabItemData(itemStack: ItemStack): List<String> {
```
Thunderblade73, line 46: make this a buildList.

Thunderblade73 (changes requested):

hannibal002:
old

PR #934, opened 2024-01-17
Feature: Taming 60 Cost Calculator

RayDeeUx:
## What

<details>
<summary>Images</summary>

</details>

## Changelog New Features
+ Taming 60 Cost Calculator. - Erymanthus


RayDeeUx:
todo: account for taming 60

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,86 @@
...
+                    "MYTHIC" -> "5"
+                    else -> ""
+                }
+                val origPetString = group("pet") ?: ""
+                val pet = origPetString.uppercase().replace(" ", "_").removePrefix("FROST_")
+                val petPrice = "$pet;$tier".asInternalName().getPriceOrNull() ?: continue@loop
+                totalCost += petPrice
+                updateList.add(Renderable.clickAndHover(
+                    text = " §7- ${group("fullThing")}§7: §6${petPrice.addSeparators()} coins",
+                    tips = listOf("§aClick to run §e/ahsearch [Lvl 1] $origPetString §ato find this pet on the Auction House.", "§aNotes: §eMake sure to set the rarity filter to ${group("tierColorCodes")}$origTierString §eon your own! §cBooster Cookie required!"),
```
jani270: Both commands can be shortened to /ahs, and I would do this instead: `ahs ] $origPetString`.
Sometimes, there are no Level 1 pets on the AH. I would also recommend shortening the tooltip in various ways. 
The note for the rarity is a bit incorrect, as you can give George a Legendary Pet instead of an Epic one.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,86 @@
...
+                    onClick = {
+                        LorenzUtils.sendCommandToServer("ahsearch [Lvl 1] $origPetString")
+                    }
+                ))
+            }
+        }
+        updateList.addAll(listOf<Renderable>(Renderable.string("§dTotal cost §7(§6Lowest BIN§7): §6${totalCost.addSeparators()} coins"), Renderable.string("§c§lDisclaimer:§r§c Some pets are not available on the Auction House.")))
```
jani270: Instead of the disclaimer, consider opening the new recipe if it isn't on the auction house.

jani270 (changes requested):

src/main/java/at/hannibal2/skyhanni/config/features/misc/pets/GeorgeConfig.java
```diff
@@ -0,0 +1,16 @@
...
+import io.github.moulberry.moulconfig.annotations.ConfigOption;
+
+public class GeorgeConfig {
+    @Expose
+    @ConfigOption(name = "Enabled", desc = "Show the currently lowest prices of the pets you need to increase your Taming level cap.")
+    @ConfigEditorBoolean
+    public boolean enabled = false;
+
+    @Expose
+    public Position pos = new Position(100, 90, false, true);
```
hannibal002: Please rename to `position`

src/main/java/at/hannibal2/skyhanni/config/features/misc/pets/GeorgeConfig.java
```diff
@@ -0,0 +1,16 @@
+package at.hannibal2.skyhanni.config.features.misc.pets;
+
+import at.hannibal2.skyhanni.config.core.config.Position;
+import com.google.gson.annotations.Expose;
+import io.github.moulberry.moulconfig.annotations.ConfigEditorBoolean;
+import io.github.moulberry.moulconfig.annotations.ConfigOption;
+
+public class GeorgeConfig {
+    @Expose
+    @ConfigOption(name = "Enabled", desc = "Show the currently lowest prices of the pets you need to increase your Taming level cap.")
```
hannibal002: please add to buy
..you need (to buy) to increase..

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,86 @@
...
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+class GeorgeDisplay {
+
+    private val config get() = SkyHanniMod.feature.misc.pets.george
+
+    private val ROW_OFFSET: Int = 9 * 4
+    private val INDEX_OFFSET: Int = 6 - 1
+
+    private val patternGroup = RepoPattern.group("george.tamingcap")
```
hannibal002, line 36: Why a group for one element?

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,86 @@
...
+    private val neededPetPattern by patternGroup.pattern(
+        "needed.pet.loreline",
+        " +(?<fullThing>(?<tierColorCodes>§.)*(?<tier>(?:[uU][nN])?[cC][oO][mM][mM][oO][nN]|[rR][aA][rR][eE]|[eE][pP][iI][cC]|[lL][eE][gG][eE][nN][dD][aA][rR][yY]|[mM][yY][tT][hH][iI][cC]|[dD][iI][vV][iI][nN][eE]|(?:[vV][eE][rR][yY] )?[sS][pP][eE][cC][iI][aA][lL]|[uU][lL][tT][iI][mM][aA][tT][eE]|[sS][uU][pP][rR][eE][mM][eE]|[aA][dD][mM][iI][nN]) (?<pet>[\\S ]+))"
```
hannibal002: Please consider using a readable regex.
This means no whatever this is: `[sS][pP][eE][cC][iI][aA][lL]|[uU][lL][tT][iI][mM][aA][tT][eE]|[sS][uU][pP][rR][eE][mM][eE]|[aA][dD][mM][iI][nN]`
Also  `(?i)` exists.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,86 @@
...
+    )
+
+    private val display = mutableListOf<Renderable>()
+
+    @SubscribeEvent
+    fun onInventoryOpen(event: InventoryOpenEvent) {
+        if (!LorenzUtils.inSkyBlock) return
```
hannibal002: Please move this inside isEnabled()

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,86 @@
...
+import at.hannibal2.skyhanni.utils.StringUtils.matchMatcher
+import at.hannibal2.skyhanni.utils.renderables.Renderable
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+class GeorgeDisplay {
+
+    private val config get() = SkyHanniMod.feature.misc.pets.george
+
+    private val ROW_OFFSET: Int = 9 * 4
```
hannibal002: Is there a reason you split those values up and calculate them? just save the exact slot id instead

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,86 @@
...
+    fun onInventoryOpen(event: InventoryOpenEvent) {
+        if (!LorenzUtils.inSkyBlock) return
+        if (!isEnabled()) return
+        if (event.inventoryName != "Offer Pets") return
+        val stack = event.inventoryItems[ROW_OFFSET + INDEX_OFFSET] ?: return
+        if (stack.cleanName() != "+1 Taming Level Cap") return
+        val updateList: MutableList<Renderable> = mutableListOf<Renderable>(Renderable.string("§d§lTaming 60 Cost:"))
+        var totalCost: Double = 0.0
+        loop@ for (line in stack.getLore()) {
+            neededPetPattern.matchMatcher(line) {
```
hannibal002, line 72: Please extract the check inside the for loop into a function

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,86 @@
...
+                val origTierString = group("tier")
+                val tier = when (origTierString.uppercase()) {
```
hannibal002: Consider using a new function for this when block

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,86 @@
...
+                        LorenzUtils.sendCommandToServer("ahsearch [Lvl 1] $origPetString")
+                    }
+                ))
+            }
+        }
+        updateList.addAll(listOf<Renderable>(Renderable.string("§dTotal cost §7(§6Lowest BIN§7): §6${totalCost.addSeparators()} coins"), Renderable.string("§c§lDisclaimer:§r§c Some pets are not available on the Auction House.")))
+        updateList.update()
+    }
+
+    private fun List<Renderable>.update() {
```
hannibal002: No need for this function. Consider building the list instead in a build list function.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,86 @@
...
+        display.clear()
+        display.addAll(this)
+    }
+
+    @SubscribeEvent
+    fun onRenderOverlay(event: GuiRenderEvent.ChestGuiOverlayRenderEvent) {
+        if (!LorenzUtils.inSkyBlock) return
```
hannibal002: remove (see comment above)

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,86 @@
...
+        if (!isEnabled()) return
+        if (InventoryUtils.openInventoryName() != "Offer Pets") return
+        config.pos.renderRenderables(display, posLabel = "Taming 60 Progress")
```
hannibal002: Please use the same GUI label name as the config toggle name

src/main/java/at/hannibal2/skyhanni/config/features/misc/pets/GeorgeConfig.java
```diff
@@ -0,0 +1,16 @@
+package at.hannibal2.skyhanni.config.features.misc.pets;
+
+import at.hannibal2.skyhanni.config.core.config.Position;
+import com.google.gson.annotations.Expose;
+import io.github.moulberry.moulconfig.annotations.ConfigEditorBoolean;
+import io.github.moulberry.moulconfig.annotations.ConfigOption;
+
+public class GeorgeConfig {
```
hannibal002, line 8: Do we really need a new sub category for one setting? Are you planning to add more George stuff soon? If not, please keep it in pet config.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: shortened + rephrased:
```kt
tips = listOf(
    "§aClick to run §e/ahs ] $origPetString §ato find it on the Auction House.",
    "§aNotes: §eSet the rarity filter yourself. §cBooster Cookie required!"
),
```

src/main/java/at/hannibal2/skyhanni/config/features/misc/pets/GeorgeConfig.java
RayDeeUx: done

src/main/java/at/hannibal2/skyhanni/config/features/misc/pets/GeorgeConfig.java
RayDeeUx: done

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: done

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: > Is there a reason

given hypixel's way of moving things around in menus without mod developer consent it's easier to split the slot ID into individual math expressions. ie: `9 * 4` to skip the first four rows, and `6 - 1` because the spawn egg is on the sixth slot of that row (and minus one because of "starts at zero" index)

src/main/java/at/hannibal2/skyhanni/config/features/misc/pets/GeorgeConfig.java
RayDeeUx: outdated, per @jani270's comments above (as well as similar feedback from other pinknames in the relevant suggestion post)

confirmed: adding one more toggle for searching for other rarities.

maybe: a toggle for a separate section of pets that dont have prices and need to be crafted?

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: done

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: `(?i) +(?<fullThing>(?<tierColorCodes>§.)*(?<tier>(?:un)?common|rare|epic|legendary|mythic|divine|(?:very )?special|ultimate|supreme|admin) (?<pet>[\S ]+))`

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: better safe than sorry; you never know when hypixel's gonna mess us up by making dynamic item names or something worse like requiring a second lore line to check

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: was gonna do that, but here's the cinch:

george's whole schtick is to request pets that can't be obtained from crafting. so i settled for a compromise instead in two parts:
```kt
private val notAHOrCraft: List<String> = listOf<String>("ENDERMAN", "BLACK_CAT", "SLUG", "SPIRIT", "GIRAFFE", "JELLYFISH", "BAL", "BABY_YETI", "RIFT_FERRET", "WISP")
```
```kt
val command: String = if (pet !in notAHOrCraft) "viewrecipe" else "wiki"
val whatCommandDoes: String = if (pet !in notAHOrCraft) "its crafting recipe" else "how to obtain it"
updateList.add(Renderable.clickAndHover(
    text = " §7- $displayPetString§7: §eNot on AH; ${if (pet !in notAHOrCraft) "try crafting it" else "find its Wiki article here"}.",
    tips = listOf("§4Click to run §e$command $pet §4to view $whatCommandDoes."),
    onClick = { LorenzUtils.sendCommandToServer("$command $pet") }
))
```
Still gotta rephrase the disclaimer though, give me a sec.

RayDeeUx: (the idea of the list of strings in the first codeblock is to account for the very low likelihood that George requests a craftable pet)

RayDeeUx:
`.getPriceOrNull() ?: Double.MAX_VALUE` is used to reduce the LOC necessary to determine cheapest pet

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: actually nevermind, changed the disclaimer to mention not including kat costs

also applied blanket "`/wiki PET`" treatment given george's whole schtick (see earlier)

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: outdated per listbuilding function

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: gonna mark this convo as resolved for now

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: alright im going to hardcode the item slot but if the slot ID changes, don't egg my house

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,117 @@
...
+import at.hannibal2.skyhanni.utils.StringUtils.matchMatcher
+import at.hannibal2.skyhanni.utils.renderables.Renderable
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+class GeorgeDisplay {
+
+    private val config get() = SkyHanniMod.feature.misc.pets.george
+
+    private val SEPARATOR: String = ";"
```
hannibal002: Does not need `: String`

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    private val patternGroup = RepoPattern.group("george.tamingcap")
```
hannibal002, line 36: remove group

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,117 @@
...
+    fun onInventoryOpen(event: InventoryOpenEvent) {
+        if (!isEnabled()) return
+        if (event.inventoryName != "Offer Pets") return
+        val stack = event.inventoryItems[41] ?: return
+        if (stack.cleanName() != "+1 Taming Level Cap") return
+        display.clear()
+        display.addAll(listBuilding(stack.getLore()))
+    }
+
+    private fun listBuilding(lore: List<String>): MutableList<Renderable> {
```
hannibal002: Split the logic up in multiple functions

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,117 @@
...
+        val updateList: MutableList<Renderable> = mutableListOf<Renderable>(Renderable.string("§d§lTaming 60 Cost: §r§d(${if (config.otherRarities) "cheapest" else "exact"} rarity)"))
```
hannibal002: The line is way too long

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,117 @@
...
+        var totalCost: Double = 0.0
```
hannibal002: Unnecessary `: Double`

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
hannibal002: Unnecessary `<Renderable>`

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,117 @@
...
+        for (line in lore) {
+            neededPetPattern.matchMatcher(line) {
+                val origTierString = group("tier") ?: ""
+                val tier = RarityToTier.entries.find { it.rarityInternal == origTierString.uppercase() }?.tier ?: -1
+                val origPetString = group("pet") ?: ""
+                val pet = origPetString.uppercase().replace(" ", "_").removePrefix("FROST_")
+                val petPrices: MutableList<Double> = mutableListOf<Double>()
```
hannibal002: <Double>

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,117 @@
...
+                if (config.otherRarities || petPriceOne == -1.0) {
+                    val lowerTier = "$pet$SEPARATOR${tier - 1}".asInternalName().getPriceOrNull() ?: Double.MAX_VALUE
+                    petPrices.add(lowerTier)
+                    if (tier != 5) {
+                        val higherTier = "$pet$SEPARATOR${tier + 1}".asInternalName().getPriceOrNull() ?: Double.MAX_VALUE
+                        petPrices.add(higherTier)
+                    }
+                }
+                val petPrice = petPrices.min()
+                val tierUsed = when (petPrices.indexOf(petPrice)) {
```
hannibal002: Things like this can be a standalone function

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    private val patternGroup = RepoPattern.group("george.tamingcap")
+    private val neededPetPattern by patternGroup.pattern(
+        "needed.pet.loreline",
+        "(?i) +(?<fullThing>(?<tierColorCodes>§.)*(?<tier>(?:un)?common|rare|epic|legendary|mythic|divine|(?:very )?special|ultimate|supreme|admin) (?<pet>[\\S ]+))"
+    )
+
+    private val display = mutableListOf<Renderable>()
+
+    private enum class RarityToTier(
```
hannibal002: Delete and use LorenzRarity instead

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @SubscribeEvent
+    fun onInventoryOpen(event: InventoryOpenEvent) {
+        if (!isEnabled()) return
+        if (event.inventoryName != "Offer Pets") return
+        val stack = event.inventoryItems[41] ?: return
+        if (stack.cleanName() != "+1 Taming Level Cap") return
+        display.clear()
+        display.addAll(listBuilding(stack.getLore()))
```
hannibal002: make display var and immutable list.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    private fun listBuilding(lore: List<String>): MutableList<Renderable> {
+        val updateList: MutableList<Renderable> = mutableListOf<Renderable>(Renderable.string("§d§lTaming 60 Cost: §r§d(${if (config.otherRarities) "cheapest" else "exact"} rarity)"))
+        var totalCost: Double = 0.0
+        for (line in lore) {
+            neededPetPattern.matchMatcher(line) {
+                val origTierString = group("tier") ?: ""
```
hannibal002: Can tier be empty?

hannibal002 (changes requested):
Too much stuff to do, ill mark this as a draft.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: done

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: done

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: rm'd

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: done

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: unclear

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,119 @@
...
+
+class GeorgeDisplay {
+
+    private val config get() = SkyHanniMod.feature.misc.pets.george
+
+    private val SEPARATOR = ";"
+
+    private val neededPetPattern by RepoPattern.pattern(
+        "george.tamingcap.needed.pet.loreline",
+        "(?i) +(?<fullThing>(?<tierColorCodes>§.)*(?<tier>(?:un)?common|rare|epic|legendary|mythic|divine|(?:very )?special|ultimate|supreme|admin) (?<pet>[\\S ]+))"
```
Thunderblade73: Please use a ```${enumJoinToPattern<LorenzRarity>{rawName.lowercase()}}``` instead of "common|rare|epic|legendary|mythic|divine|(?:very )?special|ultimate|supreme|admin"

RayDeeUx: done

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: done

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: done

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: done

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
hannibal002: < Double >

this can be removed. some formatting thing thought its a cool idea to simply remove it from this description block lol

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,113 @@
...
+        "george.tamingcap.needed.pet.loreline",
+        "(?i) +(?<fullThing>(?<tierColorCodes>§.)*(?<tier>${enumJoinToPattern<LorenzRarity>{it.rawName.lowercase()}}) (?<pet>[\\S ]+))"
+    )
+
+    private var display = listOf<Renderable>()
+
+    @SubscribeEvent
+    fun onInventoryOpen(event: InventoryOpenEvent) {
+        if (!isEnabled()) return
+        if (event.inventoryName != "Offer Pets") return
```
Thunderblade73: please make a repoPattern for "Offer Pets"

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,113 @@
...
+        val stack = event.inventoryItems[41] ?: return
```
Thunderblade73: Declare the constant 41 as a real constant with name. There should be no reason to guess what 41 means.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,113 @@
...
+        if (stack.cleanName() != "+1 Taming Level Cap") return
```
Thunderblade73: please a repoPattern for "+1 Taming Level Cap"

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+        updateList.add(Renderable.string("§dTotal cost §7(§6Lowest BIN§7): §6${totalCost.addSeparators()} coins"))
+        if (config.otherRarities) updateList.add(Renderable.string("§c§lDisclaimer:§r§c Total does not include costs to upgrade via Kat."))
+        return updateList
+    }
+
+    private fun findCheapestTier(pet: String, originalTier: Int): Pair<Double, Int> {
+        val petPriceOne = "$pet$SEPARATOR$originalTier".getPetPrice()
+        val petPrices: MutableList<Double> = mutableListOf(petPriceOne)
```
Thunderblade73:
```suggestion
        val petPrices = mutableListOf(petPriceOne)
```

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,113 @@
...
+                }
+            }
+        }
+        updateList.add(Renderable.string("§dTotal cost §7(§6Lowest BIN§7): §6${totalCost.addSeparators()} coins"))
+        if (config.otherRarities) updateList.add(Renderable.string("§c§lDisclaimer:§r§c Total does not include costs to upgrade via Kat."))
+        return updateList
+    }
+
+    private fun findCheapestTier(pet: String, originalTier: Int): Pair<Double, Int> {
+        val petPriceOne = "$pet$SEPARATOR$originalTier".getPetPrice()
```
Thunderblade73: "$pet$SEPARATOR$originalTier" please extract that as a function. Those strings are really hard to read, and at 3 places.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,113 @@
...
+    @SubscribeEvent
+    fun onInventoryOpen(event: InventoryOpenEvent) {
+        if (!isEnabled()) return
+        if (event.inventoryName != "Offer Pets") return
+        val stack = event.inventoryItems[41] ?: return
+        if (stack.cleanName() != "+1 Taming Level Cap") return
+        display = listBuilding(stack.getLore())
+    }
+
+    private fun listBuilding(lore: List<String>): MutableList<Renderable> {
```
Thunderblade73: Use a build list please
```suggestion
    private fun listBuilding(lore: List<String>) = buildList<Renderable>{
```
You need to replace ```updateList``` with ```this``` and then most of it will simply work

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,113 @@
...
+                if (config.otherRarities) "cheapest" else "exact"
+            } rarity)")
+        )
+        var totalCost = 0.0
+        for (line in lore) {
+            neededPetPattern.matchMatcher(line) {
+                val origTierString = group("tier")
+                val origPetString = group("pet")
+                val pet = origPetString.uppercase().replace(" ", "_").removePrefix("FROST_")
+                val originalTier = LorenzRarity.entries.find { it.name == origTierString.uppercase() }!!.id
```
Thunderblade73: There is already a function for this. And very important NEVER EVER USE "!!" !!!!
```suggestion
                val originalTier = LorenzRarity.getByName(origTierString.uppercase().replace(" ", "_") )?.id
```

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,113 @@
...
+    private fun findCheapestTier(pet: String, originalTier: Int): Pair<Double, Int> {
+        val petPriceOne = "$pet$SEPARATOR$originalTier".getPetPrice()
+        val petPrices: MutableList<Double> = mutableListOf(petPriceOne)
+        if (config.otherRarities || petPriceOne == -1.0) {
+            petPrices.add("$pet$SEPARATOR${originalTier - 1}".getPetPrice(otherRarity = true))
+            if (originalTier != 5) petPrices.add("$pet$SEPARATOR${originalTier + 1}".getPetPrice(otherRarity = true))
+        }
+        val petPrice = petPrices.min()
+        val cheapestTier = petPrices.cheapestTierIndex(petPrice, originalTier)
+        return Pair(petPrice, cheapestTier)
```
Thunderblade73: replace it with
```suggestion
return petPrices.withIndex().minBy { it.value }
```
and delete the cheapestTierIndex function.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,113 @@
...
+        for (line in lore) {
+            neededPetPattern.matchMatcher(line) {
+                val origTierString = group("tier")
+                val origPetString = group("pet")
+                val pet = origPetString.uppercase().replace(" ", "_").removePrefix("FROST_")
+                val originalTier = LorenzRarity.entries.find { it.name == origTierString.uppercase() }!!.id
+                val (petPrice, cheapestTier) = findCheapestTier(pet, originalTier)
+                val displayPetString =
+                    if (cheapestTier == originalTier) group("fullThing")
+                    else "${LorenzRarity.entries.find { it.id == cheapestTier }!!.formattedName} $origPetString"
```
Thunderblade73:
```suggestion
                    else "${LorenzRarity.getById(cheapestTier )?.formattedName} $origPetString"
```

Thunderblade73 (changes requested):
Ery, this PR has horrible code quality. Sure there simply somethings that you didn't know, that's ok. But I would like to see those mistakes never again from you (in this amount at least).
A good tip is to look around in the codebase for similar problems or simply at the source (eg. LorenzRarity). 
One thing I can't wrap my head around is why you used "!!" it is literally in the contributor guidelines that we are disallowed to use it. And if you want to know why, it is because it throw an exception and it is not clear when reading it where the code would end up next after the exception (worst case it will be catchend be the event handler). You can always use "?." it does 99% of the time the same without an exception that needs to be handled. And if you really want to exit the code at that point at "?: return" simple as that and it is understandable.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
lineargraph: Maybe also declare it as `9 * 4 + 5`?

hannibal002: ery did that already. I did not like it, since I still don't know what item is in that slot.

RayDeeUx: vetoed by hanni

hannibal002: You can still add a comment above the line what is going on here.

lineargraph: I think breaking up the constant into rows and coloumns, as well as using a const val instead of a comment would be better here than a comment. `const val TAMING_LEVEL_CAP_INFO_SLOT = 4 * 9 + 5`

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: did both steps; the HUD simply didn't show up. sticking with `MutableList` declaration

Thunderblade73: then you simply did something wrong. But I will give it a go

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: broke the rarity display in the HUD, reverted

Thunderblade73: the order might be wrong. It is index, value but I think you have value, index

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
Thunderblade73: I looked at your code and yes you miss used the buildList

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: it's not a matter of the L to R order of value declarations, it's that the HUD is now showing me ~53M coins for a *common* enderman [use this artifact and see for yourself](https://github.com/RayDeeUx/SkyHanni/actions/runs/7603597902)

(to see buildList fail at making the HUD show up, use this artifact)[https://github.com/RayDeeUx/SkyHanni/actions/runs/7603478706]

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
RayDeeUx: @nea89o t'was what i wanted to do, but alas: https://github.com/hannibal002/SkyHanni/pull/934#discussion_r1456284789

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
Thunderblade73: idk, why this happends but it is not the cause of the changes. I used the version from https://github.com/hannibal002/SkyHanni/pull/934/commits/eace810048e2da55016525880134dfa90356ecc2 and got 53m as well.
image
image

RayDeeUx: compromise:
```kt
val tierAndIndex = petPrices.withIndex().minBy { it.value }
return Pair(tierAndIndex.cheapestTierIndex(originalTier), tierAndIndex.value)
```
```kt
private fun IndexedValue<Double>.cheapestTierIndex(originalTier: Int) = if (this.index == 1) originalTier - 1 else if (this.index == 2) originalTier + 1 else originalTier
```

Thunderblade73: I found the bug. Sorry my fault. Didn't look at the order of stuff. The compromise is good just change the function to a when instead of an if then it is fine.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
Thunderblade73: This is how it should look like. The return at the end broke it.
```Kotlin
private fun drawDisplay(lore: List<String>) = buildList<Renderable> {
        this.add(
            Renderable.string(
                "§d§lTaming 60 Cost: §r§d(${
                    if (config.otherRarities) "cheapest" else "exact"
                } rarity)"
            )
        )
        var totalCost = 0.0
        for (line in lore) {
            neededPetPattern.matchMatcher(line) {
                val origTierString = group("tier")
                val origPetString = group("pet")
                val pet = origPetString.uppercase().replace(" ", "_").removePrefix("FROST_")
                val originalTier = LorenzRarity.getByName(origTierString.uppercase().replace(" ", "_"))?.id ?: 0
                val (cheapestTier, petPrice) = findCheapestTier(pet, originalTier)
                val displayPetString =
                    if (cheapestTier == originalTier) group("fullThing")
                    else "${LorenzRarity.getById(cheapestTier)?.formattedName} $origPetString"
                if (petPrice != -1.0) {
                    totalCost += petPrice
                    this@buildList.add(
                        Renderable.clickAndHover(
                        text = " §7- $displayPetString§7: §6${petPrice.addSeparators()} coins",
                        tips = listOf(
                            "§aClick to run §e/ahs ] $origPetString §ato find it on the Auction House.",
                            "§aNotes: §eSet the rarity filter yourself. §cBooster Cookie required!"
                        ),
                        onClick = { LorenzUtils.sendCommandToServer("ahs ] $origPetString") }
                    ))
                } else {
                    this@buildList.add(
                        Renderable.clickAndHover(
                        text = " §7- $displayPetString§7: §eNot on AH; view its Wiki article here.",
                        tips = listOf("§4Click to run §e/wiki $pet §4to view how to obtain it."),
                        onClick = { LorenzUtils.sendCommandToServer("wiki $pet") }
                    ))
                }
            }
        }
        this.add(Renderable.string("§dTotal cost §7(§6Lowest BIN§7): §6${totalCost.addSeparators()} coins"))
        if (config.otherRarities) this.add(Renderable.string("§c§lDisclaimer:§r§c Total does not include costs to upgrade via Kat."))
    }
```

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,124 @@
...
+    private fun IndexedValue<Double>.cheapestTierIndex(originalTier: Int) = when (this.index) {
+        1 -> originalTier - 1
+        2 -> originalTier + 1
+        else -> originalTier
+    }
+
+    @SubscribeEvent
+    fun onRenderOverlay(event: GuiRenderEvent.ChestGuiOverlayRenderEvent) {
+        if (!isEnabled()) return
+        if (InventoryUtils.openInventoryName() != "Offer Pets") return
```
Thunderblade73: use the pattern.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,124 @@
...
+        }
+        val tierAndIndex = petPrices.withIndex().minBy { it.value }
+        return Pair(tierAndIndex.cheapestTierIndex(originalTier), tierAndIndex.value)
+    }
+
+    private fun IndexedValue<Double>.cheapestTierIndex(originalTier: Int) = when (this.index) {
+        1 -> originalTier - 1
+        2 -> originalTier + 1
+        else -> originalTier
+    }
```
Thunderblade73: This I think is the best approach. Ery, this is no absolute, but this is how I would do it.
```suggestion
    private fun findCheapestTier(pet: String, originalTier: Int) = buildMap<Int, Double> {
        this[originalTier] = petInternalName(pet, originalTier).getPetPrice()
        if (config.otherRarities || this[originalTier] == -1.0) {
            this[originalTier - 1] = petInternalName(pet, originalTier - 1).getPetPrice(otherRarity = true)
            if (originalTier != 5) {
                this[originalTier + 1] = petInternalName(pet, originalTier + 1).getPetPrice(otherRarity = true)
            }
        }
    }.minBy { it.value }
```

Thunderblade73 (changes requested):
Done with the second round, will give it a night or two and then I will look at it again.
The majority is done. The only thing I don't like currently is the drawDisplay function but I also don't know of any better way of doing it now.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,115 @@
...
+
+    private fun findCheapestTier(pet: String, originalTier: Int) = buildMap<Int, Double> {
+        this[originalTier] = petInternalName(pet, originalTier).getPetPrice()
+        if (config.otherRarities || this[originalTier] == -1.0) {
+            this[originalTier - 1] = petInternalName(pet, originalTier - 1).getPetPrice(otherRarity = true)
+            if (originalTier != 5) this[originalTier + 1] = petInternalName(pet, originalTier + 1).getPetPrice(otherRarity = true)
+        }
+    }.minBy { it.value }
+
```
Thunderblade73, line 128: How I would implement that. May has some compile errors, since I wrote it in plain text.
```suggestion

    @SubscribeEvent
    fun onRenderOverlay(event: InventoryCloseEvent) {
    if(!isEnabled() && !event.reopenSameName) return
    display = emptyList()
    }
```

Thunderblade73: sorry, missclick. Please wait for the rest

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,115 @@
...
+    @SubscribeEvent
+    fun onRenderOverlay(event: GuiRenderEvent.ChestGuiOverlayRenderEvent) {
+        if (!isEnabled()) return
+        if (!offerPetsChestPattern.matches(InventoryUtils.openInventoryName())) return
```
Thunderblade73: Opitmaztion we can do
```suggestion
        if (display.isNotEmpty()) return
```

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,115 @@
...
+        if (config.otherRarities) this.add(Renderable.string("§c§lDisclaimer:§r§c Total does not include costs to upgrade via Kat."))
+    }
+
+    private fun findCheapestTier(pet: String, originalTier: Int) = buildMap<Int, Double> {
+        this[originalTier] = petInternalName(pet, originalTier).getPetPrice()
+        if (config.otherRarities || this[originalTier] == -1.0) {
+            this[originalTier - 1] = petInternalName(pet, originalTier - 1).getPetPrice(otherRarity = true)
+            if (originalTier != 5) this[originalTier + 1] = petInternalName(pet, originalTier + 1).getPetPrice(otherRarity = true)
+        }
+    }.minBy { it.value }
```
Thunderblade73: No need to be a map (I know I've made the misstake)
```suggestion
    private fun findCheapestTier(pet: String, originalTier: Int) = buildList<Pair<Int, Double>> {
        this.add(originalTier to petInternalName(pet, originalTier).getPetPrice())
        if (config.otherRarities || this.first().second == -1.0) {
            this.add(originalTier - 1 to petInternalName(pet, originalTier - 1).getPetPrice(otherRarity = true))
            if (originalTier != 5) this.add(originalTier + 1 to petInternalName(pet, originalTier + 1).getPetPrice(otherRarity = true))
        }
    }.minBy { it.second}
```

Thunderblade73 (changes requested):
I noticed an logical oversight. Once the display is set it is never unset after closing. Meaning it occupies memory space when it isn't needed (or Kotlin/Java makes some runtime wizardry to optimize that, but I'm not familiar with that).
So what I would suggested is that onInventoryClose, set display to a empty map. That's it. (note that you may want to check the reopenSameName value since hypixel does sometimes do funny things with inventories)

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,115 @@
...
+                    if (config.otherRarities) "cheapest" else "exact"
+                } rarity)"
+            )
+        )
+        var totalCost = 0.0
+        for (line in lore) {
+            neededPetPattern.matchMatcher(line) {
+                val origTierString = group("tier")
+                val origPetString = group("pet")
+                val pet = origPetString.uppercase().replace(" ", "_").removePrefix("FROST_")
```
Thunderblade73: ```.uppercase().replace(" ", "_")```
make that a StringUtils function, since it could be useful in many places.
For the name, idk.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,115 @@
...
+                val originalTier = LorenzRarity.getByName(origTierString.uppercase().replace(" ", "_"))?.id ?: 0
+                val (cheapestTier, petPrice) = findCheapestTier(pet, originalTier)
+                val displayPetString =
+                    if (cheapestTier == originalTier) group("fullThing")
+                    else "${LorenzRarity.getById(cheapestTier)?.formattedName} $origPetString"
```
Thunderblade73: I don't see the need for the if. Sure it might be slightly faster, but I would just use the string concat regardless of the if. Make the code more readable. (Also don't forget to remove the "fullThing" group from the pattern)

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,115 @@
...
+                }
+            }
+        }
+        this.add(Renderable.string("§dTotal cost §7(§6Lowest BIN§7): §6${totalCost.addSeparators()} coins"))
+        if (config.otherRarities) this.add(Renderable.string("§c§lDisclaimer:§r§c Total does not include costs to upgrade via Kat."))
+    }
+
+    private fun findCheapestTier(pet: String, originalTier: Int) = buildMap<Int, Double> {
+        this[originalTier] = petInternalName(pet, originalTier).getPetPrice()
+        if (config.otherRarities || this[originalTier] == -1.0) {
```
Thunderblade73: ``` this[originalTier] == -1.0```
do understand it correctly. If it doesn't find the originalTier in ah. But since -1 is the lowest value only the orginalTier will be used, so why does this condition even exist?

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,115 @@
...
+    private fun drawDisplay(lore: List<String>) = buildList<Renderable> {
+        this.add(
+            Renderable.string(
+                "§d§lTaming 60 Cost: §r§d(${
+                    if (config.otherRarities) "cheapest" else "exact"
+                } rarity)"
+            )
+        )
+        var totalCost = 0.0
+        for (line in lore) {
```
Thunderblade73: You can reduce the complexity here by remove the {} from the for. Normal;y I wouldn't do it but for this specific use it is better to read.

Thunderblade73 (commented):
Last bit of nitpicks. Once everything is done, I will have a look again and then approve this.

nea89o (commented):
Looks good, except for the wiki. That just does not work for me `/wiki WISP` fails with an invalid id.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
Thunderblade73: Could I get an answer on this one?

RayDeeUx: fallback on other rarities of a pet if the exact rarity of the pet george wants is unavailable

Thunderblade73: But the min always will use the -1. So the rest won't be used ever.

RayDeeUx:
> Looks good, except for the wiki. That just does not work for me `/wiki WISP` fails with an invalid id.

oof! i had skyhanni's fandom wiki toggled on so i didnt notice it'd fail. hmmm, maybe i'll replace with a link instead

Thunderblade73 (approved):
The code looks fine.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,132 @@
...
+                    if (config.otherRarities) "cheapest" else "exact"
+                } rarity)"
+            )
+        )
+        var totalCost = 0.0
+        for (line in lore)
+            neededPetPattern.matchMatcher(line) {
+                val origTierString = group("tier")
+                val origPetString = group("pet")
+                val pet = origPetString.convertToInternalNameString().removePrefix("FROST_")
```
hannibal002: Please extract the inside of the matchMatcher block into a function to keep indentations shorter

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,132 @@
...
+                    otherRarity = true
+                )
+            )
+        }
+    }.minBy { it.second }
+
+    @SubscribeEvent
+    fun onRenderOverlay(event: GuiRenderEvent.ChestGuiOverlayRenderEvent) {
+        if (!isEnabled()) return
+        if (!offerPetsChestPattern.matches(InventoryUtils.openInventoryName())) return
```
hannibal002: Check this once on inventory open and use a boolean inInventory, instead of checking the regex once per render

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,132 @@
...
+import at.hannibal2.skyhanni.utils.renderables.Renderable
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+class GeorgeDisplay {
+
+    private val config get() = SkyHanniMod.feature.misc.pets.george
+    private val useFandomWiki get() = SkyHanniMod.feature.commands.betterWiki.useFandom
+
+    private val SEPARATOR = ";"
```
hannibal002: no need for this variable

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,132 @@
...
+            neededPetPattern.matchMatcher(line) {
+                val origTierString = group("tier")
+                val origPetString = group("pet")
+                val pet = origPetString.convertToInternalNameString().removePrefix("FROST_")
+                val originalTier = LorenzRarity.getByName(origTierString.uppercase().replace(" ", "_"))?.id ?: 0
+                val (cheapestTier, petPrice) = findCheapestTier(pet, originalTier)
+                val displayPetString = "${LorenzRarity.getById(cheapestTier)?.formattedName} $origPetString"
+                if (petPrice != -1.0) {
+                    totalCost += petPrice
+                    this@buildList.add(
```
hannibal002: You can move the add logic above the if else block. so that you dont need the second buildlist.add below as well.

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,132 @@
...
+                                "§aClick to run §e/ahs ] $origPetString §ato find it on the Auction House.",
+                                "§aNotes: §eSet the rarity filter yourself. §cBooster Cookie required!"
+                            ),
+                            onClick = { LorenzUtils.sendCommandToServer("ahs ] $origPetString") }
+                        ))
+                } else {
+                    val typeOfWiki = if (useFandomWiki) "Fandom" else "Official Hypixel"
+                    this@buildList.add(
+                        Renderable.clickAndHover(
+                            text = " §7- $displayPetString§7: §eNot on AH; view its $typeOfWiki article here.",
```
hannibal002: overly long/worded line. maybe use "click to open wiki" instead?

hannibal002 (changes requested):
grafik
grafik

text in those two hover views is correct, the line size is not. please also change the color codes.

Also, is the order important? cant we sort by cheapest pet in this list?

src/main/java/at/hannibal2/skyhanni/config/features/misc/pets/GeorgeConfig.java
```diff
@@ -0,0 +1,22 @@
...
+
+    @Expose
+    @ConfigOption(name = "Enabled", desc = "Show the currently lowest prices of the pets you need to buy to increase your Taming level cap.")
```
CalMWolfs, line 11: this sentence sounds weird to me idk

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @SubscribeEvent
+    fun onRenderOverlay(event: InventoryCloseEvent) {
+        if (!isEnabled() && !event.reopenSameName) return
+        display = emptyList()
+    }
+
+    private fun String.getPetPrice(otherRarity: Boolean = false): Double = this.asInternalName().getPriceOrNull() ?: if (otherRarity) Double.MAX_VALUE else -1.0
+    private fun petInternalName(pet: String, originalTier: Int) = "$pet$SEPARATOR$originalTier"
```
CalMWolfs: I feel like adding this SEPARATOR constant makes this harder to read

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -299,4 +299,6 @@ object StringUtils {
     fun String.isRoman(): Boolean {
         return isRomanPattern.matches(this)
     }
+
+    fun String.convertToInternalNameString(): String = this.uppercase().replace(" ", "_")
```
CalMWolfs, line 565: This could have a better name

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,132 @@
...
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+class GeorgeDisplay {
+
+    private val config get() = SkyHanniMod.feature.misc.pets.george
+    private val useFandomWiki get() = SkyHanniMod.feature.commands.betterWiki.useFandom
+
+    private val SEPARATOR = ";"
+
+    private val patternGroup = RepoPattern.group("george.tamingcap")
```
CalMWolfs, line 36:
```suggestion
    private val patternGroup = RepoPattern.group("misc.george.tamingcap")
```

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
```diff
@@ -0,0 +1,132 @@
...
+                        Renderable.clickAndHover(
+                            text = " §7- $displayPetString§7: §eNot on AH; view its $typeOfWiki article here.",
+                            tips = listOf("§eClick to open the $typeOfWiki Wiki article for the $displayPetString §eto view how to obtain it."),
+                            onClick = {
+                                val urlCompliantPet = displayPetString.removeColor().replace(" ", "%20")
+                                if (useFandomWiki) OSUtils.openBrowser("https://hypixel-skyblock.fandom.com/wiki/Special:Search?query=$urlCompliantPet&scope=internal")
+                                else OSUtils.openBrowser("https://wiki.hypixel.net/index.php?search=$urlCompliantPet")
+                            }
+                        ))
+                }
```
CalMWolfs: How often do pets not have an ah price, this seems like they have an issue connecting to neu lowestbin api and that issue should be adressed properly instead of adding a weird work around

CalMWolfs (changes requested):
a few things

src/main/java/at/hannibal2/skyhanni/features/misc/GeorgeDisplay.kt
hannibal002: I think this is for the soulbounded pets

CalMWolfs:

PR #938, opened 2024-01-19
shcopyfoundburrowlocations

hannibal002:
/shcopyfoundburrowlocations


src/main/java/at/hannibal2/skyhanni/config/Storage.java
```diff
@@ -468,6 +468,8 @@ public static class DianaStorage {
             // TODO renmae
             public MythologicalCreatureTracker.Data mythologicalMobTracker = new MythologicalCreatureTracker.Data();
 
+            @Expose
+            public List<LorenzVec> foundBurrowLocations = new ArrayList<>();
```
CalMWolfs: Should this be in profile specific or just in general, in general makes more sense to me, also if it is expected to get very long I feel it should be separated to a new config file

CalMWolfs (changes requested):
I think that is the only thing thing that needs addressing

src/main/java/at/hannibal2/skyhanni/config/Storage.java
hannibal002: Done.
When the list gets too long, we can write a smart "extract on file io actions" logic without damaging the usage of the config structure. (Something we should have used for known features and Jacob contests, and friends as well)

CalMWolfs (approved):
Code looks good to me

PR #942, opened 2024-01-21
feature: TALBOTS_THEODOLITE support

RayDeeUx:


hannibal002:
This is neither a micro feature nor is this "maybe". This is a normal PR:

This PR size is normal. Your other PRs are just much too big, so you are not used to writing smaller PRs. But this is totally fine for a PR.

If you don't know if the feature actually works because you have not tested it yet, write so in the PR description, not the name.
Also, please don't make PRs as ready to review when you have not yet tested them yourself at least once.

hannibal002 (approved):
Works perfectly fine.

PR #943, opened 2024-01-21
Bring back the deleted test and fix the issue causing it

CalMWolfs:
There is a more complex way to fix this but I do not have the skills to do this

### DIFF
--- a/src/test/java/at/hannibal2/skyhanni/test/ItemModifierTest.kt
+++ b/src/test/java/at/hannibal2/skyhanni/test/ItemModifierTest.kt
@@ -1,0 +1,26 @@
+package at.hannibal2.skyhanni.test
+
+import at.hannibal2.skyhanni.utils.ItemUtils.isEnchanted
+import at.hannibal2.skyhanni.utils.SkyBlockItemModifierUtils.getEnchantments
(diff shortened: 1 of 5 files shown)


src/test/java/at/hannibal2/skyhanni/test/ItemModifierTest.kt
```diff
@@ -0,0 +1,26 @@
...
+    fun testUpgradeLevelMasterStars() {
+        val itemStack = TestExportTools.getTestData(TestExportTools.Item, "10starnecronhead")
+        assert(!itemStack.isRecombobulated())
+        assert(itemStack.getReforgeName() == "ancient")
+        assert(itemStack.getItemUuid() == "2810b7fe-33af-4dab-bb41-b4815f5847af")
+        assert(itemStack.isEnchanted())
+        assert(itemStack.getHotPotatoCount() == 15)
+        assert(itemStack.getEnchantments()?.size == 11)
+        assert(itemStack.hasArtOfPeace())
+//        assert(itemStack.getDungeonStarCount() == 10)
```
hannibal002, line 24: Why is this still deprecated? Was this previously the case?

hannibal002 (approved):

PR #945, opened 2024-01-22
Feature: UnclickableAuctions

RayDeeUx:
## What
Prevents bidding/buying tierboosted pets or recombobulated items, as well as hiding the auctioned item. Comes with customizable bypass key.

<details>
<summary>Images</summary>

</details>

## Changelog New Features
+ Prevent bidding on/buying recombobulated items or tier-boosted pets. - Erymanthus


RayDeeUx:
in hindsight, this could use a whitelist from the repo. but the good news is this is off by default, so only people who choose to enable this will be playing themselves—and at worst, it'll be a minor inconvenience.

RE adding repo support: maybe in a later PR by someone else? i'm not sure what the state of the SB economy is; would rather hand that task off to someone more knowledgeable

hannibal002:
old

PR #947, opened 2024-01-22
Debug: CopyBossbarCommand

RayDeeUx:
output:
`BUILD BATTLE BUG FIX PATCH RELEASED` — without formatting codes
`§2§lB§6§lU§a§lILD BATTLE BUG FIX PATCH RELEASED§r` — with formatting codes

examples:
`/shcopybossbar` — copies boss bar name with formatting codes
`/shcopybossbar true` — copies boss bar name without formatting codes

known:
\- for some reason minecraft caches the last known bossbar name instead of setting it empty. not sure if this is a bug, but i'd insist on keeping it. i think it'll come in clutch when some dev warps out of a mining lobby and wants bossbar name w/o having to warp back


hannibal002:
Already got merged as part of https://github.com/hannibal002/SkyHanni/pull/969

PR #951, opened 2024-01-24
Hide non-clickable items in Basket of Seeds, Nether Wart Pouch & Trick or Treat Bag

lunaynx:


lunaynx:
I added Basket of Seeds and Nether Wart Pouch to this PR since it's still open and I noticed those didn't have such a thing yet either.

hannibal002 (approved):

PR #958, opened 2024-01-27
Add shader reload capabilities

lineargraph:

PR #960, opened 2024-01-28
Backend: Non Textured Chroma Shader

VixidDev:
## Dependencies

- #851 
- #958  

Adds a chroma shader to be used on non-textured GUI elements.


Thunderblade73:
Please write the other PRs it depends on at the start of the description.

hannibal002 (approved):

PR #925, opened 2024-01-14
BackEnd: API Additions & Changes

j10a1n15:
Basically copied every data related thing from my scoreboard pr #893 into a new, seperate one, for better reviewing

Includes:
## Additions:
- MaxwellAPI
- QuiverAPI
- BitsAPI
- /shcopyactionbar
- text rendering method with options to align on width
- List<String>.removeNextAfter()

## Changes
- Renamed MayorElection to MayorAPI
- Skyblocktime.formatted() options
- Some Patterns -> RepoPatterns

PR #966, opened 2024-01-30
request elections api less frequently

CalMWolfs:
- request elections api a little less
- use simple time mark
- remove nested if
- convert two `logError()` to `logErrorWithData()`


Thunderblade73:
This looks like the changes done in #925. Except the less frequent thingy.

CalMWolfs:
> This looks like the changes done in #925. Except the less frequent thingy.

Damn didnt even see that, well this would probably be a more temporary fix as that pr looks like it does much more but ill leave this open as it does remove two depreciated things and improves the api being requested too frequently for now

hannibal002 (approved):

PR #968, opened 2024-01-31
Backend: GetFromSackAPI

Thunderblade73:
Adds an API for calling the hypixel command "getfromsacks".
It queues up the commands so no command will be missed.
A get can be used to inputed directly, use a chat message to get it or use a click to a slot in an inventory (currently right click but open for change) for calling the hypixel command.

Also naturaly queues up user inputed "gfs" or "getfromsacks".


CalMWolfs (changes requested):
We already have SackApi which theoretically has a list of all the user's items in sacks and the amount so maybe you could somehow integrate with that especially for getting the sack item list as not all the repo names match up to internal names and this way you can already tell if they dont have enough in their sack and then you can prompt them to get from bazaar. The sacks data in SackApi was made specifically for features like this one

Thunderblade73:
@CalMWolfs hope c5773c4 implements your suggestions. The checking with the repo I didn't remove since it needs to be up to date anyways and they are internal names just without _ and not uppercase (yeah I know it is scuffed but hey it is in the repo so no easy way to fix that)

CalMWolfs:
Yeah ok sounds good, I think if the sack status is outdated you could also say to open /sacks and update data (as the user has the sack for that item it just has messed up data) but if it's missing then assume they don't have the sack so do what you do.

src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
```diff
@@ -0,0 +1,22 @@
+package at.hannibal2.skyhanni.api
+
+import at.hannibal2.skyhanni.data.GetFromSackData
+import at.hannibal2.skyhanni.utils.LorenzUtils
+import at.hannibal2.skyhanni.utils.PrimitiveItemStack
+import net.minecraft.inventory.Slot
+
+object GetFromSackAPI {
```
hannibal002, line 30: Please use more descriptive function names. instead of get, use something like requestFromSacks, instead of chat, use sendCommandInChat or similar.

src/main/java/at/hannibal2/skyhanni/config/commands/Commands.kt
```diff
@@ -414,6 +415,9 @@ object Commands {
         registerCommand("pt", "Transfer the party to another party member") { PartyCommands.transfer(it) }
         registerCommand("pp", "Promote a specific party member") { PartyCommands.promote(it) }
         registerCommand("pd", "Disbands the party") { PartyCommands.disband() }
+
+        registerCommand("gfs", "Get from Sack (queued)") { GetFromSackData.commandHandler(it) }
```
hannibal002: I dont think we should override existing hypixel commands this way. please use MessageSendToServerEvent instead. (having a disabled feature that still has an command that simply calls the original command feels very odd to me)

src/main/java/at/hannibal2/skyhanni/config/commands/Commands.kt
```diff
@@ -414,6 +415,9 @@ object Commands {
...
+        registerCommand("getFromSacks", "Get from Sack (queued)") { GetFromSackData.commandHandler(it) }
```
hannibal002: please use lower case for consistency. Mayve we should camel case all commands in this file,but at least make this change in another pr

src/main/java/at/hannibal2/skyhanni/config/features/inventory/InventoryConfig.java
```diff
@@ -201,4 +200,15 @@ public String toString() {
     @FeatureToggle
     public boolean shiftClickBrewing = false;
 
+    @Expose
+    @ConfigOption(name = "Queued GfS", desc = "If /gfs or /getfromsacks is used it ques up the commands so all items are guarantied to receive the items")
```
hannibal002: Typo in "ques" i think?
please color the commands in descrption yellow for consistency.
Please use a dot at the end of a description.

src/main/java/at/hannibal2/skyhanni/config/features/inventory/InventoryConfig.java
```diff
@@ -201,4 +200,15 @@ public String toString() {
...
+    @ConfigEditorBoolean
+    @FeatureToggle
+    public boolean queuedGFS = true;
```
hannibal002: use a config group for gfs with queue and bazaar in it

src/main/java/at/hannibal2/skyhanni/data/GetFromSackData.kt
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    fun commandHandler(args: Array<String>) {
+        if (!config.queuedGFS) {
+            LorenzUtils.sendCommandToServer("gfs ${args.joinToString(" ")}")
+            return
+        }
+
+        if (args.size != 2) {
+            LorenzUtils.chat("§cMissing arguments! Usage: /getfromsacks <name/id> <amount>", prefix = false)
```
hannibal002: please use LorenzUtils.userError() instead

src/main/java/at/hannibal2/skyhanni/data/GetFromSackData.kt
```diff
@@ -0,0 +1,130 @@
...
+            return
+        }
+
+        val item = args[0].asInternalName()
+
+        if (!sackList.contains(item)) {
+            LorenzUtils.chat("§cCouldn't find an item with this name or identifier!", prefix = true)
```
hannibal002: use userError

src/main/java/at/hannibal2/skyhanni/data/GetFromSackData.kt
```diff
@@ -0,0 +1,130 @@
...
+            return
+        }
+
+        val amountString = args[1]
+
+        if (!amountString.isInt()) {
+            LorenzUtils.chat("§cInvalid amount!", prefix = false)
```
hannibal002: use userError

src/main/java/at/hannibal2/skyhanni/data/GetFromSackData.kt
```diff
@@ -0,0 +1,130 @@
...
+            return
+        }
+
+        val amount = amountString.toInt()
+
+        if (config.bazaarGFS && !LorenzUtils.noTradeMode) {
+            val sackInfo = SackAPI.fetchSackItem(item)
+            if (sackInfo.getStatus() != SackStatus.CORRECT && sackInfo.getStatus() != SackStatus.ALRIGHT) {
+                LorenzUtils.clickableChat("§fUnsure if items are available in Sack, §eCLICK§f to open bazaar", "bz ${item.asString()}")
```
hannibal002: Maybe use a different color. white doesnt match SkyHannis color style and clickable messages

src/main/java/at/hannibal2/skyhanni/data/GetFromSackData.kt
```diff
@@ -0,0 +1,130 @@
...
+
+    @SubscribeEvent
+    fun onInventoryClose(event: InventoryCloseEvent) {
+        inventoryMap.clear()
+    }
+
+    @SubscribeEvent
+    fun onSlotClicked(event: GuiContainerEvent.SlotClickEvent) {
+        if (!LorenzUtils.inSkyBlock) return
+        if (event.clickedButton != 1) return // right click
```
hannibal002: The comment feels confusing to me. do we filter right clicks here or do we only want right clicks below this line?

src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
hannibal002: New idea: delethe the whole class and use get from sacks data instead.

src/main/java/at/hannibal2/skyhanni/data/GetFromSackData.kt
```diff
@@ -0,0 +1,130 @@
...
+import at.hannibal2.skyhanni.utils.NumberUtil.isInt
+import at.hannibal2.skyhanni.utils.PrimitiveItemStack
+import at.hannibal2.skyhanni.utils.PrimitiveItemStack.Companion.makePrimitiveStack
+import at.hannibal2.skyhanni.utils.SimpleTimeMark
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+import java.util.LinkedList
+import java.util.Queue
+import kotlin.time.Duration.Companion.seconds
+
+object GetFromSackData {
```
hannibal002: from looking at the code and trying to test the feature ingame i dont know what the queue or the slot id is used for.

src/main/java/at/hannibal2/skyhanni/config/features/inventory/InventoryConfig.java
```diff
@@ -201,4 +200,15 @@ public String toString() {
...
+
+    @Expose
+    @ConfigOption(name = "Bazaar GfS", desc = "If you don't have enough items in sack get a prompt to buy them from bazaar.§e Queued GfS needs to be turned on!")
```
hannibal002: move the empty space before the color code, not after (for consistency)

hannibal002 (changes requested):
From looking at the code and trying to test the feature ingame, I dont know what the queue or the slot id is used for.

Also this wording is super confusing to me:
"so all items are guarantied to receive the items"

Thunderblade73:
Converted to draft since everything broke (idk why)

src/main/java/at/hannibal2/skyhanni/config/commands/Commands.kt
Thunderblade73: This suggestion broke everything. Because when the MessageSendToServerEvent is used it becomes impossible to send the command to the server. Because it recursively calls itself and blocks itself.
Eg. Only typed the command once
image

Thunderblade73:
Works again

src/main/java/at/hannibal2/skyhanni/config/commands/Commands.kt
hannibal002: Find a way to differentiate between commands sent from users and from mods.
i think we have a way to do this already. when im wrong, just keep the current command as a workaround ig.

hannibal002 (changes requested):
Either default disable this feature or only show the bazaar prompt when hypixel has sent the no item in sacks message:
image

src/main/java/at/hannibal2/skyhanni/config/commands/Commands.kt
Thunderblade73: implemented

hannibal002 (approved):

hannibal002 (approved):

PR #969, opened 2024-02-01
Send Mining Island Events

CalMWolfs:
- Sending crystal hollows events to soopy's api
- BossbarData (better than https://github.com/hannibal002/SkyHanni/pull/947) (sorry ery)
- Getting the lobby's server id
- Better posting to apis
- Filter out a few more console errors
- Some useful json utils
- Prob something else idk

All the settings do nothing currently as they all require the upcoming events api to be online. I have made errors not print for now because of this. Ideally this is merged sooner rather than later so that we can get more data for the actual feature but if not I will update this pr but there is a higher chance it wont work properly.
This new version of upcoming mining events is better than the last one:
- includes dwarven mines as well as crystal hollows
- should put less stress on his server
- will be more accurate (shows if the same event happens twice in a row)
- will provide cool metrics about what percent of lobbies currently have a certain event

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/mining/eventtracker/MiningEventTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/mining/eventtracker/MiningEventTracker.kt
@@ -2,6 +2,7 @@ package at.hannibal2.skyhanni.features.mining.eventtracker
 
 import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.config.ConfigManager
+import at.hannibal2.skyhanni.data.BossbarData
 import at.hannibal2.skyhanni.data.IslandType
 import at.hannibal2.skyhanni.data.ScoreboardData
 import at.hannibal2.skyhanni.events.BossbarUpdateEvent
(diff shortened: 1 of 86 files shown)


src/main/java/at/hannibal2/skyhanni/test/command/CopyBossbarCommand.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import at.hannibal2.skyhanni.data.BossbarData
+import at.hannibal2.skyhanni.utils.LorenzUtils
+import at.hannibal2.skyhanni.utils.OSUtils
+import at.hannibal2.skyhanni.utils.StringUtils.removeColor
+
+object CopyBossbarCommand {
+    fun command(args: Array<String>) {
+        val noFormattingCodes = args.size == 1 && args[0] == "true"
+        val (bossbarName, status) = if (noFormattingCodes) Pair(BossbarData.getBossbar().removeColor(), "without") else Pair(BossbarData.getBossbar(), "with")
```
hannibal002: Since status is only used in the else branch below, you could split the pair up into two separate ifs. one in this line, one in the else branch.

src/main/java/at/hannibal2/skyhanni/test/command/CopyBossbarCommand.kt
```diff
@@ -0,0 +1,19 @@
...
+        if (bossbarName.isBlank() || bossbarName.isEmpty()) {
```
hannibal002: isEmpty can go

src/main/java/at/hannibal2/skyhanni/features/mining/eventtracker/MiningEventTracker.kt
```diff
@@ -0,0 +1,155 @@
...
+                "eventName" to eventName,
+                "bossbar" to BossbarData.getBossbar(),
+                "serverType" to LorenzUtils.skyBlockIsland,
+                "fromChat" to (time == null)
+            )
+            return
+        }
+        lastSentEvent = eventType
+
+        val timeRemaining = if (time == null) {
```
hannibal002, line 106: Maybe use `val timeRemaining = time?.let { TimeUtils.getDuration(it) } ?: eventType.defaultLength` here

src/main/java/at/hannibal2/skyhanni/utils/LorenzUtils.kt
```diff
@@ -254,6 +259,29 @@ object LorenzUtils {
         }
     }
 
+    @SubscribeEvent
+    fun onWorldChange(event: LorenzWorldChangeEvent) {
+        serverId = null
```
hannibal002: Server id should be checked/managed in its own class.

hannibal002 (changes requested):
Only code formatting changes.

hannibal002 (approved):

PR #973, opened 2024-02-02
Feature: Optifine CIT Cache

CalMWolfs:
Written by @nea89o 

Gain 20-50% extra performance while using a skyblock cit pack
Will not work in dev environment!
`/shoptifinecache` shows stats about the cache over the last 100 ticks

Test in hub 1 looking forward while using furfsky
No cit pack
no cit

With cit pack
cit

With cit pack and option on
with option

Conclusion is that the setting works well but cit packs are still horrible for performance


src/main/java/at/hannibal2/skyhanni/config/features/misc/MiscConfig.java
```diff
@@ -98,6 +98,12 @@ public class MiscConfig {
     @Accordion
     public PetCandyDisplayConfig petCandy = new PetCandyDisplayConfig();
 
+    @Expose
+    @ConfigOption(name = "Optifine CIT Cache", desc = "Caches optifine CIT data to increase performance when using it.")
```
lunaynx:
```suggestion
    @ConfigOption(name = "OptiFine CIT Cache", desc = "Caches OptiFine CIT data to increase performance when using it.")
```

src/main/java/at/hannibal2/skyhanni/features/misc/OptifineCitCache.kt
```diff
@@ -0,0 +1,95 @@
...
+        map.clear()
+    }
+
+    fun onCommand(args: Array<String>) {
+        var amount = Int.MAX_VALUE
+        if (args.isNotEmpty() && args[0].isInt()) {
+            amount = args[0].toInt()
+        }
+
+        LorenzUtils.chat("Optifine Cache Size:")
```
lunaynx:
```suggestion
        LorenzUtils.chat("OptiFine Cache Size:")
```

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -146,6 +146,10 @@ object StringUtils {
         }
     }
 
+    fun Int.trailingS(): String {
```
lunaynx: Terrible name. Also I would just use `canBePlural`/`optionalPlural` (what the hell are these names?), maybe extend it to to default to `${singular}s`.

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
lineargraph: I prefer the original name: `s`

lunaynx: #975

lineargraph: https://git.nea.moe/nea/neuhax/src/commit/c2467b478705ccb3a2d0800beff9b1b4895c43c3/src/main/kotlin/moe/nea/sky/util/chatutils.kt#L83

RayDeeUx: seeing as we're not playing code golf, #975

lineargraph (changes requested):
Finally went through this PR and I can't say I am too happy about it.

I've voiced some of these concern in private already, but since this PR is still open, I will voice them here as well. I am generally very happy to help people with problems they might face while coding, but this PR isn't that. It is a 1:1 copied feature from one of my mods with effectively 0 user facing credit.

While i have technically given consent, I was extremely pressured to do so by hannibal and calmwolf. Hannibal just opened up a feature request describing the problem my feature solves and tagged me as a "potential help", knowing fully well that i had already solved this problem on my own and dedicated it to another mod.

I love helping other people, but I do not like being seen as a resource that can be taken for granted. The implementation of this PR additionally does not even see me as a resource that is worth being mentioned. Other features in this mod that reimplement features of other mods usually put their own spin on it (this does not) and explicitly credit the other mod in the settings. See the chroma config or the soopy burrow guess for example.

While technically the license allows for this PR (assuming proper attribution, which currently is not the case), i retract my former explicit consent for this feature to be in SkyHanni.

CalMWolfs:
Closed by the request of @nea89o

lunaynx:
Fair enough, but it's unfortunate that it's in a cheat mod, so we can't exactly go around recommending it to people.

lineargraph:
> Fair enough, but it's unfortunate that it's in a cheat mod, so we can't exactly go around recommending it to people.

Ill port it with another performance improvement patch i am working on into its own standalone mod

CalMWolfs:
> Ill port it with another performance improvement patch i am working on into its own standalone mod

I feel that this is the best solution and sorry for any problems I caused through this pull request that was never my intention.

jani270:
> > Fair enough, but it's unfortunate that it's in a cheat mod, so we can't exactly go around recommending it to people.
> 
> Ill port it with another performance improvement patch i am working on into its own standalone mod

More standalone mods is always a good idea.

PR #974, opened 2024-02-02
Feature: Skyblock Guide Highlights

Thunderblade73:
Highlights many different things in the "Ways to gain Skyblock Xp" thingy, which still can be done. 

I'm well aware that some of those highlights are quite silly, eg; Bank upgrades. But I added those for completeness sake and consistency in the menus.
## Preview
image
image
image
image


src/main/java/at/hannibal2/skyhanni/config/features/inventory/InventoryConfig.java
```diff
@@ -9,7 +9,6 @@
 import io.github.moulberry.moulconfig.annotations.ConfigEditorBoolean;
 import io.github.moulberry.moulconfig.annotations.ConfigEditorDraggableList;
 import io.github.moulberry.moulconfig.annotations.ConfigOption;
-
```
hannibal002, line 12: My auto reformat tool in IntellIj adds this empty line back. Was this a manual change or a reformat tool?

src/main/java/at/hannibal2/skyhanni/config/features/inventory/InventoryConfig.java
```diff
@@ -151,13 +154,15 @@ public String toString() {
     @Expose
     @ConfigOption(name = "Missing Tasks",
         desc = "Highlight missing tasks in the SkyBlock Level Guide inventory.")
+    // TODO move( , "inventory.highlightMissingSkyBlockLevelGuide", "inventory.skyblockGuideConfig.highlightMissingSkyBlockLevelGuide")
```
hannibal002, line 157: Do you want to move those elements in this pr or is this planned for later? in both cases, the new location should be named simpler (no duplicate names necessary)

src/main/java/at/hannibal2/skyhanni/features/inventory/SkyblockGuideHighlightFeature.kt
```diff
@@ -0,0 +1,159 @@
...
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+import org.intellij.lang.annotations.Language
+
+val group = RepoPattern.group("skyblockguide.highlight")
+
+private const val keyPrefixInventory = "inventory"
+private const val keyPrefixCondition = "condition"
+
+class SkyblockGuideHighlightFeature private constructor(
```
hannibal002, line 25: This looks like a good use case for an enum.

src/main/java/at/hannibal2/skyhanni/features/inventory/SkyblockGuideHighlightFeature.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    private constructor(
+        config: () -> Boolean,
+        key: String,
+        @Language("RegExp")
+        inventory: String,
+        loreCondition: RepoPattern,
+        onSlotClicked: (GuiContainerEvent.SlotClickEvent) -> Unit = {},
+        onTooltip: (LorenzToolTipEvent) -> Unit = {},
+    ) : this(config, group.pattern("$key.$keyPrefixInventory", inventory), loreCondition, onSlotClicked, onTooltip)
```
hannibal002, line 58: are you sure that repo pattern will work in this case? other instances where enums use the repo pattern inizialize every single enum instance in skyhanni main class on load event manually

src/main/java/at/hannibal2/skyhanni/features/inventory/SkyblockGuideHighlightFeature.kt
```diff
@@ -0,0 +1,159 @@
...
+            if (!isEnabled()) return
+            val current = activeObject ?: return
+            if (!missing.contains(event.slotId)) return
+            current.onSlotClicked.invoke(event)
+        }
+
+        @SubscribeEvent
+        fun onBackgroundDrawn(event: GuiContainerEvent.BackgroundDrawnEvent) {
+            if (!isEnabled()) return
+            val current = activeObject ?: return
```
hannibal002: unused

hannibal002 (changes requested):
I cant get this to work in my dev env. Also do we really need that many options for every single category? Sure, i like customizations, but what are the odds a user only wants to have some of those items highlighted and not everything/nothing?

hannibal002:
This does not mean "you have to delete the options", this just means "im not sure if we need them. lets discuss your reasoning"

src/main/java/at/hannibal2/skyhanni/config/features/inventory/InventoryConfig.java
Thunderblade73: Automatic, do not know which option you have turned on to enforce thism

Thunderblade73: solved in dm's

src/main/java/at/hannibal2/skyhanni/features/inventory/SkyblockGuideHighlightFeature.kt
Thunderblade73: I can't get it to work. Because it has multiple constructor + values in the companion object (which isn't initialized yet)

src/main/java/at/hannibal2/skyhanni/features/inventory/SkyblockGuideHighlightFeature.kt
Thunderblade73: it works when called in the init of an object which is loaded. eg. the companion object

src/main/java/at/hannibal2/skyhanni/config/features/inventory/InventoryConfig.java
Thunderblade73: a todo for the future. I don't like to make the move. With the name I just kept it as is.

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SkyblockGuideConfig.java
CalMWolfs, line 1: All of these are missing punctuation at the end of the line

hannibal002 (changes requested):
I still can't get this feature to work. Are you sure the repo patterns are getting loaded properly?

Why do we have that many different Options? Why can't it be one single toggle for everything? I don't think users will disable this in one specific part of the SkyBlock Guide inventory, and want to still show it in others.

Thunderblade73:
The argument for the options:
* minion / essence - could be irritating for some people in those menus
* abiphone - should be separat because this one could be very useful for some people (and the may not want the other highlights)
* story - useful to find missed content (and the rest shouldn't be mandatory to have enabled)
* oneTimeCompleteion - more of a completion thing. Isn't really that useful but gives the menus a more consistent look. (People may want to disable this)
* jacob - idk, maybe for garden only's
*  Consumable - do not have any argument why this should be a separate option but also do not know in which other option it should be combined with

CalMWolfs (changes requested):
didnt test in game but the rest looked fine

CalMWolfs (approved):
Worked for me

hannibal002:
Is it intended that this does not get highlighted?

image

hannibal002:
or here
image

hannibal002:
or here
image

hannibal002:
or here
image

hannibal002:
image
or here

hannibal002:
or here
image

hannibal002:
or here
image

hannibal002 (changes requested):
add 7 missing spots

Thunderblade73:
Added way more highlights then the 9 suggested. 
So basically everything in the ways to level up skyblock xp has an highlight now.
Exception are:
open ended tasks (eg. accessory bag size upgrade)
dungeon classes (does not feel right in this menu)
the first page (since it is difficult to make it work with the recently viewed things)

hannibal002 (approved):

PR #975, opened 2024-02-02
Refactor: Clean up pluralization methods

lunaynx:


hannibal002:
Why can the second string parameter "plural" be null?

lunaynx:
If it's null, it just falls back to `${singular}s`. It felt weird to make the default an empty string, but I can do that instead if you prefer.

hannibal002:
I have not see this


plural ?: "${singular}s"


This is fine, leave it default null

CalMWolfs (approved):
lgtm

PR #993, opened 2024-02-10
Added hoverable for hidden cheap items in ItemTracker

Mikecraft1224:
Hovering on cheap hidden items in the ItemTracker now shows a tooltip with the hidden items listed


hannibal002 (changes requested):
suggested chagnes in dms

hannibal002 (approved):

PR #1006, opened 2024-02-14
Wrong/broken Changelog: Constant shared references

lineargraph:
Add typesafe shared constant references to objects (mostly item stacks).

This PR proposes to wrap objects which are intended to be immutable shared references into a `Const` class enforcing these contracts at a type system level.


hannibal002:
old

PR #1010, opened 2024-02-15
Docs: is this a rat

hannibal002:
preview: https://github.com/hannibal002/SkyHanni/blob/is-this-a-rat/docs/IS_THIS_A_RAT.md

exclude_from_changelog

### DIFF
--- a/docs/IS_THIS_A_RAT.md
+++ b/docs/IS_THIS_A_RAT.md
@@ -1,0 +1,25 @@
+There are websites that will claim SkyHanni is a rat.
+
+Those websites are wrong.
+
+Those websites might claim to know the mod has malicious code.
+
+Looking closer at the parts in the code they claim are malicious, we notice that their detection algorhytmus has a lot of flaws:
+
+for example:
+
+example 1
+
+Those websites might even claim their findings are powered by artifical intelligence. this is obviously wrong.
+
+Please dont trust websites that claim to know wether or not any mod can rat you or harm your computer.
+
+To download a mod safely, search for `github <mod name>` on the web.
+
+Or you use mod lists like those:
+
+- https://hypixel-skyblock.fandom.com/wiki/Community_Resources:Mods
+
+- https://sbmw.ca/mod-lists/skyblock-mod-list/
+
+Some bigger discord servers also have lists of discord servers of trusted mods.
--- a/docs/IS_THIS_A_RAT.md
+++ b/docs/IS_THIS_A_RAT.md
@@ -4,7 +4,7 @@ Those websites are wrong.
 
 Those websites might claim to know the mod has malicious code.
 
-Looking closer at the parts in the code they claim are malicious, we notice that their detection algorhytmus has a lot of flaws:
+Looking closer at the parts in the code they claim are malicious, we notice that their detection algorithms has a lot of flaws:
 
 for example:
 
--- a/docs/IS_THIS_A_RAT.md
+++ b/docs/IS_THIS_A_RAT.md
@@ -12,7 +12,7 @@ example 1
 
 Those websites might even claim their findings are powered by artifical intelligence. this is obviously wrong.
 
-Please dont trust websites that claim to know wether or not any mod can rat you or harm your computer.
+Please don't trust websites that claim to know whether or not any mod can rat you or harm your computer.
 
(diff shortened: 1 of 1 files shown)


docs/IS_THIS_A_RAT.md
```diff
@@ -0,0 +1,25 @@
...
+
+for example:
+
+example 1
+
+Those websites might even claim their findings are powered by artifical intelligence. this is obviously wrong.
+
+Please dont trust websites that claim to know wether or not any mod can rat you or harm your computer.
+
+To download a mod safely, search for `github <mod name>` on the web.
```
lineargraph, line 17: This is not strictly true. Especially NEU gets deranked, since forks aren't displayed as prominently. I would recommend instead linking to a trusted community hub or referring to the hypixel forums, since those are usually well moderated and link the official github.

docs/IS_THIS_A_RAT.md
```diff
@@ -0,0 +1,25 @@
...
+
+Looking closer at the parts in the code they claim are malicious, we notice that their detection algorhytmus has a lot of flaws:
+
+for example:
+
+example 1
+
+Those websites might even claim their findings are powered by artifical intelligence. this is obviously wrong.
+
+Please dont trust websites that claim to know wether or not any mod can rat you or harm your computer.
```
lineargraph: Grammar:
```suggestion
Please don't trust websites that claim to know whether or not any mod can rat you or harm your computer.
```

docs/IS_THIS_A_RAT.md
```diff
@@ -0,0 +1,25 @@
+There are websites that will claim SkyHanni is a rat.
+
+Those websites are wrong.
+
+Those websites might claim to know the mod has malicious code.
+
+Looking closer at the parts in the code they claim are malicious, we notice that their detection algorhytmus has a lot of flaws:
```
lineargraph: Grammar: 
```suggestion
Looking closer at the parts in the code they claim are malicious, we notice that their detection algorithms has a lot of flaws:
```

lineargraph (changes requested):
Might look at writing my own text, but here are some more surface level criticisms

docs/IS_THIS_A_RAT.md
hannibal002: The hypixel forum post that comes up first is https://hypixel.net/threads/notenoughupdates-neu-skyblock-mod-release.3135465/.

There the same links are provided as the first 3-4 google search results, at least for me.

CalMWolfs: I still agree with nea here as in the past we have seen fake repositories take over real ones and while this may not happen for some mods like skyhanni or neu right now there is no guarantee that it doesnt happen in the future or to another legit mod

docs/IS_THIS_A_RAT.md
```diff
@@ -0,0 +1,25 @@
...
+
+for example:
+
+example 1
+
+Those websites might even claim their findings are powered by artifical intelligence. this is obviously wrong.
+
+Please don't trust websites that claim to know whether or not any mod can rat you or harm your computer.
+
+To download a mod safely, search for `github <mod name>` on the web.
```
CalMWolfs, line 17: This line should be removed because it isn't necessarily true

CalMWolfs (changes requested):
Also we should have part of the is this a rat tag that we have on GitHub in here so that people can read that info from here and don't have to be in the discord for it

docs/IS_THIS_A_RAT.md
hannibal002: While this sentence is true for every known mod that exists for couple months, I agree, its still too broad.
what else would you say we should tell users how to find legit mods? just trust their big discord server, or search on Modrinth?

CalMWolfs: > While this sentence is true for every known mod that exists for couple months, I agree, its still too broad. what else would you say we should tell users how to find legit mods? just trust their big discord server, or search on Modrinth?

I think i only meant to comment on this line "To download a mod safely, search for `github <mod name>` on the web."
But yea those two ideas are better than github search which is prone to suggesting a rat repo

hannibal002:
After 8 months of not working on the pr, I think this pr is no longer needed as we already have copypasta information on discord already.

PR #1016, opened 2024-02-16
Item Name Rewrite and Cache

hannibal002:
Fixed enchantment names and pet names in the chest value feature
Fixed pet names in item profit trackers

Added cache to item stack → neu internal name
Added cache to internal name → item name
Added debug option to show SkyHanni item name in item lore

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/PetAPI.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/PetAPI.kt
@@ -5,22 +5,21 @@ import at.hannibal2.skyhanni.utils.StringUtils.matches
 import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
 
 object PetAPI {
-    private val group = RepoPattern.group("misc.pet")
-    private val petMenuPattern by group.pattern(
+    private val patternGroup = RepoPattern.group("misc.pet")
+    private val petMenuPattern by patternGroup.pattern(
         "menu.title",
         "Pets(?: \\(\\d+/\\d+\\) )?"
     )
-
-    private val petItemName by group.pattern(
+    private val petItemName by patternGroup.pattern(
         "item.name",
         "§.\\[Lvl (?<level>\\d)] (?<name>.*)"
     )
-
-    private val neuRepoPetItemName by group.pattern(
+    private val neuRepoPetItemName by patternGroup.pattern(
         "item.name.neu.format",
         "(§f§f)?§7\\[Lvl 1➡(100|200)] (?<name>.*)"
     )
 
+``` To better match the format of patterns in the rest of the mod
     fun isPetMenu(inventoryTitle: String): Boolean = petMenuPattern.matches(inventoryTitle)
 
     // Contains color code + name and for older SkyHanni users maybe also the pet level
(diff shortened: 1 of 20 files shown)


src/main/java/at/hannibal2/skyhanni/data/PetAPI.kt
```diff
@@ -1,14 +1,26 @@
...
+    private val petItemName by group.pattern(
+        "item.name",
+        "§.\\[Lvl (?<level>\\d)] (?<name>.*)"
+    )
+
+    private val neuRepoPetItemName by group.pattern(
+        "item.name.neu.format",
+        "(§f§f)?§7\\[Lvl 1➡(100|200)] (?<name>.*)"
+    )
+
```
CalMWolfs:
```suggestion
    private val patternGroup = RepoPattern.group("misc.pet")
    private val petMenuPattern by patternGroup.pattern(
        "menu.title",
        "Pets(?: \\(\\d+/\\d+\\) )?"
    )
    private val petItemName by patternGroup.pattern(
        "item.name",
        "§.\\[Lvl (?<level>\\d)] (?<name>.*)"
    )
    private val neuRepoPetItemName by patternGroup.pattern(
        "item.name.neu.format",
        "(§f§f)?§7\\[Lvl 1➡(100|200)] (?<name>.*)"
    )

``` To better match the format of patterns in the rest of the mod

**Review by CalMWolfs — CHANGES_REQUESTED** [review]
Read through the code, didnt get a chance to test in game but it all seemed like it would work, just this formatting change needed

**Review by CalMWolfs — APPROVED** [review]
Lgtm

File: src/main/java/at/hannibal2/skyhanni/data/PetAPI.kt [hunk 2]
```diff
@@ -1,13 +1,23 @@
...
         "Pets(?: \\(\\d+/\\d+\\) )?"
     )
+    private val petItemName by patternGroup.pattern(
+        "item.name",
+        "§.\\[Lvl (?<level>\\d)] (?<name>.*)"
+    )
+    private val neuRepoPetItemName by patternGroup.pattern(
+        "item.name.neu.format",
+        "(§f§f)?§7\\[Lvl 1➡(100|200)] (?<name>.*)"
+    )
```
Thunderblade73, line 20: Do we really need two patterns?
Since it could simply be done with this pattern:
```
"(?:§.)*\\[Lvl (?:1➡)?(?<level>\\d)] (?<name>.*)"
```

Thunderblade73 (changes requested):
The code looks good. Only one issue with the Patterns in the PetAPI.

src/main/java/at/hannibal2/skyhanni/data/PetAPI.kt
hannibal002: I split it into two patterns, as we don't know the level from the second one, and having a nullable group felt wrong to me. But thinking more about it, it's fine, and even cleaner for the future. I plan on merging all pet related patterns on SkyHanni (search for `petLevelPattern`, there are 2 other classes using the same regex) into the PetAPI class.

hannibal002: your one regex solution breaks here: https://regex101.com/r/4uYqMg/1

CalMWolfs: I think keeping it as two and then eventually moving it all to petAPI is fine

PR #1024, opened 2024-02-19
Code Cleanup: In rounded rectangles

hannibal002:
Depends on:
- #851

this is just a small code cleanup


hannibal002:
This is not unnecessary

PR #1027, opened 2024-02-19
Code Cleanup: Removed Pet Repo Duplications

hannibal002:


CalMWolfs (approved):
Everything looks fine, while testing noticed it only worked for pets level 1-9 and thats becasue the pattern was missing a + and the colourcodes, added that and now it works, also you were getting the wrong group which would've caused a number format  exception except the pattern was already wrong

PR #1036, opened 2024-02-22
Fix: Some npc price stuff

CalMWolfs:
- Make error message more correct as it isnt bazaar data
- Print out what the input api response from hypixel was
- Potentially eliminate some json parsing errors

From https://discord.com/channels/516977525906341928/1210138446522220585 (Not really related but saw error and was confused and its because the error message was wrong and also wasn't printing the right data)


hannibal002:
I don't like the get... functions for json objects. They open up the possibility to work with json objects at all. But this is just a workaround. 
the real fix is to replace it with an java object (something like HypixelItem, id, npc_sell_price and motes_sell_price as nullable fields. inside. Then the whole itemsData map can be wrapped to neu internal name -> Hypixel Item

But this is something for another pr. in there we can deprecate those functions in JsonUtils.
For now this pr makes the code better, will review in a sec

hannibal002 (approved):

PR #1050, opened 2024-02-25
Fix: Skill Progress

superhize:
moved to
- #1058


hannibal002:
How can I reproduce those errors? i dont understand the stuff fixed in the first commit

PR #1053, opened 2024-02-25
Use neu repo instead of skyhanni repo

CalMWolfs:


src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
```diff
@@ -222,12 +223,25 @@ object GetFromSackAPI {
...
+        try {
+            val sacksData = ConfigManager.gson.fromJson<NeuSacksJson>(data).sacks
+            val uniqueSackItems = mutableSetOf<NEUInternalName>()
+
+            sacksData.values.forEach { sackInfo ->
+                sackInfo.contents.forEach { content ->
+                    uniqueSackItems.add(content)
+                }
+            }
+
```
Thunderblade73, line 238: Missing the set of the InternalNames set
```suggestion

            sackListInternalNames = uniqueSackItems.map { it.asString() }.toSet()
```

src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
```diff
@@ -222,12 +223,25 @@ object GetFromSackAPI {
...
+            sackListNames = uniqueSackItems.map { it.itemNameWithoutColor.uppercase() }.toSet()
 
-        DelayedRun.runNextTick {
-            sackListNames = sackListInternalNames.map { it.asInternalName().itemNameWithoutColor.uppercase() }.toSet()
+        } catch (e: Exception) {
+            ErrorManager.logErrorWithData(
+                e, "Error getting NEU sacks data, make sure your neu is updated.",
```
hannibal002: maybe tell "your neu repo" instead. and as long as we depend on neu, we could even run `Utils.showOutdatedRepoNotification()`

src/main/java/at/hannibal2/skyhanni/features/misc/items/EstimatedItemValue.kt
```diff
@@ -51,18 +50,12 @@ object EstimatedItemValue {
...
-            gemstoneUnlockCosts =
-                ConfigManager.gson.fromJson(
-                    data,
-                    object : TypeToken<HashMap<NEUInternalName, HashMap<String, List<String>>>>() {}.type
-                )
-        else
-            ChatUtils.error("Gemstone Slot Unlock Costs failed to load!")
+    fun onNeuRepoReload(event: NeuRepositoryReloadEvent) {
+        val data = event.getConstant("gemstonecosts") ?: run {
+            ErrorManager.skyHanniError("Gemstone Slot Unlock Costs failed to load!")
```
hannibal002: maybe add "from neu repo" at the end of line

src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
```diff
@@ -222,12 +223,26 @@ object GetFromSackAPI {
...
+            val uniqueSackItems = mutableSetOf<NEUInternalName>()
+
+            sacksData.values.forEach { sackInfo ->
+                sackInfo.contents.forEach { content ->
+                    uniqueSackItems.add(content)
+                }
+            }
+
+            sackListInternalNames = uniqueSackItems.map { it.asString() }.toSet()
+            sackListNames = uniqueSackItems.map { it.itemNameWithoutColor.uppercase() }.toSet()
```
hannibal002, line 240: Why do we need uppercase here?

hannibal002, line 240: Instead of having `sackListInternalNames` and `sackListNames` separate, how about using one map instead?

Maybe also use a list/map of item names split by _ instead of empty lines, so we don't need to do this here
 https://github.com/hannibal002/SkyHanni/blob/9d696771b9c608c3afee2811dbebb1a3869b9f8f/src/main/java/at/hannibal2/skyhanni/features/commands/tabcomplete/GetFromSacksTabComplete.kt#L19

Thunderblade73 (approved):

hannibal002 (changes requested):
Code cleanup, not tested

src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
CalMWolfs: ask thunderblade

Thunderblade73: we need two set. The code would else not work. Since the behavior of Names and InternalNames handling is completely different do to the underline Replace

CalMWolfs: this should also be something done by thunderblade as idk what this file does i'm just converting from sh repo to neu repo

Thunderblade73: to make it case unsensitve. I made everything uppercase in the checking logic. Since InternalNames normlay are uppercase

hannibal002: Isnt it more intuitive to make the list neu internal name instead of string? We only need to make the user input to a NeuInternalName before comparing

CalMWolfs: > Isnt it more intuitive to make the list neu internal name instead of string? We only need to make the user input to a NeuInternalName before comparing

I feel it would be better to have just one yes, but I also feel this is out of the scope for current pr as im unfamiliar with what most of this pr does so I dont think I should touch too much other code here

Thunderblade73: The list is only used for comparing. And as a set of string it is a just a lookup from a hashmap not a search algorithm

hannibal002: yeah agreed, out of scope of pr

hannibal002 (changes requested):
Running `/gfs` without parameters causes this:
```
SkyHanni 0.23.Beta.24: Caught an IndexOutOfBoundsException at MessageSendToServerEvent in GetFromSackAPI: 'Index: 1, Size: 1'
 
Caused by java.lang.IndexOutOfBoundsException: Index: 1, Size: 1
	at java.util.Collections$SingletonList.get(Collections.java:4815)
	at SH.features.commands.tabcomplete.GetFromSacksTabComplete.handleUnderlineReplace(GetFromSacksTabComplete.kt:30)
	at SH.api.GetFromSackAPI.onMessageToServer(GetFromSackAPI.kt:128)
	at SH.events.LorenzEvent.postAndCatch(LorenzEvent.kt:19)
	at SH.data.ChatManager.onSendMessageToServerPacket(ChatManager.kt:101)
	at SH.events.LorenzEvent.postAndCatch(LorenzEvent.kt:19)
	at SH.mixins.hooks.NetHandlerPlayClientHookKt.onSendPacket(NetHandlerPlayClientHook.kt:8)
```

Thunderblade73:
not a problem of this PR. I now fixed it in #1055

hannibal002 (approved):

PR #1054, opened 2024-02-25
Geyser fixes

Obsidianninja11:
- Added a max distance to geyser box
- Added option to only show geyser box while holding lava rod
- holdingLavaRod and holdingWaterRod in fishingAPI

https://discord.com/channels/997079228510117908/1211104046845530122

PR #1056, opened 2024-02-25
Internal Change: Code cleanup in FishingAPI

hannibal002:
## Dependencies
- #1054

Using repo data to detect NEUInternalName.isFishingRod()

+ Optimized item.isFishingRod logic. - hannibal2


src/main/java/at/hannibal2/skyhanni/features/fishing/FishingAPI.kt
```diff
@@ -52,11 +61,18 @@ object FishingAPI {
...
+        holdingLavaRod = event.newItem.isLavaRod()
+        holdingWaterRod = event.newItem.isWaterRod()
     }
 
-    fun isLavaRod() = InventoryUtils.getItemInHand()?.getLore()?.any { it.contains("Lava Rod") } ?: false
+    @SubscribeEvent
+    fun onRepoReload(event: RepositoryReloadEvent) {
+        val data = event.getConstant<ItemsJson>("Items")
+        lavaRods = data.lava_fishing_rods ?: error("§clava_fishing_rods is missing from repo.")
+        waterRods = data.water_fishing_rods ?: error("§cwater_fishing_rods is missing from repo.")
```
CalMWolfs: We dont need these errors as it will already throw an error and say the repo status isnt ideal if we try to set these lists to a null element

src/main/java/at/hannibal2/skyhanni/features/fishing/trophy/GeyserFishing.kt
```diff
@@ -69,7 +71,11 @@ class GeyserFishing {
 
     @SubscribeEvent
     fun onRenderWorld(event: LorenzRenderWorldEvent) {
-        if (!config.drawBox || !IslandType.CRIMSON_ISLE.isInIsland()) return
+        if (!config.drawBox) return
+        if (!IslandType.CRIMSON_ISLE.isInIsland()) return
+        if (config.onlyWithRod && !FishingAPI.holdingLavaRod) return
+        if (geyser?.distanceToPlayerIgnoreY()!! > 96) return
```
CalMWolfs: don't like this assertion

CalMWolfs (changes requested):
found two things that I think need to be changed

src/main/java/at/hannibal2/skyhanni/features/fishing/FishingAPI.kt
hannibal002: But without those error messages, we don't know what part exactly of the JSON file is wrong

src/main/java/at/hannibal2/skyhanni/features/fishing/trophy/GeyserFishing.kt
hannibal002: me neither.

src/main/java/at/hannibal2/skyhanni/features/fishing/FishingAPI.kt
CalMWolfs: It still produces an error, which will be presented to the user. Also these feel out of place as we don't have many other places where we error if one part of repo is missing. And if we do i'd like to remove those as they feel out of place. Getting an error that says something in ItemsJson is not correct should be all we need to debug and I feel the error would even say which part of the repo doesnt exist

hannibal002: the error doesnt know anything about the json itself. but the error will show the stack trace, pointing to this line. so its fine and your arguments about consistency are valid.

src/main/java/at/hannibal2/skyhanni/features/fishing/trophy/GeyserFishing.kt
hannibal002: fixed

CalMWolfs (approved):
lgtm

PR #1057, opened 2024-02-25
Fixed creating too many timers

hannibal002:
Fixed creating too many skill track timers.


lunaynx:
Looks good to me, so far no more OOMs with this.

PR #1063, opened 2024-02-26
Fix: Box of Seeds not working with /shcroptime

lunaynx:
This is an attempt to fix it in a way that doesn't break the fix made in hannibal002/SkyHanni@8161cf6.


hannibal002 (changes requested):
Does not work for me
image

lunaynx:
Works for me.

`/shcroptime 3 box of seeds`
Screenshot showing crop speed for Box of Seeds

Also works with `3 box` and `3 seeds`.

hannibal002 (approved):
Yeah, this resolves my problems. Thanks

PR #1064, opened 2024-02-26
Feature: Hide Far Entities

hannibal002:
## What
This debug feature increases my fps by 5% to 150%. Depending on the situation and value set.

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added Hide Far Entities. - hannibal2
    * Can perhaps increase FPS for some users by 5% to 150%.
    * Options to change the distance and number of mobs to always show.
    * Option to disable in garden.


src/main/java/at/hannibal2/skyhanni/config/features/dev/HideFarEntitiesConfig.java
```diff
@@ -0,0 +1,18 @@
...
+public class HideFarEntitiesConfig {
+    @Expose
+    @ConfigOption(name = "Enabled", desc = "Hide all entities from rendering except the nearest ones.")
+    @ConfigEditorBoolean
+    public boolean enabled = false;
+
+    @Expose
+    @ConfigOption(name = "Amount", desc = "Keep showing this amount of nearest entities.")
+    @ConfigEditorSlider(minValue = 5f, maxValue = 150f, minStep = 5f)
+    public float amount = 50f;
```
CalMWolfs: Why not just have this be an int straight away?

CalMWolfs (changes requested):
While I feel this is good for performance I feel there still needs to be some stuff like always render if within x blocks unless you cant see it or something. If this is accidentally turned on by a user selecting turn all features on they may think it is a bug with the mod

src/main/java/at/hannibal2/skyhanni/config/features/dev/HideFarEntitiesConfig.java
martimavocado: i'm guessing because sliders sometimes can act weird with non floats

lineargraph: Sliders support ints just fine.

src/main/java/at/hannibal2/skyhanni/features/misc/HideFarEntities.kt
```diff
@@ -0,0 +1,40 @@
...
+import at.hannibal2.skyhanni.utils.CollectionUtils.sorted
+import at.hannibal2.skyhanni.utils.EntityUtils
+import at.hannibal2.skyhanni.utils.LocationUtils.distanceToPlayer
+import at.hannibal2.skyhanni.utils.LorenzUtils
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+class HideFarEntities {
+    private val config get() = SkyHanniMod.feature.dev.hideFarEntities
+
+    private var ignored = emptyList<Int>()
```
Thunderblade73: make it a set

src/main/java/at/hannibal2/skyhanni/features/misc/HideFarEntities.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    @SubscribeEvent
+    fun onTick(event: LorenzTickEvent) {
+        if (isEnabled()) {
+            val maxAmount = config.maxAmount.coerceAtLeast(1)
+            val minDistance = config.minDistance.coerceAtLeast(3)
```
Thunderblade73: why coerce when the config limits it?

Thunderblade73 (changes requested):
So small issues. I do not like the way how ignored is obtained but I also can't find a better solution right now.

src/main/java/at/hannibal2/skyhanni/features/misc/HideFarEntities.kt
hannibal002: The config only limits the moveable bar, not the actual value you can put in the text box.
Also, you could change the values in the config manually. Or with /shconfig. I just want to be sure we never go below those values.

src/main/java/at/hannibal2/skyhanni/features/misc/HideFarEntities.kt
hannibal002: The `.map{}` function returns a list. I would have to make the list to a set. But we already have the list, why not use it directly?
grafik

hannibal002:
@Thunderblade73 Are your objections around the calculation of `ignored` only how the code looks, how we technically do it, or what kind of pseudocode logic we want to use at all?

src/main/java/at/hannibal2/skyhanni/features/misc/HideFarEntities.kt
Thunderblade73: for the render checks. Since a contains for a set is O(1) where a contains for a list is O(n). Sure the toSet operation is not nothing but it runs only every tick where as I understand the renderchecking runs every frame for each entity,

hannibal002: Did not know sets are that much faster compared to lists. In that case, for the current scenario (and looking at the refresh time vs the query time) its useful to create additional sets each refresh)

(Who had thought not saving an index saves "contians" lookups this much)

Thunderblade73:
It is about how we technically do it, readability is good. But let me think about, maybe I find a better solution.

src/main/java/at/hannibal2/skyhanni/features/misc/HideFarEntities.kt
```diff
@@ -0,0 +1,40 @@
...
+
+        val maxAmount = config.maxAmount.coerceAtLeast(1)
+        val minDistance = config.minDistance.coerceAtLeast(3)
+
+        ignored = EntityUtils.getAllEntities()
+            .map { it to it.distanceToPlayer() }
+            .toMap()
+            .filter { it.value > minDistance }
+            .sorted().keys.drop(maxAmount)
+            .map { it.entityId }.toSet()
```
Thunderblade73: I have two versions to improve this. First one is 30% faster and keeps the same readability. Second one is 40% faster and makes a bit of a mess (readability wise).
You choose.
```suggestion
        ignored = EntityUtils.getAllEntities()
            .map { it.entityId to it.distanceToPlayer() }
            .filter { it.second > minDistance }
            .sortedBy { it.second }.drop(maxAmount)
            .map { it.first }.toSet()
```
```suggestion
        ignored = EntityUtils.getAllEntities()
            .fold(mutableListOf<Pair<Int, Double>>())
            { list, it ->
                val distance = it.distanceToPlayer()
                if (distance > minDistance) {
                    list.add(it.entityId to distance)
                }
                list
            }.sortedBy { it.second }.drop(maxAmount)
            .map { it.first }.toSet()
```

src/main/java/at/hannibal2/skyhanni/config/features/dev/HideFarEntitiesConfig.java
Thunderblade73, line 1: Why is this in the dev Categorie?
Isn't it a feature?

hannibal002: good idea. i was thinking of a test thingy. until its actually not breaking stuff. but sure we can have this as an actual feature haha

Thunderblade73: if we make it a feature please make an option to exclude garden. Since there it wouldn't make sense for the feature

hannibal002: Same for private island?

CalMWolfs (changes requested):
With the current implementation of feature toggle turning all off features on, i cannot approve this

hannibal002:
> With the current implementation of feature toggle turning all off features on, i cannot approve this

this has changed now.

CalMWolfs (approved):
worked for me, doubled fps on showing all within 20 blocks and at most 50 entities, but maybe the default values are a little low

PR #1065, opened 2024-02-27
Backend: Remove log error

CalMWolfs:
## Dependencies
- #1028

Features:
- Removes usage of `logError` everywhere in the code and replaces it with `logErrorWithData`
- Removed `logErrorState` as it is no longer used in existing code and was rarely used in old code

+ Removes usage of logError everywhere in the code and replaces it with logErrorWithData. - CalMWolfs
+ Removed ErrorManager.logErrorState(). - CalMWolfs


src/main/java/at/hannibal2/skyhanni/features/garden/farming/FarmingWeightDisplay.kt
```diff
@@ -431,7 +431,6 @@ class FarmingWeightDisplay {
                     "url" to url,
                     "apiResponse" to apiResponse
                 )
-                error()
```
hannibal002, line 434: Why do we no longer this error message here?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/farming/FarmingWeightDisplay.kt
CalMWolfs: They already get the regular error and stacktrace and this api request isnt necessary unlike the other one, which is why the other one still tells them to rejoin the garden

hannibal002: 👍

hannibal002 (approved):

PR #1066, opened 2024-02-27
Fix: Anita Shop's Medal Profit reports the wrong items

martimavocado:


src/main/java/at/hannibal2/skyhanni/features/garden/AnitaMedalProfit.kt
```diff
@@ -6,19 +6,13 @@ import at.hannibal2.skyhanni.events.InventoryCloseEvent
 import at.hannibal2.skyhanni.events.InventoryFullyOpenedEvent
 import at.hannibal2.skyhanni.features.garden.visitor.VisitorAPI
 import at.hannibal2.skyhanni.test.command.ErrorManager
-import at.hannibal2.skyhanni.utils.ChatUtils
+import at.hannibal2.skyhanni.utils.*
```
CalMWolfs: This shouldn't have a wildcard import, make sure you are using the correct editor config

src/main/java/at/hannibal2/skyhanni/features/garden/AnitaMedalProfit.kt
```diff
@@ -76,7 +86,7 @@ class AnitaMedalProfit {
     }
 
     private fun readItem(item: ItemStack, table: MutableMap<Pair<String, String>, Pair<Double, NEUInternalName>>) {
-        val itemName = item.nameWithEnchantment ?: return
+        var itemName = item.nameWithEnchantment ?: return
```
CalMWolfs: 'nameWithEnchantment: String?' is deprecated. outdated

src/main/java/at/hannibal2/skyhanni/features/garden/AnitaMedalProfit.kt
```diff
@@ -87,7 +97,14 @@ class AnitaMedalProfit {
 
         val (name, amount) = ItemUtils.readItemAmount(itemName) ?: return
 
-        var internalName = NEUItems.getInternalNameOrNull(name)
+        var internalName: NEUInternalName? = null
+        for ((wrongName, correctName) in farmingToolIDs) {
+            if (wrongName == NEUItems.getInternalNameOrNull(name)) internalName = correctName
```
CalMWolfs: 'getInternalNameOrNull(String): NEUInternalName?' is deprecated. moved

src/main/java/at/hannibal2/skyhanni/features/garden/AnitaMedalProfit.kt
```diff
@@ -87,7 +97,14 @@ class AnitaMedalProfit {
...
+        }
+        for ((wrongName, correctName) in farmingToolNames) {
+            if (itemName == wrongName) itemName = correctName
+        }
```
CalMWolfs: Instead of having a second loop here, in the previous loop when it is found also get the display name with `correctName.itemName` this ensures that if hypixel changes some display format in the future none of this breaks

CalMWolfs (changes requested):
a few changes needed

hannibal002:
I do not like the extra hash map. The previous solution was working fine until I changed the logic behind `nameWithEnchantment`. I have added a simple workaround to fix this instead: https://github.com/hannibal002/SkyHanni/commit/694d5334afbc3bfed8baf6e4f20de9c322f3423e

PR #1072, opened 2024-02-27
Backend: EntityClick cancellability

Thunderblade73:
Fixes the not working cancellability from the entityclick.
Also cleanup of the ItemClickData

+ Fix entity click not being canceled - Thunderblade73
+ Cleanup ItemClickData. - Thunderblade73


Thunderblade73:
does not need to be in 0.23. Since nothing uses it in 0.23

hannibal002 (approved):

PR #1073, opened 2024-02-28
Fix: Weight display

CalMWolfs:
- I dont really have high enough weight to properly test this so would be good if someone that does could try this. Also i just removed some code about something and idk if it causes error but also cant really test that :| todo with profile switching

Also converted the api requests to use gson serialization and added better error messages to them
Also removed unusued stuff in ProfileStorageData as I saw them when investigating this issue


hannibal002 (approved):
works fine in game for me

hannibal002 (approved):

PR #1078, opened 2024-02-29
Internal Change: String to number final changes

hannibal002:
+ Creating number utils functions for string -> int, long, double, with either error throwing or wrong usage to user message. - hannibal2


CalMWolfs (approved):
looks fine

PR #1079, opened 2024-02-29
Improve performance of removeColor

lineargraph:


src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -34,30 +34,61 @@ object StringUtils {
...
+            // If the next formatting sequence's code indicates a non-color format and we should keep those
+            if (keepFormatting && nextFormattingSequence + 1 < length && this[nextFormattingSequence + 1] in formattingChars) {
+                // Set the readIndex to the formatting indicator, so that the next loop will start writing from that paragraph symbol
+                readIndex = nextFormattingSequence
+                // Find the next § symbol after the formatting sequence
+                nextFormattingSequence = indexOf('§', startIndex = readIndex + 1)
+            } else {
+                // If this formatting sequence should be skipped (either a color code, or !keepFormatting or an incomplete formatting sequence without a code)
+                // Set the readIndex to after this formatting sequence, so that the next loop will skip over it before writing the string
+                readIndex = nextFormattingSequence + 2
```
Thunderblade73, line 79: When using the suggestion (and deleting the out of bound check below) there is the possibility to remove a branch. Does not require to use min could simply be a the bound if from below. I'm unsure what is more performant since technically min could be implemented branchless (but unlikely). Either way move the boundscheck inside here, since the other branch can't get out of bounds
```suggestion
                readIndex = min(nextFormattingSequence + 2, this.length)
```

hannibal002: `(nextFormattingSequence+2).coerceAtLeast(this.length)`

Thunderblade73: btw. it is coerceAtMost 
But this is literally the wrapper for the boundscheck below. Min *could* be implemented better performant then a branch. And since performance is important here (else we could simply use the old removeColor) we should use the more optimal method for this, any other case I would use coerceAtMost here for readability.

lineargraph: > And since performance is important here (else we could simply use the old removeColor) we should use the more optimal method for this, any other case I would use coerceAtMost here for readability.

I have to strongly disagree. While performance is the matter of this PR, there is a difference between significant and measurable improvements and potentially maybe saving time on the order of a few clock cycles at best. "We can discard an improvement of more than 50% because we aren't the improvement that potentially saves us less than 1% also i haven't done any measurements on this" is an idiotic take.

lineargraph: I also want to point out here that the compiler is king for small stuff like this. Optimizing and inlining functions like this is often something the compiler is very good at. Very likely that whatever intrinsic Math.min has, there is a similar optimization for a simple tertiary in JIT. In fact i could make an argument (just as bogus, *especially* without benchmarking) that Math.min has worse performance, since calling the function requires overhead for the stack of the Math.min call to be allocated (which is a quite a few instructions). Sure you could inline the function, but even when inlining functions certain attributes about stacktraces must be preserved, meaning that there is still some runtime stacktrace overhead. In practice none of this matters without actual benchmarking.

Thunderblade73: Someone could go this way, but any reasonable person would say no. I suggested the move of the bounds check and also a way to write it (```min```). I do not care if it is a ```min```, a ```coerceAtMost```, a ```if```, a ```trinary``` or some bit operation to get the same result. I would maybe care if I tested and found a definite best but since I don't have any good testing environment and I can't be bothered to setup one for this thing, I do not care. The ```min``` is just what I know as the best way to do it normally (outside of java).

Thunderblade73: Also approve your commit (5582cd0254ac61b92e955f41fee640abd8d2c553). 
And thx for the testing.

src/test/java/test.kt
Thunderblade73, line 1: test class is still here

Thunderblade73 (approved):

src/test/java/test.kt
lineargraph: This is what happens when i use intellij for commits.

hannibal002 (approved):
My amateurish tests with real liife data cant confirm or deny the actual benefit of the new logic. at least the stats tell me the factor is not over 100% off of the current logic. Worst situation: we have a twice as high load, but this still doesnt matter. I believe in your logic.

hannibal002:
please dont ask

PR #1091, opened 2024-03-02
Fix: Make crafted minions a set instead of a list

CalMWolfs:
Not sure how it is still getting filled up more than it should but this will prevent any future things from breaking this as well

## Changelog Fixes
+ Fixed Bingo Minion Craft Helper sometimes causing performance problems. - CalMWolfs

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/bingo/FirstMinionTier.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/bingo/FirstMinionTier.kt
@@ -1,5 +1,6 @@
 package at.hannibal2.skyhanni.features.bingo
 
+import at.hannibal2.skyhanni.test.command.ErrorManager
 import at.hannibal2.skyhanni.utils.ItemUtils.name
 import at.hannibal2.skyhanni.utils.NEUInternalName
 import at.hannibal2.skyhanni.utils.NEUInternalName.Companion.asInternalName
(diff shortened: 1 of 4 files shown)


src/main/java/at/hannibal2/skyhanni/features/bingo/MinionCraftHelper.kt
```diff
@@ -283,14 +281,21 @@ class MinionCraftHelper {
             }
             element
         }
+        event.transform(26, "#player.bingoSessions") { element ->
```
hannibal002, line 278: We could delete the transform call above that got added at config version 19. Not sure if we want to set a president in deleting migrations this way, but the result is the very same, right?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/bingo/MinionCraftHelper.kt
CalMWolfs: I agree in not deleting migrations, but as these migrations are the same and don't actually move anything this is completely fine

src/main/java/at/hannibal2/skyhanni/features/bingo/FirstMinionTier.kt
```diff
@@ -51,19 +52,28 @@ object FirstMinionTier {
...
+        tierOneMinionsDone: MutableSet<NEUInternalName>,
     ) {
         for (minionId in tierOneMinionsFiltered) {
             val prefix = minionId.asString().dropLast(1)
             if (minions.any { it.value.startsWith(prefix) }) {
-                tierOneMinionsDone.add(minionId)
+                val successful = tierOneMinionsDone.add(minionId)
+                if (!successful) {
+                    ErrorManager.logErrorWithData(
+                        IllegalStateException("Attempted to add $minionId to tierOneMinionsDone when it already exists"),
```
hannibal002, line 63: Attempted -> Failed

hannibal002 (approved):

PR #1096, opened 2024-03-02
refactor melody quick restart

Zickles:
## Changelog Fixes
+ Fixed Harp Quick Restart not working when Harp GUI Scale is disabled. - Zickles


hannibal002:
Please give your PRs descriptions/titles that help us understand what you are doing without immediately looking at the code.
(questions like "is this a bug", is this just code cleanup", etc)

CalMWolfs (changes requested):
I feel that the when statements here can be removed and instead you can put and statements inside of the if statements so the code is more readable

e.g. ` if (config.quickRestart && isMenuGui(event.inventoryName) ) {`

Zickles:
> Please give your PRs descriptions/titles that help us understand what you are doing without immediately looking at the code. (questions like "is this a bug", is this just code cleanup", etc)

Sorry about that swear i added one tbh

hannibal002 (approved):

PR #1102, opened 2024-03-04
Change: Add Goal to CollectionTracker

Thunderblade73:
# What
Adds a goal value to the tracker. The goal is fully optional so when the argument of the goal amount is left empty it looks the same as before.
## Preview
image

## Changelog Improvements
+ Added option to add a collection track goal. - Thunderblade73
    * /shtrackcollection <item name> [goal amount]
    * Shows a chat message once the goal has been reached.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/CollectionTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/CollectionTracker.kt
@@ -44,7 +44,7 @@ class CollectionTracker {
         fun command(args: Array<String>) {
             if (args.isEmpty()) {
                 if (internalName == null) {
-                    ChatUtils.userError("/shtrackcollection <item name> ?<goal amount>")
+                    ChatUtils.userError("/shtrackcollection <item name> [goal amount]")
                     return
                 }
                 ChatUtils.chat("Stopped collection tracker.")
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/features/misc/CollectionTracker.kt
```diff
@@ -40,15 +44,30 @@ class CollectionTracker {
         fun command(args: Array<String>) {
             if (args.isEmpty()) {
                 if (internalName == null) {
-                    ChatUtils.userError("/shtrackcollection <item name>")
+                    ChatUtils.userError("/shtrackcollection <item name> ?<goal amount>")
```
hannibal002: use [] as format for optional parameters instead of ?<>

src/main/java/at/hannibal2/skyhanni/features/misc/CollectionTracker.kt
```diff
@@ -121,18 +141,22 @@ class CollectionTracker {
...
-            val format = LorenzUtils.formatInteger(itemAmount)
+            val format = itemAmount.addSeparators()
 
             var gainText = ""
             if (recentGain != 0) {
-                gainText = "§a+" + LorenzUtils.formatInteger(recentGain)
+                gainText = "§a+" + recentGain.addSeparators()
             }
 
+            val goal = if (goalAmount == -1L) "" else " §f/ §b$goalAmount §f(§a${
```
hannibal002: add `goalAmount.addSeparators()`

hannibal002 (changes requested):
looks fine in game. maybe removing the goal info when the goal is reached and send a chat message?

hannibal002 (approved):

PR #1107, opened 2024-03-05
Backend: Use more isInIsland

CalMWolfs:
## What
- Used more isInIsland()
- Made the config in reputation helper private in favour of using the existing config variables in the other classes

+ Used the isInIsland function more. - CalMWolfs

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/test/DebugCommand.kt
+++ b/src/main/java/at/hannibal2/skyhanni/test/DebugCommand.kt
@@ -6,6 +6,7 @@ import at.hannibal2.skyhanni.data.IslandType
 import at.hannibal2.skyhanni.events.DebugDataCollectEvent
 import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.LorenzUtils
+import at.hannibal2.skyhanni.utils.LorenzUtils.isInIsland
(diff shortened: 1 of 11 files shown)


src/main/java/at/hannibal2/skyhanni/test/DebugCommand.kt
```diff
@@ -102,7 +103,7 @@ object DebugCommand {
             event.addData("not on SkyBlock, but on Hypixel")
             return
         }
-        if (LorenzUtils.skyBlockIsland == IslandType.UNKNOWN) {
+        if (IslandType.UNKNOWN.isInIsland()) {
```
hannibal002: Im against this change here. its not a real island, making the wording "is in island" wrong. lets keep the original islandtype == unknown here

hannibal002 (changes requested):
Changes do look fine. not tested in game

src/main/java/at/hannibal2/skyhanni/test/DebugCommand.kt
CalMWolfs:
```suggestion
        if (LorenzUtils.skyBlockIsland == IslandType.UNKNOWN) {
```

hannibal002 (approved):

PR #1114, opened 2024-03-06
Backend: Use less item name

CalMWolfs:
## What
Use internal name more in ItemDisplayOverlayFeatures

+ Use internal name more in ItemDisplayOverlayFeatures - CalMWolfs


hannibal002 (approved):
Can confirm everything except BINGO_GOAL_RANK works.

PR #1124, opened 2024-03-08
add feature toggle to full auto update

CalMWolfs:
## What
Made it a feature toggle.
This does not need to be in the changelog but idk if the program can exclude it

exclude_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/About.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/About.java
@@ -1,5 +1,6 @@
 package at.hannibal2.skyhanni.config.features;
 
+import at.hannibal2.skyhanni.config.FeatureToggle;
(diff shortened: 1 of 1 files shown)


hannibal002:
This does not need to be in the changelog but idk if the program can exclude it

Right now it can't exclude prs. When you don't include the changelog part in the comments, the vhangelog Bilder will mark this pr as error. We can currently set a flag to also create the changelots even if some prs are with errors, but for the future we can simply add a keyword or formatting to mark this pr as excluded from changelog

CalMWolfs:
> Right now it can't exclude prs. When you don't include the changelog part in the comments, the vhangelog Bilder will mark this pr as error. We can currently set a flag to also create the changelots even if some prs are with errors, but for the future we can simply add a keyword or formatting to mark this pr as excluded from changelog

I think it would be good to have a flag to not include it in changelog for small code cleanup or fixes to things that don't need to be part of the changelog

PR #1123, opened 2024-03-08
Improvement: Load Trophy Fishing from NEU

hannibal002:
## What
Loading Trophy Fish information from NEU PV via `ProfileDataLoadedEvent`.
image

## Changelog Improvements
+ Loading Trophy Fish information from NEU PV. - hannibal2 & CalMWolfs
## Changelog Technical Details
+ Added NeuProfileDataLoadedEvent. - hannibal2

PR #1128, opened 2024-03-09
Backend: ChatClickActionManager

hannibal002:
## Dependencies
- #1123

## What
Utils function for sending chat message to the user that they can click and run an action
also directly using this feature at "load trophy fish data from neu pv"
image

+ Added StringUtils.generateRandomId(). - hannibal2
+ Added ChatUtils.clickableChat support for runnable action. - hannibal2


src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
```diff
@@ -465,6 +466,7 @@ class SkyHanniMod {
         loadModule(BossbarData)
         loadModule(EntityUtils)
         loadModule(ChatUtils)
+        loadModule(ChatClickActionManager)
```
CalMWolfs: pretty sure this is only needed if the class/object needs to be registered for events or for repo patterns

CalMWolfs (changes requested):
will review the actual getting from pv when i wake up but just saw this

CalMWolfs (approved):
worked for me, also cleaned up extra lines

PR #1140, opened 2024-03-11
Improvement: Allow teleporting when in an infested plot

Obsidianninja11:
## What
Adds a toggle to allow teleporting with the pest teleport hotkey while already in an infested plot.

## Changelog Improvements
+ Allow using the teleport hotkey when in an infested plot. - Obsidian


hannibal002 (changes requested):
I like the option, but the name and description is a bit confusing imo. maybe toggle the logic, and make it default enabled. (block while in invested block)

Obsidianninja11:
> I like the option, but the name and description is a bit confusing imo. maybe toggle the logic, and make it default enabled. (block while in invested block)

Idk what i'd name it with this

CalMWolfs (approved):
code looks good

hannibal002 (approved):

PR #1150, opened 2024-03-12
Backend: Widget abstractions

Thunderblade73:
## Dependencies
- #1733 
## What
Wraps all widgets under an enum (I have a solution to update it remotely see 3b5edd80153922c5f2ce344a5b2a1a4b740c28b8).
It also adds an widget update event that should reduce the usage of checking the whole tablist for every feature.

Widget info is also added to shcopytablist

+ Added Tab Widget abstraction. - Thunderblade73


Thunderblade73:
open to review but garden patterns are still missing. Since Garden is closed at the moment.

Thunderblade73:
@nea89o Would you mind taking a look at this PR and providing your feedback?

src/main/java/at/hannibal2/skyhanni/utils/repopatterns/RepoPatternManager.kt
```diff
@@ -156,4 +156,10 @@ object RepoPatternManager {
         }
         return RepoPatternImpl(fallback, key).also { usedKeys[key] = it }
     }
+
+    fun getUnusedPatterns(prefix: String): List<Pattern> {
```
lineargraph: Can you add a javadoc here indicating that this is a dangerous method (since it circumvents the exclusivity map)?

src/main/java/at/hannibal2/skyhanni/utils/repopatterns/RepoPatternManager.kt
```diff
@@ -156,4 +156,10 @@ object RepoPatternManager {
...
+        if (config.forceLocal.get()) return emptyList()
```
lineargraph: Maybe verify the prefix here to be of the normal pattern plus a `.`

lineargraph (changes requested):

hannibal002:
Is somethign already using this?

Thunderblade73:
Nope, only listed in copytablist

Thunderblade73:
> Is something already using this?
See #1240 for that

Thunderblade73:
Can a maintainer move this to 0.26.
I won't be able to fix the issue with the Pr before 0.25 release.

CalMWolfs (approved):
worked for me

hannibal002 (changes requested):
Two problems:

<details>
<summary>Event too complex</summary>

The current system needs this workaround:

<details>
<summary>Image</summary>

image

</details>

(See https://github.com/hannibal002/SkyHanni/pull/1977 for more info)

I'd suggest, we remove `NewValues` and `Clear` and instead fire `Event(EnumType, List)`, with either changed values or an empty list if not found.

</details>

<details>
<summary>No event when no widget</summary>

Another problem is that the event is not getting sent when the widget is disabled when word switch. I don't want to assume the widgets are always disabled if the event is not sent (and then show an "open /tab and enable widget" display for the first second after world switch).

</details>

CalMWolfs (approved):
changes lgtm

hannibal002 (approved):

PR #1155, opened 2024-03-12
Improvement: Disable Lane Switch notifications while flying

lunaynx:
## What
Disable lane switch notifications while flying.

## Changelog Improvements
+ Disable lane switch notifications while flying. - Alexia Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/farming/LaneSwitchNotification.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/farming/LaneSwitchNotification.kt
@@ -15,6 +15,7 @@ import at.hannibal2.skyhanni.utils.RenderUtils.renderString
 import at.hannibal2.skyhanni.utils.SimpleTimeMark
 import at.hannibal2.skyhanni.utils.SoundUtils
 import at.hannibal2.skyhanni.utils.SoundUtils.playSound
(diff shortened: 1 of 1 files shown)


ItsEmpa:
not a huge fan of this, what if at some point theres a farm design you need to fly in? i feel like it would be better to disable the notifications if you arent holding any farming tool

hannibal002:
I thought we already checked this by isFarming (> 0 bps).
Maybe we should check in isFarming if the time since farming is greater than one second or something.

Maybe users want to fly through their farms, then this feature should work fine as well imo.
Not that we actively support fly farming, i just dont think this is the right approach to fixing this

lunaynx:
Flying has never been meta since the old lazy farm days. It's messy because you don't accelerate to full speed immediately and you can't really control your speed beyond sprinting or not sprinting. It's highly unlikely for it to become meta again, and if it somehow does, we can change it then.

Simply checking for holding a farming tool could still lead to false notifications while people are trying to get positioned, especially since the crop start location only shows while holding a tool, for example.

ItsEmpa:
well then probably like hanni said, using the isCurrentlyFarming() check

hannibal002:
No longer necessary since https://github.com/hannibal002/SkyHanni/pull/1208 fixed those problems

PR #1157, opened 2024-03-13
Improvement: Show Town Board with accepted Rescue Mission quest

lunaynx:
## What
Show Town Board when there is an accepted Rescue Mission quest, since you need to talk to an NPC at the Town Board before starting the rescue.

## Changelog Improvements
+ Show Town Board waypoint when there is an accepted Rescue Mission quest. - Alexia


hannibal002 (approved):
looks fine to me

PR #1165, opened 2024-03-14
Backend: Create NEUItems.getPrimitiveMultiplier

hannibal002:
## What
Using PrimitiveItemStack at getMultiplier

+ Using PrimitiveItemStack at getMultiplier. - hannibal2


Thunderblade73 (approved):
Looks fine. There is actual no logic change here so nothing should break. Not tested but the code is semantically the same.

CalMWolfs (approved):

PR #1167, opened 2024-03-14
Fix: crimson isle detection after tab list widget

hannibal002:
## What
fixed crimson isle quests detection after tab list widget

## Changelog Fixes
+ Fixed wrong crimson isle quests detection after tab list widget update. - hannibal2


CalMWolfs (approved):
lgtm, though i feel this is more of a bandaid fix to code that can be improved

lunaynx:
This looks wrong, the color won't always be yellow, only for accepted but not completed quests.

Also worth noting that the non-accepted quest color is now different for Barbarian and Mage. I'll get the exact color codes in a bit.

src/main/java/at/hannibal2/skyhanni/features/nether/reputationhelper/CrimsonIsleReputationHelper.kt
```diff
@@ -42,6 +43,11 @@ class CrimsonIsleReputationHelper(skyHanniMod: SkyHanniMod) {
     private var display = emptyList<List<Any>>()
     private var dirty = true
 
+    val tabListQuestPattern by RepoPattern.pattern(
+        "crimson.reputation.tablist",
+        " §r§d.*"
```
lunaynx:
```suggestion
        " §r§[cdea].*"
```
This should work.
* `c` - Barbarian Not Accepted
* `d` - Mage Not Accepted
* `e` - Accepted
* `a` - Completed

PR #1171, opened 2024-03-14
Backend: Storage Config Moving

hannibal002:
## What
extracted player and profile specific storage into their own classes in a new packet
This will cause many merge conflicts with new features, but they should be easy to fix.

+ Extracted player and profile specific storage into their own classes in a new packet. - hannibal2


CalMWolfs (approved):
Didn't seem to reset my config on the profile I tested it on :)

PR #1176, opened 2024-03-14
Fix: Revert shader version for Mac compatibility

VixidDev:
## What
Reverts shader versions back to 120 since on Mac, LWJGL uses the legacy profile for the OpenGL context by default, which only supports up to GLSL version 1.20

+ Revert shader version for Mac compatibility. - Vixid

### DIFF
--- a/src/main/resources/assets/skyhanni/shaders/rounded_rect.fsh
+++ b/src/main/resources/assets/skyhanni/shaders/rounded_rect.fsh
@@ -1,4 +1,4 @@
-#version 130
+#version 120
 
 uniform float scaleFactor;
 uniform float radius;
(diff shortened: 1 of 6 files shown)


hannibal002 (approved):
I don't understand what those fsh scripts do precisely and I can't test this bug (windows moment), but the changes look fine.

VixidDev:
> I don't understand what those fsh scripts do precisely and I can't test this bug (windows moment), but the changes look fine.

The .fsh files are just GLSL fragment shaders, (https://www.khronos.org/opengl/wiki/Fragment_Shader if you care for more info), and yes I can't test this either since also windows moment, so it would be useful to have someone on mac test the artifact from this just to double check it works.

PR #1178, opened 2024-03-15
Improvement: Clear plot sprays when Portable Washer is used

lunaynx:
## What
Clear plot sprays when Portable Washer is used.

## Changelog Improvements
+ Clear plot sprays when Portable Washer is used. - Alexia Luna

+ Added a `removeSpray` method to `Plot`. - Alexia Luna


src/main/java/at/hannibal2/skyhanni/features/garden/GardenPlotAPI.kt
```diff
@@ -97,7 +101,7 @@ object GardenPlotAPI {
         getData()?.apply { sprayHasNotified = true }
     }
 
-    private fun Plot.setSpray(spray: SprayType, duration: Duration) {
+    private fun Plot.setSpray(spray: SprayType?, duration: Duration) {
```
hannibal002: instead of nullable, create a new function `removeSpray` that doesnt have duration as parameter

src/main/java/at/hannibal2/skyhanni/features/garden/GardenPlotAPI.kt
```diff
@@ -33,6 +33,10 @@ object GardenPlotAPI {
         "spray.target",
         "§a§lSPRAYONATOR! §r§7You sprayed §r§aPlot §r§7- §r§b(?<plot>.*) §r§7with §r§a(?<spray>.*)§r§7!"
     )
+    private val portableWasherPattern by patternGroup.pattern(
+        "spray.cleared",
```
hannibal002: in the future there could be other means to clear a spray. use a more precise pattern key. e.g. spray.cleared.portablewasher

src/main/java/at/hannibal2/skyhanni/features/garden/GardenPlotAPI.kt
```diff
@@ -155,6 +159,14 @@ object GardenPlotAPI {
 
             plot?.setSpray(spray, 30.minutes)
         }
+
+        portableWasherPattern.matchMatcher(event.message) {
+            for (plot in plots) {
+                if (plot.currentSpray != null) {
```
hannibal002, line 255: no need for this extra check. just reset everything

hannibal002 (changes requested):
not tested in game, just code clenaup ideas

src/main/java/at/hannibal2/skyhanni/features/garden/GardenPlotAPI.kt
lunaynx: I added this because otherwise it notified that the spray was cleared on all plots from 1 to 24. But we're going to be removing that anyway in favor of Hypixel's own notification, so I guess it can be removed.

hannibal002: Oh, I have overseen this extra behavior.

lunaynx: I've kept this for now because it seems like we have not removed the notification yet.

hannibal002 (approved):

PR #1179, opened 2024-03-15
Fix: Entity Render Event

hannibal002:
## What
Fixed entity render event not hiding on cancel.
This caused the damage indicator to no longer block vanilla names nearby.
This also breaks other features like terracotta hider, ashfang minis hider, teleport pad name hider, and every other feature that uses `SkyHanniRenderEntityEvent` to hide entities.

## Changelog Fixes
+ Fixed Damage Indicator not hiding vanilla names. - hannibal2


CalMWolfs (approved):
Worked

PR #1146, opened 2024-03-12
Backend: Improved the data loading to use kotlin serialisation and fixed bug

CalMWolfs:
## What
Felt like too much of a change to just merge on top of your existing code.
- Fixes issue in the existing pr that doesn't save the data of trophy fish if you have not caught at least 1 of them with SkyHanni yet

## Changelog Technical Details
+ Added JSON objects for Hypixel player API data, which can be used for other stuff later. - CalMWolfs

PR #1191, opened 2024-03-16
Backend: Remove Neu Constant

Thunderblade73:
## What
Removes the direct use of the Neu Constants. Uses instead the NeuRepositoryReloadEvent. This include the usage of more JsonObject instead of a custom Json parser.

Also cleanup of SkillAPI & SkillUtil. Since the used direct JsonObject parsing for some reason instead of a List, 

+ Removed use of notenoughupdates.util.Constants. - Thunderblade73


Thunderblade73:
@CalMWolfs now it is ready to be reviewed

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/neu/NeuReforgeStoneJson.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    @Expose
+    val internalName: String,
+
+    @Expose
+    val reforgeName: String,
+
+    @Expose
+    @SerializedName("itemTypes")
+    val rawItemTypes: Any,
```
CalMWolfs: I don't particullarly like the idea of any but if it can't be changed its fine to keep. I dont think this is used in neu/skyhanni so it would be fine to change the format in neu repo to maybe have itemTypes lead to a map of itemTypes -> list of types and internalNames -> list of internal names. If wanted for extra precaution for not giving users with broken repos NPEs you could change "itemTypes" in neu repo to something else e.g. rawItemTypes.

This same logic can be applied to reforgeAbility where instead of having it be either a string or map of rarities to abilities you could have it map a general ability and then a per rarity one. But i think there is a better way to do this like maybe with a custom TypeAdapter like i do in #1146

CalMWolfs (changes requested):
I like the banning of the import. If it is possible for you to change the format of the reforgeStones json so it doesn't have Anys and instead uses something more specific i feel that would be good but don't spend too long on that. if you have any question about it dm me.

Two things,
1) for formatting you can format
```kt
   @Expose
    @SerializedName("itemTypes")
    val rawItemTypes: Any,
```
as
```kt
   @Expose @SerializedName("itemTypes") val rawItemTypes: Any,
```

2) due to this being part of  neu repo and not skyhanni repo even though i doubt that there can be changes that will break skyhanni i think the readConstant function should have its own try catch and when an error occurs it adds the raw json text as one of the bits of data.

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/neu/NeuReforgeStoneJson.kt
```diff
@@ -0,0 +1,70 @@
...
+    @Expose @SerializedName("itemTypes") val rawItemTypes: Any,
+
+    @Expose val requiredRarities: List<LorenzRarity>,
+
+    @Expose val reforgeCosts: Map<LorenzRarity, Long>,
+
+    @Expose val reforgeStats: Map<LorenzRarity, Map<String, Double>>,
+
+    @Expose @SerializedName("reforgeAbility") val rawReforgeAbility: Any?,
+) {
```
CalMWolfs:
```suggestion
data class NeuReforgeStoneJson(
    @Expose val internalName: NEUInternalName,
    @Expose val reforgeName: String,
    @Expose @SerializedName("itemTypes") val rawItemTypes: Any,
    @Expose val requiredRarities: List<LorenzRarity>,
    @Expose val reforgeCosts: Map<LorenzRarity, Long>,
    @Expose val reforgeStats: Map<LorenzRarity, Map<String, Double>>,
    @Expose @SerializedName("reforgeAbility") val rawReforgeAbility: Any?,
) {
```

CalMWolfs (changes requested):
the rest of the code seemed to work for me just this one formatting change

CalMWolfs (approved):

PR #1195, opened 2024-03-16
Fix: OpenBrowser Crash

hannibal002:
## What
Fixed crashing when openBrowser gets called from inside a moulconfig button.

## Changelog Fixes
+ Fixed game crash when open browser button in config fails. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/test/command/ErrorManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/test/command/ErrorManager.kt
@@ -115,7 +115,7 @@ object ErrorManager {
     ) {
         if (!ignoreErrorCache) {
             val pair = if (throwable.stackTrace.isNotEmpty()) {
-                throwable.stackTrace[0].let { (it.fileName ?: "null") to it.lineNumber }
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/test/command/ErrorManager.kt
```diff
@@ -113,18 +113,17 @@ object ErrorManager {
...
         vararg extraData: Pair<String, Any?>,
     ) {
-        val error = Error(message, throwable)
-        error.printStackTrace()
-        Minecraft.getMinecraft().thePlayer ?: return
-
         if (!ignoreErrorCache) {
             val pair = if (throwable.stackTrace.isNotEmpty()) {
-                throwable.stackTrace[0].let { it.fileName!! to it.lineNumber }
+                throwable.stackTrace[0].let { (it.fileName ?: "null") to it.lineNumber }
```
lineargraph: I am pretty sure the standard way to present an absent fileName would be `"<unknown>"`

CalMWolfs (approved):
can't reproduce the crash in game but code looks like it addresses it and adds more error logging

PR #1196, opened 2024-03-16
Fix: SuperBoom TNT wrong itemNameQuery

Thunderblade73:
## What
Solve the wrong item query of "SUPERBOOM TNT" (btw. "Superboom TNT" would work).
The issue is on NEUs side but I wouldn't know a solution, since the mangelnd logic is fundamental wrong as for this specific case

## Discord
https://discord.com/channels/997079228510117908/1218607214135083180

## Changelog Fixes
+ Fixed Superboom TNT not working with Queued GFS. - Thunderblade73


CalMWolfs (approved):
yeah i guess this is just a weird case with that, you could add a small comment to explain it

hannibal002 (approved):

PR #1197, opened 2024-03-16
Feature: Architect Notifications when puzzle fails

Conutik:
## What
Sends a title screen and a chat notification overwriting the one for puzzle fail, reminding the user to use an architect, and provides a button that can be clicked to retrieve it from sacks.

<details>
<summary>Demonstration</summary>
https://conutik.catto.pics/748fad3b
</details>

## Changelog New Features
+ Added notifications for architect on puzzle fail. - Conutik
    * Shows Title.
    * Shows button in chat to retrieve from sack.
    * Only works when having enough Architect First Drafts in the sack.

+ Fixed all regexes in DungeonCopilot.kt. - Conutik
    * Now uses `RepoPattern` with good key naming conventions.


CalMWolfs (changes requested):
I feel this feature should be a sepearate feature outside of the dungeon co-pilot as afaik this feature is not fully finished and impacts performance a lot, I'll do further reviews once this is changed

src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonCopilot.kt
```diff
@@ -12,21 +12,39 @@ import at.hannibal2.skyhanni.events.LorenzWorldChangeEvent
...
 import at.hannibal2.skyhanni.utils.RenderUtils.renderString
 import at.hannibal2.skyhanni.utils.StringUtils.matchMatcher
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
 import net.minecraft.entity.item.EntityArmorStand
 import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
 
 class DungeonCopilot {
 
-    private val config get() = SkyHanniMod.feature.dungeon.dungeonCopilot
+    private val config get() = SkyHanniMod.feature.dungeon
```
CalMWolfs: this change can be reverted now i think

src/main/java/at/hannibal2/skyhanni/config/features/dungeon/DungeonConfig.java
```diff
@@ -56,6 +56,12 @@ public class DungeonConfig {
     @FeatureToggle
     public boolean highlightTeammates = true;
 
+    @Expose
+    @ConfigOption(name = "Architect Notifier", desc = "Notifies you to use the Architect in Dungeons when a puzzle is failed.")
+    @ConfigEditorBoolean
+    @FeatureToggle
+    public boolean architectNotifier = false;
```
CalMWolfs:
```suggestion
    public boolean architectNotifier = true;
```
I think this is fine to auto enable

src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonArchitectFeatures.kt
```diff
@@ -0,0 +1,57 @@
...
+        quizPuzzleFailPattern.matchMatcher(event.message) {
+            val key = group("name")
+            ChatUtils.clickableChat(
+                "§c§lPUZZLE FAILED! §r§b$key §r§ehas failed a puzzle.\n§3§l[CLICK HERE TO GET ARCHITECT]",
+                "/gfs ARCHITECT_FIRST_DRAFT 1",
+                false
+            )
+            LorenzUtils.sendTitle("§c§lPUZZLE FAILED!", 3.seconds)
+            event.blockedReason = "puzzle_fail"
+        }
```
CalMWolfs: I'd like to see this repeated code moved into a function.
Also using SkyHanni's sack api it would be cool to in brackets at the end of the line say how many of the item are left. Also as you move it to a new line I think you can write the whole item name instead of just "ARCHITECT"

CalMWolfs (changes requested):
I still would like to see this fully separated from dungeon copilot, not needing that to be enabled to use this and that not being the pattern key.

CalMWolfs (commented):
Haven't tested in game but I fixed the remaining issues with the code

CalMWolfs (approved):

hannibal002 (changes requested):
Im not a fan of the repopattern key names. the dungeon chat messages should start with dungeon.chat.xyz, and not contain the feature name. we might add new features in the future that require the same chat messages, then the name is weird.

also for `normal.puzzle.fail` and `quiz.puzzle.fail`:
please rename to puzzle.fail.normal and puzzle fail.quiz

hannibal002 (changes requested):
Please include the information what puzzle is failed, or disable this feature per default.

Thunderblade73 (approved):
This pr is fine. Adding the name of the puzzle is not necessary in my opinion since you could look that up via tab or simply know because of what player failed.

hannibal002 (approved):

PR #1206, opened 2024-03-17
Backend: Removal of Neu's SlotClickEvent

Thunderblade73:
## What
Removes all uses of NEU's `SlotClickEvent`

+ Removal of NEU's SlotClickEvent. - Thunderblade73


CalMWolfs (approved):
code looks fine to me but would it be better for the makePickblock function to be inside the SlotClickEvent data class instead of inside LorenzUtils

hannibal002 (approved):

PR #1213, opened 2024-03-18
Improvement: More pet drop messages

ItsEmpa:
## What
Adds more messages to the pet rarity drop message feature, and fix formatting.

## Changelog Improvements
+ Added more messages to Pet Drop Rarity. - Empa


CalMWolfs (approved):
would be cool to see a utils method to either get "a" or "an"  (I don't like the imporper caught an .... in LorenzEvent) but that seems out of the scope of the pr.

Patterns worked fine for me in regex101

hannibal002:
I'm not a fan of this one overcomplicated regex. We don't need to forcefully try to detect multiple different messages with the same regex.
Can we instead split them up into multiple patterns?
Like:
petDrop
petFished
petClaimed
petObtained

src/main/java/at/hannibal2/skyhanni/features/chat/RareDropMessages.kt
```diff
@@ -4,16 +4,48 @@ import at.hannibal2.skyhanni.SkyHanniMod
...
+        "pet.petclaimedmessage",
+        "(?<start>(?:§.)*You claimed a )(?:§.)*§(?<rarityColor>.)(?<petName>[^§(.]+)(?<end>.*)"
+    )
+
+    /**
+     * REGEX-TEST: §b[MVP§r§c+§r§b] Empa_§r§f §r§ehas obtained §r§a§r§7[Lvl 1] §r§6Bal§r§e!
+     */
+    private val petObtainedPattern by RepoPattern.pattern(
+        "pet.petobtainedmessage",
+        "(?<start>(?:§.)*(?:\\[.*])? ?(?:§.)?[a-zA-Z0-9_]{2,16}(?:§.)* (?:§.)*has obtained (?:§.)*\\[Lvl 1] )(?:§.)*§(?<rarityColor>.)(?<petName>[^§(.]+)(?<end>.*)"
```
hannibal002: I still don't like those overly complicated regexes. Instead of checking for the exact player name with [rank] and all, we can just use `.*` for all this. This makes the regex shorter, and focuses more on the actually important parts.
```suggestion
        "(?<start>.*has obtained (?:§.)*\\[Lvl 1] )(?:§.)*§(?<rarityColor>.)(?<petName>[^§(.]+)(?<end>.*)"
```

Maybe use the same logic in the other regexes as well.

src/main/java/at/hannibal2/skyhanni/features/chat/RareDropMessages.kt
```diff
@@ -4,16 +4,48 @@ import at.hannibal2.skyhanni.SkyHanniMod
...
 class RareDropMessages {
 
-    private val petDropPattern by RepoPattern.pattern(
-        "pet.petdropmessage",
-        "(?<typeOfDrop>§6§lPET DROP!|§5§lGREAT CATCH! §r§bYou found a §r§7\\[Lvl 1]) (?:§r)?§(?<rarityColor>.)(?<petName>[^§(.]+)(?<magicFindOrFarmingFortune>.*)"
+    /**
+     * REGEX-TEST: §6§lPET DROP! §r§5Baby Yeti §r§b(+§r§b168% §r§b✯ Magic Find§r§b)
+     * REGEX-TEST: §6§lPET DROP! §r§5Slug §6(§6+1300☘)
+     */
+    private val petDroppedPattern by RepoPattern.pattern(
```
hannibal002: use repo pattern group "pet" instead of repeating "pet.xcy" in every pattern key.

src/main/java/at/hannibal2/skyhanni/features/chat/RareDropMessages.kt
```diff
@@ -4,16 +4,48 @@ import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.events.LorenzChatEvent
 import at.hannibal2.skyhanni.utils.LorenzUtils
 import at.hannibal2.skyhanni.utils.LorenzUtils.colorCodeToRarity
-import at.hannibal2.skyhanni.utils.StringUtils.matchMatcher
+import at.hannibal2.skyhanni.utils.StringUtils.matchMatchers
 import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
 import net.minecraft.util.ChatComponentText
 import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
 
 class RareDropMessages {
```
hannibal002, line 12: Include the word Pet in the class name i guess?

src/main/java/at/hannibal2/skyhanni/features/chat/RareDropMessages.kt
```diff
@@ -4,16 +4,48 @@ import at.hannibal2.skyhanni.SkyHanniMod
...
+    private val petDroppedPattern by RepoPattern.pattern(
+        "pet.petdroppedmessage",
+        "(?<start>(?:§.)*PET DROP! )(?:§.)*§(?<rarityColor>.)(?<petName>[^§(.]+)(?<end> .*)"
+    )
+
+    /**
+     * REGEX-TEST: §5§lGREAT CATCH! §r§bYou found a §r§7[Lvl 1] §r§aGuardian§r§b.
+     */
+    private val petFishedPattern by RepoPattern.pattern(
+        "pet.petfishedmessage",
```
hannibal002, line 29: Change all the regex keys from `pet.pet<xyz>` to `pet.chatdrop.<xyz>`

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/chat/RareDropMessages.kt
ItsEmpa: i feel like this was the only regex where this made it kind of unreadable, the other regex are pretty clear in what they do, so i dont really see a point in changing the others

src/main/java/at/hannibal2/skyhanni/features/chat/RareDropMessages.kt
ItsEmpa: i was planning on expanding this class to also have some other features i had in mind related to rare drop messages, thats why i named it like that. if its needed i can just rename it to PetDropMessages for now and change it in the future when i get around to doing that

hannibal002: No big thing either way. If you plan on adding non-pet stuff later, then there is no need to rename it twice.

PR #1215, opened 2024-03-18
Improvement: Add hover effect to messages in /shchathistory

Zickles:
## What
adds original message hover effect to the message in /shchathistory

- [x] Added og message hover
- [ ] Add background hover effect so you know which message you are hovering over

## Changelog Improvements
+ Add Hover Effects In /shchathistory. - Zickles
    * adds original message hover effect if it had one


hannibal002:
old

PR #1218, opened 2024-03-19
Fix: getting slayer spawn cost from bank

hannibal002:
## What
Previously, we grabbed the slayer spawn cost for the slayer profit tracker from the change of purse in the scoreboard.
But what if your purse is empty?
Hypixel grabs the coins from the bank, but we had no way to accurately detect the change since the bank value in the tab list is rounded (20m - 50k = 20m)
We also thought about using the static price per slayer type and tier (since we detect this correctly), but there is a % reduction in slayer price on different scenarios. So I decided against showing wrong information and instead don't show any information.

But recently, Hypixel had added a chat message telling us how much coins you just spent on slayer quests using the bank:
`§7Took 1.9k coins from your bank for auto-slayer...`
This PR uses this chat line to accurately calculate the slayer spawn cost when all money is in the bank.

Suggestion on Discord: https://discord.com/channels/997079228510117908/1219560956380577853

## Changelog Fixes
+ Fixed Slayer Profit Tracker not detecting the slayer spawn cost when taking money from the bank. - hannibal2


CalMWolfs (approved):
code looks fine

PR #1226, opened 2024-03-21
Fix: NEUInternalName is null for item name 'Mushroom'

CalMWolfs:
## What
Added more debug to show current inventory name when getting internal name from item name fails which helps to debug a few different cases. Also fixed the error where item name 'Mushroom' has no internal name. It is currently just mapped to RED_MUSHROOM but all features that use this list don't support mushroom anyways.
- Also reworded something in shdebug because it sounded weird

## Changelog Fixes
+ Fixed NEUInternalName is null for item name 'Mushroom'. - CalMWolfs

+ Added more debug when internal name from item name is null. - CalMWolfs


src/main/java/at/hannibal2/skyhanni/api/CollectionAPI.kt
```diff
@@ -35,21 +36,26 @@ object CollectionAPI {
 
     val collectionValue = mutableMapOf<NEUInternalName, Long>()
 
+    private val incorrectCollectionNames = mapOf(
+        "Mushroom" to "RED_MUSHROOM".asInternalName()
```
hannibal002, line 40: repo. at least for  the future. or a // todo

hannibal002 (approved):

PR #1236, opened 2024-03-22
Improvement: Improve mining tracker

CalMWolfs:
## Dependencies
- #1235 

## What
Changes some code code around in mining event tracker so that as soon as the mineshaft releases events for there show up on the tracker. Also cleaned up the code slightly with this.

Still waiting on Hypixel to hopefully fix the bug on alpha where it doesnt show events in mineshaft properly because it hasnt been confirmed that they are different from dwarven mines. So if anyone has info on that feel free to comment it here or lmk on discord

## Changelog Improvements
+ Add the ability to see Mineshaft mining events when that releases to main. - CalMWolfs

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/mining/MiningEventConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/mining/MiningEventConfig.java
@@ -38,7 +38,7 @@ public class MiningEventConfig {
     public boolean passedEvents = false;
 
     public enum ShowType {
-        ALL("Both Mining Islands"),
+        ALL("All Mining Islands"),
         CRYSTAL("Crystal Hollows Only"),
         DWARVEN("Dwarven Mines Only"),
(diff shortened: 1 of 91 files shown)


CalMWolfs:
Still don't know if Dwarven Mines events are separate from Mineshaft events but even if they aren't this is still better than what there currently is.

src/main/java/at/hannibal2/skyhanni/config/features/mining/MiningEventConfig.java
```diff
@@ -38,9 +38,10 @@ public class MiningEventConfig {
     public boolean passedEvents = false;
 
     public enum ShowType {
-        BOTH("Both Mining Islands"),
+        ALL("Both Mining Islands"),
```
CalMWolfs:
```suggestion
        ALL("All Mining Islands"),
```

hannibal002 (approved):

PR #1243, opened 2024-03-23
Fix: Crop milestones with pest drops

ItsEmpa:
## What
Makes drops from pests get added to crop milestones

## Changelog Fixes
+ Made drops from pests get added to crop milestones. - Empa


ItsEmpa:
havent tested this with a rare drop (like pesterminator) that shouldnt get added to milestones, but i dont think that should cause any problem

hannibal002 (approved):

PR #1247, opened 2024-03-23
Fix: OneConfig listing SkyHanni twice in their modlist

martimavocado:
## What
fixes oneconfig showing skyhanni twice
image
still doesn't show the logo but that's an issue on their end

## Changelog Fixes
+ Fixes OneConfig listing SkyHanni twice. - martimavocado

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/ConfigGuiForgeInterop.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/ConfigGuiForgeInterop.kt
@@ -1,5 +1,6 @@
 package at.hannibal2.skyhanni.config
 
+import cc.polyfrost.oneconfig.utils.IgnoredGuiFactory
 import io.github.moulberry.moulconfig.gui.GuiScreenElementWrapper
 import net.minecraft.client.Minecraft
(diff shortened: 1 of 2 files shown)


hannibal002:
I dont like this workaround. should be fixed by oneconfig instead.
since you know how to fix this in skyhanni, and oneconfig is open source, can you fix this for them directly?

martimavocado:
i talked with them, they linked me to [IgnoredGuiFactory](https://github.com/Polyfrost/OneConfig/blob/develop/versions/src/main/java/cc/polyfrost/oneconfig/utils/IgnoredGuiFactory.java) which looks like their "intended" solution according to the comments, i don't know what there is to fix on their side about this, except maybe not letting the same modid create a config twice? but this is used in a bunch of their mods

hannibal002:
I'm against OneConfig soft-depend support for their config manager. And in the same way, I'm against supporting OneConfig by fixing their bugs with this solution.

After merging this, one might look at the SkyHanni packets and think we are linked to OneConfig in any way. I don't want that.

martimavocado:
that's fair

PR #1250, opened 2024-03-23
Fix: Event level up chat filter

hannibal002:
## What
All 3 regex patterns miss the `§f` and the many empty lines at the start

## Changelog Fixes
+ Fix event level up chat filter. - hannibal2


CalMWolfs (approved):
Can't test in game but looks ok

PR #1251, opened 2024-03-23
Fix: Visitor Tooltip and Tab List

hannibal002:
## What
fixed visitor tooltip getting disabled after 2 seconds when tab list does not contain visitor data

## Changelog Fixes
+ Fixed visitor tooltip breaking when visitor data is disabled in tab list. - hannibal2

+ Removed VisitorToolTipEvent. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/visitor/VisitorListener.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/visitor/VisitorListener.kt
@@ -26,7 +26,6 @@ import at.hannibal2.skyhanni.utils.LorenzUtils
 import at.hannibal2.skyhanni.utils.RenderUtils.exactLocation
 import at.hannibal2.skyhanni.utils.StringUtils.matches
 import at.hannibal2.skyhanni.utils.StringUtils.removeColor
-import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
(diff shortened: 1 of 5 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/visitor/VisitorListener.kt
```diff
@@ -41,6 +44,12 @@ class VisitorListener {
     private var lastClickedNpc = 0
     private val logger = LorenzLogger("garden/visitors/listener")
 
+    private val patternGroup = RepoPattern.group("garden.visitor")
+    private val tablistExistPattern by patternGroup.pattern(
+        "tablist.enabled",
+        "§b§lVisitors:.*"
+    )
+
```
CalMWolfs: instead of this why not just use `VisitorApi.visitorCountPattern.matches`
which is: `"§b§lVisitors: §r§f\\((?<info>.*)\\)"`

CalMWolfs (changes requested):
rest looks fine

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/VisitorListener.kt
hannibal002: good catch!

CalMWolfs (approved):
code lgtm

PR #1261, opened 2024-03-24
Fix: Fixed Cookie Time with Effect Widget in Custom Scoreboard

j10a1n15:
## What
This Pull Request fixes the booster cookie time in custom scoreboard, when the Effects widget is enabled

## Changelog Fixes
+ Fixed Cookie time with in custom scoreboard when effects widget is enabled. - j10a1n15


CalMWolfs (approved):
I dont have a cookie so can't test this but at some point this should be moved to a data/api as i also need these new changes for compact scoreboard.

You don't have to do this but i think it would be good to have this done before we release 0.24

j10a1n15:
> moved to a data/api

absolutely agree, this should at some point be just a simpletimemark so its a lot better

hannibal002 (approved):

PR #1269, opened 2024-03-24
Backend: Banned Classes

hannibal002:
## What
Added a list of banned classes per mod versions in the repo that either hide events or listener classes.
Useful for hot fixing bugs, illegal features, or other Hypixel things.

## How to test

<details>
<summary>Test Tutorial</summary>

Create the repo file BannedClasses.json, and write this inside:
```
{
    "banned_classes": {
        "0.24.Beta.20": [
            "at.hannibal2.skyhanni.features.garden.composter.ComposterDisplay",
            "at.hannibal2.skyhanni.events.LorenzChatEvent"
        ]
    }
}
```
Notice how both all chat filters no longer work, and the composter display does no longer show.

</details>

+ Added banned classes. - hannibal2
    * Stop banned event classes from getting fired.
    * Stop banned listener classes from accepting events.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/BannedClasses.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/BannedClasses.kt
@@ -14,8 +14,7 @@ object BannedClasses {
         list = event.getConstant<BannedClassesJson>("BannedClasses").bannedClasses[SkyHanniMod.version] ?: emptyList()
     }
 
-    fun isBanned(provider: () -> Class<*>?): Boolean =
-        list.isNotEmpty() && isBanned(provider())
+    fun isBanned(provider: () -> Class<*>?): Boolean = list.isNotEmpty() && isBanned(provider())
 
     fun isBanned(javaClass: Class<*>?): Boolean = javaClass?.name?.let(list::contains) ?: false
 }
--- a/src/main/java/at/hannibal2/skyhanni/data/BannedClasses.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/BannedClasses.kt
@@ -14,5 +14,8 @@ object BannedClasses {
         list = event.getConstant<BannedClassesJson>("BannedClasses").bannedClasses[SkyHanniMod.version] ?: emptyList()
     }
 
-    fun isBanned(javaClass: Class<*>): Boolean = list.contains(javaClass.name)
+    fun isBanned(provider: () -> Class<*>?): Boolean =
+        list.isNotEmpty() && isBanned(provider())
+
+    fun isBanned(javaClass: Class<*>?): Boolean = javaClass?.name?.let(list::contains) ?: false
 }
--- a/src/main/java/at/hannibal2/skyhanni/data/BannedClasses.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/BannedClasses.kt
@@ -1,0 +1,18 @@
+package at.hannibal2.skyhanni.data
+
+import at.hannibal2.skyhanni.SkyHanniMod
+import at.hannibal2.skyhanni.data.jsonobjects.repo.BannedClassesJson
+import at.hannibal2.skyhanni.events.RepositoryReloadEvent
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+object BannedClasses {
+
+    private var list = listOf<String>()
+
+    @SubscribeEvent
+    fun onRepoReload(event: RepositoryReloadEvent) {
+        list = event.getConstant<BannedClassesJson>("BannedClasses").bannedClasses[SkyHanniMod.version] ?: emptyList()
+    }
+
+    fun isBanned(javaClass: Class<*>): Boolean = list.contains(javaClass.name)
+}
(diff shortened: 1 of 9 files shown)


src/main/java/at/hannibal2/skyhanni/data/BannedClasses.kt
```diff
@@ -0,0 +1,20 @@
+package at.hannibal2.skyhanni.data
+
+import at.hannibal2.skyhanni.SkyHanniMod
+import at.hannibal2.skyhanni.data.jsonobjects.repo.BannedClassesJson
+import at.hannibal2.skyhanni.events.RepositoryReloadEvent
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+object BannedClasses {
+
+    private var list = listOf<String>()
```
Thunderblade73, line 10: use a set instead of a list. The set is about ~25% faster.
<details>
<summary>Test reuslt</summary>
Top being list and bottom being set

image
image
</details>
<details>
<summary>Code</summary>

Tested twice once with list first and once with set first. Couldn't test on the same string since the compiler optimized my test away when timing both for the same string.
```Kotlin
package at.hannibal2.skyhanni.data

import at.hannibal2.skyhanni.SkyHanniMod
import at.hannibal2.skyhanni.data.jsonobjects.repo.BannedClassesJson
import at.hannibal2.skyhanni.events.RepositoryReloadEvent
import at.hannibal2.skyhanni.utils.CollectionUtils.editCopy
import at.hannibal2.skyhanni.utils.LorenzDebug
import kotlin.time.Duration
import kotlin.time.measureTime

object BannedClasses {

    private var list = listOf<String>()
    private var set = setOf<String>()

    private var listTime = Duration.ZERO
    private var setTime = Duration.ZERO

    private var ab = false

    @SubscribeEvent
    fun onRepoReload(event: RepositoryReloadEvent) {
        list = event.getConstant<BannedClassesJson>("BannedClasses").bannedClasses[SkyHanniMod.version] ?: emptyList()
        set = list.editCopy { add(0, "") }.drop(1).toSet()
    }

    fun isBanned(provider: () -> Class<*>?): Boolean = list.isNotEmpty() && isBanned(provider())

    fun isBanned(javaClass: Class<*>?): Boolean {
        val result: Boolean
        val name = javaClass?.name ?: return false
        if (ab) {
            listTime += measureTime {
                result = list.contains(name)
            }
        } else {
            setTime += measureTime {
                result = set.contains(name)
            }
        }
        ab = !ab
        LorenzDebug.log("Times:\n$listTime\n$setTime")
        return result
    }
}
```
</details>

Thunderblade73 (changes requested):
I'm unhappy about the performance impact. The isBanned call takes up 3% of the runtime, I'm aware that some calls are async therefore it isn't a direct 3% performance impact, but still it takes up not an unsignification amount of processing time.

I don't really know how to fix that. Since the problem is that we have a stupid amount of event listener and therefore call this function a lot. 
<details>
<summary>Test</summary>

The test run about 220 seconds. It started with 1% and got worse over time. Until it settled at about 3%. Not a stress test, just joined Skyblock normally and where in the hub, garden and private island.
image

```Kotlin
object BannedClasses {

    private var set = setOf<String>()

    var setTime = Duration.ZERO
    var start = SimpleTimeMark.farPast()

    @SubscribeEvent
    fun onRepoReload(event: RepositoryReloadEvent) {
        set = event.getConstant<BannedClassesJson>("BannedClasses").bannedClasses[SkyHanniMod.version]?.toSet()
            ?: emptySet()
        start = SimpleTimeMark.now()
    }

    fun isBanned(provider: () -> Class<*>?): Boolean = set.isNotEmpty() && isBanned(provider())

    fun isBanned(javaClass: Class<*>?): Boolean =
        measureTimedValue { javaClass?.name?.let(set::contains) ?: false }.also { setTime += it.duration }.value
}
```
```Kotlin
        fun testCommand(args: Array<String>) {
            SoundUtils.playBeepSound()

            LorenzDebug.log(
                "${(BannedClasses.setTime / BannedClasses.start.passedSince()) * 100}%\n" +
                    BannedClasses.setTime.toString()
            )
            }
```

</details>
<details>
<summary>Spark</summary>

https://spark.lucko.me/zaiLgZe9xk
I run it as is without modified code. (So without the test above)
For Tick event the isBanned call cost us 6% more performance then usual.
image

</details>
For both Test and Spark I used the example for the bannedClasses.

lineargraph:
> I'm unhappy about the performance impact. The isBanned call takes up 3% of the runtime, I'm aware that some calls are async therefore it isn't a direct 3% performance impact, but still it takes up not an unsignification amount of processing time.
> 
> I don't really know how to fix that. Since the problem is that we have a stupid amount of event listener and therefore call this function a lot.
> Test
> Spark
> For both Test and Spark I used the example for the bannedClasses.

How is the perf impact with an empty list tho? Because the case *with* banned classes is going to be fairly rare.

hannibal002:
Caching.
for events we could add a boolean blocked as member in LorenzEvent and this gets changed on repo load, and for listeners we could ditch the whole check and simply unregister/register the classes on demand?

CalMWolfs:
#2008 implemented this

PR #1277, opened 2024-03-26
Improvement: Get current speed without tab list

lunaynx:
## What
Get the player's current speed in the Garden without relying on the tab list widget.

## Changelog Improvements
+ Added the ability to get your current speed on the Garden even if it isn't shown in the tab list. - Alexia Luna
    * This also means the speed now updates faster.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
@@ -42,7 +42,7 @@ class GardenOptimalSpeed {
 
     @SubscribeEvent
     fun onTick(event: LorenzTickEvent) {
-        currentSpeed = (Minecraft.getMinecraft().thePlayer.capabilities.getWalkSpeed() * 1000).toInt()
+        currentSpeed = (Minecraft.getMinecraft().thePlayer.capabilities.walkSpeed * 1000).toInt()
 
         if (sneaking) {
             currentSpeed = (currentSpeed * 0.3).toInt()
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
@@ -29,21 +29,11 @@ class GardenOptimalSpeed {
 
     private val config get() = GardenAPI.config.optimalSpeeds
 
-    private val currentSpeedPattern by RepoPattern.pattern(
-        "garden.optimalspeed.currentspeed",
-        " Speed: §r§f✦(?<speed>.*)"
-    )
-
     private val configCustomSpeed get() = config.customSpeed
     private var sneakingTime = 0.seconds
     private val sneaking get() = Minecraft.getMinecraft().thePlayer.isSneaking
     private val sneakingPersistent get() = sneakingTime > 5.seconds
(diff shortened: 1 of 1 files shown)


hannibal002 (changes requested):
Walking speed on Hypixel is capped at 500. But you can have more than 500 speed stat. I don't think this approach is superior to tab list, just a nice extra. Maybe use this internal speed value for quick updates, but allow for tab list changes to override it still? Not sure

lunaynx:
This gets your actual effective speed, capped by your Speed Cap. It is how fast the server tells the client to move. If it was faster than the actual cap, you would be getting constant lagbacks.

hannibal002:
Well, this depends on what we want to use this speed value for!
Technically, the user cannot exceed 500 speed at Hypixel. Just the tab list shows you more speed as the speed stat. For farming, this extra speed is not necessary to show, as the user cant move faster than 500 any way.

Testing this with black cat and cactus knife:

with beta:
image
with this pr:
image


This looks weird when the tab list shows 600 speed, but our feature is only detecting 500 
image


The correct speed from tab list is necessary for e.g. damage formula. I don't know if we want this change, since it's not necessary for garden and not relying on the tab list is a good thing in general.

lunaynx:
Well, the tab list is simply wrong in that case. If the client is being told to move at 500% speed, it will not move any faster (sprinting speed boost aside, but that is irrelevant). Why would you not want the *correct* value, which is in fact the value returned by this method, instead of believing whatever the tab list says? 

Also, how does speed affect damage in any way? I have never heard of that. Please explain.

Either way, the code I modified is Garden-specific, so nothing else should break.

hannibal002:
I agree with the PR for garden.
Will merge

You DO have 600 speed. This is simply the 400 base speed + 100 black cat + 100 cactus knife. Why should this not be the case? Nowhere in the speed stat breakdown, it tells you there is a additional cap of 500. This 500 limit is only a cap for the walking speed and has nothing to do with speed stat.
Why does this matter at all? See warden heart. And there might be other items I don't know of right now that are impacted by your actual speed.
oh, and the sword that gives you +10 speed per right click for 10ish seconds (already nerfed) gave you really high speed values. This was meta for one tapping mobs with warden helmet.

src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
```diff
@@ -52,22 +42,16 @@ class GardenOptimalSpeed {
 
     @SubscribeEvent
     fun onTick(event: LorenzTickEvent) {
+        currentSpeed = (Minecraft.getMinecraft().thePlayer.capabilities.getWalkSpeed() * 1000).toInt()
```
hannibal002: Use property access syntax instead of the getter method.
```suggestion
        currentSpeed = (Minecraft.getMinecraft().thePlayer.capabilities.walkSpeed * 1000).toInt()
```

hannibal002 (changes requested):
Works fine in game. Just a small code cleanup

lunaynx:
I see. That's its own niche case. Generally, when you want to get the speed, you want the effective walk speed. If this information ever gets abstracted away and made available outside this class, and some features need the actual value of the Speed stat instead, we can always just have two separate variables, like `walkSpeed` and `speedStat`. But since this change is confined to Garden-specific code, it's fine the way it is for now.

hannibal002 (approved):

PR #1282, opened 2024-03-26
Backend: always use local repo patterns when in dev env

hannibal002:
## What
always use local repo patterns when in dev env

+ Always use local repo patterns when in dev env. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/repopatterns/RepoPatternManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/repopatterns/RepoPatternManager.kt
@@ -58,6 +58,8 @@ object RepoPatternManager {
             }
         }
 
+    val localLoading: Boolean get() = config.forceLocal.get() || LorenzUtils.isInDevEnviromen()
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/utils/repopatterns/RepoPatternManager.kt
```diff
@@ -97,8 +98,9 @@ object RepoPatternManager {
      * Reload patterns in [usedKeys] from [regexes] or their fallbacks.
      */
     private fun reloadPatterns() {
+        val localLoading = config.forceLocal.get() || LorenzUtils.isInDevEnviromen()
```
Thunderblade73: Add this to the RepoPatternMaanager as a val get. A more central place to get the value from. And also I have an open PR that would needs this value.

CalMWolfs:
Excited for this pr. No more will I spend too long wondering why my new repo patterns don't work after the config for my dev environment resets and this option resets to being off

PR #1293, opened 2024-03-27
Fix: Black Hotbar Slots

Thunderblade73:
## What
Fixed the issue with the black hotbar slots for good, be "reseting" the state as a last thing for the GuiOverlayRenderEvent event. It could make issue if the inistial state where diffrent, but can't fix that because it's hard to get the state out of the GlStateManger since everything is private and there are no geters only setters.

## Changelog Fixes
+ Fixed Hotbar loosing its transparency. - Thunderblade73
    * This was the case if the skill progress none textured bar was rendered.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/GuiEditManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/GuiEditManager.kt
@@ -47,10 +47,18 @@ class GuiEditManager {
         openGuiPositionEditor(hotkeyReminder = false)
     }
 
+    @SubscribeEvent(priority = EventPriority.HIGHEST)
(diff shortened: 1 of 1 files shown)


CalMWolfs (commented):
Why do we do this here instead of just doing it where the event is called. That seems like it would make more sense to me

Thunderblade73:
Because there are multiple places (2) where the event is called. Easier to do it inside the event as last thing.

CalMWolfs (changes requested):
Maybe instead of calling one event function the proper name and the other "e" you can call it the eventNamePre and eventNamePost

Thunderblade73:
Wait, e is not intended to be here. 😅.
it was just my Debug entry point to figure the default state of GLStatManager out.

CalMWolfs (approved):

PR #1299, opened 2024-03-27
Backend: error manager only for beta users

hannibal002:
## What
Error Manager now has a parameter to only show for beta users.

Instead of putting this behind a toggle in the settings, this is now a form of differentiating between beta users and full version users. Beta users are meant to see more "behind the scenes" stuff or the mod, and also more error messages so we can easier fix bugs.
At the same time, we don't want too many duplicate bug reports from many more full version users and don't want to bother them with error messages as much.

+ Error Manager now has a parameter to only show for beta users. - hannibal2
    * Less spam for full version users, still enough reports from beat users.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/gui/customscoreboard/ScoreboardElements.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/gui/customscoreboard/ScoreboardElements.kt
@@ -804,16 +804,19 @@ private fun getFooterDisplayPair() = listOf(
 private fun getExtraDisplayPair(): List<ScoreboardElementType> {
     if (unknownLines.isEmpty()) return listOf("<hidden>" to HorizontalAlignment.LEFT)
 
-    if (amountOfUnknownLines != unknownLines.size && config.unknownLinesWarning) {
+    val size = unknownLines.size
+    if (amountOfUnknownLines != size && config.unknownLinesWarning) {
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/features/gui/customscoreboard/ScoreboardElements.kt
```diff
@@ -804,16 +804,18 @@ private fun getFooterDisplayPair() = listOf(
 private fun getExtraDisplayPair(): List<ScoreboardElementType> {
     if (unknownLines.isEmpty()) return listOf("<hidden>" to HorizontalAlignment.LEFT)
 
-    if (amountOfUnknownLines != unknownLines.size && config.unknownLinesWarning) {
+    val size = unknownLines.size
+    if (amountOfUnknownLines != size && config.unknownLinesWarning) {
+        val message = "CustomScoreboard detected $size unknown line${if (size > 1) "s" else ""}"
```
CalMWolfs: Don't we have some sort of utils to pluralise stuff

CalMWolfs (changes requested):
Looks good apart from this

src/main/java/at/hannibal2/skyhanni/features/gui/customscoreboard/ScoreboardElements.kt
hannibal002: already got fixed in another pr

CalMWolfs (approved):

PR #1333, opened 2024-04-01
Backend: Improved Scoreboard Line joining with special formatting codes

j10a1n15:
## What

+ Improved Scoreboard Line joining with special formatting codes. - j10a1n15


hannibal002:
I would rather prefer a char array check. i don't think regex is faster than isInArray check.
image


Or do we need to ignore case check here? We have not needed it before, and this is not fixing a bug. I assume Hypixel never uses upper case in scoreboard for color codes

j10a1n15:
> Or do we need to ignore case check here? We have not needed it before, and this is not fixing a bug. I assume Hypixel never uses upper case in scoreboard for color codes

https://discord.com/channels/997079228510117908/1162844830360146080/1224391447185199135 
image
 they do :pensive:

hannibal002 (approved):

PR #1334, opened 2024-04-01
Feature: Craftable Item List

hannibal002:
## Depends on
- #886

<details>
<summary>Image</summary>

grafik

</details>

## What
Add Craftable Items List.

## Changelog New Features
+ Shows a list of items that can be crafted with the items in your inventory when inside the crafting menu. - hannibal2
    * Click on an item to open /recipe.
    * Option to include items from inside sacks.
    * Option to include vanilla items.


src/main/java/at/hannibal2/skyhanni/config/features/inventory/CraftableItemListConfig.java
```diff
@@ -0,0 +1,37 @@
...
+    @ConfigOption(
+        name = "Include Sacks",
+        desc = "Include items from inside the sacks.")
+    @ConfigEditorBoolean
+    public boolean includeSacks = false;
+
+    @Expose
+    @ConfigOption(
+        name = "Include Vanilla Items",
+        desc = "Include vanilla items.")
```
CalMWolfs: from just reading this in settings I cannot tell what exactly this option does.

src/main/java/at/hannibal2/skyhanni/features/inventory/craft/CraftableItemList.kt
```diff
@@ -0,0 +1,129 @@
...
+            for (recipe in recipes) {
+                if (recipe !is CraftingRecipe) continue
+                val neededItems = neededItems(recipe, internalName.toString())
+                // Just a fail save, should not happen normally
+                if (neededItems.isEmpty()) continue
+                val amount = canCraftAmount(neededItems, avaliableMaterial)
+                if (amount > 0) {
+                    pricePer[internalName] = pricePer(neededItems)
+                    lines[internalName] = Renderable.clickAndHover("§8x$amount ${internalName.itemName}",
+                        tips = listOf("Click to craft ${internalName.itemNameWithoutColor}!"),
```
CalMWolfs: this feels plain without colour, i think it should have it

src/main/java/at/hannibal2/skyhanni/features/inventory/craft/CraftableItemList.kt
```diff
@@ -0,0 +1,129 @@
...
+    fun onInventoryOpen(event: InventoryOpenEvent) {
+        if (!isEnabled()) return
+        if (!craftItemPattern.matches(event.inventoryName)) return
+        inInventory = true
+
+        val pricePer = mutableMapOf<NEUInternalName, Double>()
+        val lines = mutableMapOf<NEUInternalName, Renderable>()
+        val avaliableMaterial = readItems()
+        for (internalName in NEUItems.allInternalNames) {
+            if (config.excludeVanillaItems && !internalName.isVanillaItem()) continue
```
CalMWolfs:
```suggestion
            if (config.excludeVanillaItems && internalName.isVanillaItem()) continue
```
Either this is a small logic mistake from you or I don't understand what this setting does

CalMWolfs (changes requested):
I really like the idea of this feature but right now the Include Sacks option needs some work. When enabling it i get a list that is 176 items long. (I also dont have much in my sacks or many sacks so for others it would be much more). It would be cool to have ways to sort this like we do for chest value. {price per (asc, desc) and amount craftable (asc, desc)} Also a way to "scroll" down that list if everything does not fit on the screen.

src/main/java/at/hannibal2/skyhanni/features/inventory/craft/CraftableItemList.kt
hannibal002: i changed include vanilla items to exclude vanilla items and have not properly changed the logic there. thanks for the catch!

src/main/java/at/hannibal2/skyhanni/features/inventory/craft/CraftableItemList.kt
hannibal002: agreed

hannibal002:
added scrollable list support from https://github.com/hannibal002/SkyHanni/pull/886 and moved back to 0.25

hannibal002:
The display now sorts by price per item, not multiplying per amount craftable.
Also, we allow scrolling now.

hannibal002:
grafik

PR #1340, opened 2024-04-02
Fix: All Names Party Tab

hannibal002:
## What
Added names of lobby players to /p tab complete

## Changelog Fixes
+ Added names of lobby players to /p tab complete. - hannibal2


CalMWolfs (commented):
Code lgtm

PR #1342, opened 2024-04-03
Improvement + Fix: Tick event being called twice a tick

CalMWolfs:
## What
Makes tick event only get called once per tick. Offset some of the tick events.

## Changelog Improvements
+ Improved mod performance. - CalMWolfs, ThatGravyBoat

## Changelog Fixes
+ Fixed tick event being called twice a tick. - CalMWolfs
+ Made the fire pit warning actually make a sound. - CalMWolfs

+ Added offset to tick event. - CalMWolfs
+ Made dungeon milestone use repo instead of local patterns. - CalMWolfs


hannibal002 (approved):

hannibal002:
I don't see the need for the tick offset check, also this pr removes many "is it the right tick" checks, causing more checks at the same time, not improving performance. It just makes the features feel more reactive/faster.
But overall, the changes are healthy.

PR #1351, opened 2024-04-03
Fix: Supercraft /gfs not working for 1,000+ items

lunaynx:
## What
Fixed the Supercraft /gfs prompt not showing when you crafted 1,000 or more items.

## Changelog Fixes
+ Fixed the Supercraft /gfs prompt not showing when you crafted 1,000 or more items. - Alexia Luna


CalMWolfs (changes requested):
You have to make a new patter as this will cause an NPE for users on old versions that super craft more than 1000 items

lunaynx:
Right, I forgot about that.

lunaynx:
Any suggestions on the new pattern name? We need some sort of convention for this. There's also another pull request I was going to make that has the same issue.

CalMWolfs (approved):

hannibal002 (approved):

PR #1378, opened 2024-04-07
Backend: Switch from LorenzUtils.inDungeons to DungeonAPI.inDungeon()

j10a1n15:
## What
Since IslandType.CATACOMBS.isInIsland has been working for quite a while, its finally time to actually start using it through DungeonAPI. This Pull Request replaces all instances

+ Switch from LorenzUtils.inDungeons to DungeonAPI.inDungeon(). - j10a1n15


src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonTeammateOutlines.kt
```diff
@@ -22,7 +24,7 @@ class DungeonTeammateOutlines {
         }
     }
 
-    private fun isEnabled() = LorenzUtils.inSkyBlock && LorenzUtils.inDungeons && config.highlightTeammates
+    private fun isEnabled() = LorenzUtils.inSkyBlock && IslandType.CATACOMBS.isInIsland() && config.highlightTeammates
```
CalMWolfs: can remove the inSkyblock check here as isInIsland already checks that

CalMWolfs (changes requested):
Also I feel to keep consistancy with RiftApi and GardenApi we should use `DungeonsApi.inDungeons()` instead of `IslandType.CATACOMBS.isInIsland` but dungeon api just use the `IslandType.CATACOMBS.isInIsland` instead of checking the dungeon floor. 

If you/others dont want to do this then just the one change

CalMWolfs (approved):
also optimised imports across the whole project (only 4 files got changed)

hannibal002 (approved):

PR #1382, opened 2024-04-07
Backend: Bump MoulConfig to 3.0.0 betas

lineargraph:
This changes the package from being io.github.moulberry.moulconfig to io.github.notenoughupdates.moulconfig. Before, there was a mix of both package names, now it is just one. This will probably cause some merge conflicts, but just running a global search/replace with those two will fix almost all of them.

There are also some minor changes in the way ConfigProcessorDriver works, it now takes a structure reader as constructor parameter.

Lastly some components have been rewritten. Notably the colour component has been rewritten, but it should behave identically to before.

This is a beta version of moulconfig because some features are not fully rewritten to be compatible with some of the new settings (like stretching out descriptions, instead of scaling them down).

Some of the new features (like portals and jumps) are needed for an upcoming PR of mine.

<details>
<summary>Images</summary>

n/a. Should be visually mostly identical to 2.5.0,

</details>

+ Bump MoulConfig to 3.0.0 betas - !nea
    * Adds jumps, links and portals.
    * Changes the package from a mix of `io.github.moulberry.moulconfig` and `io.github.notenoughupdates.moulconfig` to just `io.github.notenoughupdates.moulconfig`


lineargraph:
Here is one new thing you can do: 

```java
    @ConfigOption(name = "Hallo",desc = "link hallo")
    @ConfigLink(owner = About.class, field = "autoUpdates")
    public transient Void link;
```

This allows you to add a link to another config option elsewhere.

PR #1390, opened 2024-04-08
Fix: Renderable.itemStack itemScale 

Thunderblade73:
## What
This addresses the issue where the bounding box does not update when applying an itemScale different from 1.
Also the scale values is remapped so that 1.0 is the minecraft item scale instead of 1.0 being the item font size. The item font size is now added as a default instead and globally accessible. Also the item font size was updated from being 0.56 to 0.67 to remove mixels, thx to Empa.

This may change some original scaling values unintentional. I went through all uses of any itemScale and changed those values accordingly but I may missed some.

Skyhanni items no longer have mixels (on default size) - Empa

<details>
<summary>Images</summary>

image

</details>

## Changelog Fixes
+ Rendered Items in SkyHanni GUIs no longer have partially broken textures. (on default size) - Empa

+ Fixed itemScale for Renderable.itemStack. - Thunderblade73


hannibal002:
before reading code i thought "well then we should also fix currently stored user values for the scale, but this would be overkill, not important enough". Then i saw this and thought "perfection"
image

hannibal002:
1.7
image
0.7
image

hannibal002 (approved):

PR #1403, opened 2024-04-09
Fix: Item Render in Next Contest Overlay

Thunderblade73:
## What
Fixed the next contest overlay render weird because of #1390. Also removed unnecessary space in the next visitor overlay.

<details>
<summary>Images</summary>

Before #1390
image
Before Fix

After Fix
image

</details>

## Changelog Fixes
+ Next contest overlay now looks like before. - Thunderblade73


hannibal002 (approved):
It doesnt feel exactly like the old scale, but it looks fine.

PR #1405, opened 2024-04-09
Backend: isKeyClicked

Thunderblade73:
## What
Fixes the need to add an delay to Renderable.clickable to insure that holding won't trigger onClick. The issue was that most of the time the renderable was newly instaniated after an press, therefore not saving the keyWasDown state.

This way the "clicked" state is saved globally and can only trigger one caller for one keyDown.

+ Added isKeyClicked. - Thunderblade73
    * This cleans up the Renderable.clickable.


hannibal002:
Can this go as well now?
https://github.com/hannibal002/SkyHanni/blob/7960d2ad27ac68ea9dcfaf848bd611b13db0af87/src/main/java/at/hannibal2/skyhanni/features/garden/farming/FarmingWeightDisplay.kt#L556

Thunderblade73:
Technically yes. But it shouldn't, since it prevents spamming the website (with an auto clicker, yes I tested it with one, was funny),

hannibal002 (approved):

PR #1406, opened 2024-04-09
Fix: Pest Count (again)

ItsEmpa:
## What
(still WIP, need to test a lot more)
Tries to fix various issues with pest count.

## Changelog Fixes
+ Fixed issues with Pest Count. - Empa


hannibal002 (commented):
works fine for me

PR #685, opened 2023-11-05
Improvement: Price per copper based visitor blocking

Obsidianninja11:
## What
Added option to block refusing/accepting a visitor above/below a certain price per copper
Reworked blocking refusing visitor logic

## Changelog Improvements
+ Improved visitor reward blocking. - Obsidian
    * Added option to block accepting/refusing expensive/cheap copper.
    * Added option to prevent blocking never accepted visitors.
    * Reworked visitor blocking logic to make it more like not clickable items.

PR #1417, opened 2024-04-10
Improvement: Better Visitor Reward Warning

Obsidianninja11:
## What
Added option to block refusing/accepting a visitor above/below a certain price per copper
Reworked blocking refusing visitor logic

## Changelog Improvements
+ Improved visitor reward blocking. - Obsidian
    * Added option to block accepting/refusing expensive/cheap copper.
    * Added option to prevent blocking never accepted visitors.
    * Reworked visitor blocking logic to make it more like not clickable items.


hannibal002 (approved):
This is a copy of #685

PR #1426, opened 2024-04-11
Backend: Adding Hypixel Mod API v0.1.6

hannibal002:
## What
Hypixel is testing out a new [Mod API](https://github.com/HypixelDev/ModAPI), and we want to test this out as well!
We use the currently available version `0.1.6`

[Forum post](https://hypixel.net/threads/hypixel-mod-api-developer-preview-feedback.5635119/)

+ Adding Hypixel mod API. - hannibal2


jani270 (changes requested):
# "Before we go into more detail we want to make it clear to developers that this is a preview and may be disabled at any time and we heavily advise against using this on any production mod."

lineargraph:
> # "Before we go into more detail we want to make it clear to developers that this is a preview and may be disabled at any time and we heavily advise against using this on any production mod."

While i hear you, I also think that the features that are in the API right now are all very minimal. If the location package is discontinued, it would take minutes at most to change out the code for it.

hannibal002:
Copying from discord:
This pull request just imports the library.
This way we can easier and faster Stadt playing around with test features.
Nothing in this pr does impact users.

Actual features we can implement right now (e.g. detect party members) could be made to just work in dev env. I don't plan on giving users unstable features. This is no feature.

hannibal002:
I still think we can already merge this, and then start working on features in another pr (e.g. party detection)

lineargraph:
> I still think we can already merge this, and then start working on features in another pr (e.g. party detection)

Supplemental party detection would definitely be nice. And if hypixel ever stops responding, we still have our old working code. Maybe only use this after joining the server to get ourselves up to speed once?

hannibal002:
> Supplemental party detection would definitely be nice. And if hypixel ever stops responding, we still have our old working code.

Yes exactly.

> Maybe only use this after joining the server to get ourselves up to speed once?

What do you mean?

lineargraph:
> > Maybe only use this after joining the server to get ourselves up to speed once?
> 
> What do you mean?

Currently the party info is broken if you join the game while already being in a party. We would need to run `/party list` or something to get info about our party status.

CalMWolfs (approved):

PR #1438, opened 2024-04-13
Feature: Cold Overlay

j10a1n15:
## What
Adds a textured cold overlay to glacite tunnels.

<details>
<summary>Images</summary>

[Showcase)](https://youtu.be/KZ7UxZUdq1U?t=114)

https://github.com/hannibal002/SkyHanni/assets/45315647/4a6920ee-40bd-4fcf-bcc3-f55e2015ef28

image
image

</details>

## Changelog New Features
+ Added a textured cold overlay to Glacite Tunnels. - j10a1n15, Empa
    * Change at what cold level the texture should appear.

+ Added MiningAPI which includes all GlaciteTunnels related stuff. - j10a1n15, Empa
+ Added ColdUpdateEvent. - j10a1n15


src/main/java/at/hannibal2/skyhanni/features/mining/ColdOverlay.kt
```diff
@@ -0,0 +1,64 @@
...
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+import org.lwjgl.opengl.GL11
+
+
+class ColdOverlay {
+
+    private val areaPattern by RepoPattern.pattern("mining.coldoverlay.area", "Glacite Tunnels")
+
+    fun getCold(): Int? = ScoreboardData.sidebarLinesFormatted.matchFirst(ScoreboardPattern.coldPattern) {
+        return group("cold").replace("-", "").toInt()
```
martimavocado: just a nitpick, i would do 
```suggestion
        return group("cold").toInt().absoluteValue
```

hannibal002: My idea: move a cold value into clacite tunnel api, and then use ColdChangeEvent instead. (we use it in scoreboard, custom mining notifications, and now here as well)

martimavocado: > glacite tunnel api

what if we just did a general mining api instead?

j10a1n15: > > glacite tunnel api
> 
> what if we just did a general mining api instead?

sounds better, I agree

hannibal002: There is already a FossilExcavatorAPI, not sure if we want to move everything into one big mining api. but since we dont have one yet, we might start with one already. we can always split it up later if it gets too big.

src/main/java/at/hannibal2/skyhanni/data/MiningAPI.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    private val group = RepoPattern.group("data.miningapi")
+    private val glaciteAreaPattern by group.pattern("area.glacite", "Glacite Tunnels")
+
+    private var oldCold = 0
+
+
+    fun inGlaciteArea() = glaciteAreaPattern.matches(HypixelData.skyBlockArea) || IslandType.MINESHAFT.isInIsland()
+
+    fun getCold(): Int? = ScoreboardData.sidebarLinesFormatted.matchFirst(ScoreboardPattern.coldPattern) {
```
hannibal002: why dont we share "oldCold" here?

src/main/java/at/hannibal2/skyhanni/data/MiningAPI.kt
```diff
@@ -0,0 +1,38 @@
...
+        return group("cold").toInt().absoluteValue
+    }
+
+    @SubscribeEvent
+    fun onScoreboardChangeEvent(event: ScoreboardChangeEvent) {
+        val newCold = getCold() ?: return
+
+        if (newCold - oldCold != 0) {
```
hannibal002: use `newCold != oldCold` instead

hannibal002 (changes requested):
works fine in game. we should not check the cold value via regex from scoreboard each frame, we already have a locally stored value, use it!

hannibal002 (approved):

PR #1452, opened 2024-04-15
Fix: Bazaar Copy null

hannibal002:
## What
Fixed a bug that caused Bazaar copy cancelled buy order to show null.

## Changelog Fixes
+ Fixed Bazaar copy not working and showing null instead. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/BazaarCancelledBuyOrderClipboard.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/BazaarCancelledBuyOrderClipboard.kt
@@ -24,8 +24,7 @@ class BazaarCancelledBuyOrderClipboard {
      */
     private val lastAmountPattern by patternGroup.pattern(
         "lastamount",
-        "§6§7from §a(?<amount>.*)§7x §7missing item" +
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/BazaarCancelledBuyOrderClipboard.kt
```diff
@@ -16,16 +18,21 @@ import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
...
 
     private val patternGroup = RepoPattern.group("bazaar.cancelledorder")
+
+    /**
+     * REGEX-TEST: §6§7from §a59§7x §7missing items.
+     */
     private val lastAmountPattern by patternGroup.pattern(
         "lastamount",
-        "§a(?<amount>.*)§7x"
+        "§6§7from §a(?<amount>.*)§7x §7missing item" +
```
j10a1n15: Is there any particular reason for the + "" part

hannibal002: No! Good catch.

PR #1454, opened 2024-04-15
Backend: Graph Editor

Thunderblade73:
## Dependencies
- #1405

## What
Adds an editor in which you can make a graph ingame.
Also has an implementation of Dijkstras Algorithm (was to lazy for A*).

<details>
<summary>Images</summary>

image

</details>

+ DevTool Graph Editor. - Thunderblade73


CalMWolfs (changes requested):
Seemed to mostly work from my testing, except it took a few minutes to work out how I was supposed to use it. Some instructions would be nice, possibly togglable with a keybind and would show in the editing gui. Also I would love chat feedback, as i had no clue what was happening. 
Lastly, the default keybind of t for text just opened the chat for me until I rebound it

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
```diff
@@ -840,6 +841,7 @@ class SkyHanniMod {
         loadModule(GardenStartLocation)
         loadModule(PetCandyUsedDisplay())
         loadModule(ServerRestartTitle)
+        loadModule(GraphEditor)
```
ThatGravyBoat: Use `SkyHanniModule` annotation instead

ThatGravyBoat (changes requested):

CalMWolfs (approved):

PR #1456, opened 2024-04-15
Backend: Add unimined build system

lineargraph:
## What
This changes the build system from architectury loom to unimined.

This has some advantages and disadvantages.

### For Loom

- Dependency configurations and tasks get generated during plugin initialization
	- This means you can directly refer to configurations and (most) tasks using type safe accessors
- Logging is set up automatically (including minecrell)
- A decent amount of launching is handled after handing off to the runtime. The build system writes all the config to a shared config file, meaning new run configurations are created quite easily.

### For Unimined

- Non-main source sets can receive minecraft dependencies.
	- This means that configurations for that (and the main source set) are not available as type safe accessors
	- You can create new segregated source sets to limit which parts of the code can access certain dependencies
- Remapping in theory is faster for mixins, but not really. Benchmarks roughly the same. 
- We have more control over logging.
- You can create new run configs only via custom source sets. This does mean that it is slightly more effort to do this, but you also have more capabilities to configure those runs.

### Migration

In theory the only migration that is needed is to delete the `Minecraft Client` run configuration. It will then regenerate as needed by unimined. It will also allow you to directly use the `runClient` gradle task.

### Motive

The main motive behind this is the individual source set support. This allows us to easier export a public (stable) API to allow other mods to interact with us. Another thing is to limit NEU to only a certain subset of the code. Yet another thing would be to potentially allow breaking out parts of the code into another source set for a sort of "libskyhanni" library that other mods can shade and relocate to jumpstart their mod with rich libraries.

+ Switched build system to unimined. - !nea
+ Changed auto mixins to be gathered at compile time, rather than runtime. - !nea

exclude_from_changelog
since this pr gets reverted with https://github.com/hannibal002/SkyHanni/commit/e76b3561bf60f62afcf4a4d667492beb85f35f85

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ gradle.properties
 
 .devauth/*
 !.devauth/config.toml
+
+.env
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,12 +1,13 @@
-import org.apache.commons.lang3.SystemUtils
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+import xyz.wagyourtail.unimined.api.minecraft.MinecraftConfig
+import xyz.wagyourtail.unimined.api.minecraft.task.RemapJarTask
+import xyz.wagyourtail.unimined.api.runs.RunConfig
 import java.io.ByteArrayOutputStream
 
 plugins {
     idea
     java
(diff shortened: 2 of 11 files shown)


hannibal002 (approved):
Reduces build process after one class changed from ~1m5s to 45s for me

PR #1059, opened 2024-02-25
Feature: Hotm Features + Hotm API/Data

Thunderblade73:
Point of focus of this PR ist the Hotm API, which gives every data that could be needed from the hotm tree. The features are only there so that the PR can be test (If more detailed testing is need revert 243a4c2e2353a8fa615b1abc6e6538538f711994) and since there also useful for the user they are full features.
## Preview
image
image

## Changelog New Features
+ Added HOTM perks level and Enable Display. - Thunderblade73

## Changelog Technical Details
+ Added HotmAPI. - Thunderblade73 &  j10a1n15 & martimavocado

PR #1460, opened 2024-04-15
Feature: Total Powder Cost in Perk Lore

j10a1n15:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/1059

## What
This Pull Request adds the option to show the "powder spend/total needed" per HotM Perk.

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added HotM Perks Powder features. - j10a1n15
    * Added total powder cost to HotM perks.
    * Added powder cost for 10 levels.

+ Added getPerkByName(name) and calculateTotalCost(level) to HotmAPI. - j10a1n15


src/main/java/at/hannibal2/skyhanni/features/mining/PowderPerHotmPerk.kt
```diff
@@ -0,0 +1,65 @@
...
+        if (!isEnabled()) return
+
+        val itemName = event.itemStack.displayName
+        val perkEnum = HotmData.getPerkByNameOrNull(itemName.removeColor()) ?: return
+
+        if (perkEnum.getLevelUpCost() == null) return
+
+        val currentPowderSpend = perkEnum.calculateTotalCost(perkEnum.activeLevel)
+        val maxPowderNeeded = perkEnum.calculateTotalCost(perkEnum.maxLevel)
+        val percentage = ((currentPowderSpend.toDouble() / maxPowderNeeded.toDouble()) * 100).round(2)
```
Thunderblade73: use fractionOf

src/main/java/at/hannibal2/skyhanni/features/mining/PowderPerHotmPerk.kt
```diff
@@ -0,0 +1,65 @@
...
+class PowderPerHotmPerk {
+
+    private val config get() = SkyHanniMod.feature.mining.hotmConfig
+
+    @SubscribeEvent
+    fun onTooltip(event: LorenzToolTipEvent) {
+        if (!isEnabled()) return
+
+        val itemName = event.itemStack.displayName
+        val perkEnum = HotmData.getPerkByNameOrNull(itemName.removeColor()) ?: return
```
Thunderblade73: just call it perk.

src/main/java/at/hannibal2/skyhanni/features/mining/PowderPerHotmPerk.kt
```diff
@@ -0,0 +1,65 @@
...
+
+        if (perkEnum.getLevelUpCost() == null) return
+
+        val currentPowderSpend = perkEnum.calculateTotalCost(perkEnum.activeLevel)
+        val maxPowderNeeded = perkEnum.calculateTotalCost(perkEnum.maxLevel)
```
Thunderblade73: Save the max Powder in the enum. Since it is literally an constant, so no need to calculate it every render frame.

src/main/java/at/hannibal2/skyhanni/data/HotmData.kt
```diff
@@ -0,0 +1,483 @@
...
+        }
+
+    var slot: Slot? = null
+        private set
+
+    fun getLevelUpCost() = costFun(rawLevel)
+
+    fun getReward() = rewardFun(activeLevel)
+
+    fun calculateTotalCost(desiredLevel: Int) = (0 until desiredLevel).sumOf { level -> costFun(level) ?: 0.0 }.toInt()
```
Thunderblade73: Shouldn't it start with 1 not 0?
The error is only quiet small about 10-100 powder for most perks, but still it should be correct. Use the calculator in fandom for one perk to verify the cost.
Also there would be a problem if desired level is 1, since this must always be 0 which it isn't in the way you call it here.

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/HotmData.kt
j10a1n15: I tried it with both a 0 and a 1, both numbers didnt match the data from the wiki.
Problem with the 1 is, as you stated, that it will no longer work.
What should I do then?

Thunderblade73: Ok, will have a look. Because the need to match up. So there is a problem in the data.

src/main/java/at/hannibal2/skyhanni/data/HotmData.kt
```diff
@@ -0,0 +1,487 @@
...
+            private set(value) {
+                ProfileStorageData.profileSpecific?.mining?.availableTokens = value
+            }
+
+        var heartItem: Slot? = null
+
+        init {
+            entries.forEach {
+                it.guiNamePattern
+                it.totalCostMaxLevel = it.calculateTotalCost(it.maxLevel)
```
Thunderblade73: this assignment can be done in the enum. No need to have it in the companion object. Also makes it possible to make total Cost Maxlevel a val.

j10a1n15: This would be worse because instead of being able to access the calculateTotalCost() method, I would need to write it out every single time, and I wouldn't even be able to use the maxLevel constant in the enum entry itself.
I might be overlooking some things right now, please tell me if I do so.

Thunderblade73: I don't see a problem here. Since calculateTotalCost is known at the instantiation. So why shouldn't it work? Maybe try change the order of the elments eg. calculateTotalCost and after that total cost. And inside the enum you can use the maxlevel.
So there is no issue here. Maybe I worded it missleading.
So here literally the code:
```Kotlin
val totalCostMaxLevel = calculateTotalCost(maxLevel)
```

j10a1n15: When I try to set the max level inside of the enum constructor, the method won't get detected.
Maybe I still don't understand what you mean, please explain it to me then.
image

Thunderblade73: that is not in the enum, that is inside the constructor of the enum. Inside the enum is after ;

Thunderblade73: image

Thunderblade73: \* as a val please. Forgot to change that.

j10a1n15: Oh my fucking god.
Sorry I'm stupid
Thanks for explaining!

src/main/java/at/hannibal2/skyhanni/data/HotmData.kt
Thunderblade73: ```1 until disiredLevel``` works. So simply change the 0 to 1.
image
image

Thunderblade73:
To fix the merge conflict merge [hotmAPI](https://github.com/Thunderblade73/SkyHanni/tree/hotm-API) into it

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -120,6 +120,13 @@ object StringUtils {
         return null
     }
 
+    fun List<String>.indexOfFirstMatch(pattern: Pattern): Int? {
```
CalMWolfs: should be in regex utils

CalMWolfs (changes requested):
just this one thing, feature works fine and seems cool

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
j10a1n15: What even is this, I never wrote that

Thunderblade73: It is in regexUtils. But you didn't merge the hotmAPI again. So you have two now.

CalMWolfs (approved):

PR #1464, opened 2024-04-16
Feature: Allow party members to request allinvite to be enabled

CalMWolfs:
## What
Allow party members to request allinvite to be enabled by doing !allinvite in party chat. Also added delay between doing some commands

## Changelog New Features
+ Allow party members to request allinvite to be turned on. - CalMWolfs
    * Say !allinv in party chat for them to enable all invite.


src/main/java/at/hannibal2/skyhanni/config/features/misc/PartyCommandsConfig.java
```diff
@@ -16,12 +16,17 @@ public class PartyCommandsConfig {
     @Expose
     @ConfigEditorBoolean
     @ConfigOption(name = "Party Transfer", desc = "Automatically transfer the party to people who type §b!ptme")
-    public boolean transferCommand = false;
+    public boolean transferCommand = true;
```
lineargraph: Not sure about making this one `true` by default. This is an auto opt in into something that sends commands without user input (which can be construed to be against the rules)

CalMWolfs: i guess thats fair, will change back

hannibal002 (changes requested):
Maybe we should also store if the party is currently enabled, otherwise this can have unintended behavior if sending !allinvite twice or two people sending it at the same time.

hannibal002 (approved):

PR #1468, opened 2024-04-16
Fix: Gemstones item query

Thunderblade73:
## What
Fixes the item query for gemstones. Especially for queued gfs.
I don't like the fix, but it works.

## Changelog Fixes
+ Queued gfs now works with gemstone names. - Thunderblade73

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
@@ -20,7 +20,7 @@ object StringUtils {
     private val resetPattern = "(?i)§R".toPattern()
     private val sFormattingPattern = "(?i)§S".toPattern()
     private val stringColourPattern = "§[0123456789abcdef].*".toPattern()
-    private val asciiPattern = "[^${Char(0)}-${Char(255)}]".toPattern()
+private val asciiPattern = "[^\\x00-\\x7F]".toPattern()
 
     fun String.trimWhiteSpaceAndResets(): String = whiteSpaceResetPattern.matcher(this).replaceAll("")
     fun String.trimWhiteSpace(): String = whiteSpacePattern.matcher(this).replaceAll("")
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -20,11 +20,13 @@ object StringUtils {
     private val resetPattern = "(?i)§R".toPattern()
     private val sFormattingPattern = "(?i)§S".toPattern()
     private val stringColourPattern = "§[0123456789abcdef].*".toPattern()
+    private val asciiPattern = "[^${Char(0)}-${Char(255)}]".toPattern()
```
ThatGravyBoat: This regex is not for ASCII, ASCII does not go to 255 this regex should be changed to the following. using Char(n) does not provide any benefit here.

```suggestion
private val asciiPattern = "[^\\x00-\\x7F]".toPattern()
```

ThatGravyBoat (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
Thunderblade73: Yeah, made an ups by including 128 characters more 😅. Also I didn't know how to but a specific char into a String in Kotlin, so thank you.

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -20,11 +20,13 @@ object StringUtils {
     private val resetPattern = "(?i)§R".toPattern()
     private val sFormattingPattern = "(?i)§S".toPattern()
     private val stringColourPattern = "§[0123456789abcdef].*".toPattern()
+private val asciiPattern = "[^\\x00-\\x7F]".toPattern()
```
jani270:
```suggestion
    private val asciiPattern = "[^\\x00-\\x7F]".toPattern()
```

jani270 (changes requested):
🤓

hannibal002 (approved):

PR #1469, opened 2024-04-16
Feature: Add Ignored users for Party Commands

martimavocado:
## What
Adds `/shignore` which lets you blacklist certain people from using party commands.
This has several sub-commands, including `add`, `remove`, `list` and `clear`. 
- `add` and `remove` always take another argument, being the username to add/remove. 
- `list` takes an optional argument, checking if a given username is ignored, or listing every username when none is given.
- `clear` clears the list with a confirmation prompt.
- when no subcommands are taken, it will assume a add/remove toggle for the given username. 

https://canary.discord.com/channels/997079228510117908/1229893091285008526

<details>
<summary>Images</summary>

image

image
image

image

image
image

image

</details>

## Changelog New Features
+ Added /shignore. - martimavocado
    * This lets you block users from running party commands.


src/main/java/at/hannibal2/skyhanni/features/commands/PartyChatCommands.kt
```diff
@@ -71,10 +76,103 @@ object PartyChatCommands {
         if (command.requiresPartyLead && PartyAPI.partyLeader != LorenzUtils.getPlayerName()) {
             return
         }
+        if (isBlockedUser(event.author)) {
+            ChatUtils.chat("§cIgnoring chat command from ${event.author}. Unblock them using *unblacklist command here*.")
```
CalMWolfs: seems to be missing a command here

CalMWolfs (changes requested):
also can you please do the generic changes (make functions private and change the config variable to be called config) like in #1464

src/main/java/at/hannibal2/skyhanni/features/commands/PartyChatCommands.kt
ItsEmpa: i assume the missing command is `/shignore remove <name>`. could make this a clickable message that runs that command when clicked

ItsEmpa: also i feel like there should be a toggle to disable this message

CalMWolfs (approved):
Haven't tested in game but code looks fine

hannibal002 (changes requested):
I thought we wanted to use Hypixel's  /ignore list, not create our own?

CalMWolfs:
> I thought we wanted to use Hypixel's /ignore list, not create our own?

Own ignore list is fine here I think, this is like a list of not trusted party members but they arent bad enough to be fully blocked from being in the party, or at least that is my understanding

martimavocado:
> I thought we wanted to use Hypixel's /ignore list, not create our own?

if this is the case that wasn't very clear for me from the conversation in the discord thread, but i could redo this to use hypixel's list instead of a custom one if you think that's better

hannibal002:
i dont think thats better, i just thought we decided to do it this way. if this was not clear, im fine with the current version as well

src/main/java/at/hannibal2/skyhanni/config/features/misc/PartyCommandsConfig.java
```diff
@@ -46,4 +49,11 @@ public String toString() {
         }
     }
 
+    @Expose
+    @ConfigEditorBoolean
+    @ConfigOption(name = "Show reminder", desc = "Shows a reminder when an unauthorized player tries to run a command.")
+    public boolean showIgnoredReminder = true;
+
+    @Expose
+    public List<String> blacklistedUsers = new ArrayList<>();
```
hannibal002: move to storage

hannibal002 (approved):
works now

hannibal002 (changes requested):
missclick

hannibal002 (changes requested):
merge conflicts

hannibal002 (approved):
image

PR #1470, opened 2024-04-17
Fix: Memory leak

CalMWolfs:
## What

WIP: Thunderblade says stuff needs to change a bit

Fix memory leaks caused by some entities not being removed properly.
There may still be other negative performance impacts due to mob detection but this fixes the memory leak of the player not being removed from MobData.players leading to the whole previous world still being stored.

## Changelog Fixes
+ Fixed memory leak. - CalMWolfs

+ Clear mob detection data more often. - CalMWolfs
    * Clear on config value update and world switch.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/mob/MobDetection.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/mob/MobDetection.kt
@@ -327,19 +327,13 @@ class MobDetection {
             event.addData("Mob Detection is manually disabled!")
         } else {
             event.addIrrelevant {
-                add("normal enabled")
-                add("Active Players: ${MobData.players.size}")
-                add("Active players: ${MobData.players.map { "${it.name} - ${System.identityHashCode(it)}" }}")
-                add("Active Mobs: ${MobData.players.size + MobData.displayNPCs.size + MobData.summoningMobs.size + MobData.skyblockMobs.size + MobData.special.size}")
-                add("Active Mobs: ${MobData.currentMobs.map { "${it.name} - ${System.identityHashCode(it)}" }}")
-                val inDistanceMobs = MobData.retries.count { it.value.outsideRange() }
-                add("Searching for Mobs: ${MobData.retries.size - inDistanceMobs}")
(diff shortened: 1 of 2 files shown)


hannibal002 (changes requested):
it breaks the contract that every Spawn Event also has a Despawn event

src/main/java/at/hannibal2/skyhanni/data/mob/MobDetection.kt
```diff
@@ -313,8 +302,7 @@ class MobDetection {
             // is S0EPacketSpawnObject -> addEntityUpdate(packet.entityID)
             is S37PacketStatistics -> // one of the first packets that is sent when switching servers inside the BungeeCord Network (please some prove this, I just found it out via Testing)
             {
-                shouldClear.set(true)
-                allEntitiesViaPacketId.clear()
+                clearMobs()
```
Thunderblade73, line 305: ehm, you break thread safety here. Will lead to an ConcurrentModifactionExpection.

src/main/java/at/hannibal2/skyhanni/data/mob/MobDetection.kt
```diff
@@ -354,4 +344,21 @@ class MobDetection {
...
+            clearMobs()
+        }
+    }
+
+    private fun clearMobs() {
+        mobDetectionReset()
+        MobData.previousEntityLiving.clear()
+        MobData.currentEntityLiving.clear()
+        MobData.retries.clear()
+        MobData.entityToMob.clear()
```
Thunderblade73, line 359: this breaks the contract.
Better way is to check if it is empty else show a error with the data.

Thunderblade73 (changes requested):
I like that the config is handled correctly. But the general clearing literally thows out an error, also breaks the contract that every spawn has to have a despawn. And I'm unsure what breaks because of the WorldSwitchEvent, the other event is only there when you leave hypixel, which btw is not handled anymore because of that change.

CalMWolfs:
Closing as memory leak was already fixed in #1473

PR #1471, opened 2024-04-17
Feature: add guide to dungeon hub races

catgirlseraid:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/153

## What
Add guide to dungeon hub races for all three races. Only works for Nothing; No return races

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added dungeon hub race waypoints. - seraid
    * Only works for Nothing; No return races.

+ Added NamedParkourJson. - seraid


CalMWolfs (approved):

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/NamedParkourJson.kt
CalMWolfs, line 1: This should not be a map of maps, NamedParkourJson should only be `Map<String, ParkourJson>` and then a further `DungeonHubRacesJson` data class should hold a `Map<String, NamedParkourJson>`

Another option is to scrap the idea of making a NamedParkourJson and just making a DungeonHubRacesJson and adding the relevant fields into here, could maybe make an enum for race type and use those as some of the keys

CalMWolfs (changes requested):

CalMWolfs (approved):

src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonsRaceGuide.kt
```diff
@@ -0,0 +1,91 @@
...
+            it.value.lookAhead = config.lookAhead.get() + 1
+        }
+    }
+
+    @SubscribeEvent
+    fun onRenderWorld(event: LorenzRenderWorldEvent) {
+        if (!isEnabled()) return
+        if (currentRace == null) return
+
+        parkourHelpers[currentRace]?.render(event)
```
hannibal002, line 87: can we store the ParkourHelper instance in the variable currentRace? that way we dont need to do one map lookup each frame

hannibal002 (dismissed):
works fine, except it breaks on every ability use (e.g. aote teleport). this is part of ParkourHelper, maybe you can fix it?

CalMWolfs:
> works fine, except it breaks on every ability use (e.g. aote teleport). this is part of ParkourHelper, maybe you can fix it?

In my opinion this is not a problem as the feature is not made to work for the later races. The whole feature is written to only work with Nothing at all; No return races. In these races you cannot use abilities so this will never break. This is made as just a little guide for beginners as once you are at the point of using abilities and pearls these waypoints do not provide much use.

hannibal002 (approved):

PR #1484, opened 2024-04-18
Backend: EditorConfig always wrap enums

Thunderblade73:
## What
Since we always do it manual, why not automate it. There are 2-3 enums that are not wrapped but it also does not really hurt there.

+ Force wrapping of enums. - Thunderblade73


hannibal002:
wanna pr the reformatted code as well?

Thunderblade73:
sure

PR #1496, opened 2024-04-19
Backend: f7 bosses in Mob Detection

Thunderblade73:
+ Fixed f7 bosses not detecting as boss. - Thunderblade73


hannibal002 (changes requested):
the level is wrong.
also the brackets `﴾﴿` are not getting detected with `matchMatcher`
image

Thunderblade73:
The brackets get match, the do not get highlight here since they are very weird. But looking at the right side it shows that the match. The level one is correct and I pushed a fix for that.

hannibal002 (approved):

PR #1497, opened 2024-04-19
Backend: Improves performance of item backgrounds and borders

ThatGravyBoat:
## What
This PR changes how item backgrounds are handled, previously many locations in the code used the `ItemRenderBackground` class to set the background this now is done individually as to not keep a map of itemstacks and to not move functionality to other parts of the codebase.

+ Removed `ItemRenderBackground`. - ThatGravyBoat
    * Added helpers in `RenderUtils` for drawing borders.
    * Added helpers in `RenderUtils` for drawing borders and highlights on `RenderGuiItemOverlayEvent`.
+ Changed `RenderRealOverlayEvent` to `RenderGuiItemOverlayEvent` to be more representative of what the event is for. - ThatGravyBoat


CalMWolfs (approved):
worked for me

src/main/java/at/hannibal2/skyhanni/features/itemabilities/abilitycooldown/ItemAbilityCooldown.kt
```diff
@@ -303,16 +307,31 @@ class ItemAbilityCooldown {
...
-                    opacity = 80
-                    if (!config.itemAbilityShowWhenReady) return
-                }
-                stack.background = color.addOpacity(opacity).rgb
+            var opacity = 130
+            if (color == LorenzColor.GREEN) {
+                opacity = 80
+                if (!config.itemAbilityShowWhenReady) return
             }
+            event highlight color.addOpacity(opacity)
```
hannibal002, line 336: While holding an item with active ability, the whole screen gets a gray effect.
image
image

src/main/java/at/hannibal2/skyhanni/features/itemabilities/abilitycooldown/ItemAbilityCooldown.kt
```diff
@@ -48,7 +49,7 @@ class ItemAbilityCooldown {
     )
 
     private var lastAbility = ""
-    private var items = mapOf<ItemStack, List<ItemText>>()
+    private var items = mapOf<String?, List<ItemText>>()
```
hannibal002: We always check a non-nullable key for the map, the key should not be nullable.

src/main/java/at/hannibal2/skyhanni/features/itemabilities/abilitycooldown/ItemAbilityCooldown.kt
```diff
@@ -254,7 +255,11 @@ class ItemAbilityCooldown {
             abilityItems = ItemUtils.getItemsInInventory(true).associateWith { hasAbility(it) }
         }
 
-        items = abilityItems.mapValues { kp -> kp.value.map { createItemText(it) } }
+        items = abilityItems.entries.associateByTo(
+            mutableMapOf(),
+            { it.key.getIdentifier() },
```
hannibal002, line 261: filter out when getIdentifier() returns null

hannibal002 (changes requested):
Works fine in my dev env, just the gray effect needs to go.

hannibal002 (changes requested):
while inside inventories, slots with items on them have a different background to other items. this does not happen when pressing E. maybe because of an NEU fancy overlay thing? does not happen in skyblock menu, but in other menus, e.g. bazaar
image

image

ThatGravyBoat:
I cant reproduce, you must have a feature enabled that is conflicting with the highlight changes

ThatGravyBoat:
you have the feature HideNotClickableItems on which does that to show which items you cant click

hannibal002 (changes requested):
<details>
<summary>colours in /tab look desaturated now.</summary>

image

</details>


<details>
<summary>same in inventory</summary>

image

</details>

I think this affects all 35 usages of `infix fun Slot.highlight(color: LorenzColor) {`.

ThatGravyBoat:
Okay fixed this, I by accident removed the lighting when fixing that color setting and attribute changes that were already there that caused the whole screen to have its colored changed.

hannibal002 (approved):
works fine for me

PR #1502, opened 2024-04-20
Improvement: Coin Loss Visitor Reward Warning

albarv340:
## What
An extension of #1417 based on [this suggestion](https://discord.com/channels/997079228510117908/1230117554392006727).
It adds an alternative visitor reward warning based entirely on raw coin loss instead of coins per copper. Useful for people focusing on accepting as many visitors as possible without losing too much money. 
The value of copper is estimated based on the value of a Green Thumb I book. There is some improvement potential there.

<details>
<summary>Images</summary>

image
image
image

</details>

## Changelog Improvements
+ Added option to block accepting/refusing visitors depending on coin loss. - Albin


hannibal002 (changes requested):
Looks like either reward or profit fetching or mathing is wrong here.
image

other cases are fine though:
image
image

albarv340:
> Looks like either reward or profit fetching or mathing is wrong here. image
> 
> other cases are fine though: image image

I forgot to account for the fact that profit is negative loss in the tooltip, but the calculation is right. It should show up as a positive number now:
image

(It took me way longer than I would have hoped to get a visitor with a profitable reward)

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/VisitorAPI.kt
```diff
@@ -131,6 +131,8 @@ object VisitorAPI {
     ) {
         var offersAccepted: Int? = null
         var pricePerCopper: Int? = null
+        var totalPrice: Double? = null
+        var totalReward: Double? = null
```
hannibal002, line 135: This does only use copper price, not the reward item price.

hannibal002 (changes requested):
works fine, except for rare reward items as their price is not in loss calculation

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/VisitorAPI.kt
albarv340: From my understanding [this part](https://github.com/hannibal002/SkyHanni/pull/1502/files#diff-5db3faf190c23cd99f5ac5ee568272fb5e768d41094816303417250ff554fd18R340) of GardenVisitorFeatures.kt already takes the reward value into account. And totalPrice is the variable used for this feature.
```kotlin
val price = internalName.getPrice() * amount
if (readingShoppingList) {
    totalPrice += price
    lastFullPrice += price
} else {
    foundRewards.add(internalName)
    totalPrice -= price
}
```

albarv340: image

Here it can be seen that the price of the space helmet is being considered when calculating the loss. The copper is worth around 500k and the cookies cost around 1.6b. The loss being 976m means the space helmet is being considered.

hannibal002 (approved):

PR #1505, opened 2024-04-20
Fix: Fixed transfer command

nobaboy:
## What
Fixes transfer command for party chat commands.
(Quick note, if it's actually supposed to be like that my bad, but it didn't work for many players)

## Changelog Fixes
+ Fixed Transfer Party Chat command. - nobaboy


nobaboy:
I'd appreciate if someone sees this to test it because ever since swapping to unimined, I haven't been able to load the game or compile anything

CalMWolfs:
> I'd appreciate if someone sees this to test it because ever since swapping to unimined, I haven't been able to load the game or compile anything

Hey, if you're having problems launching with the new changes can you post them in the #coding channel and ill see if i can solve them

hannibal002:
> I'd appreciate if someone sees this to test it because ever since swapping to unimined, I haven't been able to load the game or compile anything

I have had the same problem. Delete your "build" folder, then its working for me again.

hannibal002 (changes requested):
Thanks for trying to fix this bug!

While your solution will work for PartyChatEvent, the variable cleanedAuthor in AbstractChatEvent will fail for other instances, e.g. PlayerAllChatEvent
Why? See the function `nameFormat` at https://github.com/hannibal002/SkyHanni/blob/f8efe9effd27bee18bced64385fc1ad9b05b68bd/src/main/java/at/hannibal2/skyhanni/data/hypixel/chat/PlayerNameFormatter.kt#L70.
We cut the emblem from the display name via `emblemPattern` before running `cleanPlayerName`. 

Additionally, you can use `val name by lazy {//implementation }` instead of `val name get() = author.cleanPlayerName()`. this way the value will only get calculated lazy, meaning we only calculate it if we access the value at all, and cache the value. this is perfect for this use case here.

nobaboy:
> Thanks for trying to fix this bug!
> 
> While your solution will work for PartyChatEvent, the variable cleanedAuthor in AbstractChatEvent will fail for other instances, e.g. PlayerAllChatEvent Why? See the function `nameFormat` at
> 
> https://github.com/hannibal002/SkyHanni/blob/f8efe9effd27bee18bced64385fc1ad9b05b68bd/src/main/java/at/hannibal2/skyhanni/data/hypixel/chat/PlayerNameFormatter.kt#L70
> .
> We cut the emblem from the display name via `emblemPattern` before running `cleanPlayerName`.
> 
> Additionally, you can use `val name by lazy {//implementation }` instead of `val name get() = author.cleanPlayerName()`. this way the value will only get calculated lazy, meaning we only calculate it if we access the value at all, and cache the value. this is perfect for this use case here.

I mean I have a couple of solutions but I have no clue if they'd even fit...
like get rid of the emblem in the message then clean name, or match for a [A-z0-9_] in the author name which returns the player name

hannibal002:
I think the best idea would be to delete the existing events, and create one event with an channel enum.

nobaboy:
honestly not to turn you down or anything but I'd have to spend quite a lot of time looking through the classes to see which files use what, and I'm quite busy. we could partially fix party chat commands with just cleaning the name where it is required but if I do have some time later on I'll give it a chance

hannibal002:
Yeah, fair. The underlying code for the player chat events will get a partial rewrite any way.
I think for this fix in party chat command, we can just move the cleanedAuthor string from AbstractChatEvent to PartyChatEvent and it's fine.

nobaboy:
> Yeah, fair. The underlying code for the player chat events will get a partial rewrite any way. I think for this fix in party chat command, we can just move the cleanedAuthor string from AbstractChatEvent to PartyChatEvent and it's fine.

I'll just move it to PartyChatEvent for now and when you guys refactor the code you know what to do.
I didn't test all chat when putting cleanedAuthor in AbstractChatEvent, though all other classes function properly, the only two that are not returning a cleanedAuthor are show item and all chat, so that's a fault on my part

hannibal002 (approved):

PR #1511, opened 2024-04-21
Fix: Rift Blood Effigies not being properly detected

lunaynx:
## What
Fixed Rift Blood Effigies not being detected from the scoreboard because the effigy symbol we're checking for was being stripped away by the scoreboard cleaning code.

## Changelog Fixes
+ Fixed Rift Blood Effigies not being detected from the scoreboard. - Luna


lunaynx:
Marking as a draft, because while this does fix the specific issue it was meant to fix, I plan to include further fixes to Blood Effigy detection in this PR.

hannibal002:
old

PR #1520, opened 2024-04-23
Fix: Maxwell Power not updating when unlocking a new Power

j10a1n15:
## What
This Pull Request fixes the power not updating, when the player unlocks a new power.

## Changelog Fixes
+ Fixed Maxwell Power not updating when unlocking a new power. - j10a1n15


src/main/java/at/hannibal2/skyhanni/data/MaxwellAPI.kt
```diff
@@ -52,10 +52,14 @@ object MaxwellAPI {
...
 
     private val patternGroup = RepoPattern.group("data.maxwell")
-    private val chatPowerpattern by patternGroup.pattern(
+    private val chatPowerPattern by patternGroup.pattern(
         "chat.power",
         "§eYou selected the §a(?<power>.*) §e(power )?for your §aAccessory Bag§e!"
     )
+    private val chatPowerUnlockedPattern by patternGroup.pattern(
+        "chat.power.unlocked",
+        "§eYou permanently unlocked the §a(?<power>.*) §epower!"
```
lunaynx: I would prefer to use the other message, just in case Hypixel ever changes the current behavior of automatically switching to the newly unlocked power. (Not tested)
```suggestion
        "§eYour selected power was set to (?:§r)*§a(?<power>.*)(?:§r)*§e!"
```

lunaynx (changes requested):
(See below)

hannibal002 (approved):

PR #1523, opened 2024-04-23
Add Funny

CalMWolfs:
## Dependencies
- #1519

exclude_from_changelog


lineargraph (changes requested):
This feels very much like too much. Not only is the implementation of this rather shoddy, but this is also a very jarring visual change. Changing a colored name or even custom symbols don't have nearly the same impact on other players that this has. I also think the novelty of this will wear of very soon. The spinning is a hard no go, just completely unprecedented self aggrandization, and even just `Grum`/`Dinnerbone` style flipping is too much. These are massive visual (notably) vanity projects should stay limited to (at most) capes and colored names.

lineargraph:
Biblically speaking, this is a sin. It also is just pointless (and annoying) attention whoring.

hannibal002:
Please add an extra option for spin, and for upside down, separate. spin is default disabled,
for the upside down option, idk the default

hannibal002 (changes requested):
remove size change, put rotating default disabled in dev, and upside down needs another condition when to activate, idk what yet

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RendererLivingEntityHook.kt
```diff
@@ -57,11 +52,13 @@ class RendererLivingEntityHook {
 
     private fun isCoolPerson(userName: String?): Boolean {
         if (!LorenzUtils.inSkyBlock) return false
+        if (!config.flipContributors || LorenzUtils.isAprilFoolsDay) return false
```
lineargraph: Shouldnt this be `&& !LorenzUtils.isAprilFoolsDay`

lineargraph:
Also I feel like spinning and flipping should be two options in the skyhanni contributor list. (Also i really dislike spinning still. I dont think it should exists)

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RendererLivingEntityHook.kt
CalMWolfs: I'll look in the morning

hannibal002 (approved):
works fine (except config description)

PR #1522, opened 2024-04-23
Backend: Replace Internal Commands 

hannibal002:
## What
replace many internal commands with onClick feature

## Changelog Technical Details
+ Replace many internal commands with onClick() feature. - hannibal2

PR #1524, opened 2024-04-23
Improvement: Command wording update

catgirlseraid:
## Dependencies
- #1522 

## What
changed name of config reset command to make it easier to find

## Changelog Improvements
+ Changed Reset Config Command. - seraid


hannibal002 (approved):
this command was originally designed to not reset the config, but rather to reload the config render processor, to update changed config classes in the overlay after hot swapping. the warning in the command description was only a bug, not a intended feature.

but this change makes sense, until the oroginal bug gets fixed and you have two commands, one with and one without reset, but both reload the config elements

PR #1537, opened 2024-04-24
Backend: Cleanup code

CalMWolfs:
This simplifies the mixin logic used, now we only check cool person where the original mc code checks for Dinnerbone. As we dont want to use the cape logic from the original code we always return true here. If this check for the cape is occurring they are already on the cool person list (or they are Grum) so we see if the flipping logic needs to be applied then we apply it

exclude_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/mixins/transformers/renderer/MixinContributorRendererEntityLiving.java
+++ b/src/main/java/at/hannibal2/skyhanni/mixins/transformers/renderer/MixinContributorRendererEntityLiving.java
@@ -17,12 +17,13 @@ public class MixinContributorRendererEntityLiving<T extends EntityLivingBase> {
     private final RendererLivingEntityHook skyHanni$hook = new RendererLivingEntityHook();
 
     @Redirect(method = "rotateCorpse", at = @At(value = "INVOKE", target = "Ljava/lang/String;equals(Ljava/lang/Object;)Z", ordinal = 0))
-    private boolean rotateCorpse(String displayName, Object v2, T bat, float p_77043_2_, float p_77043_3_, float partialTicks) {
-        return skyHanni$hook.rotateCorpse(displayName, bat);
+    private boolean checkName(String displayName, Object v2, T bat, float p_77043_2_, float p_77043_3_, float partialTicks) {
+        return skyHanni$hook.isCoolPerson(displayName);
     }
 
     @Redirect(method = "rotateCorpse", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/player/EntityPlayer;isWearing(Lnet/minecraft/entity/player/EnumPlayerModelParts;)Z"))
-    private boolean rotateCorpse(EntityPlayer bat, EnumPlayerModelParts p_175148_1_) {
-        return skyHanni$hook.isWearing(bat, EnumPlayerModelParts.CAPE);
+    private boolean isWearing(EntityPlayer player, EnumPlayerModelParts p_175148_1_) {
+        skyHanni$hook.rotatePlayer(player);
+        return true;
     }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/mixins/transformers/renderer/MixinContributorRendererEntityLiving.java
```diff
@@ -17,12 +17,13 @@ public class MixinContributorRendererEntityLiving<T extends EntityLivingBase> {
...
+    private boolean checkName(String displayName, Object v2, T bat, float p_77043_2_, float p_77043_3_, float partialTicks) {
+        return skyHanni$hook.isCoolPerson(displayName);
     }
 
     @Redirect(method = "rotateCorpse", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/player/EntityPlayer;isWearing(Lnet/minecraft/entity/player/EnumPlayerModelParts;)Z"))
-    private boolean rotateCorpse(EntityPlayer bat, EnumPlayerModelParts p_175148_1_) {
-        return skyHanni$hook.isWearing(bat, EnumPlayerModelParts.CAPE);
+    private boolean isWearing(EntityPlayer player, EnumPlayerModelParts p_175148_1_) {
+        skyHanni$hook.rotatePlayer(player);
+        return true;
```
hannibal002, line 27: not a fan of forcefully modify minecraft, without options to disable/change

src/main/java/at/hannibal2/skyhanni/mixins/transformers/renderer/MixinContributorRendererEntityLiving.java
```diff
@@ -17,12 +17,13 @@ public class MixinContributorRendererEntityLiving<T extends EntityLivingBase> {
     private final RendererLivingEntityHook skyHanni$hook = new RendererLivingEntityHook();
 
     @Redirect(method = "rotateCorpse", at = @At(value = "INVOKE", target = "Ljava/lang/String;equals(Ljava/lang/Object;)Z", ordinal = 0))
-    private boolean rotateCorpse(String displayName, Object v2, T bat, float p_77043_2_, float p_77043_3_, float partialTicks) {
-        return skyHanni$hook.rotateCorpse(displayName, bat);
+    private boolean checkName(String displayName, Object v2, T bat, float p_77043_2_, float p_77043_3_, float partialTicks) {
```
hannibal002, line 20: can we keep using the original function name for consistency?

hannibal002 (changes requested):
works fine in game

src/main/java/at/hannibal2/skyhanni/mixins/transformers/renderer/MixinContributorRendererEntityLiving.java
CalMWolfs: This mixin is only called if we have already modified the code, (when spin contributors is on and they are on the cool person list). 
> As we dont want to use the cape logic from the original code we always return true here. If this check for the cape is occurring they are already on the cool person list (or they are Grum) so we see if the flipping logic needs to be applied then we apply it

So in the end the only two minecraft accounts this mixin will ever affect is Dinnerbone and Grum

src/main/java/at/hannibal2/skyhanni/mixins/transformers/renderer/MixinContributorRendererEntityLiving.java
CalMWolfs: I feel using this new function name is more descriptive of what part of the function is being mixed into, but if you really want I can revert the function name

src/main/java/at/hannibal2/skyhanni/mixins/transformers/renderer/MixinContributorRendererEntityLiving.java
hannibal002: fair

PR #1539, opened 2024-04-24
Improvement: Command to clear kismet

Thunderblade73:
## What
A fix to clear the kismet data. Since currently there is one way to bug out the state (finish without milestone).

See: https://discord.com/channels/997079228510117908/1232355086424997910 (does not fix the issue here, but users can get back to a usable state)

## Changelog Improvements
+ Added a command to clear the kismet state in Croesus. - Thunderblade73

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/dungeon/CroesusChestTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/dungeon/CroesusChestTracker.kt
@@ -261,6 +261,14 @@ class CroesusChestTracker {
 
         private val croesusChests get() = ProfileStorageData.profileSpecific?.dungeons?.runs
 
+        fun resetChest() = croesusChests?.let {
+            it.clear()
+            it.addAll(generateMaxChest())
+        }
+
+        fun generateMaxChest() = generateSequence { DungeonRunInfo() }.take(maxChests)
+        fun generateMaxChestAsArrayList() = generateMaxChest().toCollection(ArrayList())
+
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/features/dungeon/CroesusChestTracker.kt
```diff
@@ -261,6 +261,14 @@ class CroesusChestTracker {
 
         private val croesusChests get() = ProfileStorageData.profileSpecific?.dungeons?.runs
 
+        fun resetChest() = croesusChests?.let {
+            it.clear()
+            it.addAll(generateMaxChest())
```
hannibal002, line 267: The command has no chat feedback.

src/main/java/at/hannibal2/skyhanni/features/dungeon/CroesusChestTracker.kt
```diff
@@ -261,6 +261,14 @@ class CroesusChestTracker {
...
+        }
+
+        fun generateMaxChest() = generateSequence { DungeonRunInfo() }.take(maxChests)
+        fun generateMaxChestAsArrayList() = generateMaxChest().toCollection(ArrayList())
```
hannibal002: can we use `toList()` instead of `toCollection(ArrayList())`?

hannibal002 (changes requested):
looks good

hannibal002 (approved):

PR #1544, opened 2024-04-25
Backend: PlayerDeathEvent

hannibal002:
## What
Creating and using PlayerDeathEvent.

+ Created and implemented PlayerDeathEvent. - hannibal2


src/main/java/at/hannibal2/skyhanni/data/PlayerDeathManager.kt
```diff
@@ -0,0 +1,30 @@
...
+        deathMessagePattern.matchMatcher(message) {
+            val name = group("name")
+            val reason = group("reason").removeColor()
+            PlayerDeathEvent(name, reason, event).postAndCatch()
+
+        }
+    }
+}
+
+class PlayerDeathEvent(val name: String, val reason: String, val chatEvent: LorenzChatEvent) : LorenzEvent()
```
CalMWolfs: This should be moved to its own class like all other event

src/main/java/at/hannibal2/skyhanni/data/PlayerDeathManager.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    @SubscribeEvent
+    fun onChat(event: LorenzChatEvent) {
+        val message = event.message
+        deathMessagePattern.matchMatcher(message) {
+            val name = group("name")
+            val reason = group("reason").removeColor()
+            PlayerDeathEvent(name, reason, event).postAndCatch()
+
+        }
```
CalMWolfs:
```suggestion
            PlayerDeathEvent(name, reason, event).postAndCatch()
        }
```

src/main/java/at/hannibal2/skyhanni/data/MiningAPI.kt
```diff
@@ -57,11 +52,13 @@ object MiningAPI {
...
         }
-        coldResetDeath.matchMatcher(event.message) {
-            if (group("name") == LorenzUtils.getPlayerName()) {
-                updateCold(0)
-                lastColdReset = SimpleTimeMark.now()
-            }
+    }
+
+    @SubscribeEvent
+    fun onPlayerDeath(event: PlayerDeathEvent) {
```
CalMWolfs, line 135: we could early return if the player name is not the user but not necessary

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/data/MiningAPI.kt
hannibal002: early return is not needed and rather confusing when its the only check in the function imo

hannibal002: also this is no precondition, but THE condition

src/main/java/at/hannibal2/skyhanni/data/PlayerDeathManager.kt
hannibal002: i dont like the current design of all events in one event package. this only provides us with one long list, without order.
rather have the events where they are getting fired or use a more strict sub package strcuture, what do you thing?

CalMWolfs: I think for the sake of consistency this should be moved. I like the idea of specific sub packages and I do plan on doing something about that in the next few weeks but for now we should stick to the existing system

hannibal002: agreed

PR #1546, opened 2024-04-25
Feature: Tunnels maps

Thunderblade73:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/165

## What
A kind of google maps for the glacite tunnels.
Supports all gemstones and tungsten, umber and mithril. Does not support glacite since this is literally everywhere.
Brings you only to the general location of the targeted ore/gem not directly to the veins.
The target can also be set via the commission inventory for collect commissions.

<details>
<summary>Images</summary>

image
Unbenannt
image image|
image
image
image
image
image

</details>

## Changelog New Features
+ Added Tunnel Maps. - Thunderblade73
    * Provides a way to navigate inside the new Glacite Tunnels.


hannibal002:
```
SkyHanni 0.25.Beta.21: Caught an NoSuchElementException in TunnelsMaps at InventoryFullyOpenedEvent: Collection contains no element matching the predicate.
 
Caused by java.util.NoSuchElementException: Collection contains no element matching the predicate.
	at SH.features.mining.TunnelsMaps.getGenericName(TunnelsMaps.kt:413)
	at SH.features.mining.TunnelsMaps.onInventoryFullyOpened(TunnelsMaps.kt:145)
	at SH.data.OtherInventoryData.done(OtherInventoryData.kt:95)
	at SH.data.OtherInventoryData.onChatPacket(OtherInventoryData.kt:88)
	at SH.mixins.hooks.NetworkManagerHookKt.onReceivePacket(NetworkManagerHook.kt:8)
```

src/main/java/at/hannibal2/skyhanni/data/EntityMovementData.kt
```diff
@@ -41,6 +46,17 @@ class EntityMovementData {
         }
     }
 
+    private val warpingPattern by RepoPattern.pattern("warping", "§7Warping...")
```
ThatGravyBoat: There are multiple warp messages that are possible that this will miss, see: [ChatFilter.kt#L71](https://github.com/hannibal002/SkyHanni/blob/beta/src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt#L71)

I also dont really understand the point of a warp event when the `LorenzWorldChangeEvent` exists. If its because hypixel calls it 3 times then do what I did in [DiscordRpcManager.kt](https://github.com/hannibal002/SkyHanni/blob/beta/src/main/java/at/hannibal2/skyhanni/features/misc/discordrpc/DiscordRPCManager.kt#L177) and cache the time when it will run the runnable and then check if its in the future and not try to run a delayed task again.

Thunderblade73: A WorldChange mustn't be the case here, this also triggers when you warp locally in the same lobby. The other messages shouldn't matter since they only happen if you warp to an island that your not on. Sure the event isn't perfect since it also triggers when world changing but it works.

ThatGravyBoat: As to not cause issues in the future if someone were to use this and expect for it to work on all warps this should support the other messages too.

hannibal002 (approved):

PR #1552, opened 2024-04-25
Improvement + Fix: Item tooltips disappearing / changing

VixidDev:
## What
Fixes item tooltips disappearing / changin when having the coloring toggle off, causing the code to end up in a situation where `cachedLoreAfter` was never set, but the lore was still considered cached and tried to read an empty `cachedLoreAfter`, also made sure to set the `cachedLoreAfter` in a specific situation where it would cause irrelevant/incorrect tooltips to be shown for other items.

Also makes it so enchant books always show their enchant descriptions.

## Changelog Improvements
+ Enchant books always show descriptions regardless of if 'Hide enchant description' is enabled. - Vixid

## Changelog Fixes
+ Fixed tooltips disappearing with the Enchant Parsing feature. - Vixid

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/items/enchants/EnchantParser.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/items/enchants/EnchantParser.kt
@@ -166,6 +166,7 @@ object EnchantParser {
                 if (chatComponent != null) editChatComponent(chatComponent, loreList)
                 return
             }
+            loreCache.updateAfter(loreList)
             return
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/misc/items/enchants/EnchantParser.kt
```diff
@@ -344,6 +356,11 @@ object EnchantParser {
         return if (removeGrayEnchants) -1 else lastGrayEnchant
     }
 
+    // Surely Hypixel won't change Enchanted Book display names right... right?
+    private fun itemIsBook() : Boolean {
```
Thunderblade73, line 362: use itemCategory here.
```
currentItem?.itemCategoryOrNull() == ItemCategory.ENCHANTED_BOOK
```
(not 100% correct, but close enough)

hannibal002 (approved):

PR #1556, opened 2024-04-26
Backend: Use less deprecated LorenzVec functions

CalMWolfs:
## What
Use less deprecated LorenzVec functions.
- There were two usages of LorenzVec.subtract in SoopyGuessBurrow that i did not change due to that whole file being a mess and doing a lot of null assertion
- Cleanup of duplicate code in EntityUtils
- Also optimised imports for the whole project
- Got rid of multiplyZeroSave  because there is no need for it

+ Use fewer deprecated LorenzVec functions. - CalMWolfs
+ Cleanup of duplicate code in EntityUtils. - CalMWolfs


Thunderblade73 (approved):
Looked at all changes and the semantics are the same. I might have missed stuff in the Entity Utils, since quite a lot changed there. With the multiply zero save change, Idk, since there might be a reason why we needed, but I can't tell.

CalMWolfs:
> With the multiply zero save change, Idk, since there might be a reason why we needed, but I can't tell.

I want to say hannibal just randomly thought of it two years ago, I cannot think of a reason either, at first I thought it was maybe for the sign of the number but then again 0.0 == -0.0 so i have no clue

PR #1565, opened 2024-04-27
Backend: Update neu version in recommended version

CalMWolfs:
exclude_from_changelog


j10a1n15:
You could also bump the devenv version

CalMWolfs:
> You could also bump the devenv version

I guess, the tag was weird with jitpack because it was deleted and remade so the jitpack has the one month old version. But yea ill do it gimme a second

hannibal002 (approved):

PR #1568, opened 2024-04-27
Fix: Guild chat rank detection

hannibal002:
## What
Fixed party chat rank detection.

## Changelog Fixes
+ Fixed some guild chat messages having wrong format. - hannibal2


jani270 (approved):
e

PR #1570, opened 2024-04-27
Fix: Cleanup RPC

ThatGravyBoat:
## What
This PR cleans up and improves parts of the RPC manager.
- Fixes 1 issue with the button PR that caused a error if disabled due when I published the version to maven it was published from the git version and not what I had locally on my system.
- I have also updated IPC 1 more time add make it easier to handle buttons.
- Cleaned up unused code
- Fixed a bug where every world swap 3 executor services were created
- Changed connected boolean to check the socket state instead.

## Changelog Fixes
+ Fixed issue where RPC would error if SkyCrypt button was turned off. - ThatGravyBoat

+ Made DelayedRun return its SimpleTimeMark. - ThatGravyBoat
    * This is useful as to not over extraneously run an operation when another is already queued.


src/main/java/at/hannibal2/skyhanni/features/misc/discordrpc/DiscordRPCManager.kt
```diff
@@ -60,66 +54,55 @@ object DiscordRPCManager : IPCListener {
...
-                    consoleLog(ex.toString())
-                    ChatUtils.clickableChat(
-                        "Discord Rich Presence was unable to start! " +
-                            "This usually happens when you join SkyBlock when Discord is not started. " +
-                            "Please run /shrpcstart to retry once you have launched Discord.",
-                        onClick = {
-                            startCommand()
-                        }
-                    )
+                client = IPCClient(APPLICATION_ID).apply {
```
hannibal002: extract code inside apply into a new function. Too many intendations

hannibal002 (changes requested):
code cleanup. not tested in game yet

src/main/java/at/hannibal2/skyhanni/features/misc/discordrpc/DiscordRPCManager.kt
ThatGravyBoat: Ive moved it to its own method and cleaned up some misnaming of a method

hannibal002 (changes requested):
Could not resolve com.jagrosh:DiscordIPC:0.5.2.

hannibal002 (approved):
works for me

PR #1578, opened 2024-04-28
Feature: Update command

ItsEmpa:
## What
Adds a `/shupdate` command, that can be used to directly start downloading an update. By default it will use the current update stream, but that can be changed by using `/shupdate <beta/release>`.

## Changelog New Features
+ Added `/shupdate` command. - Empa
    * Can be used like `/shupdate <beta/full>` to download updates from a specific update stream.


ItsEmpa:
tested it now, seems to be working so im marking it ready for review

hannibal002 (approved):

PR #1585, opened 2024-04-28
Improvement: Hoppity Menu Shortcut appearing in islands that it's not supposed to be used in.

raaaaaven:
## What
Removed the Hoppity Menu Shortcut in areas that cannot access the Chocolate Factory.

## Changelog Improvements
+ Removed the Hoppity Menu Shortcut in The Rift, Kuudra, Catacombs and Mineshafts. - raven
    * You cannot use the chocolate factory in these areas, resulting in the button being removed.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/event/chocolatefactory/ChocolateFactoryShortcut.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/event/chocolatefactory/ChocolateFactoryShortcut.kt
@@ -1,9 +1,9 @@
 package at.hannibal2.skyhanni.features.event.chocolatefactory
 
+import at.hannibal2.skyhanni.data.IslandType
 import at.hannibal2.skyhanni.events.GuiContainerEvent
 import at.hannibal2.skyhanni.events.InventoryCloseEvent
 import at.hannibal2.skyhanni.events.InventoryFullyOpenedEvent
-import at.hannibal2.skyhanni.features.rift.RiftAPI
 import at.hannibal2.skyhanni.utils.HypixelCommands
 import at.hannibal2.skyhanni.utils.LorenzUtils
 import at.hannibal2.skyhanni.utils.NEUInternalName.Companion.asInternalName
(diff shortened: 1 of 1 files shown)


CalMWolfs (approved):

src/main/java/at/hannibal2/skyhanni/features/event/chocolatefactory/ChocolateFactoryShortcut.kt
```diff
@@ -36,7 +36,7 @@ class ChocolateFactoryShortcut {
 
     @SubscribeEvent
     fun onInventoryOpen(event: InventoryFullyOpenedEvent) {
-        if (RiftAPI.inRift()) return
+        if (LorenzUtils.inAnyIsland(IslandType.THE_RIFT, IslandType.KUUDRA_ARENA, IslandType.CATACOMBS, IslandType.MINESHAFT)) return
         if (!LorenzUtils.inSkyBlock) return
```
walkerselby, line 39: The not in SkyBlock precondition should happen before the Skyblock Island specific precondition; I know it was inverted before you started editing but while you are here, could you flip it?

walkerselby (changes requested):

src/main/java/at/hannibal2/skyhanni/features/event/chocolatefactory/ChocolateFactoryShortcut.kt
CalMWolfs: Better yet. InAntIsland checks if you are in skyblock already do that line can be completely removed

raaaaaven: Done.

CalMWolfs (approved):

hannibal002 (changes requested):
now we only return when the user is in skyblock AND in one of those islands. we do not return when the user is not in skyblock.

raaaaaven:
> now we only return when the user is in skyblock AND in one of those islands. we do not return when the user is not in skyblock.

Should be fixed now since I added the skyblock check once again.

hannibal002 (approved):
Looks fine now

PR #1588, opened 2024-04-28
Update CONTRIBUTING.md

101dal:
## What
Added a direct link to the official Oracle website to download the java jdk version needed to make hot swap work


lineargraph:
Those links invalidate after some time i am pretty sure (also i dont think this is legal). Just make an account it really is not that hard

101dal:
ok I close

PR #1517, opened 2024-04-23
Feature: Chocfactory assorted

catgirlseraid:
## Dependencies
- #1514

## What
edited a few hoppity features to make them better,
added copying chocolate factory stats to clipboard when hovering
added a highlight to unbought items in the shop

## Changelog New Features
+ Copy Chocolate Factory Stats to clipboard. - seraid
+ Highlight unbought items in Hoppity shop. - seraid

## Changelog Improvements
+ Improved various Hoppity features. - seraid

PR #1593, opened 2024-04-29
Backend: more performance

hannibal002:
## Dependencies
- #1517

## What
more performance

exclude_from_changelog


CalMWolfs (approved):
worked for me

PR #1595, opened 2024-04-30
Fix: VisitorRewardWarning only blocking normal clicks

Obsidianninja11:
## What
Fixes being able to bypass the reward warning with middle click (or drop or hotbar hotkeys) https://discord.com/channels/997079228510117908/1233943553529610320

## Changelog Fixes
+ Fixed visitor reward warning only blocking normal clicks. - Obsidian


hannibal002 (approved):
works for me

PR #1603, opened 2024-04-30
Fix: Defer renderable tooltips to end of rendering

ThatGravyBoat:
## What
This PR moves the actual rendering of renderable tooltips to the end of all rendering as to not have issues with overlays in guis rendering over them. i.e. highlights

## Changelog Fixes
+ Fixed slot highlight showing over item tooltip. - ThatGravyBoat

+ Renderable tooltips are now deferred in end of RenderTick. - ThatGravyBoat
    * Tooltips should only be set in between RenderTickEvent start and RenderTickEvent end. i.e. screen render events.


hannibal002 (approved):
image

PR #1626, opened 2024-05-01
Improvement: disable mark low levels

catgirlseraid:
## What
if mark low levels is  set to 0 the feature is disabled

## Changelog Improvements
+ Added a way to disable marking low-level parties in dungeon finder by setting it to 0. - seraid


CalMWolfs (approved):

ThatGravyBoat (approved):
LGTM

PR #1636, opened 2024-05-01
Feature: Changelog viewer

Thunderblade73:
## What
Shows the changelog ingame before upadeting and after.

<details>
<summary>Images</summary>

image
image

</details>

## Changelog New Features
+ Added In-Game Changelog Viewer. - Thunderblade73
    * Clickable chat option shows on new update or first join with new version.
    * Works with `/shchangelog <old version> <new version>`.


src/main/java/at/hannibal2/skyhanni/data/jsonobjects/other/ChangelogJson.kt
```diff
@@ -0,0 +1,12 @@
+package at.hannibal2.skyhanni.data.jsonobjects.other
+
+import com.google.gson.annotations.Expose
+import com.google.gson.annotations.SerializedName
+
+class ChangelogJson(
```
ThatGravyBoat: use a data class, be consistent with all other json objects and put serializedName and expose on the same line as the property.

src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangelogViewer.kt
```diff
@@ -0,0 +1,362 @@
...
+        return true
+    }
+
+    private fun String.toVersionTag(): VersionTag {
+        val split = this.split('.')
+        val ints = split.filter { it.isInt() }.map { it.toInt() }
+        return VersionTag(ints, split.contains("Beta"))
+    }
+
+    private fun getChangelog(currentVersion: VersionTag, targetVersion: VersionTag) {
```
ThatGravyBoat: I dont think doing this in the mod is the right choice, I think this should be done in a workflow and similar to the regexes the minecraft formatted changelog should be uploaded there. 

People already have issues with 429 errors on github I dont think calling more github apis is a good idea when it can be automatically uploaded to the repo that has to be downloaded on startup anyway.

src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangelogViewer.kt
```diff
@@ -0,0 +1,362 @@
...
+                            }
+                        }.toMap()
+                }
+            } catch (e: Exception) {
+                ErrorManager.logErrorWithData(e, "Changelog Loading Failed")
+            }
+        }
+    }
+
+    fun handelCommand(args: Array<String>) {
```
ThatGravyBoat:
```suggestion
    fun handleCommand(args: Array<String>) {
```

src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangelogViewer.kt
```diff
@@ -0,0 +1,362 @@
...
+        }
+        getChangelog(currentVersion, targetVersion)
+        openChangelog()
+    }
+
+    private fun openChangelog() {
+        if (Minecraft.getMinecraft().currentScreen !is ChangeLogViewer) SkyHanniMod.screenToOpen = ChangeLogViewer()
+    }
+
+    class ChangeLogViewer : GuiScreen() {
```
ThatGravyBoat: this is way too big to be an inner class, this should be moved to its own file.

src/main/java/at/hannibal2/skyhanni/utils/CollectionUtils.kt
```diff
@@ -313,4 +313,7 @@ object CollectionUtils {
     fun <K, V : Any> Map<K?, V>.filterNotNullKeys(): Map<K, V> {
         return filterKeys { it != null } as Map<K, V>
     }
+
+    fun <K, V> Map<K, V>.containsKey(vararg keys: K) = keys.all { this.keys.contains(it) }
```
ThatGravyBoat: this easily conflicts with the existing `Map.containsKey` this is also misnamed as its a vararg of keys so it should be `containsKeys` also when changing to `containsKeys` change `this.keys.contains(it)` to `this.containsKey(it)`

src/main/java/at/hannibal2/skyhanni/features/misc/update/GuiOptionEditorUpdateCheck.kt
```diff
@@ -31,25 +34,28 @@ class GuiOptionEditorUpdateCheck(option: ProcessedOption) : GuiOptionEditor(opti
...
 
-        val widthRemaining = width - button.width - 10
-
         GlStateManager.scale(2F, 2F, 1F)
-        val currentVersion = SkyHanniMod.version
         val sameVersion = currentVersion.equals(nextVersion, true)
         TextRenderUtils.drawStringCenteredScaledMaxWidth(
-            "${if (UpdateManager.updateState == UpdateManager.UpdateState.NONE) GREEN else RED}$currentVersion" +
-                if (nextVersion != null && !sameVersion) "➜ ${GREEN}${nextVersion}" else "",
+            "${if (UpdateManager.updateState == UpdateManager.UpdateState.NONE) GREEN else RED}$currentVersion" + if (nextVersion != null && !sameVersion) "➜ ${GREEN}${nextVersion}" else "",
```
ThatGravyBoat: please use the color codes and fix the existing things in this class from using EnumChatFormatting as to keep consistent with almost every other chatformatting usages in the codebase.

ThatGravyBoat (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangelogViewer.kt
Thunderblade73: TLDR: I would want to stay at the approach of using the github api but inform the user about the rate limit.

This is completely optional for player so when the get rate limited the would not care that much. I can add a extra case for the rate limited state, so that we are transparent. Also I don't like to add a changelog to the repo each update since that while increase the repo to stupid amounts (over a long period of time. Which about 1,8KB per day (uncompressed) and that brings the repo since to 0,5MB over a year just from changelogs and the repo sit at the moment at 0,38MB for all things for comperisson). Sure there is no effektiv limit on size on the repo but I don't want to grow the repo infidelity.

Also most of the user will one use 1 token (from 60 per hour). So I don't see it that rate limiting is a issue here, since it isn't automatic like everything else that uses tokens. And the calls to the github api are completely cached (get's cleared after 30min of not using the feature) meaning even if the open it multiple times the only use 1 token. The only case where the use multiple tokens is when it the current version is really old and then it is just 1 token per 100 versions.

src/main/java/at/hannibal2/skyhanni/features/misc/update/GuiOptionEditorUpdateCheck.kt
Thunderblade73: I didn't change anything here, the formatting just removed the wrap and to put it simply the rendering of that class uses only NEU calls. So the would need to be reworked as a whole.

ThatGravyBoat: I know you didnt change anything here but I am saying while you're here fix it, because this is the only class in the whole codebase that actually uses the EnumChatFormatting inline in a chat unlike everywhere else it uses the section symbols and the color codes

src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangelogViewer.kt
ThatGravyBoat: I know it has to be done by a user but that is not excuse to allow the player to get rate limited many mods rely on the GitHub API including SkyHanni so if you get them rate limited needlessly then for an hour they cant use the GitHub API which will cause many problems for several mods. 

People already get rate limited already due to how many mods use the GitHub API literally while trying to test this I got rate limited on the first request.

Thunderblade73:
Currently stale as I don't have an clear answer if it should use github api or repo.
https://discord.com/channels/997079228510117908/1323678704303345725

hannibal002:
forgot this pr exists. its on me

Thunderblade73:
TODO:
- [x] fix version comparison
- [x] auto fill the 0 at the end if not found
- [x] a "help" for the command arguments
- [x] use DrawContext instead of glStateManger
- [x] fix detekt

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
```diff
@@ -1566,6 +1568,50 @@ object RenderUtils {
...
+        RoundedRectangleShader.halfSize = floatArrayOf(widthIn / 2f, heightIn / 2f)
+        RoundedRectangleShader.centerPos = floatArrayOf(xIn + (widthIn / 2f), yIn + (heightIn / 2f))
+
+        GlStateManager.pushMatrix()
+        ShaderManager.enableShader(ShaderManager.Shaders.ROUNDED_RECTANGLE)
+
+        drawGradientRect(x - 5, y - 5, x + width + 5, y + height + 5, topColor, bottomColor)
+
+        ShaderManager.disableShader()
+        GlStateManager.popMatrix()
```
CalMWolfs: can we use DrawContextUtils for these

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
```diff
@@ -1566,6 +1568,50 @@ object RenderUtils {
...
+        x: Int,
+        y: Int,
+        width: Int,
+        height: Int,
+        topColor: Int,
+        bottomColor: Int,
+        radius: Int = 10,
+        smoothness: Int = 1,
+    ) {
+        val scaledRes = ScaledResolution(Minecraft.getMinecraft())
```
CalMWolfs: Instead of Minecrafts ScaledResolution please use GuiScreenUtils

src/main/java/at/hannibal2/skyhanni/features/misc/update/GuiOptionEditorUpdateCheck.kt
```diff
@@ -32,25 +35,28 @@ class GuiOptionEditorUpdateCheck(option: ProcessedOption) : GuiOptionEditor(opti
...
+        if (UpdateManager.updateState != UpdateManager.UpdateState.NONE) {
+            changelog.render(getChangelogPosition(width), 30)
+        }
+
+        val widthRemaining = width - button.width - 10
+
         if (UpdateManager.updateState == UpdateManager.UpdateState.DOWNLOADED) {
-            TextRenderUtils.drawStringCentered(
-                "${GREEN}The update will be installed after your next restart.",
+            TextRenderUtils.drawStringCenteredScaledMaxWidth(
```
CalMWolfs: Please use GuiRenderUtils

src/main/java/at/hannibal2/skyhanni/features/misc/update/GuiOptionEditorUpdateCheck.kt
```diff
@@ -81,6 +85,20 @@ class GuiOptionEditorUpdateCheck(option: ProcessedOption) : GuiOptionEditor(opti
             }
             return true
         }
+        if (Mouse.getEventButtonState() && (mouseX - getChangelogPosition(width) - x) in (0..changelog.width) && (mouseY - 30 - y) in (0..changelog.height)) {
```
CalMWolfs: Please use MouseCompat

src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangelogViewer.kt
```diff
@@ -0,0 +1,189 @@
...
+                    showChangelog(tag, tag)
+                } else {
+                    showChangelog(current, tag)
+                }
+            }
+
+            2 -> {
+                val target = ModVersion.fromString(args[0])
+                if (!target.isValid()) {
+                    ChatUtils.userError("Version shape requirement failed, first argument")
```
CalMWolfs: I think saying shape requirement can be confusing to users
maybe instead: "Invalid version: \<version> versions are in the following format: \<either random version number or current skyhanni version number>"

CalMWolfs (changes requested):
I know you are still working on this and are also waiting on hannibal to choose between github api/repo but these are a few things I have noticed

src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangelogViewer.kt
Thunderblade73: I will redo the whole command part as an complex command, which gives me more control and also makes the responses more accurate to write.

src/main/java/at/hannibal2/skyhanni/features/misc/update/GuiOptionEditorUpdateCheck.kt
Thunderblade73: That's inside moulconfig. So idk if it would be correct to use GuiRenderUtils.

CalMWolfs: > That's inside moulconfig. So idk if it would be correct to use GuiRenderUtils.

We put the exact moulconfig method in guiRenderUtils

hannibal002 (changes requested):
please use either to use github api (preferred) or the repo

hannibal002:
image

forgot this comment existed, its on me

hannibal002 (changes requested):
image

hannibal002:
wouldn't it be more useful to show the changelog when a new udpate is found?
image

to reproduce, i downgraded the version to 3.4.0 before running this pr

hannibal002 (changes requested):
please review my changes and feedback on the idea of showing the changelog viewer when update is available

Thunderblade73:
> wouldn't it be more useful to show the changelog when a new udpate is found? image
> 
> to reproduce, i downgraded the version to 3.4.0 before running this pr

There is a button on the update thingy in config. That should be enough.

hannibal002:
> There is a button on the update thingy in config. That should be enough.

i have not noticed this button while testing.
either way, having a additional chat message "click here to see the changes in the new version" next to the "your sh is outdated" messgae is def a good addition.

hannibal002:
last time i tested the pr i didnt notice the button in the config.
will test again later today.

either way - i still heavily think we shoud also show the "changelog preview/button to click on" in chat together with the "sh fould an upodate" message.

hannibal002 (changes requested):
i deleted the sh folder and dropped the version by one. there is no changelog visivle anywhere in the chat.
image

hannibal002:
@Thunderblade73 bump

PR #1638, opened 2024-05-02
Feature: added chocolate until next prestige

catgirlseraid:
## What
Added Chocolate until next prestige

image

is off by default

## Changelog New Features
+ Added amount of chocolate until next prestige to stats display. - seraid


CalMWolfs (changes requested):
can be negative if you have more than prestige, also shouldnt show when you are max prestige (5)

CalMWolfs (approved):
changes looking good to me

hannibal002 (approved):

PR #1641, opened 2024-05-02
Improvement: Show stars on time tower and coach jackrabbit

CalMWolfs:
## What
- Star now shows on time tower and coach jackrabbit.
- Maxed upgrades in chocolate factory now show a checkmark.
- All upgrades in chocolate factory are now stored as ChocolateFactoryUpgrade. 
- Rename some events to the correct thing.
- Store when next best upgrade can be bought.

image
image

## Changelog Improvements
+ Star now shows on time tower and coach jackrabbit. - CalMWolfs
+ Maxed upgrades in Chocolate Factory now show a checkmark. - CalMWolfs

+ All upgrades in chocolate factory are now stored as ChocolateFactoryUpgrade. - CalMWolfs

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryAPI.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryAPI.kt
@@ -6,34 +6,19 @@ import at.hannibal2.skyhanni.config.features.inventory.chocolatefactory.Chocolat
 import at.hannibal2.skyhanni.config.storage.ProfileSpecificStorage.ChocolateFactoryStorage
 import at.hannibal2.skyhanni.data.ProfileStorageData
 import at.hannibal2.skyhanni.data.jsonobjects.repo.HoppityEggLocationsJson
-import at.hannibal2.skyhanni.events.InventoryCloseEvent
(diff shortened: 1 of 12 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryAPI.kt
```diff
@@ -156,19 +171,18 @@ object ChocolateFactoryAPI {
     }
 
     private fun updateInventoryItems(inventory: Map<Int, ItemStack>) {
```
hannibal002: 97 lines function. refactor in multiple smaller functions

hannibal002 (changes requested):
not tested in game.

ChocolateFactoryAPI no longer feels like an API class, but rather like a "do everything" class. It should just provide the data, no feature logic. Everything else should get moved to utils classes or smaller feature-specific classes. If the same math is needed for multiple features, then we should still move it to utils. If the same data+math is needed for multiple features, then we can have a SMALL function in the API that connects data with utils functions. 300+ lines for one class is definitely too long.
If a lot of the code in the API is just "reading and processing data", then move that into a "data loader" class, and again extract utils into a new chocolate factory utils class.

hannibal002 (approved):

PR #1643, opened 2024-05-02
Improvement: Option to not flip yourself (Fancy Contributor)

Obsidianninja11:
## What
Adds an option to not flip yourself in Fancy Contributor section.
I put it in technical because it doesn't really apply to most users.

+ Option to not flip yourself (Fancy Contributors). - Obsidian

exclude_from_changelog


hannibal002:
Im against this pr.

Rather just remove yourself in repo file instead.
Why should others see you spinny if you don't like to see it yourself?
If this option magically also toggles the repo value/change your spinny for others, sure.
Since this is client side only, I don't see the reason to have this option.

Additional note: we intentionally don't have "features only for contributions" in the mod. A new developer that has not contacted us yet should be able to work with all options in the dev category independently.

PR #1647, opened 2024-05-02
Improvement: Pest line middle of plot

hannibal002:
## What
Show middle of plot while pest line is showing

## Changelog Improvements
+ Displays the middle of the plot while the pest line is showing. - hannibal2
    * This makes it clearer when the guess points to the middle of the plot rather than to a pest.


CalMWolfs:
While testing this didnt seem useful at all, what is the main point of this?

CalMWolfs (approved):

PR #1651, opened 2024-05-02
Fix: Max Player Size in Mineshaft

j10a1n15:
## What
Max Player Count is 4 in there.

## Changelog Fixes
+ Fixed wrong max lobby size in Custom Scoreboard size while in Mineshaft. - j10a1n15

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/HypixelData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/HypixelData.kt
@@ -182,11 +182,12 @@ class HypixelData {
             ScoreboardData.sidebarLinesFormatted.matchFirst(scoreboardVisitingAmoutPattern) {
                 return group("maxamount").toInt()
             }
-            val islandSpecific = mapOf(
-                IslandType.MINESHAFT to 4,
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/data/HypixelData.kt
```diff
@@ -183,8 +182,11 @@ class HypixelData {
             ScoreboardData.sidebarLinesFormatted.matchFirst(scoreboardVisitingAmoutPattern) {
                 return group("maxamount").toInt()
             }
-            if (IslandType.CRYSTAL_HOLLOWS.isInIsland()) return 24
-            return if (serverId?.startsWith("mega") == true) 80 else 26
+            val islandSpecific = mapOf(
```
Thunderblade73: please use a when here. Instead of what ever this is. The null case is just the else branch of the when

Thunderblade73 (changes requested):

Thunderblade73 (approved):

hannibal002 (approved):

PR #1658, opened 2024-05-02
Fix upside down mixin and split up spinny from upside down

lineargraph:
## Pseudo Dependency
Works entirely without https://github.com/hannibal002/SkyHanni-REPO/pull/182, but that PR would be useful.

## What
Fixes the upside down mixin being incompatible with other mods and also splits upside down and spinny into two options in the contributors.json

exclude_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/mixins/hooks/RendererLivingEntityHook.kt
+++ b/src/main/java/at/hannibal2/skyhanni/mixins/hooks/RendererLivingEntityHook.kt
@@ -41,7 +41,6 @@ object RendererLivingEntityHook {
      */
     @JvmStatic
     fun rotatePlayer(player: EntityPlayer) {
-        if (!config.rotateContributors) return
         if (!LorenzUtils.inSkyBlock) return
         if (!config.rotateContributors && !LorenzUtils.isAprilFoolsDay) return
         val name = player.name ?: return
--- a/src/main/java/at/hannibal2/skyhanni/mixins/hooks/RendererLivingEntityHook.kt
+++ b/src/main/java/at/hannibal2/skyhanni/mixins/hooks/RendererLivingEntityHook.kt
@@ -8,9 +8,10 @@ import net.minecraft.client.renderer.GlStateManager
 import net.minecraft.entity.EntityLivingBase
 import net.minecraft.entity.player.EntityPlayer
 
-class RendererLivingEntityHook {
+object RendererLivingEntityHook {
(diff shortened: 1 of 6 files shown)


src/main/java/at/hannibal2/skyhanni/mixins/transformers/renderer/MixinRendererLivingEntityHookSBA.java
```diff
@@ -15,19 +12,11 @@
...
     @Inject(method = "equals", at = @At("HEAD"), cancellable = true, remap = false)
     private static void onEquals(String displayName, Object otherString, CallbackInfoReturnable<Boolean> cir) {
-        if (skyHanni$hook.isCoolPerson(displayName)) {
+        if (RendererLivingEntityHook.shouldBeUpsideDown(displayName)) {
             cir.setReturnValue(true);
         }
     }
 
-    @Inject(method = "isWearing", at = @At("HEAD"), cancellable = true, remap = false)
-    private static void onIsWearing(EntityPlayer player, EnumPlayerModelParts p_175148_1_, CallbackInfoReturnable<Boolean> cir) {
```
ThatGravyBoat, line 29: Isnt this still needed? as SBA used a transformer to replace the vanilla isWearing with this method

lineargraph: Oh yeah, looks like SBA is just straight up replacing the call instead of just putting in an `||`.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RendererLivingEntityHook.kt
```diff
@@ -27,18 +28,24 @@ class RendererLivingEntityHook {
...
+        return ContributorManager.shouldBeUpsideDown(name)
     }
 
     /**
-     * Player is already on the cool person list so rotate them if the option is on.
+     * Check if the player should spin and rotate them if the option is on.
      */
+    @JvmStatic
     fun rotatePlayer(player: EntityPlayer) {
         if (!config.rotateContributors) return
```
ThatGravyBoat: This can be removed as it is checked below now

ThatGravyBoat (approved):
Worked in-game, besides the extra checked that was kept in `rotatePlayer` everything looks good.

CalMWolfs (approved):

PR #1676, opened 2024-05-03
Feature: Middle click in Trades

Reebix:
## What
When the config it toggled on (InventoryConfig/Middle Click in Trade), it changes the normal clicks into middle clicks for less cooldown and effectively higher selling speed

</details>

## Changelog New Features
+ Middle click in Trades. - Rebix


Reebix:
TODO: exclude the tradable items from middle click
(kinda forgot that)

PR #1669, opened 2024-05-03
Improvement: line to hoppity guess

hannibal002:
## What
line to hoppity guess

<details>
<summary>Images</summary>

image

</details>

## Changelog Improvements
+ Show a line to Hoppity guess point. - hannibal2

PR #1681, opened 2024-05-04
Improvement: All Hoppity Waypoints with Egglocator

Obsidianninja11:
## What
Added an option to show all waypoints even with Egglocator in inventory.

## Changelog Improvements
+ Added an option to show all Hoppity waypoints with Egglocator. - Obsidian

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/event/HoppityEggsConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/event/HoppityEggsConfig.java
@@ -17,10 +17,23 @@ public class HoppityEggsConfig {
     public boolean waypoints = true;
 
     @Expose
+    @ConfigOption(name = "Waypoint Line", desc = "Draw a line from your eyes to the Hoppity's Hunt Waypoint.")
+    @ConfigEditorBoolean
+    public boolean drawLine = true;
+
+    @Expose
     @ConfigOption(name = "Show All Waypoints", desc = "Show all possible egg waypoints for the current lobby. §e" +
         "Only works when you don't have an Egglocator in your inventory.")
     @ConfigEditorBoolean
     public boolean showAllWaypoints = false;
+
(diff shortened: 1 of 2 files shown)


hannibal002:
is the improvement of line needed?
check out https://github.com/hannibal002/SkyHanni/pull/1669

src/main/java/at/hannibal2/skyhanni/config/features/event/HoppityEggsConfig.java
```diff
@@ -16,12 +16,25 @@ public class HoppityEggsConfig {
     @FeatureToggle
     public boolean waypoints = true;
 
+    @Expose
+    @ConfigOption(name = "Waypoint Line", desc = "Draw a line from your eyes to the Hoppity's Hunt Waypoint.")
+    @ConfigEditorBoolean
+    public boolean drawLine = true;
```
CalMWolfs: i feel this should be default off

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/config/features/event/HoppityEggsConfig.java
CalMWolfs: nvm will use whatever #1669 is

Obsidianninja11: Well there was a line already, so i had it default on. But i'm removing this from my pr since hanni already has a pr for the same thing.
although #1669 doesn't have a toggle at all. maybe it should?

hannibal002:
why? i dont think this is needed

CalMWolfs:
I also dont see the point in this

Obsidianninja11:
Ig not really that needed
Especially now that it uses locator in hotbar instead of inv

PR #1691, opened 2024-05-04
Fix: Adding Gemstones to /ff

hannibal002:
## What
Trying to add gemstone support to /ff

Does not work :(

## Changelog Fixes
+ Fix Gemstones in /ff. - hannibal2


Thunderblade73:
Outdated, was added by another Pr already.

PR #1696, opened 2024-05-04
Improvement: Cleanup HoppityEggPlayerOpacityChanger

ThatGravyBoat:
## What
This cleans up the `HoppityEggPlayerOpacityChanger` class now named `HoppityEggDisplayManager` to not do extra operations needlessly and to not store the armor and to fix issues where if a player had other layers it would not apply correctly.

The name change is so that in the future it can be used for similar display like code for hoppity eggs. ie. showing hitbox of egg

The check for the opacity no longer checks through every possible egg even when one is found and will escape early.
It will also now check if the opacity is greater than or equal to 100 instead of only 100 and if the opacity is less than or equal to 0 for hiding.

+ Added EntityRenderLayersEvent for enabling and disabling an entities layers. - ThatGravyBoat
    * Useful for hiding extra layers of a player such as armor, capes, Items and so on.


hannibal002:
testing in 2 days

hannibal002 (changes requested):
this pr breaks opacity.
see this comparison without changed config

<details>
<summary>Images</summary>

without this pr
image
with this pr
image


</details>

ThatGravyBoat:
My bad, I put that check after the event as to not waste checks and I misplaced it. Works now

hannibal002 (approved):
works again
image

PR #1129, opened 2024-03-09
Remind Command

Zickles:
# WILL MAKE DESCRIPTION WHEN PR IS READY
## Dependencies
- pr_number_or_link_here

## What
Describe what this pull request does, including technical details, screenshots, links to discord, etc.

## Changelog New Features
+ Cool new feature. - your_name_here
    * extra info

# WILL MAKE DESCRIPTION WHEN PR IS READY

exclude_from_changelog

PR #1708, opened 2024-05-05
Feature: Remind Command

ThatGravyBoat:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/1718

## What
A continuation of #1129, finishing it. This adds the remind command along with some component utilities.

<details>
<summary>Images</summary>

</details>

## Changelog New Features
+ Added Remind command. - ThatGravyBoat & Zickles
    * Use `/shremind <time> <note>` to set reminders for your future self.


ThatGravyBoat:
Im going to pull out the text class into its own PR as I think its a little out of scope to keep it in here and I want to depend on it in other PRs, so while I do that ill make this as a draft

ThatGravyBoat:
Now updated to depend on the text api PR

CalMWolfs (approved):
lgtm

src/main/java/at/hannibal2/skyhanni/features/misc/reminders/ReminderManager.kt
```diff
@@ -0,0 +1,256 @@
...
+    private const val REMINDERS_LIST_ID = -546745
+    private const val REMINDERS_ACTION_ID = -546746
+    private const val REMINDERS_MESSAGE_ID = -546747
+
+    private val storage get() = SkyHanniMod.feature.storage.reminders
+    private val config get() = SkyHanniMod.feature.misc.reminders
+
+    private var listPage = 1
+
+    private fun newId(): String {
```
hannibal002: Can't we simply use UUID.Random().toString() here? The Base62 and storage approach feels like overkill

src/main/java/at/hannibal2/skyhanni/features/misc/reminders/ReminderManager.kt
```diff
@@ -0,0 +1,256 @@
...
+        "§e/shremind help - §bShows this help message".asComponent().send()
+        createDivider().send()
+    }
+
+    @SubscribeEvent
+    fun onSecondPassed(event: SecondPassedEvent) {
+        val remindersToSend = mutableListOf<IChatComponent>()
+
+        for ((id, reminder) in getSortedReminders()) {
+            if (!reminder.shouldRemind(config.interval.minutes)) break
```
hannibal002: break feels wrong here. how about continue?

src/main/java/at/hannibal2/skyhanni/features/misc/reminders/ReminderManager.kt
```diff
@@ -0,0 +1,256 @@
...
+                        command = "/shremind remove $id"
+                    }.wrap("§8[", "§8]"),
+                    " ",
+                    "§e§l⟳".asComponent {
+                        hover = "§7Click to move".asComponent()
+                        suggest = "/shremind move $id 1m"
+                    }.wrap("§8[", "§8]"),
+                )
+            } else {
+                storage.remove(id)
```
hannibal002, line 220: Are you sure this will not cause concurrent modification exceptions? editing a map while iterating over it. I have not encountered some while testing though

hannibal002 (changes requested):
Works fine!
Maybe we should show the help list when just doing "/shremind" instead of the one liner:
image

src/main/java/at/hannibal2/skyhanni/features/misc/reminders/ReminderManager.kt
hannibal002: Never mind, just checked, `sortedWith()` creates a copy

src/main/java/at/hannibal2/skyhanni/features/misc/reminders/ReminderManager.kt
ThatGravyBoat: its sorted by time so if it cant remind that means all other reminders after this point will also not be able to be reminded of

src/main/java/at/hannibal2/skyhanni/features/misc/reminders/ReminderManager.kt
ThatGravyBoat: yeah I wrote it with users also being able to manually type it out in mind but since the change to a nicer chat interface it can just be changed to a random of anything really

src/main/java/at/hannibal2/skyhanni/features/misc/reminders/ReminderManager.kt
ThatGravyBoat: wait actually nvm I forgot about repeat reminders

src/main/java/at/hannibal2/skyhanni/features/misc/reminders/ReminderManager.kt
```diff
@@ -42,7 +42,7 @@ object ReminderManager {
     private fun newId(): String {
         var id: String
         do {
-            id = Base62.encode((Math.random() * Int.MAX_VALUE).toInt())
+            id = UUID.randomUUID().toString()
         } while (storage.containsKey(id))
```
hannibal002: Please remove this storage check here.
Even with one million reminders in the list, the probability of a collision when creating a new random UUID is around `9.4 x 10^-26`.
Instead, we could remove the whole function and replace it with https://github.com/hannibal002/SkyHanni/blob/40faf50e3f109d0028e67b6dd916ff79f2cb5604/src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt#L386

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/reminders/ReminderManager.kt
hannibal002: I've maybe worded that a bit wrong. Fixed it here https://github.com/hannibal002/SkyHanni/pull/1708/commits/3101b6883fe95b4c6dbf71776083e63794c7b0b2

PR #1717, opened 2024-05-06
Improvement: Enchant Parsing SBA Convert Roman Numerals and NEU Inventories compat

VixidDev:
## What
Makes enchant parsing work when SBA's Convert Roman Numerals feature is enabled and in NEU's inventories, such as /pv and storage overlay.

This will most likely need testing before approval since there is probably a certain case where my regex or code doesn't like a certain enchant format on some item / enchant / lore description. I've tested pretty much every item I have and other peoples items through /pv and they all work, but more testing from others would be appreciated.

## Changelog Improvements
+ Improved Enchant Parsing compatibility with other mods. - Vixid
    * Now works with SBA's Convert Roman Numerals feature.
    * Now works with NEU's inventories, such as `/pv` and storage overlay.


hannibal002:
ItemHoverEvent already does the same thing. just make its list modifyable, ig

VixidDev:
> ItemHoverEvent already does the same thing. just make its list modifyable, ig

Oh I see, idk how I missed that event. Yeah I'll probably just use that and make it modifiable.

Mikecraft1224:
Doesn't work with NEU enchant colors
The regex in line 37 doesn't understand the enchants when there are other color codes applied

hannibal002 (changes requested):
Mikecraft1224's comment

VixidDev:
Moving the ItemHoverEvent mixin inject ensures it is never affected by NEU enchant colors so it doesn't break and die. If anyone tests this make sure your repo is using the updated regex otherwise the tooltip will be missing random enchants. Works with roman numerals to numbers from both sba and sh, but the roman to numbers from sh doesn't work in NEU pv, but I think thats for @Mikecraft1224 to amend.

CalMWolfs (approved):

PR #1718, opened 2024-05-06
Backend: Text API

ThatGravyBoat:
## What
This PR adds the `Text` object which is a file that contains numerous utilities for creating chat components. In addition to this it makes it easier to complete some todos that were related to components. This also cleans up the chat action manager to be a map to action to make sure lookups at O(1) instead of looping through all, it also moves it to the utilities package with the other components utils into a new package `chat`.

+ Added Text object that provides various helpers, DSL, and utilities for interaction with chat components. - ThatGravyBoat


hannibal002 (changes requested):
Please dont delete/break existing functions in ChatUtils. rather mark them as deprecated and delete them in 1-2 weeks in another pr.

ThatGravyBoat:
> Please dont delete/break existing functions in ChatUtils. rather mark them as deprecated and delete them in 1-2 weeks in another pr.

The methods that were removed were never used/already deprecated or were only ever used once, the only function that wasnt deprecated or was never used was createHoverable which was only ever used in commands and I updated that usage. I dont see the point in keeping dead code for it to be deprecated for only a week 

Deprecating code and then deleting it a week later makes no difference than just doing it now as it will just delay the possible PRs that could be using them to break in a week instead of now.

hannibal002 (commented):
My thoughts were just so that open PRs with this code can one click fix the deprecated code instead of looking up and writing the necessary lines manually. 
but those functions aren't often used anyway, it seems, so it's fine

CalMWolfs (approved):

hannibal002 (approved):

PR #1719, opened 2024-05-06
Improvement: Chocolate Factory

hannibal002:
## What
split upgrades into highlight best upgrade and highlight affordable upgrades

## Changelog Improvements
+ Chocolate Factory changes. - hannibal2
    * Split upgrades into highlight best upgrade and highlight affordable upgrades.
    * Highlight best upgrade when affordable.
    * Highlight best slot green when affordable, even when "highlight affordable upgrades" is disabled.


CalMWolfs (changes requested):
1. i dont really see the point in this, the current system is working fine for everyone in my opinion
2. i dont think it works as nothng is ever highlighted gold

PR #1724, opened 2024-05-06
Improvement: Multiple chat clicks

hannibal002:
## What
made normal clickable chat message multiple times clickable, and disable this for some cases

## Changelog Improvements
+ Allow multiple clicks on the same SkyHanni chat message. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/commands/Commands.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/commands/Commands.kt
@@ -629,7 +629,8 @@ object Commands {
                 "Are you sure you want to switch to beta? These versions may be less stable.",
                 onClick = {
                     UpdateManager.checkUpdate(true, updateStream)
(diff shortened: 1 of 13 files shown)


ThatGravyBoat:
I improved how these click events work in #1718 in terms of performance and usability, should I include this change in there by changing the default and changing the few that need oneTimeClick and then that can supersede this instead of getting merge conflicts for it to be instantly changed again?

hannibal002:
yes, good catch. Should i keep the "set to only one time" changes in this pr or do you wanna merge it entirely into your pr?

ThatGravyBoat:
> yes, good catch. Should i keep the "set to only one time" changes in this pr or do you wanna merge it entirely into your pr?

Ive merged it into my PR

hannibal002:
yield for https://github.com/hannibal002/SkyHanni/pull/1718

PR #1541, opened 2024-04-24
Backend: Unit Test Repo Pattern List

Thunderblade73:
Superseded by #1733 

Test are in [src/test/java/at/hannibal2/skyhanni/test/RepoPatternTest.kt](https://github.com/hannibal002/SkyHanni/pull/1541/files#diff-df42cd3aaf4abeec910c1a81a174ce91737c72c7c58f75f5196924ca5f81b003R53-R151)

PR #901, opened 2024-01-10
Wrong/broken Changelog: Add pattern list support

lineargraph:
Superseded by https://github.com/hannibal002/SkyHanni/pull/1733

PR #1733, opened 2024-05-07
Backend: RepoPattern Exclusive Group & RepoPattern List

Thunderblade73:
## Supersedes
Merge this one instead of both of the ones, since most of the stuff gets changed here
- #901
- #1541

## What
A way to reserve a key namespace for patterns.

+ Added RepoPattern.list. - !nea
    * A way to store x amount of patterns.
+ Added unit test for repo pattern lists. - Thunderblade73
+ Added RepoPattern.exclusiveGroup. - Thunderblade73
    * Reserves a key namespace, that is only accessible from this group.


CalMWolfs (approved):
new tests look good!

PR #1768, opened 2024-05-11
Improvement: sh gui 20s

hannibal002:
## What

The GUI position editor can now edit all GUIs that were visible in the last 20 seconds, also those inside inventories.

## Changelog Improvements
+ The GUI position editor can now edit all GUIs that were visible within the last 20 seconds, including those inside inventories. - hannibal2


CalMWolfs (approved):
works

PR #1771, opened 2024-05-11
Backend: performance improvement trevor trapper

hannibal002:
## What
performance improvement trevor trapper

exclude_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/trevor/TrevorFeatures.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/trevor/TrevorFeatures.kt
@@ -4,12 +4,12 @@ import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.config.ConfigUpdaterMigrator
 import at.hannibal2.skyhanni.config.features.misc.TrevorTheTrapperConfig.TrackerEntry
 import at.hannibal2.skyhanni.data.IslandType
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/features/misc/trevor/TrevorFeatures.kt
```diff
@@ -323,6 +324,11 @@ object TrevorFeatures {
         resetTrapper()
     }
 
+    @SubscribeEvent
+    fun onTick(event: LorenzTickEvent) {
+        inTrapperDen = LorenzUtils.skyBlockArea == "Trapper's Den"
```
j10a1n15: Probably update it on scoreboard change + use repopattern

CalMWolfs (changes requested):
j10a1n15's comment

src/main/java/at/hannibal2/skyhanni/features/misc/trevor/TrevorFeatures.kt
hannibal002: repo pattern is good.
scorboard change event is tricky: we either need to delay the LorenzUtils.skyBlockArea check by 1-2 ticks, or we need to check the scoreboard data directly. both is not ideal.

CalMWolfs: we can do the scoreboarch change later

CalMWolfs (approved):

PR #1772, opened 2024-05-12
Backend: Make MayorJson be kotlin instead of java

CalMWolfs:
## What
Make MayorJson be kotlin instead of java.

+ Make MayorJson Kotlin instead of Java. - CalMWolfs


j10a1n15 (approved):
works ingame

hannibal002 (approved):

PR #1774, opened 2024-05-12
Fix: Chocolate Factory Time Tower math

hannibal002:
## What
Fixed Chocolate Factory time remaining calculations while Time Tower is active.

## Changelog Fixes
+ Fixed Chocolate Factory time remaining calculations while the Time Tower is active. - hannibal2


CalMWolfs (approved):
worked for me

PR #1798, opened 2024-05-15
Fix: White Enchantments

CalMWolfs:
## What
Fix max enchantments showing as white when the user has sh (and sba) chroma off.

## Changelog Fixes
+ Fixed enchantment colours showing as white when SkyHanni chroma is not enabled. - CalMWolfs


lunaynx:
Hmm, won't this make both "great" and "perfect" enchants gold by default with chroma off?

CalMWolfs:
> Hmm, won't this make both "great" and "perfect" enchants gold by default with chroma off?

Yes, but both gold is better than one as just white and having people complaining about skyhanni breaking their enchantment colours, the other solutions are to not have chroma by default or to have skyhanni chroma on by default or having enchant parsing off by default. But out of these four, having them both gold seemed like the best option

lunaynx:
Hmm, why not make the perfect fallback another color like purple, or gold and bold? (Though bold is already used for ultimate enchants.)

Obsidianninja11:
> Hmm, why not make the perfect fallback another color like purple, or gold and bold? (Though bold is already used for ultimate enchants.)

I think gold and bold is good.

CalMWolfs:
made it gold and bold

PR #1812, opened 2024-05-16
Improvement: Added a way to detect sheep and rabbits in the oasis.

AviKohn7:
## What
<details>
<summary>Added a way to detect sheep and rabbits in the oasis by checking if the first word of the mob name matches certain prefixes.</summary>
This will only work in English.
</details>

## Changelog Improvements
+ Trevor Mob Detection now works with Oasis mobs. - Awes


src/main/java/at/hannibal2/skyhanni/features/misc/trevor/TrevorSolver.kt
```diff
@@ -78,7 +83,12 @@ object TrevorSolver {
             foundID = -1
         }
     }
-
+    private fun isTrevorMob(mob: Mob) : Boolean{
+        val name = mob.name
+        val firstWord = name.substringBefore(" ")
+        if(possibleNames.any{prefix: String -> prefix == firstWord}) return true
+        return false
```
Thunderblade73:
```suggestion
        return possibleNames.any{prefix: String -> prefix == firstWord}
```

Thunderblade73: Better
```suggestion
        return possibleNames.contains(firstWord)
```

AviKohn7:
By the way, finnegan may be soon, so this would likely be a useful improvement to add before that.

hannibal002 (approved):

PR #1823, opened 2024-05-17
Fix: FF ioobe

Thunderblade73:
## What
FIx the IndexOutOfBoundsException insde ff do to wierd regex behaviour.

## Changelog Fixes
+ Fixed an error in /ff when you don't have a gemstone. - Thunderblade73

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/FarmingFortuneDisplay.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/FarmingFortuneDisplay.kt
@@ -347,7 +347,7 @@ object FarmingFortuneDisplay {
 
         for (line in tool?.getLore()!!) {
             tooltipFortunePattern.matchMatcher(line) {
-                displayedFortune = group("display")!!.toDouble()
+                displayedFortune = group("display")?.toDouble() ?: 0.0
                 reforgeFortune = groupOrNull("reforge")?.toDouble() ?: 0.0
                 gemstoneFortune = groupOrNull("gemstone")?.toDouble() ?: 0.0
             } ?: continue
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/FarmingFortuneDisplay.kt
```diff
@@ -345,9 +347,9 @@ object FarmingFortuneDisplay {
 
         for (line in tool?.getLore()!!) {
             tooltipFortunePattern.matchMatcher(line) {
-                displayedFortune = group(1)!!.toDouble()
-                reforgeFortune = group(2)?.toDouble() ?: 0.0
-                gemstoneFortune = group(3)?.toDouble() ?: 0.0
+                displayedFortune = group("display")!!.toDouble()
```
lunaynx: While we're at it, let's change this too:
```suggestion
                displayedFortune = group("display")?.toDouble() ?: 0.0
```

martimavocado: does this work without exploding? i thought you had to use `toDoubleOrNull()`

src/main/java/at/hannibal2/skyhanni/features/garden/FarmingFortuneDisplay.kt
```diff
@@ -53,7 +54,7 @@ object FarmingFortuneDisplay {
     )
     private val tooltipFortunePattern by patternGroup.pattern(
         "tooltip",
```
CalMWolfs:
```suggestion
        "tooltip.new",
```

so that people on old versions do not get errors with the new pattern that was added before

CalMWolfs (changes requested):

hannibal002 (approved):

PR #1824, opened 2024-05-17
Backend: Keep visual formatting appearance with CharSequence.removeColor(keepFormatting=true)

appable0:
## What
Change to `CharSequence#removeColor(keepFormatting=true)` so that the formatting is reset when there are color codes that reset the applied formatting styles. Added additional test cases to cover these situations. Does not change behavior with `keepFormatting=false`.

### Previous behavior
Removed all color codes and kept all non-color formatting codes in the message. 

`Original: §5§o§a§lPROMOTE §8➜ §7[158§7] §5Manager`
`Modified: §o§lPROMOTE ➜ [158] Manager`

Screenshot 2024-05-17 at 12 22 25 PM

### New behavior
Keeps all non-color formatting codes. If a non-color formatting code is encountered, replaces the next occurrence of a color code (if one exists) with §r to reset the active formatting. Otherwise, removes all color codes.

`Original: §5§o§a§lPROMOTE §8➜ §7[158§7] §5Manager`
`Modified: §o§r§lPROMOTE §r➜ [158] Manager`

Screenshot 2024-05-17 at 12 24 30 PM

+ Modified removeColor for visual formatting consistency. - appable
    * Keeps formatting consistency in the presence of color codes acting as implicit reset codes.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
@@ -42,6 +42,7 @@ object StringUtils {
     }
 
     private val formattingChars = "kmolnrKMOLNR".toSet()
+    private val colorChars = "abcdefABCDEF0123456789".toSet()
 
     /**
      * Removes color and optionally formatting codes from the given string, leaving plain text.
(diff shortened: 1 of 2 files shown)


CalMWolfs (approved):
lgtm

PR #1826, opened 2024-05-17
Improvement: Show all the best rabbit upgrades that should be bought instead of just one

flxwly:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/194

## What
Added numbers that show how many times you need to upgrade each rabbit. (Needs testing!)
[Suggestion on SkyHanni-Discord](https://discord.com/channels/997079228510117908/1234597863066177656)

Changed the yellow insignia to a border so it doesn't overlap with the number for the best loadout.

Cost calculation is based on [this](https://discord.com/channels/997079228510117908/1234043917792579604/1234715506305400863)

<details>
<summary>Images</summary>

many_upgrades

two_left

one_left

good_setup

</details>

## Changelog Improvements
+ Show all the best possible upgrades instead of only one. - flxwly
    * Currently only works for rabbits.
    * The number of how many times a rabbit should be upgraded is displayed above the current level number.


flxwly:
Found a bug today, but since my upgrade costs are so high I can't reproduce it.
Basically the number of upgrades on certain rabbits changed by one after upgrading any rabbit.

flxwly:
The bug seems to persist and the origin might be in a state missmatch between some ChocolateFactoryAPI variables and the calculated prices/effective costs. If there are many upgrades possible recalculating them takes time. For now it should only recalculate everything if the user upgrades any upgrade above the optimal level, if the user gained enough chocolate to upgrade one more upgrade, if the user spends chocolate for other things than upgrading or if it is the first time opening chocolate factory. I'll add a toggle for the feature and say it's done.

Debugging is really painfull because I can't collect enough data on only one minecraft account. I commented everything if anyone wants to take a look at it.

CalMWolfs:
I want to wait for the changes on alpha to be on the main server before this is looked at in case that changes anything, or there is something new to account for (the divine rabbits)

flxwly:
Still can't figure out why it sometimes shows one less upgrade than possible. Also sometimes the number goes up by one for a short period of time. Rabbit coach should work in theory but I can't test it.

src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryDataLoader.kt
```diff
@@ -408,6 +439,217 @@
...
+        val currentUpgrades = ArrayList(
+            list.filter { it.extraPerSecond != null && it.extraPerSecond > 0 }
+                .filter { it.slotIndex != ChocolateFactoryAPI.timeTowerIndex }
+        )
+
+        // check if any of the current upgrades is of higher level then previous calculated best upgrades
+        val isCurrentHigher = currentUpgrades.any { current ->
+            val upgrade = ChocolateFactoryAPI.allBestPossibleUpgrades[current.slotIndex]?.lastOrNull()
+            val lvlOnLastRun = ChocolateFactoryAPI.lastUpgradesWhenChecking[current.slotIndex]
+            return@any /*Last level was not set (e.g., first run after clear)*/lvlOnLastRun == null
```
hannibal002: pleasedont use in between comments. rather extract variables with descriptive names

src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryDataLoader.kt
```diff
@@ -408,6 +439,217 @@ object ChocolateFactoryDataLoader {
...
+
+            ChocolateFactoryAPI.lastBestNotAffordableUpgrade = bestUpgrade
+            return remainingChocolate
+        }
+
+        // ---------------- simulate making the upgrade ----------------
+
+        // Keep track of total base increase and multiplier increase after upgrades.
+        val nextBaseIncrease = baseIncreaseAfterUpgrades + (ChocolateFactoryAPI.rabbitSlots[bestUpgrade.slotIndex] ?: 0)
+        val nextMultiplierIncrease = multiplierIncreaseAfterUpgrades + 0.01 * (if (bestUpgrade.slotIndex == ChocolateFactoryAPI.coachRabbitIndex) 1 else 0)
```
hannibal002: too long line

src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryDataLoader.kt
```diff
@@ -408,6 +439,217 @@ object ChocolateFactoryDataLoader {
...
+     * @param baseIncreaseAfterUpgrades total base increase after upgrades
+     * @param multiplierIncreaseAfterUpgrades total multiplier increase after upgrades
+     */
+    private fun findAllBestUpgradesImpl(
+        list: ArrayList<ChocolateFactoryUpgrade>,
+        allUpgrades: MutableMap<Int, MutableList<ChocolateFactoryUpgrade>>,
+        remainingChocolate: Long = profileStorage?.currentChocolate ?: 0,
+        baseIncreaseAfterUpgrades: Int = 0,
+        multiplierIncreaseAfterUpgrades: Double = 0.0
+    ): Long {
```
hannibal002, line 523: function too long please extract or compact

src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryDataLoader.kt
```diff
@@ -407,6 +438,219 @@ object ChocolateFactoryDataLoader {
...
+        return price
+    }
+
+    /** Constructs the rabbit with level = 0 after [upgrade] if upgrade is a rabbit.
+     *  Does not recalculate [ChocolateFactoryUpgrade.effectiveCost] and [ChocolateFactoryUpgrade.extraPerSecond] but
+     *  instead just copies it from [upgrade]
+     *
+     * @param upgrade the current upgrade
+     */
+    private fun getNextRabbit(upgrade: ChocolateFactoryUpgrade): ChocolateFactoryUpgrade? {
```
hannibal002: functions like this do not belong into a "data loader" class

src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryDataLoader.kt
```diff
@@ -186,6 +180,46 @@ object ChocolateFactoryDataLoader {
         processInventory(list, inventory)
 
         findBestUpgrades(list)
+        if (config.showAllBestUpgrades) findAllBestUpgrades(list)
+
+        // TODO: Remove (ill leave it in for now since the bug still persists for now - flxwly)
```
hannibal002: move those commented out test cases into a new functionn

hannibal002 (changes requested):
just looked over the code, not tested inn game yet

src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryDataLoader.kt
flxwly: Should it be moved to the API class then? Or just the factory upgrade class? I originally put it there because only `ChocolateFactoryDataLoader.findAllBestUpgradesImpl` depends on it but I see where that isn't enough justification to put it there.

hannibal002: `ChocolateFactoryAPI` already gets quite full. Ideally, we would entirely restructure each function over all classes in the chocolate package. But for the scope of this pr, I, think it's fine when you move the new functions either into existing utils classes or create a new RabbitUtils class. Up to you.

src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/RabbitUtils.kt
```diff
@@ -0,0 +1,70 @@
...
+            // Use upgrade cost per level if it exists, otherwise use the formula.
+            if ((ChocolateFactoryAPI.upgradeCostPerLevel[slotIndex]?.size ?: 0) > level) {
+                val nextRaw = ChocolateFactoryAPI.upgradeCostPerLevel[slotIndex]?.get(level) ?: 0
+
+                price = floor(nextRaw * prestigeMultiplier * zetaMultiplier).toLong()
+            } else {
+                val base = ChocolateFactoryAPI.upgradeCostFormulaConstants[slotIndex]?.get("base") ?: 0.0
+                val multiplier = ChocolateFactoryAPI.upgradeCostFormulaConstants[slotIndex]?.get("exp") ?: 0.0
+
+                price = floor(round(base * multiplier.pow((level + 1)) * prestigeMultiplier) * zetaMultiplier).toLong()
```
Bloxigus: I know I'm not a contributor here, but this should be 
``` 
price = floor(floor(round( base * multiplier.pow(level + 1)) * prestigeMultiplier) * zetaMultiplier).toLong()
```
to prevent quite a few off-by-one errors in prestiges that dont already have round prestigeMultipliers, such as the cost of level 12 Rabbit Sis at Chocolate Factory 4 with zeta rabbit, which is calculated as 12 069 by this formula but is actually 12 068.

flxwly: Thank you for taking your time reviewing the formulae. However the change you requested should not make a difference since x.toLong() will yield the same number as floor(x) in this case. But you made me recognize that there's an opening bracket to much in `multiplier.pow((level + 1))` that should instead be at the beginning.

Bloxigus: That's on me for incorrectly assuming the role of `.toLong()` in that line, otherwise the math checks out with your latest changes

CalMWolfs:
@flxwly are you planning on updating this pr or can I close it

PR #1828, opened 2024-05-17
Fix: Bestiary Display when in search results

appable0:
## What
Allows the bestiary display to display in search mode (`/be [query]`) when Overall Progress is enabled. Previously, it was always disabled and stated that Overall Progress was not enabled whenever in search mode.

Unsolvable issue: when in search mode and the only results are already-maxed bestiaries, it is impossible to determine whether Overall Progress is enabled. It simply assumes that Overall Progress is enabled in this case.

<details>
<summary>Images</summary>
Screenshot 2024-05-17 at 4 15 54 PM
</details>

## Changelog Fixes
+ Fixed bestiary display always being disabled in search mode. - appable

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/combat/BestiaryData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/combat/BestiaryData.kt
@@ -39,11 +39,19 @@ object BestiaryData {
 
     private val patternGroup = RepoPattern.group("combat.bestiary.data")
 
+    /**
+     * REGEX-TEST: §7Progress to Tier 14: §b26%
+     * REGEX-TEST: §7Progress to Tier XV: §b57.1%
+     */
     private val tierProgressPattern by patternGroup.pattern(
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/combat/BestiaryData.kt
```diff
@@ -36,6 +38,17 @@ object BestiaryData {
     private val config get() = SkyHanniMod.feature.combat.bestiary
 
     private val patternGroup = RepoPattern.group("combat.bestiary.data")
+
+    private val tierProgressPattern by patternGroup.pattern(
```
hannibal002, line 46: please include regex-test for those two

hannibal002 (changes requested):
works in game

jani270:
I LOVE YOU

jani270 (approved):
🙏 May your pillow be cold on both sides.

PR #1788, opened 2024-05-13
Backend: Use SkillAPI in SkillExpGainEvent

superhize:
## What
Update the SkillExpGainEvent to use SkillAPI data.

## Changelog Technical Details
+ SkillExpGainEvent now uses SkillAPI data. - HiZe

PR #1830, opened 2024-05-18
Feature: added leaderboard display

catgirlseraid:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/195
- #1788

## What
Added an overlay to show the xp needed until you pass the person in front of you, same with farming collections. this uses the elite bot api.

image

## Current known bugs 

skill display only takes the current xp until next level instead of total level upon skill gain but is correct initally

## Changelog New Features
+ Collection leaderboard. - seraid
+ Skill XP leaderboard. - seraid
+ Pest kills leaderboard. - seraid


ptlthg:
Thanks for working on this! I have some notes about the current implementation:

- There can be some code deduplication done here. My leaderboard rank endpoints work the same no matter which leaderboard is called, so there can be a generic function for all three uses of it (weight, xp, collection).
- I mentioned it on Discord, but the graph endpoints aren't guaranteed to send the newest data point, so it doesn't really make sense to use them here for getting data on world load/profile swap.
   - There's now an `amount` field on the leaderboard rank endpoint, so the player's current weight/collection/xp will be returned as well. This can be used instead.
- Personally, I don't see the value in having the display show "You have passed #5000" all the time, not sure if that's supposed to be final. I have the same thought with the refresh countdown, it doesn't really feel needed. I think the current state of the Farming Weight display is a good starting point in terms of looks, this is subjective though

catgirlseraid:
> Thanks for working on this! I have some notes about the current implementation:
> 
> * There can be some code deduplication done here. My leaderboard rank endpoints work the same no matter which leaderboard is called, so there can be a generic function for all three uses of it (weight, xp, collection).
> * I mentioned it on Discord, but the graph endpoints aren't guaranteed to send the newest data point, so it doesn't really make sense to use them here for getting data on world load/profile swap.
>   
>   * There's now an `amount` field on the leaderboard rank endpoint, so the player's current weight/collection/xp will be returned as well. This can be used instead.
> * Personally, I don't see the value in having the display show "You have passed #5000" all the time, not sure if that's supposed to be final. I have the same thought with the refresh countdown, it doesn't really feel needed. I think the current state of the Farming Weight display is a good starting point in terms of looks, this is subjective though

there is a toggle for the refresh timer, currently it's on by default but that can be changed.
switched it away from the graph api endpoint and now it only uses the leaderboard endpoint, thanks for adding that
the "you have passed #5000" was a bug, it was meant to show the rank in front of you and is fixed now

thanks <3

catgirlseraid:
image
is meant to look like this, will manually refresh if the player clicks on it

src/main/java/at/hannibal2/skyhanni/api/EliteBotAPI.kt
```diff
@@ -0,0 +1,47 @@
...
+    fun onRepoReload(event: RepositoryReloadEvent) {
+        val data = event.getConstant<EliteAPISettingsJson>("EliteAPISettings")
+        checkDuration = data.refreshTimeMinutes.minutes
+        disableFetchingWhenPassed = data.disableFetchingWhenPassed
+        disableRefreshCommand = data.disableRefreshCommand
+    }
+
+    @SubscribeEvent
+    fun onChat(event: LorenzChatEvent) {
+        if (event.message.startsWith("§8Profile ID: ")) {
```
walkerselby, line 35: Flip if check, to make it more readable.

src/main/java/at/hannibal2/skyhanni/api/EliteBotAPI.kt
```diff
@@ -0,0 +1,47 @@
...
+            val id = event.message.removePrefix("§8Profile ID: ")
+            val newID = try {
+                UUID.fromString(id)
+            } catch (_: Exception) {
+                null
+            }
+            if (profileID != newID) {
```
walkerselby, line 42: inline this imo

src/main/java/at/hannibal2/skyhanni/features/garden/farming/EliteFarmingCollectionDisplay.kt
```diff
@@ -0,0 +1,334 @@
...
+        var difference: Long
+
+        do {
+            rank = collectionRanks[lastFetchedCrop]?.let { if (it == -1) 5001 else it } ?: return
+            nextRank = rank - 1
+            personToBeat = placements[nextRank]?.first ?: ""
+            amountToBeat = placements[nextRank]?.second ?: 0
+            difference = amountToBeat - collection
+
+
```
walkerselby, line 188: remove excess line

src/main/java/at/hannibal2/skyhanni/features/garden/farming/EliteFarmingCollectionDisplay.kt
```diff
@@ -0,0 +1,334 @@
...
+                rank--
+            }
+            collectionPlacements[crop] = placements
+            lastFetchedCrop = crop
+
+            if (data.amount != 0L) {
+                currentCollections[crop] = data.amount
+            }
+
+
```
walkerselby, line 285: remove extra line

PR #1832, opened 2024-05-18
Improvement + Fix: Quiver Warning

ItsEmpa:
## What
Fixes low quiver warning
 triggering when switching to no arrow type, and also makes the low arrows after run reminder specify what arrows to use again. also some cleanup and simplification around other quiver related things.

## Changelog Improvements
+ Re-added different arrow types in Quiver Reminder after a run. - Empa

## Changelog Fixes
+ Fixed Low Quiver Warning incorrectly appearing when switching to no arrows. - Empa


hannibal002 (approved):
works fine for me

PR #1844, opened 2024-05-20
Backend: MobDetection Memory Leak

Thunderblade73:
## What
Fixed the hash of mobs to be consitent. Which fixes the issue that HashSets broke do to not being consitent.

+ Fixed a small Memory Leak in MobData. - Thunderblade73


hannibal002:
Cant we simply use `baseEntity.uniqueID` as base for the hash?

Thunderblade73:
No, since the the problem is that the baseEntity can change

hannibal002 (approved):

PR #1846, opened 2024-05-20
Fix: Hover and clicks in guis

CalMWolfs:
## What
Fix gui click and hover things

## Changelog Fixes
+ Fixed some issues where skyhanni guis would have click and hover events when they shouldn't. - CalMWolfs


Thunderblade73:
The changes should be inside Renderable.shouldAllowLink

hannibal002 (approved):

PR #1856, opened 2024-05-21
Backend: Better Mob Highlight

Thunderblade73:
## What
Fixes the highlighting of mobs if there are stacked. E.g: Sea Emperor

+ Fixed stacked mob highlights. - Thunderblade73

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/mob/Mob.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/mob/Mob.kt
@@ -5,7 +5,7 @@ import at.hannibal2.skyhanni.data.mob.MobFilter.summonOwnerPattern
 import at.hannibal2.skyhanni.events.MobEvent
 import at.hannibal2.skyhanni.mixins.hooks.RenderLivingEntityHelper
 import at.hannibal2.skyhanni.utils.CollectionUtils.toSingletonListOrEmpty
-import at.hannibal2.skyhanni.utils.ColorUtils.withAlpha
+import at.hannibal2.skyhanni.utils.ColorUtils.addAlpha
 import at.hannibal2.skyhanni.utils.EntityUtils.canBeSeen
 import at.hannibal2.skyhanni.utils.EntityUtils.cleanName
 import at.hannibal2.skyhanni.utils.EntityUtils.isCorrupted
@@ -114,20 +114,32 @@ class Mob(
     fun isInvisible() = if (baseEntity !is EntityZombie) baseEntity.isInvisible else false
 
     private var highlightColor: Color? = null
+
+    /** If no alpha is set or alpha is set to 255 it will set the alpha to 127 */
     fun highlight(color: Color) {
-        highlightColor = color
+        highlightColor = if (color.alpha == 255) {
+            color.addAlpha(127)
+        } else {
+            color
+        }
         internalHighlight()
     }
 
     private fun internalHighlight() {
-        highlightColor?.let {
-            RenderLivingEntityHelper.setEntityColorWithNoHurtTime(baseEntity, it.withAlpha(127)) { true }
+        highlightColor?.let { color ->
+            RenderLivingEntityHelper.setEntityColorWithNoHurtTime(baseEntity, color.rgb) { true }
+            extraEntities.forEach {
+                RenderLivingEntityHelper.setEntityColorWithNoHurtTime(it, color.rgb) { true }
+            }
         }
     }
 
     private fun internalRemoveColor() {
         if (highlightColor == null) return
         RenderLivingEntityHelper.removeCustomRender(baseEntity)
+        extraEntities.forEach {
+            RenderLivingEntityHelper.removeCustomRender(it)
+        }
     }
 
     val boundingBox: AxisAlignedBB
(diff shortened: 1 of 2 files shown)


CalMWolfs (dismissed):

src/main/java/at/hannibal2/skyhanni/data/mob/Mob.kt
```diff
@@ -166,42 +178,44 @@ class Mob(
...
-        }
-
-        armorStand?.entityId ?: Int.MIN_VALUE -> armorStand = entity as EntityArmorStand
-        else -> {
-            extraEntitiesList.remove(entity)
-            extraEntitiesList.add(entity)
-            Unit // To make return type of this branch Unit
+    internal fun internalUpdateOfEntity(entity: EntityLivingBase) {
+        internalRemoveColor()
+        when (entity.entityId) {
```
hannibal002, line 202: i dont understand this line

src/main/java/at/hannibal2/skyhanni/data/mob/Mob.kt
```diff
@@ -114,20 +114,32 @@ class Mob(
...
     private var highlightColor: Color? = null
+
+    /** If no alpha is set or alpha is set to 255 it will set the alpha to 127 */
     fun highlight(color: Color) {
-        highlightColor = color
+        highlightColor = if (color.alpha == 255) {
+            color.addAlpha(127)
+        } else {
+            color
+        }
```
hannibal002:
```suggestion
        highlightColor = color.takeIf { it.alpha == 255 }?.addAlpha(127) ?: color
```

hannibal002 (changes requested):
when entity id is not 0? im really confused here

src/main/java/at/hannibal2/skyhanni/data/mob/Mob.kt
Thunderblade73: I don't understand what you don't understand. Please elaborate.
Since this is a simple when to switch between which entity is for the update, eg basEntity, Armorstand or else it must be in the extra list.
The only thing I did new here was that I remove the highlight for the old entity and activated it again for the new ones.
And for the question why we need it: Hypixel sends sometimes 2 spawn packets for one and the same entity which is handheld from minecraft that the first one is invalid and "removes" the first entity and then spawns a new one. The issue is we pull so fast that we can get this first "wrong" entity, which doesn't matter in the values since the are literally the same but the reference gets brocken and therefore it won't  update the game state for the entity any longer.

src/main/java/at/hannibal2/skyhanni/data/mob/Mob.kt
```diff
@@ -166,42 +174,44 @@ class Mob(
...
-        armorStand?.entityId ?: Int.MIN_VALUE -> armorStand = entity as EntityArmorStand
-        else -> {
-            extraEntitiesList.remove(entity)
-            extraEntitiesList.add(entity)
-            Unit // To make return type of this branch Unit
+    internal fun internalUpdateOfEntity(entity: EntityLivingBase) {
+        internalRemoveColor()
+        when (entity.entityId) {
+            baseEntity.entityId -> {
+                baseEntity = entity
```
hannibal002, line 204: when will two entities have the same id? i dont see the need for this

src/main/java/at/hannibal2/skyhanni/data/mob/Mob.kt
```diff
@@ -166,42 +174,44 @@ class Mob(
...
+            }
+
+            armorStand?.entityId ?: Int.MIN_VALUE -> armorStand = entity as EntityArmorStand
```
hannibal002, line 207: instead of `?:` you can use `, `. 
also, for the case that armor stand is not null, isnt this the same as with base entity? why should both have the same id but be different entity objects?
what is the case where entity.entityId is min value so that the min value check can be true?

src/main/java/at/hannibal2/skyhanni/data/mob/Mob.kt
```diff
@@ -166,42 +174,44 @@ class Mob(
...
+            else -> {
+                extraEntitiesList.remove(entity)
+                extraEntitiesList.add(entity)
+                Unit // To make return type of this branch Unit
```
hannibal002, line 211: can be removed. having return value of `when` be `Any` is fine as well/ no difference to `Unit` here

hannibal002 (changes requested):
some questions, others this time

src/main/java/at/hannibal2/skyhanni/data/mob/Mob.kt
Thunderblade73: TLDR, reference breaks because of double spawning that hypixel does. 
https://github.com/hannibal002/SkyHanni/pull/1856#discussion_r1613187418

src/main/java/at/hannibal2/skyhanni/data/mob/Mob.kt
Thunderblade73: But Intelji cries, so I changed it. Since it should be Unit semanticly.

hannibal002 (approved):

PR #1858, opened 2024-05-21
Improvement + Fix: Sea Creature Alert

ItsEmpa:
## Dependencies
- #1856 
- #1857 

## What
Adds an option to show the name of the sea creature as the title instead of just "RARE SEA CREATURE" or "RARE CATCH", makes it so that thunders, jawbus, and other bosses get highlighted by it if damage indicator is disabled. also makes it use mobdetection instead.

## Changelog Improvements
+ Added option to show the name of the sea creature in the title notification. - Empa

## Changelog Fixes
+ Fixed Thunder and Jawbus not being highlighted by Highlight Rare Sea Creatures. - Empa


CalMWolfs (approved):
code looks fine

hannibal002:
I don't feel comfortable moving more features over to the mob detection API as long as there are still random errors in detection (heavy pearls)

Also, please dont remove the enum, we want to use this to allow users to change what sea creatures to warn for.

hannibal002 (changes requested):
see above

Thunderblade73:
I just want to say that it works 99,999% of the time. Heavy Pearl is a edge case where I can't be bother to fix it (Since I need to write an exception for the "Matriach" Mob which is made out of ~50 entity's, and I never captured the mob 100%)
Also you sad a long time ago that if something breaks for an feature the feature has to fix that (in the detection). Since you argued that it is impossible to get a edge cases. And that beta users should "test" the detection (indirectly via features).

PR #1865, opened 2024-05-22
Backend: Add dungeon phases

martimavocado:
## What
Adds dungeon phase detection

+ Added dungeon phase detection. - martimavocado


src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonAPI.kt
```diff
@@ -343,4 +347,97 @@ object DungeonAPI {
             fun getByInventoryName(inventory: String) = entries.firstOrNull { it.inventory == inventory }
         }
     }
+
+    enum class DungeonPhase {
```
Thunderblade73: you can embed the patterns directly into the enum. Look at Skyblock Guide Highlight as an example of how (Not a good example but the first that came to mind)

catgirlseraid:
change phasePatternGroup to patternGroup 🥺

martimavocado:
> change phasePatternGroup to patternGroup 🥺

there's already a `patternGroup` though do i just make `phasePatternGroup` a subgroup inside `patternGroup`? or i could just add `boss.message` to every key in the new patterns i added and ditch `phasePatternGroup`

hannibal002 (changes requested):
please add regex tests to all patterns

martimavocado:
> please add regex tests to all patterns

is there a point in doing this for most of these patterns? only one isn't checking for an exact string. i don't mind doing this, just curious as to why

hannibal002 (changes requested):
Please move the code from DungeonAPI into a new class DungeonBossAPI or similar, and change the name of the event and the enum to indlicate that this is not the dungeon phase but the boss phase.

hannibal002 (changes requested):
the phase doesnt reset to null after leaving dungeon

hannibal002 (approved):

PR #1866, opened 2024-05-23
Backend: Make skyhanni ReplaceItemEvent

CalMWolfs:
## What
No longer use an neu event to replace items, and instead use our own.

+ No longer use a NEU event to replace items, and instead use our own. - CalMWolfs


src/main/java/at/hannibal2/skyhanni/mixins/transformers/gui/MixinInventoryBasic.java
```diff
@@ -0,0 +1,28 @@
...
+    @Shadow
+    private ItemStack[] inventoryContents;
+
+    @Inject(method = "getStackInSlot", at = @At("HEAD"), cancellable = true)
+    public void getStackInSlot(int index, CallbackInfoReturnable<ItemStack> cir) {
+        ItemStack stack = index >= 0 && index < this.inventoryContents.length ? this.inventoryContents[index] : null;
+        if (stack == null) return;
+        ReplaceItemEvent event = new ReplaceItemEvent((InventoryBasic) (Object) this, stack, index);
+        event.postAndCatch();
+        if (event.getReplacement() != null) {
```
hannibal002: good spot for ?.let

src/main/java/at/hannibal2/skyhanni/mixins/transformers/gui/MixinInventoryBasic.java
```diff
@@ -0,0 +1,28 @@
...
+
+@Mixin(InventoryBasic.class)
+public class MixinInventoryBasic {
+
+    @Shadow
+    private ItemStack[] inventoryContents;
+
+    @Inject(method = "getStackInSlot", at = @At("HEAD"), cancellable = true)
+    public void getStackInSlot(int index, CallbackInfoReturnable<ItemStack> cir) {
+        ItemStack stack = index >= 0 && index < this.inventoryContents.length ? this.inventoryContents[index] : null;
```
hannibal002: please use a hook

hannibal002 (changes requested):
looks good, not tested in game

src/main/java/at/hannibal2/skyhanni/mixins/transformers/gui/MixinInventoryBasic.java
ThatGravyBoat: Can we stop with hooks they are useless after the feature/event has been created and just create a bunch of files with a single method that always do the same call a single function or event that provide 0 benefit. The project already has enough clutter of single method files all over the place I dont think continuing with hooks is good for project maintainability.

hannibal002: Why do you think we use hooks in the first place? Why should this argument not bet correct here?

ThatGravyBoat: Hooks in MC dev are an old concept that were carried over from ASM/patches in older mods because it was cleaner back then to have the ASM injection call a function which you controlled externally as to not clutter up the ASM injection like many things these things continue through several mods even when not relevant, likely you probably saw them in sk1ers mods or other mods that exclusively used ASM instead of mixin or Skyblock mods that saw them use it and also did it for their mixins, in such thinking they were for things like allowing for hot swapping when in reality this was never the reason for them. 

Both cases for hooks in previous eras of mc dev had reasons from either not cluttering up patches which is a problem or not cluttering manual ASM, in the era of mixins they provide 0 benefit.

They do not make it any cleaner now and just clutter up things no which defeats their original purpose when they were originally implemented by mod developers.

hannibal002: Interesting stuff about ASM, thanks for explaining this.
My arguments for hooks are slightly different: since all mixin classes in SkyHanni are java, I want to leave the java scope as soon as possible by hooking into a Kotlin file. This brings us many benefits:

- General "Kotlin versus Java" code cleanness
- Support for break points/debugging
- Errors in the hook are readable in the stack trace without the need to use ASM dumps for mixins (not sure if this is the correct term)
- Support for hot swap without complex magic (I don't know if it's possible at all with the current system we use)
- Overall fewer java code lines compared to Kotlin, cool for statistics
- Errors that are getting thrown that happen further down the stack are cleanly detectable that they include a mixin

I can imagine there are workarounds for some of those arguments that allow us to not use hooks. But given that the existing codebase exclusively uses hooks for mixins, I don't think we should stop this code style abruptly.

We can surely discuss alternatives, but this pr doesn't provide any.

ThatGravyBoat: You can breakpoint/debug mixin files, its quite easy actually just put a breakpoint and it will work.
You can also hot swap mixins too which is also quite easy its actually part of mixin itself.
Errors are populated back to the mixin method aswell.
Errors that have a mixin in the stacktrace have a special identifier infront of them if they are not an @Overwrite mixin which you should never use anyway.

hannibal002: > You can breakpoint/debug mixin files, its quite easy actually just put a breakpoint and it will work.

How does IntelliJ know the location/line numbers of the code in the final jar? the minecraft dev intellij addon cant access the mixins of other mods in my prism profile.
and for dev env, i have not yet got it working without setting up anything special (can you reproduce this? if so, please share what you do)

> You can also hot swap mixins too which is also quite easy its actually part of mixin itself.

How? If it's simple, feel free to add support for it. I thought mixins compile only once on startup, and change mixins later requires the ASM/backend logic to first remove the original mixin, then add the modified one. this feels very complicated to me.

> Errors are populated back to the mixin method aswell.

I never said the opposite, just that it looks weird when looking at the stack trace at a glance. having a  "buffer" call in between clearly states in the stack trace that this is a mixin

> Errors that have a mixin in the stacktrace have a special identifier infront of them

What do you mean? never seen this.

ThatGravyBoat: 1. For debugging when mixins are compiled with the debug mode which is auto enabled for skyhanni they are put into the .mixin-out folder in the run folder this is where all the mixins go, you can breakpoint in here, this is also where everything is that is actually loaded in-game so if another mod mixins or another uses asm it will also be in here as to not have debugging break you breakpoint in there will also fix issues where line numbers dont add up because NEU mixins into a place.

2. Mixins come with a hotswap agent that allows for class redefinition during run you just need to add -javaagent:<absolute path to mixin jar> to vm arguments

3/4. Mixins when injected into a class create a copy of the method into it but renames it to `handler$<your_method_name>$zzo000` this is easily identifiable as a mixin, this can also maybe be changed by the mixin plugin to add skyhanni infront of it if not can be manually done if needed or can be done automatically another way, but either way it is still visible and do not need anything special.

hannibal002: how does IntelliJ know about the .mixin-out folder of my prism profile?

ThatGravyBoat: intellij menus -> run -> attach to process

hannibal002: I agree, we should stop creating files literally called "hook" with only one function. Those are useless and cause clutter.
But I don't like having more than one line or data processing/logic code in the mixins. We can simply move that logic over to the feature class that was previously using the hook.

For this example, cal will move the logic into a companion object in the ReplaceItemEvent class

hannibal002 (approved):

PR #1845, opened 2024-05-20
Backend: Create repo patterns and list of patterns

catgirlseraid:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/1733

## What
Change anything that should be a repo pattern to a repo pattern.
This includes chat messages and item lore since hypixel can and does change these.
This pr will not add patterns to inventory names or mob names.

v2 pattern groups are added since functionality is changed from match to find for better performance but if they use the same repo pattern every old version of the mod would break

## Changelog Technical Details
+ Added more repo patterns. - seraid

PR #1874, opened 2024-05-24
Feature: Highlight rabbits with requirement

superhize:
## What
Highlight rabbits that have requirements in the Hoppity Collection menu.

## Changelog New Features
+ Highlight rabbits that have requirements in the Hoppity's Collection menu. - HiZe
    * Green: Requirement met.
    * Red: Requirement not met.


src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityCollectionStats.kt
CalMWolfs, line 1: in spirit of prs like https://github.com/hannibal002/SkyHanni/pull/1845 can we use patterns for this please

CalMWolfs (changes requested):

CalMWolfs:
Can we also get an option to only show the red on not unlocked ones and otherwise show nothing, also this whole setting should probably be default off as there are probably quite a few people that dont have m7 completion that dont want to see red highlighted slots by default

CalMWolfs (approved):
worked fine

PR #1884, opened 2024-05-24
Feature: Default amount for /gfs with no args

minhperry:
Add ability to set default amout for /gfs if no amount argument is provided. Feature is togglable.

Also improving on handling the amount before sending to server. (Amount = 0 or negative, or is not integer)

This could have been splitted into further commits but I was too into this that I forgot to switch branch and committing.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
@@ -145,7 +145,7 @@ object GetFromSackAPI {
 
         when (result) {
             CommandResult.VALID -> getFromSack(stack ?: return)
-            CommandResult.WRONG_ARGUMENT -> ChatUtils.userError("Missing arguments! Usage: /getfromsacks <name/id> <amount>")
+            CommandResult.WRONG_ARGUMENT -> ChatUtils.userError("Missing arguments! Usage: /getfromsacks <name/id> [amount]")
             CommandResult.WRONG_IDENTIFIER -> ChatUtils.userError("Couldn't find an item with this name or identifier!")
             CommandResult.WRONG_AMOUNT -> ChatUtils.userError("Invalid amount!")
             CommandResult.INTERNAL_ERROR -> {}
--- a/src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
@@ -18,7 +18,7 @@ import at.hannibal2.skyhanni.utils.LorenzUtils
 import at.hannibal2.skyhanni.utils.NEUCalculator
 import at.hannibal2.skyhanni.utils.NEUInternalName
 import at.hannibal2.skyhanni.utils.NEUInternalName.Companion.asInternalName
-import at.hannibal2.skyhanni.utils.NumberUtil.isDouble
+import at.hannibal2.skyhanni.utils.NumberUtil.isInt
 import at.hannibal2.skyhanni.utils.PrimitiveItemStack
 import at.hannibal2.skyhanni.utils.PrimitiveItemStack.Companion.makePrimitiveStack
 import at.hannibal2.skyhanni.utils.SimpleTimeMark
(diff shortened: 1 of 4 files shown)


minhperry:
Would you guys prefer this approach, or another approach of setting it to 0 to disable instead?

ItsEmpa:
better to have it be a separate toggle to disable it, that way it can be disabled with `/shdefaultoptions` with the @FeatureToggle

hannibal002:
Please include the changelog format in the pr description

minhperry:
> Please include the changelog format in the pr description

Alright I will edit once I have access again.

Also would invalid input for the number (e.g
 negative numbers or non-numbers) be dealt with by the config or do I have to process that myself?

ItsEmpa:
you only have to deal with negative numbers

src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
```diff
@@ -165,16 +165,29 @@ object GetFromSackAPI {
...
     private fun commandValidator(args: List<String>): Pair<CommandResult, PrimitiveItemStack?> {
-        if (args.size <= 1) {
-            return CommandResult.WRONG_ARGUMENT to null
+        var amountString: String
+        // If only 1 arg (item name due to how it is impl) -> (if disabled -> old way, enabled -> default), else amount = last
+        amountString = if (args.size <= 1) {
+            if (!config.gfsGetDefault) {
+                return CommandResult.WRONG_ARGUMENT to null
+            } else {
+                config.gfsGetDefaultAmount.toString()
```
ItsEmpa, line 174:
```suggestion
                config.gfsGetDefaultAmount.coerceAtLeast(1).toString()
```

src/main/java/at/hannibal2/skyhanni/config/features/inventory/GetFromSackConfig.java
```diff
@@ -23,4 +24,15 @@ public class GetFromSackConfig {
     @ConfigEditorBoolean
     @FeatureToggle
     public boolean superCraftGFS = true;
+
+    @Expose
+    @ConfigOption(name = "GfS Default", desc = "When enabled, using §e/gfs §7without specifying an amount will give a default amount of the item.")
```
ItsEmpa, line 29:
```suggestion
    @ConfigOption(name = "GfS Default", desc = "When enabled, using §e/getfromsacks §7without specifying an amount will give a default amount of the item.")
```

src/main/java/at/hannibal2/skyhanni/config/features/inventory/GetFromSackConfig.java
```diff
@@ -23,4 +24,15 @@ public class GetFromSackConfig {
...
+    @ConfigEditorBoolean
+    @FeatureToggle
+    public boolean gfsGetDefault = false;
+
+    @Expose
+    @ConfigOption(name = "GfS Default Amount", desc = "The default amount of items to get when §e/gfs §7is used without amount.")
```
ItsEmpa, line 35:
```suggestion
    @ConfigOption(name = "GfS Default Amount", desc = "The default amount of items to get when §e/getfromsacks §7is used without amount.")
```

src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
```diff
@@ -165,16 +165,29 @@ object GetFromSackAPI {
     )
 
     private fun commandValidator(args: List<String>): Pair<CommandResult, PrimitiveItemStack?> {
-        if (args.size <= 1) {
-            return CommandResult.WRONG_ARGUMENT to null
+        var amountString: String
+        // If only 1 arg (item name due to how it is impl) -> (if disabled -> old way, enabled -> default), else amount = last
+        amountString = if (args.size <= 1) {
```
ItsEmpa, line 170:
```suggestion
        // If only 1 arg (item name due to how it is impl) -> (if disabled -> old way, enabled -> default), else amount = last
        var amountString = if (args.size <= 1) {
```

ItsEmpa (changes requested):

src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
ItsEmpa, line 170: intellij fucked this up oops

minhperry:
Close for now since this is not ready

src/main/java/at/hannibal2/skyhanni/config/features/inventory/GetFromSackConfig.java
jani270: Dont think thats a good one. Just makes the text longer for no reason

ItsEmpa: fair, i just thought it would be better to be consistent with the other option that has the command

minhperry:
Improved at #1893

PR #1886, opened 2024-05-25
add module ordering

CalMWolfs:
done using intellij's sort lines

I will fix any merge conflicts this here causes

exclude_from_changelog


Thunderblade73:
You know that this will lead to a merge conflict with 99% of the PRs (I would say 110).

CalMWolfs:
> You know that this will lead to a merge conflict with 99% of the PRs (I would say 110).

I will fix them all

jani270:
but like why, this is useless

hannibal002:
The idea is to add something similar to NEU's `NEUAutoSubscribe`, not cause more merge conflicts in SkyHanniMod.
Currently, I try to avoid merge conflicts via adding new classes next to similar classes.
No need for this PR entirely.

ThatGravyBoat:
> The idea is to add something similar to NEU's `NEUAutoSubscribe`, not cause more merge conflicts in SkyHanniMod.
> 
> Currently, I try to avoid merge conflicts via adding new classes next to similar classes.
> 
> No need for this PR entirely.

I don't see a point in doing the NEUAutoSubscribe at all imo it's unneeded and just hides away the functionality of the actual registering to an annotation processor that has to then generate the code for registering, the current system works fine as is and is easily understandable by most, hiding away the registering an auto generated class/object is messy & and waste of time in both compiling and actually adding the annotation processor, the only pro I see doing it that way is it saves on the 3 seconds it takes to resolve a merge conflict when another module gets added.

PR #1890, opened 2024-05-25
Fix: Small Memory Leak

hannibal002:
## What
Fixed smaller memory leak problems when staying on one island for a long time

## Changelog Fixes
+ Fixed small memory leaks when staying in one island for extended periods. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/slayer/HideMobNames.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/slayer/HideMobNames.kt
@@ -15,8 +15,8 @@ import kotlin.time.Duration.Companion.minutes
 
 class HideMobNames {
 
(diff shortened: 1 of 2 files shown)


CalMWolfs (approved):

ThatGravyBoat:
This can be better solved by just not storing the entity at all and storing the int id of the entity so there is 0 chance for the entity to not be dereferenced. You can also do a time cache but using the int should solve it pretty much in whole.

hannibal002:
> This can be better solved by just not storing the entity at all and storing the int id of the entity so there is 0 chance for the entity to not be dereferenced. You can also do a time cache but using the int should solve it pretty much in whole.

Not really relevant to the actual memory leak (would only drastically reduce the impact of memory leak, not fix it), but I still like the idea.

CalMWolfs (approved):

PR #1893, opened 2024-05-25
Feature: Default amount for /gfs with no args v2

minhperry:
## What
* Adds a toggle to enable /gfs default amount when no arguments are provided
* Improves amount check before sending the command to the server.

Please note that since this is only a draft PR, there are still remaining debug artifacts.

## Changelog New Features
* Added GfS Default and GfS Default Amount - minhperry


minhperry:
error with /gfs gemstone amount

Thunderblade73:
>error with /gfs gemstone amount

You need to fix this function
image

PR #1894, opened 2024-05-25
Fix: Scoreboard not showing player count in dungeons

ThatGravyBoat:
## What
Fixes that in dungeons the players title is replaced with a party title so the player count isnt calculated.

## Changelog Fixes
+ Fixed scoreboard player count not displaying in dungeons. - ThatGravyBoat

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/HypixelData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/HypixelData.kt
@@ -68,6 +68,9 @@ class HypixelData {
             "playeramount.guesting",
             "^\\s*(?:§.)*Guests (?:§.)*\\((?<amount>\\d+)\\)\\s*$"
         )
+        /**
+         * REGEX-TEST:           §r§b§lParty §r§f(4)
+         */
         private val dungeonPartyAmountPattern by patternGroup.pattern(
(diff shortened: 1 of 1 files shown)


lunaynx:
In theory, couldn't this be wrong if some of the party members leave the dungeon during the run?

ThatGravyBoat:
> In theory, couldn't this be wrong if some of the party members leave the dungeon during the run?

it updates accordingly, when I tested someone disconnected so that should mean its the same for someone just doing /hub or something

CalMWolfs (approved):
worked

src/main/java/at/hannibal2/skyhanni/data/HypixelData.kt
```diff
@@ -67,6 +68,10 @@ class HypixelData {
             "playeramount.guesting",
             "^\\s*(?:§.)*Guests (?:§.)*\\((?<amount>\\d+)\\)\\s*$"
         )
+        private val dungeonPartyAmountPattern by patternGroup.pattern(
```
hannibal002, line 74: please add regex test

hannibal002 (changes requested):
worked fine for me

hannibal002 (approved):

PR #1903, opened 2024-05-27
Feature: Moveable and Scaleable Hotbar

Thunderblade73:
## What
Adds the hotbar to the gui editor. Will conflict with mods like Apec.
Has no fancy alignment, it is handle like every other gui element. I wouldn't rather add that option to all gui elements then specifically to that feature

<details>
<summary>Images</summary>

image
image
image

</details>

## Changelog New Features
+ Added Editable Hotbar. - Thunderblade73
    * Allows for moving and scaling in the SkyHanni GUI editor.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/gui/HotbarConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/gui/HotbarConfig.java
@@ -25,7 +25,7 @@ public class HotbarConfig {
     public Position hotbar = new Position(20, 20);
 
     @Expose
-    @ConfigOption(name = "Show Outside Skyblock", desc = "Enables the hotbar to be edited even outside of skyblock.")
+    @ConfigOption(name = "Show Outside Skyblock", desc = "Enables the hotbar to be edited even outside of SkyBlock.")
     @ConfigEditorBoolean
     public boolean showOutsideSkyblock = false;
 }
--- a/src/main/java/at/hannibal2/skyhanni/config/features/gui/HotbarConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/gui/HotbarConfig.java
@@ -1,0 +1,31 @@
+package at.hannibal2.skyhanni.config.features.gui;
+
+import at.hannibal2.skyhanni.config.FeatureToggle;
+import at.hannibal2.skyhanni.config.core.config.Position;
+import com.google.gson.annotations.Expose;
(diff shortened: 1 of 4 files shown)


src/main/java/at/hannibal2/skyhanni/config/features/gui/HotbarConfig.java
```diff
@@ -0,0 +1,31 @@
...
+    @ConfigOption(name = "§cNotice", desc = "This option will be §c§lincompatible §r§7with mods that change the hotbar. Eg: §eApec§7.")
+    @ConfigEditorInfoText
+    public String notice = "";
+
+    @Expose
+    @ConfigLink(owner = HotbarConfig.class, field = "editable")
+    public Position hotbar = new Position(20, 20);
+
+    @Expose
+    @ConfigOption(name = "Show Outside Skyblock", desc = "Enables the hotbar to be edited even outside of skyblock.")
```
CalMWolfs:
```suggestion
    @ConfigOption(name = "Show Outside Skyblock", desc = "Enables the hotbar to be edited even outside of SkyBlock.")
```

src/main/java/at/hannibal2/skyhanni/features/gui/MovableHotBar.kt
```diff
@@ -0,0 +1,38 @@
...
+import net.minecraftforge.client.event.RenderGameOverlayEvent
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+class MovableHotBar {
+
+    private val config get() = SkyHanniMod.feature.gui.hotbar
+
+    private var post = false
+
+    fun isEnabled(): Boolean = config.editable && (config.showOutsideSkyblock || LorenzUtils.inSkyBlock)
```
CalMWolfs: I feel generally in the codebase this isEnabled function is at the bottom of the file

CalMWolfs (changes requested):
feature works fine, just a bit of code cleanup. 
Additionally, right now it defaults to the top left corner which I get why that is done but something that could be really cool is when turning on the feature for the first time (by checking if the position is set to the default position) it would calculate the current hotbar position and set its position to there, not necessary but would be cool

src/main/java/at/hannibal2/skyhanni/features/gui/MovableHotBar.kt
hannibal002: ive fixed it

hannibal002 (approved):

PR #1904, opened 2024-05-28
Backend: Auto load annotation

CalMWolfs:
## What
Added annotation to automatically load modules

Add to any module that has a repo pattern or receives subscribe events

This is just the first part of this, I only load one module so that the changelog only contains stuff useful to this annotation, but I've tested with all files and it works fine.
This only works with Kotlin Objects, a later pr will convert all the feature classes to objects and add the annotation to them. This means all future features should be made in Kotlin Objects, not Classes.
There will also be a live plugin related to this annotation, but that will come once I start adding the annotation to more things.

I also made it crash when you load the same module twice (if you add the annotation and forget to remove it from load module)

+ Added annotation to automatically load modules. - CalMWolfs
    * Just add the @SkyHanniModule annotation to an object to auto load it.

### DIFF
--- a/annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/ModuleProcessor.kt
+++ b/annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/ModuleProcessor.kt
@@ -32,12 +32,12 @@ class ModuleProcessor(private val codeGenerator: CodeGenerator, private val logg
         }
 
         if (symbol !is KSClassDeclaration) {
-            logger.error("@LoadModule is only valid on class declarations", symbol)
+            logger.error("@SkyHanniModule is only valid on class declarations", symbol)
             return null
         }
 
         if (symbol.classKind != ClassKind.OBJECT) {
-            logger.error("@LoadModule is only valid on kotlin objects", symbol)
+            logger.error("@SkyHanniModule is only valid on kotlin objects", symbol)
             return null
         }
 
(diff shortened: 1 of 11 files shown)


annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/ModuleProcessor.kt
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private fun validateSymbol(symbol: KSAnnotated): KSClassDeclaration? {
+        if (!symbol.validate()) {
+            logger.warn("Symbol is not valid: $symbol")
+            return null
+        }
+
+        if (symbol !is KSClassDeclaration) {
+            logger.error("@LoadModule is only valid on class declarations", symbol)
```
ThatGravyBoat:
```suggestion
            logger.error("@SkyHanniModule is only valid on class declarations", symbol)
```

annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/ModuleProcessor.kt
```diff
@@ -0,0 +1,68 @@
...
+            return null
+        }
+
+        if (symbol.classKind != ClassKind.OBJECT) {
+            logger.error("@LoadModule is only valid on kotlin objects", symbol)
```
ThatGravyBoat:
```suggestion
            logger.error("@SkyHanniModule is only valid on kotlin objects", symbol)
```

ThatGravyBoat (changes requested):

ThatGravyBoat (approved):

hannibal002 (approved):
works for me

PR #1914, opened 2024-05-29
Feature: DownloadSourceChecker

ThatGravyBoat:
## What
This adds a checker that will check the validity of the jar that the mod is being ran from. 
When a jar is ran using this and the jar is downloaded from an untrusted source like Discord then a pop up will show up, this popup will inform the user about the dangers of running an untrusted jar.

As of right now there are only 2 trusted sources: 
- GitHub with this repo id
- Modrinth with the SkyHanni project id.

This should stop basic attempts of just shoving crappy malware into SkyHanni to have it look similar to SkyHanni in file contents.

There is an override/skip option that will ask for a "password", this menu will also explain again that they should not be trusting people and as a last resort the password says they are in danger.

Note: This is for windows only.

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added a source download verification checker. - ThatGravyBoat
    * Warns you when the mod has been downloaded from an untrusted source (not the official GitHub or Modrinth).

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/tweaker/DownloadSourceChecker.java
+++ b/src/main/java/at/hannibal2/skyhanni/tweaker/DownloadSourceChecker.java
@@ -1,0 +1,148 @@
+package at.hannibal2.skyhanni.tweaker;
+
+import javax.swing.JFrame;
+import javax.swing.JOptionPane;
+import javax.swing.JPanel;
+import javax.swing.UIManager;
+import java.awt.event.WindowEvent;
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.net.URL;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class DownloadSourceChecker {
+
+    private static final String GITHUB_REPO_TEXT = "repo_id=511310721";
+    private static final String MODRINTH_URL = "/data/byNkmv5G/";
+    private static final String THE_PASSWORD = "danger";
+
+    private static final String[] PASSWORD_POPUP = {
+        "If someone asks you to type in here,",
+        "",
+        "the likelihood of them rating you is high!",
+        "",
+        "Enter the password:"
+    };
+
+    private static final String[] SECURITY_POPUP = {
+        "The file you are trying to run is hosted on a non-trusted domain.",
+        "",
+        "Host: %s",
+        "",
+        "Please download the file from a trusted source.",
+        "",
+        "IF YOU DO NOT KNOW WHAT YOU ARE DOING, CLOSE THIS WINDOW!",
+        "",
+        "And download from the official link below."
+    };
+
+    public static void init() {
+        if (!TweakerUtils.isOnWindows()) return;
+        String host = isAllowedFile();
+        if (host != null) {
+            openMenu(host);
+        }
+    }
+
+    private static void openMenu(String host) {
+        try {
+            UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());
+        } catch (Exception e) {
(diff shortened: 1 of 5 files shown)


src/main/java/at/hannibal2/skyhanni/tweaker/SkyHanniTweaker.java
```diff
@@ -0,0 +1,14 @@
+package at.hannibal2.skyhanni.tweaker;
+
+import org.spongepowered.asm.launch.MixinTweaker;
+
+@SuppressWarnings("unused")
+public class SkyHanniTweaker extends MixinTweaker {
```
lineargraph: Please use `(List<String>) Launch.blackboard.get("TweakClasses")` to add the mixin tweaker class name (as a string) to the list of tweakers to process, instead of extending MixinTweaker. While the MixinTweaker itself has checks to prevent double initializations to some degree, other tweakers do not, and this sets a bad example.

src/main/java/at/hannibal2/skyhanni/tweaker/DownloadSourceChecker.java
```diff
@@ -0,0 +1,148 @@
...
+            null,
+            new JPanel[] { links, buttons },
+            links
+        );
+
+        if (!close.get()) return;
+        TweakerUtils.exit();
+    }
+
+    private static String isAllowedFile() {
```
lineargraph: This could be named something like "getEndangeringHostname"

src/main/java/at/hannibal2/skyhanni/tweaker/DownloadSourceChecker.java
```diff
@@ -0,0 +1,148 @@
...
+public class DownloadSourceChecker {
+
+    private static final String GITHUB_REPO_TEXT = "repo_id=511310721";
+    private static final String MODRINTH_URL = "/data/byNkmv5G/";
+    private static final String THE_PASSWORD = "danger";
+
+    private static final String[] PASSWORD_POPUP = {
+        "If someone asks you to type in here,",
+        "",
+        "the likelihood of them rating you is high!",
```
lineargraph: Should this be ratting instead?

lineargraph (changes requested):
Looks nice, but there are some minor nit picks

CalMWolfs (approved):

hannibal002:
Doeesnt work for me.
ive tried uploading the mod on discord, downloading it via firefox, then putting this jar into mods folder. the game started like normal

ThatGravyBoat:
> Doeesnt work for me. ive tried uploading the mod on discord, downloading it via firefox, then putting this jar into mods folder. the game started like normal

try with a different browser, Firefox is known to not implement standard systems that most browsers have.

hannibal002:
zone identifier is a windows feature, should be independent of firefox, right?

ThatGravyBoat:
> zone identifier is a windows feature, should be independent of firefox, right?

Its a standard set by windows and every browser but seemingly if it is not working for you then firefox is not doing this standard and is not going through the proper download procedure . I know for a fact it works with chrome and edge because I tested on both. Firefox accounts for less than 4% of all browsers so it being a problem isnt going to affect that many people, this is just a precautionary system it is not a full proof thing, it wont stop everything but will prevent basic attempts.

hannibal002:
Just tested with chrome, didn't work as well.

Ive copied the link from the discord message that contained the file, pasted it in chrome. let it download, moved the downloaded file from chrome downloads folder to my prism profile mods folder, started prism. now i see the start screen of mc.

hannibal002 (approved):
Worked for me. (patience is key)

PR #1916, opened 2024-05-29
Feature: Jyrre Bottle and Cacao Truffle held time in lore

Obsidianninja11:
## Dependencies
- #1920 

## What
Shows the held time in lore of Bottles of Jyrre and Dark Cacao Truffles.
Also adds the option to show the dark cacao hours held as a stack size (same as existing bottle of jyrre stack size).

I think the name of the config might need to be changed. I couldn't come up with a good name for it (Currently `Time Held in Lore`).

<details>
<summary>Images</summary>

image
image
image

</details>

## Changelog New Features
+ Added option to show time held in lore for Jyrre Bottles and Cacao Truffles. - Obsidian
+ Added Dark Cacao Truffle hours held as stack size. - Obsidian

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/HeldTimeInLore.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/HeldTimeInLore.kt
@@ -3,13 +3,16 @@ package at.hannibal2.skyhanni.features.inventory
 import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.events.LorenzToolTipEvent
 import at.hannibal2.skyhanni.utils.ItemUtils.getInternalName
+import at.hannibal2.skyhanni.utils.LorenzUtils
 import at.hannibal2.skyhanni.utils.NEUInternalName.Companion.asInternalName
 import at.hannibal2.skyhanni.utils.SkyBlockItemModifierUtils.getBottleOfJyrreSeconds
 import at.hannibal2.skyhanni.utils.SkyBlockItemModifierUtils.getSecondsHeld
-import at.hannibal2.skyhanni.utils.TimeUtils
+import at.hannibal2.skyhanni.utils.TimeUtils.format
 import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+import kotlin.time.DurationUnit
+import kotlin.time.toDuration
 
-class HeldTimeInLore {
+object HeldTimeInLore {
     private val config get() = SkyHanniMod.feature.inventory
 
     private val jyrreBottle by lazy { "NEW_BOTTLE_OF_JYRRE".asInternalName() }
@@ -17,18 +20,19 @@ class HeldTimeInLore {
 
     @SubscribeEvent
     fun onTooltip(event: LorenzToolTipEvent) {
-        if (!config.heldTimeInLore) return
-        if (event.toolTip.size < 10) return // to fix when tooltip isn't fully loaded
-
+        if (!config.timeHeldInLore) return
+        if (!LorenzUtils.inSkyBlock) return
+        if (event.toolTip.size < 10) return // Fixes conflict with NEU's sort warning
         val stack = event.itemStack
         val internalName = stack.getInternalName()
-        val timeHeld = TimeUtils.formatDuration(
+        val timeHeld = (
             when (internalName) {
                 jyrreBottle -> stack.getBottleOfJyrreSeconds() ?: return
                 cacaoTruffle -> stack.getSecondsHeld() ?: return
                 else -> return
-            }.toLong() * 1000 - 999
-        )
+            }
+            ).toDuration(DurationUnit.SECONDS).format()
+
         event.toolTip.add(10, "§7Time Held: §b$timeHeld")
     }
 }
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/HeldTimeInLore.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/HeldTimeInLore.kt
@@ -21,13 +21,14 @@ object HeldTimeInLore {
     fun onTooltip(event: LorenzToolTipEvent) {
         if (!LorenzUtils.inSkyBlock) return
         if (!config.timeHeldInLore) return
+
         val stack = event.itemStack
-        val internalName = stack.getInternalName()
-        val timeHeld = when (internalName) {
+        val timeHeld = when (stack.getInternalName()) {
             jyrreBottle -> stack.getBottleOfJyrreSeconds()
             cacaoTruffle -> stack.getSecondsHeld()
             else -> return
         } ?: return
+
         val formatted = timeHeld.seconds.format()
         
         event.toolTip.add(10, "§7Time Held: §b$formatted")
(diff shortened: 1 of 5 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/HeldTimeInLore.kt
```diff
@@ -0,0 +1,34 @@
...
+    private val cacaoTruffle by lazy { "DARK_CACAO_TRUFFLE".asInternalName() }
+
+    @SubscribeEvent
+    fun onTooltip(event: LorenzToolTipEvent) {
+        if (!config.heldTimeInLore) return
+        if (event.toolTip.size < 10) return // to fix when tooltip isn't fully loaded
+
+        val stack = event.itemStack
+        val internalName = stack.getInternalName()
+        val timeHeld = TimeUtils.formatDuration(
```
CalMWolfs: 'formatDuration(Long, TimeUnit = ..., Boolean = ..., Boolean = ..., Int = ...): String' is deprecated. Has an offset of one second

src/main/java/at/hannibal2/skyhanni/features/inventory/HeldTimeInLore.kt
```diff
@@ -0,0 +1,34 @@
...
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+class HeldTimeInLore {
+    private val config get() = SkyHanniMod.feature.inventory
+
+    private val jyrreBottle by lazy { "NEW_BOTTLE_OF_JYRRE".asInternalName() }
+    private val cacaoTruffle by lazy { "DARK_CACAO_TRUFFLE".asInternalName() }
+
+    @SubscribeEvent
+    fun onTooltip(event: LorenzToolTipEvent) {
```
CalMWolfs, line 22: No skyblock check?

src/main/java/at/hannibal2/skyhanni/features/inventory/HeldTimeInLore.kt
```diff
@@ -0,0 +1,34 @@
...
+        if (!config.heldTimeInLore) return
+        if (event.toolTip.size < 10) return // to fix when tooltip isn't fully loaded
```
CalMWolfs: i have never heard of this happening, can you provide a way to reproduce this?

src/main/java/at/hannibal2/skyhanni/features/inventory/HeldTimeInLore.kt
```diff
@@ -0,0 +1,34 @@
...
+import at.hannibal2.skyhanni.SkyHanniMod
+import at.hannibal2.skyhanni.events.LorenzToolTipEvent
+import at.hannibal2.skyhanni.utils.ItemUtils.getInternalName
+import at.hannibal2.skyhanni.utils.NEUInternalName.Companion.asInternalName
+import at.hannibal2.skyhanni.utils.SkyBlockItemModifierUtils.getBottleOfJyrreSeconds
+import at.hannibal2.skyhanni.utils.SkyBlockItemModifierUtils.getSecondsHeld
+import at.hannibal2.skyhanni.utils.TimeUtils
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+class HeldTimeInLore {
```
CalMWolfs: in preparation for #1904 it would be nice for this to be an object so i have less to do later

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/HeldTimeInLore.kt
Obsidianninja11: It kept happening when hovering over the item in the auction house.
EDIT: It was because of neu's sort warning.

src/main/java/at/hannibal2/skyhanni/features/inventory/HeldTimeInLore.kt
Obsidianninja11: Oh yeah.
but also RestorePieceOfWizardPortal.kt doesn't have it

CalMWolfs: so one bit of bad code means we should add another bit of bad code?

src/main/java/at/hannibal2/skyhanni/features/inventory/HeldTimeInLore.kt
Obsidianninja11: Yes, but not really.
What it provides still uses formatDuration, just in a more complicated format

CalMWolfs: why use deprecated function ⁉️ they are deprecated for a reason

src/main/java/at/hannibal2/skyhanni/features/inventory/HeldTimeInLore.kt
Obsidianninja11: Well it's not bad code.
It should work in neu pv, ~~and I don't see a point for another check here?~~
It can't misfire, since it requires the id to match.

added

src/main/java/at/hannibal2/skyhanni/features/inventory/HeldTimeInLore.kt
```diff
@@ -0,0 +1,38 @@
...
+    private val config get() = SkyHanniMod.feature.inventory
+
+    private val jyrreBottle by lazy { "NEW_BOTTLE_OF_JYRRE".asInternalName() }
+    private val cacaoTruffle by lazy { "DARK_CACAO_TRUFFLE".asInternalName() }
+
+    @SubscribeEvent
+    fun onTooltip(event: LorenzToolTipEvent) {
+        if (!config.timeHeldInLore) return
+        if (!LorenzUtils.inSkyBlock) return
+        if (event.toolTip.size < 10) return // Fixes conflict with NEU's sort warning
```
CalMWolfs: can you let me know if my patch fixes the need for this as I think it should, if so ill open a pr with that and otherwise this is fine

CalMWolfs (changes requested):
all changes look good, just that one thing

src/main/java/at/hannibal2/skyhanni/features/inventory/HeldTimeInLore.kt
```diff
@@ -0,0 +1,37 @@
...
+        val internalName = stack.getInternalName()
+        val timeHeld = (
+            when (internalName) {
+                jyrreBottle -> stack.getBottleOfJyrreSeconds() ?: return
+                cacaoTruffle -> stack.getSecondsHeld() ?: return
+                else -> return
+            }
+            ).toDuration(DurationUnit.SECONDS).format()
+
+        event.toolTip.add(10, "§7Time Held: §b$timeHeld")
```
CalMWolfs:
```suggestion
        val timeHeld = when (internalName) {
            jyrreBottle -> stack.getBottleOfJyrreSeconds()
            cacaoTruffle -> stack.getSecondsHeld()
            else -> return
        } ?: return
        val formatted = timeHeld.seconds.format()
        
        event.toolTip.add(10, "§7Time Held: §b$formatted")
```

I feel this code is a bit cleaner

CalMWolfs (changes requested):
just one more bit of code cleanup

CalMWolfs (approved):

Obsidianninja11:
Wait what
getting the config when not in sb could cause a crash?

CalMWolfs:
> Wait what getting the config when not in sb could cause a crash?

It can, but it would require a tooltip event to be posted before config loads which is not possible

hannibal002:
> Wait what getting the config when not in sb could cause a crash?

LorenzToolTipEvent is safe to use here. But other events aren't. So It's better assume every event can crash, so it will never happen accidentally.

PR #1918, opened 2024-05-29
Fix: Chocolate factory position not showing when line overflows

sayomaki:
## What
Currently, the chocolate factory statistics do not show leaderboard position (and instead shown as ???) for players with a low position (more than 5 or 6 digits, exact digit count not confirmed) (see attached screenshot for more information)

This is caused by the regex pattern unable to match since the optional group has to either match completely or not match at all (as the line overflows with the "Chocolate" bit ending on the next line).

<details>
<summary>Images</summary>

Screenshot 2024-05-28 at 10 35 44 PM

Text description:
```
§7§7You are §8#§b112,910 §7in all-time
§7Chocolate.
§7§8You are in the top §67.93%§8 of players!
```

</details>

## Changelog Fixes
+ Fixed a bug where the chocolate factory leaderboard position was not showing when the position had many digits. - sayomaki


hannibal002 (approved):
Ive added a regex 101 test for this new line. other than that the PR was fine. thanks for the fix!

PR #1927, opened 2024-05-29
Improvement: Default amount for /gfs

saga-00:
## What
Adds default amount for /gfs in case of no args. The default amount can be changed.

## Changelog Improvements
+ Added support for using the "/gfs <item>" command without specifying an amount. - saga
    * The default value can be modified in the config.


src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
```diff
@@ -165,31 +165,42 @@ object GetFromSackAPI {
     )
 
     private fun commandValidator(args: List<String>): Pair<CommandResult, PrimitiveItemStack?> {
-        if (args.size <= 1) {
-            return CommandResult.WRONG_ARGUMENT to null
+        if (args.isEmpty()) return CommandResult.WRONG_ARGUMENT to null
+
+        val arguments = args.toMutableList()
+        val argsNull = args.last().toIntOrNull() == null
```
Thunderblade73: this could work for the calculate issue
```suggestion
        val argsNull = args.last().last().toString().toIntOrNull() == null
```

src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
```diff
@@ -165,31 +165,42 @@ object GetFromSackAPI {
...
         amountString = NEUCalculator.calculateOrNull(amountString)?.toString() ?: amountString
 
         if (!amountString.isDouble()) return CommandResult.WRONG_AMOUNT to null
 
-        val itemString = args.dropLast(1).joinToString(" ").uppercase().replace(':', '-')
+        val itemString = arguments.dropLast(1).joinToString(" ").uppercase().replace(':', '-')
+        val replacedString = itemString.replace("_"," ")
 
-        val item = when {
+        var item = when {
```
Thunderblade73: this is still a val not a var

src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
```diff
@@ -165,31 +165,42 @@ object GetFromSackAPI {
...
             SackAPI.sackListInternalNames.contains(itemString) -> itemString.asInternalName()
-            SackAPI.sackListNames.contains(itemString) -> NEUInternalName.fromItemNameOrNull(itemString) ?: run {
-                ErrorManager.logErrorStateWithData(
-                    "Couldn't resolve item name",
-                    "Query failed",
-                    "itemName" to itemString
-                )
-                return CommandResult.INTERNAL_ERROR to null
-            }
+            SackAPI.sackListNames.contains(itemString) -> NEUInternalName.fromItemNameOrNull(itemString)
```
Thunderblade73: this call is a sub set of the next calling. TLDR: can be removed.

Thunderblade73 (changes requested):
Not a bad approach. Needs some small fixing but than it should work. But it only works when the queued gfs is turned on. So I would advice to do a similar approach, of just adding the number if it doesn't find the number, but as it's own handler. Something like Line 127. (Btw. the number adding should happen before handleUnderlineReplace)

src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
Thunderblade73: Actually idk if String supports .last() . Nevertheless it isn't that hard to implement

src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
```diff
@@ -165,31 +165,41 @@ object GetFromSackAPI {
...
-            return CommandResult.WRONG_ARGUMENT to null
+        if (args.isEmpty()) return CommandResult.WRONG_ARGUMENT to null
+
+        val arguments = args.toMutableList()
+        val argsNull = args.last().last().toString().toIntOrNull() == null
+
+        if (argsNull) {
+            if (!config.defaultGFS) return CommandResult.WRONG_ARGUMENT to null
+
+            arguments.add(config.defaultAmountGFS.toString())
```
Thunderblade73: Same thing, but without a mutable list
```suggestion
       val argsNull =  args.last().last().toString().toIntOrNull() == null
        val arguments = if(argsNull){
           if (!config.defaultGFS) return CommandResult.WRONG_ARGUMENT to null
           args + config.defaultAmountGFS.toString()
        } else {
          args
```

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/api/GetFromSackAPI.kt
```diff
@@ -165,31 +165,41 @@ object GetFromSackAPI {
...
 
+        if (item == null) run {
+            ErrorManager.logErrorStateWithData(
+                "Couldn't resolve item name",
+                "Query failed",
+                "itemName" to itemString
+            )
+            return CommandResult.INTERNAL_ERROR to null
+        }
+
```
Thunderblade73: this move isn't needed anymore. It can stay where it was

src/main/java/at/hannibal2/skyhanni/config/features/inventory/GetFromSackConfig.java
```diff
@@ -23,4 +24,14 @@ public class GetFromSackConfig {
     @ConfigEditorBoolean
     @FeatureToggle
     public boolean superCraftGFS = true;
+
+    @Expose
+    @ConfigOption(name = "Default GfS", desc = "If you don't provide an amount, a default one will be used instead.")
```
Thunderblade73: Please correct my bad writing, but add something like that.
```suggestion
    @ConfigOption(name = "Default GfS", desc = "If you don't provide an amount, a default one will be used instead. Queued Gfs needs to be turned on for this to work.")
```

Thunderblade73 (changes requested):

Thunderblade73 (approved):

hannibal002:
entirely rewrote the changelog

PR #1928, opened 2024-05-29
Feature: Crimson Isle Rescue Mission Waypoints

NetheriteMiner:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/203

## What
Uses the ParkourHelper class to create a path for each variant of the Crimson Isle "Rescue Mission" quest. Does not support "S" tier because that is to be added (also, I can't find any videos of mage S tier). This feature relies on data from the existing reputation helper so it is added as an add-on to that instead of being a standalone feature (and thus does not have a FeatureToggle annotation). 

Lots of testing is needed.

## Changelog New Features
+ Added Rescue Mission Waypoints. - NetheriteMiner + hannibal2
    * Shows paths to the Agent, Recruiter, or Hostage during Crimson Isle rescue missions.


NetheriteMiner:
Since I'm a mage, I can't test the S tier barbarian quests. While this is in beta, I need a barbarian to test both variants of the S rank quest.

PR #1934, opened 2024-05-30
Improvement: Make unclaimed eggs gui clickable

Obsidianninja11:
## What
Makes the unclaimed eggs gui clickable along with the warning chat message.

Changelog msg is kinda bad, but i'm tired and can't think of something better

## Changelog Improvements
+ Made the unclaimed eggs GUI clickable. - Obsidian


Obsidianninja11:
Should this be affected by the config that turns it on for the chat message?

Might not be ready for review if it should be.

PR #1937, opened 2024-05-30
Feature: Item Pickup Log

catgirlseraid:
## What
Item pickup log, also logs coins and sack items optionally. There are setting to choose how the user wants to align the log (ie add new items at the bottom, top, center) and options to compact the numbers, so they're easier to read. There is an option to live update (combine the + and minus amounts into a single line) or keep them seperate like sba.

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added Item Pickup Log. - catgirlseraid


src/main/java/at/hannibal2/skyhanni/config/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,189 @@
...
+import at.hannibal2.skyhanni.utils.renderables.Renderable
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+import net.minecraft.client.Minecraft
+import net.minecraft.item.ItemStack
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+import java.util.Objects
+import kotlin.time.Duration.Companion.seconds
+
+@SkyHanniModule
+object ItemPickupLog {
```
j10a1n15: You accidentally created this in the config folder

j10a1n15 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,301 @@
...
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+import java.util.Objects
+import kotlin.math.absoluteValue
+import kotlin.time.Duration.Companion.seconds
+
+@SkyHanniModule
+object ItemPickupLog {
+    private val config get() = SkyHanniMod.feature.inventory.itemPickupLogConfig
+
+    private val coinIcon = "COIN_TALISMAN".asInternalName()
```
Thunderblade73, line 61: does the skyblock coin one not work?

ThatGravyBoat: skyblock coin isnt a real item in the repo so it resolves to the fallback item which is a barrier named `Missing Repo Item`

Thunderblade73: Ok.

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,314 @@
...
+            if (old != null) {
+                itemList[hash] = old.copy(second = old.second + itemStack.stackSize)
+            } else {
+                itemList[hash] = itemStack to itemStack.stackSize
+            }
+        }
+
+        if (!worldChangeCooldown()) return
+
+        oldItemList.forEach {
```
hannibal002: this can be merged with itemList below (remove code duplication)

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,314 @@
...
+        updateItem(0, UpdatedItem("§6Coins", event.coins.absoluteValue.toLong(), coinIcon), coinIcon.getItemStack(), event.coins < 0)
+    }
+
+    @SubscribeEvent
+    fun onTick(event: LorenzTickEvent) {
+        if (!isEnabled()) return
+
+        val oldItemList = mutableMapOf<Int, Pair<ItemStack, Int>>()
+
+        itemList.forEach {
```
hannibal002: use `putAll` instead

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,314 @@
...
+        ITEM_NAME("§d[Cute] Skirtwearer's Cake Soul"),
+        CHANGE_AMOUNT("§a+256"),
+        ;
+
+        override fun toString() = display
+    }
+
+    private val config get() = SkyHanniMod.feature.inventory.itemPickupLogConfig
+    private val coinIcon = "COIN_TALISMAN".asInternalName()
+    private var itemList = mutableMapOf<Int, Pair<ItemStack, Int>>()
```
hannibal002, line 63: use PrimitiveItemStack as value

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,314 @@
...
+            itemsAddedToInventory[hash]?.let {
+                it.updateAmount(itemInfo.amount)
+                return
+            }
+            if (isBannedItem(item)) return
+            itemsAddedToInventory[hash] = itemInfo
+        }
+    }
+
+    private data class UpdatedItem(val name: String, var amount: Long, val neuInternalName: NEUInternalName?) {
```
hannibal002: what does this class represent? maybe use a more descriptive name

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,314 @@
...
+        var time = SimpleTimeMark.now()
```
hannibal002: what is time? use a more descriptive name please

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,314 @@
...
+                        DisplayLayout.CHANGE_AMOUNT -> {
+                            val formattedAmount = if (config.shorten) amount.shortFormat() else amount.addSeparators()
+                            add(Renderable.string("${prefix}${formattedAmount}"))
+                        }
+
+                        DisplayLayout.ITEM_NAME -> {
+                            add(Renderable.string(name))
+                        }
+
+                        null -> {}
```
hannibal002: can be removed

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,314 @@
...
+
+    private fun isEnabled() = LorenzUtils.inSkyBlock && config.enabled
+
+    private fun renderList(prefix: String, amount: Long, name: String, itemIcon: NEUInternalName?): Renderable {
+
+
+        return Renderable.horizontalContainer(
+            buildList {
+                for (item in config.displayLayout) {
+                    when (item) {
```
hannibal002: extract into a new function to avoid many indentations

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        fun isExpired() = time.passedSince() > config.expireAfter.seconds
+    }
+
+    private fun isEnabled() = LorenzUtils.inSkyBlock && config.enabled
+
+    private fun renderList(prefix: String, amount: Long, name: String, itemIcon: NEUInternalName?): Renderable {
+
+
```
hannibal002: remove duplicate new line

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,314 @@
...
+            },
+        )
+    }
+
+
+    private fun worldChangeCooldown(): Boolean {
+        return LorenzUtils.lastWorldSwitch.passedSince() > 2.seconds
+    }
+
+    private fun updateDisplay() {
```
hannibal002, line 261: extract logic into more functions, with names that explain their steps

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,314 @@
...
+            while (iterator.hasNext()) {
+                val item = iterator.next()
+
+                if (removedItemsToNoLongerShow.containsKey(item.key)) {
+                    val it = removedItemsToNoLongerShow[item.key]
+                    if (it != null) {
+                        val currentTotalValue = item.value.amount - it.amount
+
+                        if (currentTotalValue > 0) {
+                            display.add(renderList("§a+", currentTotalValue, item.value.name, item.value.neuInternalName))
```
hannibal002: create a utils function using  `renderList` and `item.value`, to avoid the many code duplications

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
hannibal002: the `for ((key, value) : list) {` style is more readable here

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,314 @@
...
+            } else if (it.value.second > oldItemList[it.key]!!.second) {
+                val amount = (it.value.second - oldItemList[it.key]?.second!!)
+                val item = UpdatedItem(it.value.first.displayName, amount.toLong(), it.value.first.getInternalNameOrNull())
+                updateItem(it.key, item, it.value.first, false)
+            }
+        }
+        updateDisplay()
+    }
+
+    private fun ItemStack.hash(): Int {
```
hannibal002: why does this exist? whats wrong with just using neu internal name as key?

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,314 @@
...
+    private fun updateItem(hash: Int, itemInfo: UpdatedItem, item: ItemStack, removed: Boolean) {
+        if (removed) {
+            itemsAddedToInventory[hash]?.let { added ->
+                added.time = SimpleTimeMark.now()
+            }
+            itemsRemovedFromInventory[hash]?.let {
+                it.updateAmount(itemInfo.amount)
+                return
+            }
+            if (isBannedItem(item)) return
```
hannibal002: should the banned item check happen at the beginning of the function?

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,314 @@
...
+
+    @SubscribeEvent
+    fun onPurseChange(event: PurseChangeEvent) {
+        if (!isEnabled() || !config.coins || !worldChangeCooldown()) return
+
+        updateItem(0, UpdatedItem("§6Coins", event.coins.absoluteValue.toLong(), coinIcon), coinIcon.getItemStack(), event.coins < 0)
+    }
+
+    @SubscribeEvent
+    fun onTick(event: LorenzTickEvent) {
```
hannibal002, line 116: can we use ItemAddInInventoryEvent here instead?

hannibal002 (changes requested):
I've found multiple spots where I think stuff happens unnecessary, or we already have working solutions for the problem.

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
j10a1n15: We cant with the current state of the event, since that sends the internal name of the item. This isnt ideal with items that have reforges, stars, recombs,...

hannibal002: why do we need to work with those stats? you only pickup items that are "clean"

j10a1n15: Most of the time yes, but sometimes you wont and this will happen
image

hannibal002: as discussed in discord, this is necessary to differentiate between item modifiers. Please add a comment to the tick event why we cant use `ItemAddInInventoryEvent ` here.

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
hannibal002: see onTick comment

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,314 @@
...
+                    removedItemsToNoLongerShow.remove(item.key)
+                }
+            }
+        }
+        for (item in removedItemsToNoLongerShow) {
+            display.add(renderList("§c-", item.value.amount, item.value.name, item.value.neuInternalName))
+        }
+        val renderable = Renderable.verticalContainer(display, verticalAlign = config.alignment)
+
+        this.display = Renderable.fixedSizeBox(renderable, 30, 75)
```
Thunderblade73: Use a fixed SizeColumn here. So that at least the width is dynamic for the gui editor.
```suggestion
        this.display = Renderable.fixedSizeColumn(renderable, 30)
```

catgirlseraid: image
The problem is when the item list is empty it will then show up like this. The default time out is 10 seconds which means even if the player picked up an item right before opening the edit gui screen after 10 seconds it would change back to being a narrow column. I agree my current way is flawed, and are open to other ideas but i think it is better than using a column.

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
catgirlseraid: i don't really see how this could be merged as they seem to do diffferent things, I am open to discussion if i've missed something <3

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
catgirlseraid: if this is removed intellij throws a warning about enum argument can be null in java, but exhastive when contains no null branch so i put it there so there is no warning, is it better practice to do this or not?

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
j10a1n15: PrimitiveItemStack doesnt work here since the name of the item comes directly from the internal name which has the same issue we talked about yesterday with the ItemAddInInventoryEvent
image

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
hannibal002: Don't render an empty list then. Should fix it

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
hannibal002: those two code parts do the same, except on different maps, and with one value being different.
image

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
hannibal002: true. then either write a get accessor member, or explicit type in function variable
image

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
hannibal002: fair.

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
catgirlseraid: if i don't render an empty list, if the player hasn't picked up any items the list won't show up at all in the gui editor, is this ok?

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,304 @@
...
+import net.minecraft.item.ItemStack
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+import java.util.Objects
+import kotlin.math.absoluteValue
+import kotlin.time.Duration.Companion.seconds
+
+@SkyHanniModule
+object ItemPickupLog {
+
+    enum class DisplayLayout(private val display: String) {
```
hannibal002, line 40: change the enum order to the default order of displayLayout

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
hannibal002: the gui editor shows gui elements that got rendered recently. this is the same behaviour of every other gui element in skyhanni. we dont want to render/show the gui when its not neeced.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
catgirlseraid: resolved thanks to j10a1n15

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,341 @@
...
+                itemList[hash] = itemStack to itemStack.stackSize
+            }
+        }
+
+        if (!worldChangeCooldown()) return
+
+        checkForDuplicateItems(itemList, oldItemList, false)
+        checkForDuplicateItems(oldItemList, itemList, true)
+
+        updateDisplay()
```
hannibal002: make it so that update display only gets called when the display changes, not every tick.

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -0,0 +1,341 @@
...
+        addedItems: MutableMap<Int, PickupEntry>,
+        removedItems: MutableMap<Int, PickupEntry>,
+    ) {
+        val iterator = addedItems.iterator()
+        while (iterator.hasNext()) {
+            val item = iterator.next()
+
+            if (removedItems.containsKey(item.key)) {
+                val it = removedItems[item.key]
+                if (it != null) {
```
hannibal002: remove unnecessary if

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
catgirlseraid: are you sure you want this? i could do it but i would have to save an entire list of every item and check the next list against it every tick which i think could end out being more inefficent than just updating the display every tick. if you see a better way please tell me <3

catgirlseraid: done

PR #1957, opened 2024-05-31
Feature: "Not any pests" chat filter and "No pests!" title

saga-00:
## What
Adds a filter which hides the "not any pests" message. Also adds a "No Pests!" title, which triggers when you use the Pest Tracker with no pests on your garden. Those features can be used separately.

<details>
<summary>Images</summary>

image
java_YM4kzF8yW4

</details>

## Changelog New Features
+ Added No Pests Chat Filter. - saga
    * Removed the chat message "There are no Pests on your Garden!".
+ Added No Pests Title. - saga
    * Shows a title when you use the Pest Tracker without any pests to clear.


src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
```diff
@@ -403,6 +403,10 @@ class ChatFilter {
         "§d[\\w']+ the Fairy§r§f: Have a great life!".toPattern()
     )
 
+    private val gardenPestMessage = listOf(
+        "§cThere are not any Pests on your Garden right now! Keep farming!"
```
CalMWolfs: could we make this a repo pattern now so it doesnt have to be done later

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestFinder.kt
```diff
@@ -154,6 +156,15 @@ object PestFinder {
 
     private var lastKeyPress = SimpleTimeMark.farPast()
 
+    @SubscribeEvent
+    fun onChat(event: LorenzChatEvent) {
+        if (!config.noPestTitle) return
+
+        val containsCheck = GardenAPI.inGarden() && event.message == "§cThere are not any Pests on your Garden right now! Keep farming!"
```
CalMWolfs: Same as other comment

CalMWolfs (changes requested):
The message should be made into a repo pattern, probably in PestFinder and then ChatFilter can use that pattern

src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
```diff
@@ -489,6 +490,7 @@ class ChatFilter {
         dungeonConfig.soloClass && DungeonAPI.inDungeon() && message.isPresent("solo_class") -> "solo_class"
         dungeonConfig.soloStats && DungeonAPI.inDungeon() && message.isPresent("solo_stats") -> "solo_stats"
         dungeonConfig.fairy && DungeonAPI.inDungeon() && message.isPresent("fairy") -> "fairy"
+        config.gardenNoPest && GardenAPI.inGarden() && PestFinder.noPestsChatPattern.matches(message) -> "gardenPest"
```
CalMWolfs:
```suggestion
        config.gardenNoPest && GardenAPI.inGarden() && PestFinder.noPestsChatPattern.matches(message) -> "garden_pest"
```

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestFinder.kt
```diff
@@ -154,6 +162,15 @@ object PestFinder {
 
     private var lastKeyPress = SimpleTimeMark.farPast()
 
+    @SubscribeEvent
+    fun onChat(event: LorenzChatEvent) {
+        if (!config.noPestTitle) return
+
+        val containsCheck = GardenAPI.inGarden() && noPestsChatPattern.matches(event.message)
+
+        if (event.blockedReason == "gardenPest" || containsCheck) LorenzUtils.sendTitle("§eNo pests!", 2.seconds)
```
CalMWolfs: im like 99% sure that even if the message is blocked this event will still recieve it and match the pattern, therefore there is no need to check the blocked reason, also probably return if not on garden before checking if the config option is enabled

CalMWolfs (changes requested):
changes look good, just two more things i saw

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestFinder.kt
saga-00: indeed correct, didn't think about it, my first thought was the blockedReason path. Thank you for the changes, will be pushing again.

CalMWolfs (approved):

PR #1978, opened 2024-06-02
Feature: Add chocolate production time for stray rabbits

sayomaki:
## What
Show the chocolate production time for stray/golden rabbits caught in the chocolate factory menu. 

- New config option (added below duplicate rabbit time since they are somewhat related) to toggle this feature on/off
- Added regex tests to find chocolate gained from stray/golden rabbits
- Following format from duplicated rabbits (in chat message) but instead shown as a tooltip beside the stray rabbit caught
- Not adding regex pattern for Fish the Rabbit since it adds a new rabbit instead of giving chocolates (except duplicate, but handled by duplicate rabbit time already)
- I have tested the code and confirmed that it works (except golden rabbits since hobbity hunt event is currently inactive, but the pattern has been added)

Refer to attached images for an example tooltip.

<details>
<summary>Images</summary>

Screenshot

</details>

## Changelog New Features
+ Added stray/golden chocolate rabbit production time. - sayomaki


CalMWolfs (approved):
worked for me (I did 2 small fixes)

PR #1981, opened 2024-06-03
Backend: Add Kotlin mixins

ThatGravyBoat:
## What
This adds an annotation processor to allow for a new form of mixin creation called `KMixin` these allow for a more defined specified control of mixins, unlike normal mixins these are written in Kotlin only. 

The following is the specification of the current KMixin.

- KMixin is defined by either the `@KMixin` annotation or `@KPseudoMixin`
- KMixins can only be defined on an object
- Mixins can contain other code other than injects and redirects, these unlike normal mixins wont be copied to the mixined class.
- If you need the instance of the class you are mixining into in a function you can use the `@KSelf` annotation along with a variable and the type of the class
- If you need to shadow a field you can do `@KShadow` and then the name of the field and then the type
- Additionally you can do `@KShadow(ShadowKind.METHOD)` to shadow a method, to do this you must have the type be a function of the arguments and return of the method you want to shadow.
- If an inject needs to be applied to a static method you can use the `@KStatic` annotation
- For redirects of methods you use the `@KRedirectCall` annotation
- For redirecting fields you use the `@KRedirectField` annotation, depending on the return type and the arguments it will determine if this is a redirect of a get or put field.
- For basic injects you use the `@KInject` annotation.
- For injects which need to inject a specific method you use the `@KInjectAt` annotation.

> [!NOTE]
> More injectors can easily be added by adding a new `InjectorSerializer` to the `Injectors.serializers` map

+ Added KMixin (Kotlin mixins). - ThatGravyBoat


CalMWolfs (approved):
worked for me

PR #1996, opened 2024-06-04
Backend: Add automated style checking

ThatGravyBoat:
## What
This adds a workflow that will automatically make sure the style of the code is consistent, this will only work for new PRs but overtime it should eventually get to all files, from a quick check there are about 300 existing files that dont make the style config.

This updates the editor config to configure the project style and also increases the line length to 140 which is standard in most projects.

exclude_from_changelog


CalMWolfs (approved):
👀

PR #2000, opened 2024-06-04
Feature: Enchanted Book Drop Name + Edit Chat Message

ItsEmpa:
## What
Adds a feature to show the name of the enchantment in enchanted books dropped in the chat messages instead of just saying "Enchanted Book", and sends a custom message if you get a book without any chat message.

Also adds a way to be able to edit previously sent chat messages in chat without having to rely on message ids, and without the message being moved.

Still needs some more testing.

## Changelog New Features
+ Added Enchanted Book Drop Name. - Empa
    * Shows enchantment name in rare drop messages and sends a custom one if none is provided by Hypixel.

+ Added ChatUtils.editFirstMessage. - Empa


ItsEmpa:
seems like its working properly, set it as ready for review

src/main/java/at/hannibal2/skyhanni/features/chat/RareDropMessages.kt
```diff
@@ -2,82 +2,155 @@ package at.hannibal2.skyhanni.features.chat
...
+    @HandleEvent(onlyOnSkyblock = true)
+    fun onItemAdd(event: ItemAddEvent) {
+        if (event.amount != 1 || event.source != ItemAddManager.Source.ITEM_ADD) return
+        if (!config.enchantedBook) return
+        val internalName = event.internalName
+        val category = internalName.getItemStackOrNull()?.getItemCategoryOrNull() ?: return
+        if (category != ItemCategory.ENCHANTED_BOOK) return
+        if (inAnyIsland(ignoredBookIslands)) return
+
+        val anyRecent = ChatUtils.chatLines.none {
```
hannibal002: when using patcher, this will iterate over many many lines. maybe first use `take()` to not process everything but only the last 50ish messages? (i assume the messages are in order in the list)

src/main/java/at/hannibal2/skyhanni/features/chat/RareDropMessages.kt
```diff
@@ -2,82 +2,155 @@ package at.hannibal2.skyhanni.features.chat
...
+            it.passedSinceSent() < 1.seconds &&
+                (enchantedBookPattern.matches(it.message) || internalName.itemName in it.message.removeColor())
+        }
+
+        if (anyRecent) {
+            var message = "§r§6§lRARE DROP! ${internalName.itemName}"
+            userLuck?.takeIf { it != 0f }?.let { luck ->
+                var luckString = luck.roundTo(2).toString()
+                if (luck > 0) luckString = "+$luckString"
+                message += " §a($luckString ✴ SkyHanni User Luck"
```
hannibal002: not sure i like this without opt out

hannibal002 (changes requested):
code looks good. any way we can test this easily?

src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
```diff
@@ -510,7 +510,7 @@ object ItemUtils {
 
     fun NeuInternalName.isRune(): Boolean = contains("_RUNE;")
 
-    // use when showing the item name to the user (in guis, chat message, etc.), not for comparing
+    /** Use when showing the item name to the user (in guis, chat message, etc.), not for comparing. */
```
NopoTheGamer, line 525: why are we doing multi line comments for a single line?

hannibal002: Why are we using single line comments for multi lines? We never use multi-line comments inside functions, only as docs above a function.

ItsEmpa: this is not multiline comment, its a documentation comment (javadoc in java, i think its called KDoc in kt but whatever)

difference is that with doc comments, when you hover over a function
imagen
it actually shows those comments. and i think in these cases they should be shown, since they are talking about when not to use them.

hannibal002:
Empa: 
id recommend just fishing with a totem of corruption
corruption 1 books would work

hannibal002 (approved):
Can not test this, hopefully beta users can test
image

PR #2008, opened 2024-06-05
Backend: Custom event bus

ThatGravyBoat:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/212

## What
This adds a fully custom loader independent event bus, this includes several optimizations for SkyHanni specifically such as to early escape the invoker loop if no other invokers can receive cancel, this also includes things such as proper generic events, checks for if on skyblock or a skyblock island, custom event priority using a number instead of an enum with optional constants matching the forge event priorities. 

It also provides a way for the repo to remotely disable a specific event and method an event is handled on.

In addition to the additions above unlike forges event bus the cancellable events use a interface as to not put people under the wrong assumption of a cancel being allowed on an event even though its not cancellable.

> [!IMPORTANT]
> For performance benefits this only supports events on a single class 
> ie. if EventA extends EventB but EventA is posted EventB wont also be posted.

> [!NOTE]
> As of right now I have just changed 1 single event to the new event bus to show off its capabilities, all future events should be switched to this event bus and following PR(s) will switch the existing ones.

+ Added custom event bus. - ThatGravyBoat
    * Added live plugin to show when an event method is missing its annotation.


CalMWolfs (approved):

src/main/java/at/hannibal2/skyhanni/features/event/UniqueGiftingOpportunitiesFeatures.kt
```diff
@@ -71,16 +73,16 @@ object UniqueGiftingOpportunitiesFeatures {
...
         playerColor(event)
         val entity = event.entity as? EntityArmorStand ?: return
         analyzeArmorStand(entity)
     }
 
-    private fun playerColor(event: EntityEnterWorldEvent) {
+    private fun playerColor(event: EntityEnterWorldEvent<Entity>) {
         if (event.entity is EntityOtherPlayerMP) {
-            val entity = event.entity as EntityOtherPlayerMP
+            val entity = event.entity
```
Thunderblade73, line 87: the generic should be EntityOtherPlayerMP

ThatGravyBoat: It cant be because this also calls analyzeArmorStand

Thunderblade73:
> Important
> 
> For performance benefits this only supports events on a single class ie. if EventA extends EventB but EventA is posted EventB wont also be posted.

Can we get this as a explicit functionally since we have some events that need that behavior. Eg: GuiRenderEvent, MobEvent, WorldClickEvent

Thunderblade73:
> > Important
> > For performance benefits this only supports events on a single class ie. if EventA extends EventB but EventA is posted EventB wont also be posted.
> 
> Can we get this as a explicit functionally since we have some events that need that behavior. Eg: GuiRenderEvent, MobEvent, WorldClickEvent

Never mind, we can just get similar behavior with generics and type aliases

ThatGravyBoat:
> > Important
> > For performance benefits this only supports events on a single class ie. if EventA extends EventB but EventA is posted EventB wont also be posted.
> 
> Can we get this as a explicit functionally since we have some events that need that behavior. Eg: GuiRenderEvent, MobEvent, WorldClickEvent

The explicit way is to subscribe to the events individually, I dont think its beneficial in anyway to let it try to post its super types.
If a class needs it they can subscribe to the other events.

- GuiRenderEvent most of them dont actually use the supertype.
- MobEvent the super type is only used once and the same class subscribes to ALL of the sub events so it can easily be delegating them.
- WorldClickEvent's super type is only ever used once

ThatGravyBoat:
> > > Important
> > > For performance benefits this only supports events on a single class ie. if EventA extends EventB but EventA is posted EventB wont also be posted.
> > 
> > 
> > Can we get this as a explicit functionally since we have some events that need that behavior. Eg: GuiRenderEvent, MobEvent, WorldClickEvent
> 
> Never mind, we can just get similar behavior with generics and type aliases

And yes you can also do that

.live-plugins/event/plugin.kts
```diff
@@ -0,0 +1,76 @@
...
+class HandleEventInspectionKotlin : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+
+        val visitor = object : KtVisitorVoid() {
+            override fun visitNamedFunction(function: KtNamedFunction) {
+                val hasEventAnnotation = function.annotationEntries.any { it.shortName!!.asString() == handleEvent }
+                val isEvent = function.valueParameters.firstOrNull()?.type()?.supertypes()
+                    ?.any { it.fqName?.asString() == skyhanniEvent } ?: false
+
+                if (isEvent && !hasEventAnnotation && function.valueParameters.size == 1 && function.isPublic) {
```
Thunderblade73, line 34: idk how but this is an issue
image

CalMWolfs: this is related to the other plugin not the one added in this one, and its not really wrong but ill push a commit that kind of disables this warning for now until #1983 gets merged

Thunderblade73: cal fixed the issue in beta

hannibal002 (approved):

PR #2017, opened 2024-06-06
Improvement: Add Flight duration to Custom Scoreboard

ItsEmpa:
## What
Adds flight duration back to custom scoreboard. since Hypixel decided to randomly remove it from getting displayed in scoreboard, we have to track the duration ourselves.

Continues #1415 

## Changelog Improvements
+ Re-added Flight Duration to Custom Scoreboard. - Empa, j10a1n15


ItsEmpa:
so uh

image

PR #2030, opened 2024-06-08
Improvement: SkyMart Copper Price Bazaar Options

Triggered0:
## What
Add option to choose "sell order / instant sell" for Copper price

<details>
<summary>Images</summary>

image

</details>

## Changelog Improvements
+ Added option to choose Bazaar price for Copper price. - Trîggered


hannibal002 (approved):
Works fine!

The legacy ID is only needed for old enum values, as this was needed to migrate them from the old ID system over to enum system. This is not needed for a new enum at all.

I've replaced `getPrice` with `getPriceOrNull`. The former will break the feature and show a user error when a copper price is not found. Using `orNull` with `continue` will simply skip those items. This is useful when Hypixel decides to add new items to the SkyMart that are not yet in the NEU repo. Having error messages for users at that spot doesn't benefit anyone, as the bug is not with this feature, but rather with missing neu repo items.

PR #2033, opened 2024-06-08
Improvement: Use EnumMap and fields for config

lineargraph:
## PR Reviews

When your PR is marked as ready for review, some of our maintainers will look through your code to make sure everything is good to go. In order to do this, they may request some changes you will need to do, **or fix smaller stuff (like merge conflicts) for you**. If a maintainer has reviewed your PR, make sure to **pull any of their changes** into your local project before doing more work on your code. Having maintainers fix small stuff for you helps us speed up the process of merging your PR, so if some of your systems warrant further care, be sure to let us know (preferably with a code comment).

Make sure to only mark your PR as "Ready to review" when it is. If you still want to do major changes, you can keep a draft PR open until then.

## What
The config is accessed quite often and using a hash map for each one of these accesses adds up over time. This isn't suuper bad, but it is bad enough to the point where it does show up in profiles occasionally and the fix is quite easy. Using an `EnumMap` instead of a regular map is already a bit of a performance boost, but since `SkyhanniMod.feature` is used very often, I also added a bit of kotlin reflection to automatically set a field whenever it is loaded. Thanks to kotlins property reflection syntax this is fairly easy to use and extend, as well as type checked during compilation.

### Going further

Almost every feature right now accesses the config multiple times per tick, sometimes per frame. While those accesses are just a few chained `GETFIELD`s, this could probably be improved. Config changes are rare and config changes within the same tick are something that we should simply not consider. The infrastructure for this does not exist right now, but one could imagine a future where moulconfigs properties are used extensively to reduce each config access down to at most two `GETFIELD`s each method call. Is this our future or one of our many divergent timelines? Is this future desirable? Is it worth the effort? I don't have the answer. I just know that performance can be found even in the most unlikely of places if you just open your eyes.

## Changelog Improvements
+ Improved overall performance slightly by about ~1%. - nea

+ Added mixin search scope. - nea
+ Changed `.gitignore` so that subdirectories of the `.idea` folder can be manually unignored. - nea

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
-.idea/*
+# Use **/*.* since if a directory is ignored, children of that directory *cannot* be unignored using ! again.
+.idea/**/*.*
 !.idea/icon.svg
 !.idea/dictionaries/default_user.xml
+!.idea/scopes/Mixins.xml
 .vscode/
 run/
 build/
--- a/.idea/scopes/Mixins.xml
+++ b/.idea/scopes/Mixins.xml
@@ -1,0 +1,3 @@
+<component name="DependencyValidationManager">
+  <scope name="Mixins" pattern="src:at.hannibal2.skyhanni.mixins.transformers..*" />
+</component>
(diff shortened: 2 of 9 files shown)


hannibal002 (approved):
jani hat nix anderes erwartet

PR #2046, opened 2024-06-09
Feature: Last server

j10a1n15:
## What
This Pull Request adds a feature, which reminds the player if he joins a server he joined previously (customisable time).

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added a notification when you join a server you have joined previously. - j10a1n15


hannibal002 (changes requested):
Does not work for me. `LorenzUtils.inSkyBlock` is false at the moment the world change event fires. 
I suggest an different approach:
periodically check the value of `HypixelData.serverId`, compare against last known value. If the value changes, assume server switch. This should also work faster than 5 seconds after joining. Also fixed the potential bug when a user quickly swaps servers in under 5 seconds.

hannibal002 (approved):
works fine now

PR #2053, opened 2024-06-10
Backend: Fixed Scrollables infinity scrolling

Thunderblade73:
## What
Fixed a bug with scrollabels if they are intantiated each frame which lead to infinite scrolling in the last direction.

+ Fixed infinite scrolling for scrollables. - Thunderblade73


hannibal002 (approved):
works for me

PR #2055, opened 2024-06-11
Fix: Corrupt config files not resetting to default state

CalMWolfs:
## Changelog Fixes
+ Fixed corrupted config files not being reset to the default state. - CalMWolfs


src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -203,6 +203,11 @@ class ConfigManager {
         if (output == defaultValue) {
             logger.log("Setting $fileName to be blank as it did not exist. It will be saved once something is written to it")
         }
+        @Suppress("SENSELESS_COMPARISON")
+        if (output == null) {
```
hannibal002, line 206: defaultValue cant be null, output should never be null here as well. what edge case is this?
if we call this function with defaultValue null, it should throw an error elsewhere before touching this code, right?

Maybe we should add `var output: Any? = defaultValue` above?

hannibal002 (changes requested):
see review

hannibal002 (approved):

PR #2056, opened 2024-06-11
Backend: LorenzVec up/down and drawLineToEye

ItsEmpa:
Adds a .up() and .down() methods to lorenzvec, and uses them in all cases where its useful instead of adding a vector with a specific y value. also adds drawLineToEye method, and a couple other cleanup functions related to LorenzVec

+ Cleaned up LorenzVec and added drawLineToEye. - Empa


src/main/java/at/hannibal2/skyhanni/features/rift/area/livingcave/LivingCaveDefenseBlocks.kt
```diff
@@ -146,12 +147,11 @@ object LivingCaveDefenseBlocks {
             if (time > System.currentTimeMillis()) {
                 val location = block.location
                 event.drawWaypointFilled(location, color)
-                event.draw3DLine(
-                    block.entity.getLorenzVec().add(y = 0.5),
+                event.drawLineToEye(
                     location.add(0.5, 0.5, 0.5),
```
CalMWolfs: Shouldn't this also be block center

CalMWolfs (changes requested):

CalMWolfs (approved):

PR #2058, opened 2024-06-11
Fix: Anniversary Custom Scoreboard

ItsEmpa:
## What
Fixes the anniversary line in scoreboard not being detected

## Changelog Fixes
+ Fixed Anniversary Line not being detected in Custom Scoreboard. - Empa, j10a1n15

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/CollectionUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/CollectionUtils.kt
@@ -133,6 +133,9 @@ object CollectionUtils {
         return newList
     }
 
+    /**
+     * This does not work inside a [buildList] block
+     */
     fun List<String>.addIfNotNull(element: String?) = element?.let { plus(it) } ?: this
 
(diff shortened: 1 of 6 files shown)


j10a1n15 (approved):
Undraft it, and ready to go

src/main/java/at/hannibal2/skyhanni/utils/CollectionUtils.kt
```diff
@@ -133,6 +133,9 @@ object CollectionUtils {
         return newList
     }
 
+    /**
+     * This does not work inside a [buildList] block
```
hannibal002, line 137: yes, because build list blocks are MutableLists, this is an extension function to a List.

hannibal002: Also it looks like this function is unused?

hannibal002 (approved):
the countdown in the custom scoreboard sometimes skips a second, then later stays on the same second for two seconds. not ideal. this is a problem of the general update delay, i suppose. outside of this pr.

works fine enough for now

PR #2068, opened 2024-06-12
Backend: CustomeScoreboard UnkownLinesHandler auto adds patterns from the repo

Thunderblade73:
## What
Adds the ability for the UnkownLinesHandler to get patterns form the repo that are not in his list but should. Eg: A completely new line for the scoreboard.
Does not automatically add present patterns to the list only patterns that are not present localy but remotly. Since the current exklusive pattern group implementation does not allow to get "allPatterns".

Btw. I love that the PR is more formatting as really changes.

+ New Scoreboard lines are added automatically from remote. - Thunderblade73
    * The pattern still needs to be added to the list in development.


j10a1n15:
It errored for the first 2-3s, then it stopped. So I think it works? Idk I think it should be fine?
image

Thunderblade73:
The first thing might be the delay for the repo update.

j10a1n15 (approved):
Nothing we can really do about a repoupdate delay. Im sure its fine.

hannibal002 (approved):

PR #2090, opened 2024-06-14
Fix: Golden Strays Causing New Rabbits to Detect as Dupes

DavidArthurCole:
## What
Fix an issue where Golden Strays could cause Hoppity Compact to always flag New Rabbits as dupes.
See this thread for context, and images surrounding the issue itself: https://discord.com/channels/997079228510117908/1251015915739414559

I believe what happened in my case, is that a Golden Stray opened a Dinner egg for me before this compact happened, which did not trigger the compact to fully run, but did pre-populate the variable objects with some dummy data, that was stored and used the next time compacting ran.

The fix for this was doing a full-reset of all variables when `eggFoundPattern` matches. This was moved to a private function so the same set of clearing could be applied to post-send.

## Changelog Fixes
+ Fixed a rare case where golden strays caused incorrect duplicate detection during Compact Chat for Hoppity's. - Daveed


hannibal002 (approved):
LGTM

PR #2097, opened 2024-06-14
Backend: HotmAPI Level Saving wrong with Blue Egg

Thunderblade73:
+ Fixed HotmAPI saving incorrect level values when using Blue Egg. - Thunderblade73


hannibal002 (approved):
LGTM

PR #2098, opened 2024-06-14
Backend: ConcurrentModifactionException in RepoPatternManager during Tests

Thunderblade73:
https://discord.com/channels/997079228510117908/1251234438960910428

The problem was that the logger instantiated the SkyHanniMod class. Which initialized everything inside of the test.

+ Fixed a ConcurrentModificationException in RepoPatternManager during tests. - Thunderblade73


src/main/java/at/hannibal2/skyhanni/utils/repopatterns/RepoPatternManager.kt
```diff
@@ -169,7 +169,8 @@ object RepoPatternManager {
      */
     private fun reloadPatterns() {
         val remotePatterns = remotePattern
-        for (it in usedKeys.values) {
+        val instances = usedKeys.values
+        for (it in instances) {
```
lineargraph: This does not change anything. The right hand side of `in` is only evaluated once.

hannibal002 (approved):

PR #2099, opened 2024-06-14
Backend: Fixed SkyHanni not being a mod

hannibal002:
## What
We were a tweaker only for couple commits. Now we are a tweaker and a mod. 🥳 

+ Made SkyHanni into a Mod. - hannibal2 + nea

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/tweaker/ModLoadingTweaker.java
+++ b/src/main/java/at/hannibal2/skyhanni/tweaker/ModLoadingTweaker.java
@@ -1,0 +1,51 @@
+package at.hannibal2.skyhanni.tweaker;
+
+
+import net.minecraft.launchwrapper.ITweaker;
+import net.minecraft.launchwrapper.LaunchClassLoader;
+import net.minecraftforge.fml.relauncher.CoreModManager;
+import org.spongepowered.asm.launch.MixinBootstrap;
+
+import java.io.File;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.List;
+
+/**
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/tweaker/ModLoadingTweaker.java
```diff
@@ -0,0 +1,51 @@
...
+        URL location = ModLoadingTweaker.class.getProtectionDomain().getCodeSource().getLocation();
+        if (location == null) return;
+        if (!"file".equals(location.getProtocol())) return;
+        try {
+            MixinBootstrap.getPlatform().addContainer(location.toURI());
+            String file = new File(location.toURI()).getName();
+            CoreModManager.getIgnoredMods().remove(file);
+            CoreModManager.getReparseableCoremods().add(file);
+        } catch (URISyntaxException e) {
+            System.err.println("NEU could not re-add itself as mod.");
```
jani270:
```suggestion
            System.err.println("SkyHanni could not re-add itself as mod.");
```

jani270: Could not guess that it was taken from neu

lineargraph: not neu

lineargraph (changes requested):

src/main/java/at/hannibal2/skyhanni/tweaker/ModLoadingTweaker.java
ItsEmpa: gg skyhanni is officially neu

PR #2100, opened 2024-06-14
Fix: Compact Hoppity Chat for Purchased Rabbits

DavidArthurCole:
## What
Reported by Luna here: https://discord.com/channels/997079228510117908/1251294395140145322

The issue stemmed from having no capturing for the case where a rabbit is purchased from hoppity, so the chats were allowed through. This PR addresses that while leaving all other aspects untouched.

The main issue, and reason for the complexity of the comment in the new `reset` call, is that when finding a meal egg, the "you found [meal egg]" is the first message in the chain, whereas when purchasing, the "you bought [rabbit]" message is the last in the chain, so a new flow had to be designed.

<details>
<summary>Images</summary>

image

</details>

## Changelog Fixes
+ Fixed Compact Hoppity Chat not working with rabbits purchased from the Hoppity NPC. - Daveed


hannibal002 (approved):
LGTM

PR #2101, opened 2024-06-15
Fix: Hoppity Requirement when already found.

hannibal002:
## What
Hypixel allows purchasing Rabbits from Hoppity NPC even when the requirement is not yet met.
Fixed highlight wrong missing requirement.

## Changelog Fixes
+ Fixed highlight missing requirements even when Hoppity Rabbit is already found. - hannibal2


Obsidianninja11:
I'm kinda against this.
The requirement and it being "found" (by buying) are kind of unrelated.
This could be kind of misleading if it's merged

(but also this is so minor i don't think it's a big deal if it goes either way)

CalMWolfs:
> I'm kinda against this.
> 
> The requirement and it being "found" (by buying) are kind of unrelated.
> 
> This could be kind of misleading if it's merged
> 
> 
> 
> (but also this is so minor i don't think it's a big deal if it goes either way)

There is really no reason to get the requirement once you already have the rabbit. Unless hypixel fixes the bug and removes "illegally" gotten rabbits there is no reason for this or not to be merged

CalMWolfs (approved):

PR #2104, opened 2024-06-15
Backend: Mobdetection Shiny Pig

Thunderblade73:
## What
Added shiny pig to the mob detection. It's a fix in the sense that the pigs would cost some small performance since the detection would try to match them to anything.

To Test: Enable the MobDebug Setting for Special Mobs to Name or HighlightWithName and see if the pigs get a name above them, that isn't the armorstand, so there should be the displayed twice. 

+ Added Shiny Pig to Mob Detection. - Thunderblade73


src/main/java/at/hannibal2/skyhanni/data/mob/MobFilter.kt
```diff
@@ -121,6 +121,13 @@ object MobFilter {
         "§.§lCOLLECT!",
     )
 
+    /** REGEX-TEST: SHINY PIG
```
DavidArthurCole:
```suggestion
    /**
     * REGEX-TEST: SHINY PIG
```

CalMWolfs (approved):
Works for me
image

PR #2105, opened 2024-06-15
Improvement: Add Keybinds to Custom Wardrobe

j10a1n15:
## What
This Pull Request adds custom wardrobe slot keybinds to skyhanni, only while in the custom wardrobe.

> [!Warning]
> ~~You somehow are able to use this, even when the config to use custom keybinds or even the custom wardrobe is disabled?
> It like straight up ignores the config values
> image~~
>
> Nevermind, I want to die, I had neu's keybinds enabled

## Changelog Improvements
+ Added Custom Wardrobe Slot Keybinds. - j10a1n15
    * These function independently of NEU. NEU's keybinds will not work in the Custom Wardrobe.


hannibal002 (approved):
LGTM

PR #2112, opened 2024-06-17
Improvement + Fix: Ashfang Features with Mob Detection

ItsEmpa:
## Dependencies
- #2096 

## What
Refactors basically the entirety of the Ashfang features to use mob detection and a cleaner code style, and adds things like having the text on the gravity orbs and blaze souls be adjusted with distance.

## Changelog Fixes
+ Fixed Ashfang Blazes sometimes being highlighted with the wrong color. - Empa
+ Fixed Ashfang Reset Cooldown counting in the wrong direction. - Empa

## Changelog Technical Details
+ Added EntityLeaveWorldEvent. - Empa

PR #2114, opened 2024-06-17
Backend: Mob Highlight

ItsEmpa:
## Dependencies
- #2096 
- #2112 

## What
Refactors the MobHighlight class to (mostly) use mob detection. For highlighting zealots the old EntityMaxHealthUpdateEvent is used instead, as mob detection currently does not work with mobs that are far away.

Made it possible to simply use Entity.highlight() to highlight entities, without having to remove the highlight after that (as it now gets removed with the EntityLeaveWorldEvent).

Also did some cleanup in EntityUtils and added a function to more easily know if an entity would have a certain base health if it wasnt corrupted/runic.

+ Refactor MobHighlight to use Mob Detection. - Empa
    * With the exception of Zealot-related highlights.
+ Added an Entity.highlight(color) function. - Empa
+ Added EntityLivingBase.matchesHealth() for checking if a mob matches a certain health when runic/corrupted. - Empa


CalMWolfs:

PR #2115, opened 2024-06-17
Fix: Ore Event

ItsEmpa:
## What
Fixes a couple issues with the ore mined event, like it sometimes getting stuck.

## Changelog Fixes
+ Fixed ore event sometimes getting stuck. - Empa

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/MiningAPI.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/MiningAPI.kt
@@ -70,7 +70,7 @@ object MiningAPI {
 
     private var recentClickedBlocks = mutableMapOf<LorenzVec, MinedBlock>()
     private var surroundingMinedBlocks = mutableMapOf<LorenzVec, MinedBlock>()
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/data/MiningAPI.kt
```diff
@@ -143,7 +143,7 @@ object MiningAPI {
         if (waitingForInitSound) {
             if (event.soundName in allowedSoundNames && event.pitch == 0.7936508f) {
                 val pos = event.location.roundLocationToBlock()
-                if (recentClickedBlocks.none { it.position == pos }) return
+                if (recentClickedBlocks.toList().none { it.position == pos }) return
```
hannibal002: this is useless. rather change recentClickedBlocks to either be val or be immutable if you want to fix concurrent modification exceptions

hannibal002 (changes requested):

hannibal002 (approved):

PR #2117, opened 2024-06-18
Fix: Kill Combo Chat Filter

DavidArthurCole:
## What
Reported here: https://discord.com/channels/997079228510117908/1252469848991072307
Kill combo chat filter stops working at kill combos 75 and above, due to malformed pattern.

<details>
<summary>Images</summary>

image

</details>

## Changelog Fixes
+ Fixed kill combo chat filter not working at high combo counts. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
@@ -89,9 +89,13 @@ object ChatFilter {
     )
 
     // Kill Combo
+    // To note: there is an intentional space left after the "50" kill combo test case,
+    // as said space is evidently present for 50, but no further values in the chain.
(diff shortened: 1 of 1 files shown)


hannibal002:
Can we merge the two patterns into one regex?

jani270:
I must say that i actually liked this "bugged" way before, the low combos got hidden but if you did grind for a high combo you got the cool message.

hannibal002:
I just looked at the code again. This is one of the few instances where "it's not a bug, it's a feature" holds true. I intentionally wanted to ignore higher kill combo numbers from the filter. 

Maybe we can add more settings to this filter than simply "entirely visible/hidden"?

DavidArthurCole:
Suppose if this is intentional functionality, I'll close this out. I wasn't too bothered by it - just figured it was a bug, and not intended. I'm good with "working as intended"

PR #2125, opened 2024-06-19
Fix: Double Hook not detected with thunder bottle filled message

appable0:
## What
Fixes double hook sea creatures not being properly detected when a thunder bottle is filled simultaneously. Resolves [this Discord bug report](https://discord.com/channels/997079228510117908/1252318543714717766).

**Note**: I bypassed the thunder bottle filled message instead of ignoring the first non-sea-creature message after a double hook. This is to avoid double hook being assigned to the wrong sea creature if the first sea creature isn't detected (e.g. if Hypixel adds new sea creatures).

## Changelog Fixes
+ Fixed a rare double hook detection issue. - appable


ItsEmpa:
> e.g. if Hypixel adds new sea creatures
the messages used for this are already stored in repo, so that wouldnt be much of an issue imo
not that important tho, so i think this solution is good

hannibal002 (approved):

PR #2135, opened 2024-06-21
Fix: Fishing Bait Warning

ItsEmpa:
## What
Fixes fishing bait warnings not working at all (item category was set to "FISHING_BAIT" instead of "BAIT"
). also refactored it while at it

## Changelog Fixes
+ Fixed Fishing Bait warnings not working. - Empa


hannibal002 (approved):
works fine for me

PR #2140, opened 2024-06-23
Fix: Powder loss on scoreboard not being accounted for

lunaynx:
## What
Fixed powder loss on scoreboard not being accounted for.

## Changelog Fixes
+ Fixed powder loss on scoreboard not being counted. - Luna
    * This could happen if your powder amount changed while SkyHanni wasn't there to see it, or Hypixel sent an incomplete scoreboard line, and would sometimes lead to incorrect powder amounts being displayed until you open `/hotm`.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/HotmData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/HotmData.kt
@@ -652,9 +652,10 @@ enum class HotmData(
                 val amount = group("amount").formatLong()
                 val difference = amount - type.getCurrent()
 
-                if (difference > 0) {
+                if (difference != 0L) {
                     type.gain(difference)
-                    ChatUtils.debug("Gained §a${difference.addSeparators()} §e${type.lowName} Powder")
+                    val what = if (difference > 0) "Gained" else "Lost"
+                    ChatUtils.debug("$what §a${difference.addSeparators()} §e${type.lowName} Powder")
                 }
             }
         }


Thunderblade73:
This will mess up the total powder amount. Also powder can only change in hotm (as far as I now). Better way to fix it would be to block the scoreboard when inside the hotm menu (since there it would happen, I assume)

lunaynx:
> This will mess up the total powder amount.

That's possible, I haven't thought about it too much.

> Also powder can only change in hotm (as far as I now). Better way to fix it would be to block the scoreboard when inside the hotm menu (since there it would happen, I assume)

No? It can change when opening chests or mining mithril/gemstones/glacite, and the update is shown on the scoreboard (usually, sometimes it misses some changes, e.g. if you gain mithril adn gemstone powder at once, but it tends to be resynced once you next gain that type of powder).

While yes, it can't normally *decrease* if you're not actively spending it on HOTM perks, it could have decreased since last time if you spent powder without SkyHanni active to see it.

Anyway, I'm still not entirely sure how the bug happens, I should probably wait until I can reproduce it again with more debug logging. But it happens when *outside* the `/hotm` menu and powder mining, and doesn't seem to resync until I open the `/hotm` menu again for some reason.

lunaynx:
Maybe if we detect powder loss on the scoreboard, we should just prompt the user to open `/hotm` to resync instead of trying to guess what's right.

PR #2145, opened 2024-06-24
Fix: Powder cost formulas and calculations

lunaynx:
## What
Fixed a bunch of errors and inconsistencies in the HotmAPI powder cost formulas, as well as made the powder spent calculation more logical.

## Changelog Fixes
+ Fix incorrect powder costs and powder spent amounts being displayed for some HOTM perks. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/HotmData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/HotmData.kt
@@ -34,7 +34,6 @@ import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
 import net.minecraft.inventory.Slot
 import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
 import kotlin.math.ceil
-import kotlin.math.floor
 import kotlin.math.pow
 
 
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/data/HotmData.kt
```diff
@@ -57,50 +56,50 @@ private fun calculatePeakOfTheMountainLoot(level: Int): Map<HotmReward, Double>
...
+        { level -> (level + 1.0).pow(3.05).toInt() },
         { level -> mapOf(HotmReward.MINING_FORTUNE to level * 5.0) },
     ),
     QUICK_FORGE(
         "Quick Forge",
         20,
-        { currentLevel -> (currentLevel + 2.0).pow(4) },
-        { level -> mapOf(HotmReward.FORGE_TIME_DECREASE to 10.0 + (level * 0.5)) },
+        { level -> (level + 1.0).pow(4).toInt() },
+        { level -> mapOf(HotmReward.FORGE_TIME_DECREASE to if (level >= 20) 30.0 else 10.0 + (level * 0.5)) },
```
Thunderblade73, line 79: are you sure that it does not work with blue egg?

Thunderblade73: oh, the updated the wiki formula
```
Reward Formula: min(30, 10+(Level*0.5)+floor(Level/20)*20)
```

lunaynx: Yeah, I'm pretty sure blue egg doesn't buff it further, but I can have someone confirm.

lunaynx: Confirmed.
image

PR #2147, opened 2024-06-25
Fix: launchers reading mod as libninepatch

ThatGravyBoat:
## What
This fixes launchers like Modrinth detecting the mod as LibNinePatch because it contains a shadowed fabric.mod.json file from MoulConfig which uses LibNinePatch

## Changelog Fixes
+ Fix launchers showing SkyHanni as LibNinePatch. - ThatGravyBoat


hannibal002 (approved):
This indeed fixed the bug.

PR #2151, opened 2024-06-25
Backend: RecalculatingValue as a ReadOnlyProperty

ItsEmpa:
## What
Makes RecalculatingValue be a ReadOnlyProperty, so that it can be used as a property delegate, so you dont need to call .getValue() to use it. Also makes its usages be done like this.

+ Made RecalculatingValue a ReadOnlyProperty. - Empa

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/SlayerAPI.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/SlayerAPI.kt
@@ -62,7 +62,7 @@ object SlayerAPI {
         }
 
         event.addData {
-            add("activeSlayer: ${getActiveSlayer()}")
+            add("activeSlayer: $activeSlayer")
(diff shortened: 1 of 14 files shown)


hannibal002 (changes requested):
```
SkyHanni 0.26.Beta.12: Caught an NullPointerException in SlayerAPI at LorenzTickEvent: null
 
Caused by java.lang.NullPointerException: null
	at SH.utils.RecalculatingValue.getValue(RecalculatingValue.kt:18)
	at SH.utils.RecalculatingValue.getValue(RecalculatingValue.kt:22)
	at SH.data.SlayerAPI.getActiveSlayer(SlayerAPI.kt:85)
	at SH.data.SlayerAPI.onTick(SlayerAPI.kt:126)
	at SH.data.MinecraftData.onTick(MinecraftData.kt:76)
```

PR #2157, opened 2024-06-25
Fix: Double Powder Detection

superhize:
## What
Fix detection of 2x Powder event in Powder Tracker

## Changelog Fixes
+ Fixed detection of 2x Powder Event for Powder Tracker. - HiZe
    * Now uses the Tab Widget.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/mining/powdertracker/PowderTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/mining/powdertracker/PowderTracker.kt
@@ -14,6 +14,7 @@ import at.hannibal2.skyhanni.events.LorenzWorldChangeEvent
 import at.hannibal2.skyhanni.events.SecondPassedEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.CollectionUtils.addAsSingletonList
+import at.hannibal2.skyhanni.utils.CollectionUtils.nextAfter
 import at.hannibal2.skyhanni.utils.ConditionalUtils.afterChange
 import at.hannibal2.skyhanni.utils.ConfigUtils
 import at.hannibal2.skyhanni.utils.LorenzUtils.isInIsland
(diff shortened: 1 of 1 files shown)


hannibal002 (changes requested):
the event widget is default disabled. this means we need to tell the user to enable the widget in order to fully use the powder tracker.
also this gets overwritten by other events, e.g. hoppity event. 

im not sure if replacing the original logic is a good idea. maybe keep the BossbarData logic as a workaround until we can confirm the tab widget logic works 100%

src/main/java/at/hannibal2/skyhanni/features/mining/powdertracker/PowderTracker.kt
```diff
@@ -56,6 +59,10 @@ object PowderTracker {
         "powder.bossbar",
         "§e§lPASSIVE EVENT §b§l2X POWDER §e§lRUNNING FOR §a§l(?<time>.*)§r",
     )
+    private val tablistEventDuration by patternGroup.pattern(
```
hannibal002, line 66: please add regex tests

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/mining/powdertracker/PowderTracker.kt
```diff
@@ -98,12 +110,28 @@ object PowderTracker {
...
+                    val durationLine = TabWidget.EVENT.lines.getOrNull(index + 1) ?: return
+                    tablistEventDuration.matchMatcher(durationLine) {
+                        val duration = group("duration")
+                        powderTimer = TimeUtils.getDuration(duration)
+                        tracker.update()
+                    }
+                }
+            }
+        } else {
+            doublePowder = powderBossBarPattern.matcher(BossbarData.getBossbar()).find()
```
hannibal002: this line feels like duplicate code

hannibal002 (changes requested):

hannibal002 (approved):

PR #2160, opened 2024-06-26
Backend: RegexUtils deprecation and new functions

ItsEmpa:
## What
Deprecates some regexutils methods so that the reciever parameter is always either a pattern or a list of patterns, so its more consistent. Also adds a bunch of methods that will be used in an upcoming pr.

+ Deprecated some methods in RegexUtils, and added some others. - Empa


Thunderblade73:
Make a list to a sequence is "free" while a sequence to List may be expensive. So for that reason change the order for eg. matchFirst. Order in the sense that the list method calls the sequence method instead of what it is currently.

PR #2206, opened 2024-07-09
Fix: Kuudra Key Value

Thunderblade73:
## What
Current value skyhanni gives back is: "-166.992.4"
Since it is "crafted" with 2000 coins which all have a value of -1.

So two problems are there 
* Coins didn't have a value.
* The crafting evalutation allowed negativ values (and preferd those)

## Changelog Fixes
+ Fixed kuudra key value. - Thunderblade73


Thunderblade73:
Cool, it opened twice!

PR #2207, opened 2024-07-09
Fix: Powder Tracker Pattern.

superhize:
## What
Updated pattern to work with the new update

I'll open a new pr later to group everything since all the pattern follow the same logic

## Changelog Fixes
+ Fixed Powder detection from chat messages. - HiZe


raffel080108:
Trying to use the powder tracker with the version from this pull request doesn't seem to work. It doesn't even increase the chests opened amount

woost0:
The powder tracker isn't the only thing that needed updating, the chat filter also needs to be updated because the powderMining filter doesn't work with the new pattern either.

PR #2213, opened 2024-07-10
Feature: Notify when a certain player join your lobby

superhize:
## What
Receive a message when specific users join your lobby or when you join their lobby.

## Changelog New Features
+ Receive a message when a specific player join your lobby. - HiZe

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/misc/PersonOfInterestConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/misc/PersonOfInterestConfig.java
@@ -18,4 +18,19 @@ public class PersonOfInterestConfig {
     @ConfigOption(name = "Players List", desc = "Players list you want to be notified for.\n§cCase sensitive, separated by comma.")
     @ConfigEditorText
     public String playersList = "hypixel";
+
+    @Expose
+    @ConfigOption(name = "Use Prefix", desc = "Should the [SkyHanni] prefix should be in the join/leave message ?")
+    @ConfigEditorBoolean
+    public boolean usePrefix = true;
+
+    @Expose
+    @ConfigOption(name = "Join Message", desc = "Configure the message when someone join.")
+    @ConfigEditorText
+    public String joinMessage = "&b%s &ajoined your lobby.";
+
+    @Expose
+    @ConfigOption(name = "Left Message", desc = "Configure the message when someone leave.")
+    @ConfigEditorText
+    public String leftMessage = "&b%s &cleft your lobby.";
 }
--- a/src/main/java/at/hannibal2/skyhanni/config/features/misc/PersonOfInterestConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/misc/PersonOfInterestConfig.java
@@ -17,5 +17,5 @@ public class PersonOfInterestConfig {
     @Expose
     @ConfigOption(name = "Players List", desc = "Players list you want to be notified for.\n§cCase sensitive, separated by comma.")
     @ConfigEditorText
-    public String playersList = "BinRich";
+    public String playersList = "hypixel";
 }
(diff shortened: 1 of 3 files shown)


j10a1n15:
It would be so cool if an option to alert for all devs exists 

I also thought about that but was too lazy to add it myself

src/main/java/at/hannibal2/skyhanni/features/misc/PersonOfInterest.kt
```diff
@@ -0,0 +1,72 @@
...
+        loop@ for (line in event.lines) {
+            player.matchMatcher(line) {
+                val name = group("name")
+                if (name == LorenzUtils.getPlayerName()) continue@loop
+                currentLobbyPlayers.add(name)
+            }
+        }
+
+        val playerJoined = currentLobbyPlayers.filter { it in poi && it !in notifyList }.toSet()
+        val playerLeft = poi.filter { it in notifyList && it !in currentLobbyPlayers }.toSet()
```
Thunderblade73: .toSet() is not needed here. Works the same with a list.

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/PersonOfInterest.kt
```diff
@@ -0,0 +1,78 @@
...
+        if (!isEnabled()) return
+        notifyList.clear()
+        currentLobbyPlayers.clear()
+    }
+
+    @SubscribeEvent
+    fun onTablistUpdate(event: WidgetUpdateEvent) {
+        if (!isEnabled()) return
+        if (!event.isWidget(TabWidget.PLAYER_LIST)) return
+        val poi = config.playersList.split(",")
```
ItsEmpa:
```suggestion
        val poi = config.playersList.split(",").map { it.trim() }
```
also maybe it would be better to have this be a property? so you just store a set of names, and you update it on config load

ItsEmpa (changes requested):

src/main/java/at/hannibal2/skyhanni/config/features/misc/PersonOfInterestConfig.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Expose
+    @ConfigOption(name = "Use Prefix", desc = "Should the [SkyHanni] prefix should be in the join/leave message ?")
+    @ConfigEditorBoolean
+    public boolean usePrefix = true;
+
+    @Expose
+    @ConfigOption(name = "Join Message", desc = "Configure the message when someone join.")
+    @ConfigEditorText
+    public String joinMessage = "&b%s &ajoined your lobby.";
```
ItsEmpa: make it so that to use § you have to write &&, and also write both this and the fact that you have to use %s to show the playername in the config description

ItsEmpa (changes requested):

superhize:
> It would be so cool if an option to alert for all devs exists
> 
> I also thought about that but was too lazy to add it myself

I can add that based on the contributor list in the repo if everyone is ok

j10a1n15:
> > It would be so cool if an option to alert for all devs exists
> 
> > 
> 
> > I also thought about that but was too lazy to add it myself
> 
> 
> 
> I can add that based on the contributor list in the repo if everyone is ok

I don't see why people shouldn't be fine with that, you could ask in the forum or smth idk. YYou should probably default disable it though

jani270:
> > It would be so cool if an option to alert for all devs exists
> > I also thought about that but was too lazy to add it myself
> 
> I can add that based on the contributor list in the repo if everyone is ok


God please no

hannibal002:
i like the idea, but maybe this should link to shmarkplayer?

superhize:
#2653

PR #2158, opened 2024-06-26
Fix: Farming Weight Overtake ETA

evhan3584:
## PR Reviews

When your PR is marked as ready for review, some of our maintainers will look through your code to make sure everything is good to go. In order to do this, they may request some changes you will need to do, **or fix smaller stuff (like merge conflicts) for you**. If a maintainer has reviewed your PR, make sure to **pull any of their changes** into your local project before doing more work on your code. Having maintainers fix small stuff for you helps us speed up the process of merging your PR, so if some of your systems warrant further care, be sure to let us know (preferably with a code comment).

Make sure to only mark your PR as "Ready to review" when it is. If you still want to do major changes, you can keep a draft PR open until then.

## What
This is a simple fix to the Farming Weight Overtake ETA.

## Changelog Fixes
+ Fix an issue where the farming overtake ETA was being displayed as 60 times longer than it actually was. - Evhan_

PR #2217, opened 2024-07-11
Fix: Farming ETA Duration "NaN"

DavidArthurCole:
## What
Re: #2158, which had a suggested but not-implemented fix for this issue, fix the `NaN` issue that occasionally appears when starting/stopping farming or killing pests. This was due to `diff` being 0, causing a div-by-0 error.

## Changelog Fixes
+ Fixed an error in the Farming ETA Duration calculation. - Daveed


hannibal002:
I think checking if diff is 0, and if so, block `addCrop` is a better solution

DavidArthurCole:
> I think checking if diff is 0, and if so, block `addCrop` is a better solution

Yeah, that's probably better. Done in the latest commit.

PR #2225, opened 2024-07-16
Fix: Fish The Rabbit Production Time in Lore

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1262570859470651412
Fix Fish the Rabbit not having production time added to lore, due to having different formatting.

<details>
<summary>Images</summary>

image

</details>

## Changelog Fixes
+ Fixed the detection of production time from Fish the Rabbit. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryTooltipStray.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryTooltipStray.kt
@@ -20,7 +20,7 @@ object ChocolateFactoryTooltipStray {
      */
     private val chocolateGainedPattern by ChocolateFactoryAPI.patternGroup.pattern(
         "rabbit.stray",
-        "(?:§.)?+(?:Rabbit§7, so )?(?:[Yy]ou )?(?:gained |received )?§6\\+?(?<amount>[\\d,]+)(?:\$| Chocolate§7!)"
+        "(?:§.)?+(?:Rabbit§7, so )?(?:[Yy]ou )?(?:gained |received )?§6\\+?(?<amount>[\\d,]+)(?:$| Chocolate§7!)"
     )
 
     @SubscribeEvent(priority = EventPriority.HIGH)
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryTooltipStray.kt
```diff
@@ -16,10 +16,11 @@ object ChocolateFactoryTooltipStray {
      * REGEX-TEST: §7You gained §6+2,465,018 Chocolate§7!
      * REGEX-TEST: §7gained §6+30,292 Chocolate§7!
      * REGEX-TEST: §7§6+36,330 Chocolate§7!
+     * REGEX-TEST: §9Rabbit§7, so you received §655,935,257
      */
     private val chocolateGainedPattern by ChocolateFactoryAPI.patternGroup.pattern(
         "rabbit.stray",
-        "(?:§.)+(?:You )?(?:gained )?§6\\+(?<amount>[\\d,]+) Chocolate§7!"
+        "(?:§.)?+(?:Rabbit§7, so )?(?:[Yy]ou )?(?:gained |received )?§6\\+?(?<amount>[\\d,]+)(?:\$| Chocolate§7!)"
```
CalMWolfs: pretty sure you shouldnt be escapign the $ here

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryTooltipStray.kt
DavidArthurCole: It's being string-escaped, not truly escaped in the pattern itself. It getting prepended at all was a result of pasting the regex in from r101, if I had to guess.

    private val p1 by patternGroup.pattern(
        "testpatterngroup.pattern",
        "(?:§.)?+(?:Rabbit§7, so )?(?:[Yy]ou )?(?:gained |received )?§6\\+?(?<amount>[\\d,]+)(?:\$| Chocolate§7!)"
    )
    private val p2 by patternGroup.pattern(
        "testpatterngroup.pattern2",
        "(?:§.)?+(?:Rabbit§7, so )?(?:[Yy]ou )?(?:gained |received )?§6\\+?(?<amount>[\\d,]+)(?:$| Chocolate§7!)"
    )
    
        val testString = "§9Rabbit§7, so you received §655,935,257";
        ChatUtils.chat("P1 match? " + (p1.matches(testString).toString()) + "P2 match? " + (p2.matches(testString).toString()))

image

CalMWolfs: id still prefer if we use the pattern that does match in regex 101 as that is the kinda the point of the regex tests

CalMWolfs (approved):

PR #2228, opened 2024-07-16
Improvement: Apply border radius to scoreboard texture

ThatGravyBoat:
## What
This makes it so that the rounded corner smoothness is applied to the custom background texture.

<details>
<summary>Images</summary>

</details>

## Changelog Improvements
+ Rounded corner smoothness now applies to custom Scoreboard textures. - ThatGravyBoat


Thunderblade73 (approved):
Addition works. Code is cleaner than before 😀. Shader magic was added (I still can't read it 😅).
I also like the addition very much. Makes the scoreboard way cleaner with a texture.

PR #2239, opened 2024-07-20
Fix: Stray Rabbit Production Time Tooltip

DavidArthurCole:
## What
Leading formatting of tooltips was causing the pattern to not match.

## Changelog Fixes
+ Fixed stray rabbit production time tooltip display. - Daveed


hannibal002 (changes requested):
The regex tests are working for both, the current version and your new version. What is the change supposed to fix? Is it possible you can add new regex tests that break in the current version and only work with the new version?

DavidArthurCole:
> The regex tests are working for both, the current version and your new version. What is the change supposed to fix? Is it possible you can add new regex tests that break in the current version and only work with the new version?

I've fixed the regex tests - they were missing the lore line formatting (didn't match on the patterns from #2225, they do match the new pattern)

PR #2243, opened 2024-07-20
Feature: Accessory MP as stack size

minhperry:
## What
https://discord.com/channels/997079228510117908/1263951633059156009

<details>
<summary>Images</summary>

image

image

image
</details>

## Changelog New Features
+ Added Accessory magical power display as stack size. - minhperry
    * Only works inside the Accessory Bag and Auction House.


src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,101 @@
...
+                    "LEGENDARY" -> return AccessoryRarity.LEGENDARY
+                    "MYTHIC" -> return AccessoryRarity.MYTHIC
+                    else -> return@matchMatcher null
+                }
+            }
+        }
+        return null
+    }
+
+    private enum class AccessoryRarity(val code: String) {
```
hannibal002: LorenzRarity already exists.

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,101 @@
...
+    private val MPMap = mapOf(
+        AccessoryRarity.COMMON to 3, AccessoryRarity.SPECIAL to 3,
+        AccessoryRarity.UNCOMMON to 5, AccessoryRarity.VERY_SPECIAL to 5,
+        AccessoryRarity.RARE to 8,
+        AccessoryRarity.EPIC to 12,
+        AccessoryRarity.LEGENDARY to 16,
+        AccessoryRarity.MYTHIC to 22
+    )
+
+    private val acceptedInvPattern by RepoPattern.pattern(
```
hannibal002, line 46: add regex tests

src/main/java/at/hannibal2/skyhanni/config/features/misc/MPDisplayConfig.java
```diff
@@ -0,0 +1,20 @@
+package at.hannibal2.skyhanni.config.features.misc;
+
+import at.hannibal2.skyhanni.config.FeatureToggle;
+import com.google.gson.annotations.Expose;
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorBoolean;
+import io.github.notenoughupdates.moulconfig.annotations.ConfigOption;
+
+public class MPDisplayConfig {
+    @Expose
+    @ConfigOption(name = "Enable MP Display", desc = "Show Magical Power as stack size.")
```
hannibal002: more expressive explanation.

src/main/java/at/hannibal2/skyhanni/config/features/misc/MiscConfig.java
```diff
@@ -283,4 +283,9 @@ public class MiscConfig {
     @ConfigEditorBoolean
     @FeatureToggle
     public boolean maintainGameVolume = false;
+
+    @Expose
+    @ConfigOption(name = "Magical Power Display", desc = "")
+    @Accordion
+    public MPDisplayConfig mpDisplay = new MPDisplayConfig();
```
hannibal002: move into inventory category, rename the variable and class name to MagicalPower

src/main/java/at/hannibal2/skyhanni/config/features/misc/MPDisplayConfig.java
```diff
@@ -0,0 +1,20 @@
...
+    @ConfigEditorBoolean
+    @FeatureToggle
+    public boolean enabled = false;
+
+    @Expose
+    @ConfigOption(name = "Colored", desc = "Whether to make the numbers colored.")
+    @ConfigEditorBoolean
+    @FeatureToggle
```
hannibal002: this is no feature toggle

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,101 @@
...
+            endMP *= 2
+        }
+        if (itemName == "Rift Prism") {
+            endMP = 11
+        }
+
+        event.stackTip = "${if (colored) rarity else "§7"}${endMP}"
+    }
+
+    private fun ItemStack.isSkyblockAccessory(): AccessoryRarity? {
```
hannibal002: `ItemStack.getItemRarityOrNull()` already exists

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,101 @@
...
+        AccessoryRarity.MYTHIC to 22
+    )
+
+    private val acceptedInvPattern by RepoPattern.pattern(
+        "inv.acceptable",
+        """(Accessory Bag \(\d+/\d+\)|Auctions Browser)"""
+    )
+    private val accessoryLorePattern by RepoPattern.pattern(
+        "accessory.lore",
+        """a?\s*(COMMON|UNCOMMON|RARE|EPIC|LEGENDARY|MYTHIC)\s*(DUNGEON\s*)?ACCESSORY\s*a?"""
```
hannibal002: not necessary

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,101 @@
...
+import at.hannibal2.skyhanni.utils.RegexUtils.matchMatcher
+import at.hannibal2.skyhanni.utils.RegexUtils.matches
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+import io.github.moulberry.notenoughupdates.util.stripControlCodes
+import net.minecraft.item.ItemStack
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+@SkyHanniModule
+object MagicalPowerDisplay {
+    private val config get() = SkyHanniMod.feature.misc.mpDisplay.enabled
```
hannibal002: make this variable only beSkyHanniMod.feature.misc.mpDisplay, not enabled. remove the variable below

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,101 @@
...
+    @SubscribeEvent
+    fun onRenderItemTip(event: RenderItemTipEvent) {
+        if (!isEnabled()) return
+        if (!acceptedInvPattern.matches(InventoryUtils.openInventoryName().stripControlCodes())) return
+
+        val item = event.stack
+        val rarity = item.isSkyblockAccessory() ?: return
+        val itemName = item.displayName.stripControlCodes()
+
+        var endMP = MPMap[rarity]!!
```
hannibal002: avoid using `!!` in Kotlin. Rather, handle the "unknown rarity" case by showing a user error via `ErrorManager`

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,101 @@
...
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+import io.github.moulberry.notenoughupdates.util.stripControlCodes
+import net.minecraft.item.ItemStack
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+@SkyHanniModule
+object MagicalPowerDisplay {
+    private val config get() = SkyHanniMod.feature.misc.mpDisplay.enabled
+    private val colored get() = SkyHanniMod.feature.misc.mpDisplay.colored
+    private fun isEnabled() = config && LorenzUtils.inSkyBlock
```
hannibal002: please first check if in skyblock, then check the config. also move this function at the very bottom of the class for consistency

hannibal002 (changes requested):
Thanks for this pr! the code looks good overall, still needs some polishing.

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,101 @@
...
+    fun onRenderItemTip(event: RenderItemTipEvent) {
+        if (!isEnabled()) return
+        if (!acceptedInvPattern.matches(InventoryUtils.openInventoryName().stripControlCodes())) return
+
+        val item = event.stack
+        val rarity = item.isSkyblockAccessory() ?: return
+        val itemName = item.displayName.stripControlCodes()
+
+        var endMP = MPMap[rarity]!!
+        if (itemName == "Hegemony Artifact") {
```
lunaynx: It's better to check the item ID. Someone may still have an old reforged accessory.

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,101 @@
...
+            endMP *= 2
+        }
+        if (itemName == "Rift Prism") {
+            endMP = 11
```
lunaynx: Same here (EDIT: one line above, but you get the idea).

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
minhperry: but then how would I know if it's an accessory? NBT does not have any info

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,101 @@
...
+import net.minecraft.item.ItemStack
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+@SkyHanniModule
+object MagicalPowerDisplay {
+    private val config get() = SkyHanniMod.feature.misc.mpDisplay.enabled
+    private val colored get() = SkyHanniMod.feature.misc.mpDisplay.colored
+    private fun isEnabled() = config && LorenzUtils.inSkyBlock
+
+    private val MPMap = mapOf(
```
j10a1n15: why create this when you can add the mp into the enum directly

j10a1n15 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
hannibal002: Do we want to show the numbers outside of the accessory bag as well? In that case, we need a check like this, yes.

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
minhperry: DIVINE, ULTIMATE and SUPREME does not have any MP data

lunaynx: I'm pretty sure Supreme rarity doesn't even exist at all anymore.

Neither do Divine or Ultimate accessories exist. (Ultimate probably never will, as it's reserved for cosmetics.)

hannibal002: > DIVINE, ULTIMATE and SUPREME does not have any MP data

then map those to 0

minhperry: MP values are only used for this feature so changing it in the enum itself "poisons" the enum.
I don't think it's that necesary.

ThatGravyBoat: I think if you don't want the pollute the enum it would be better to make this a function with a when statement so that the explicit not null is not needed on the getter of the map.

j10a1n15: I wrote this before seeing hanni say to remove the AccessoryRarity enum
So this review is obsolete yes

hannibal002: Function > map imo too

minhperry: > Function > map imo too

Make sense. Will change later.

minhperry:
adding support for abicases for now

minhperry:
for some reasons the "Auctions: *" regex would just never pass despite passing IJ tester

image

minhperry:
image

Is this intended it just spams my email with this

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,128 @@
...
+            } else if (stripped == "a VERY SPECIAL HATCESSORY a") {
+                return LorenzRarity.VERY_SPECIAL
+            }
+
+            if (accessoryLorePattern.matches(stripped)) {
+                return this.getItemRarityOrNull()
+            }
+        }
+        return null
+    }
```
Thunderblade73: We have ItemCategory for that,
Simply use 
````suggestion
private fun ItemStack.isAccessory(): LorenzRarity?{
   if(this.itemCategoryOrNull() != ItemCategory.ACCESSORY && this.itemCategoryOrNull() != ItemCategory.HATCESSORY) return null
   return this.getItemRarityOrNull()
}
```
The syntax might be a bit off. I did just type that out of my head without checking if the names where correct.
If the itemrarity does not work for special and very special let me know. Since that should be fixed in the rarity detection code.

Thunderblade73 (changes requested):
Sorry, I late to this. But what the isAccessory function does should already be done by the itemCategory code. If it isn't for some wierd reason / edge case, please let me know and I will fix it.

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,106 @@
...
+            ErrorManager.skyHanniError(
+                "Unknown rarity '$rarity' for item '${item.displayName}§7'"
+            )
+        }
+        if (itemID == "HEGEMONY_ARTIFACT".asInternalName())
+            endMP *= 2
+        if (itemID == "RIFT_PRISM".asInternalName())
+            endMP = 11
+        if (item.isAbicase())
+            endMP += contactAmount / 2
```
Thunderblade73: Save the InternalNames above. Saves a bit of resources (.asInternalName creates a new object every time)
```Kotlin
    private val hegemonyArtifact = "HEGEMONY_ARTIFACT".asInternalName()
    private val riftPrism = "RIFT_PRISM".asInternalName()
```
Also a bit cleaner with a when. I also don't know what to do with the last if, but this way it should be fine.
```suggestion
       when(itemID){
            hegemonyArtifact -> endMP *= 2
            riftPrism -> endMP = 11
            else -> if (item.isAbicase()) {
                endMP += contactAmount / 2
            }
        }
```

src/main/java/at/hannibal2/skyhanni/config/features/inventory/MagicalPowerConfig.java
```diff
@@ -0,0 +1,19 @@
...
+    @Expose
+    @ConfigOption(name = "Magical Power Display", desc = "Show Magical Power as stack size inside Accessory Bag and Auction House.")
+    @ConfigEditorBoolean
+    @FeatureToggle
+    public boolean enabled = false;
+
+    @Expose
+    @ConfigOption(name = "Colored", desc = "Whether to make the numbers colored.")
+    @ConfigEditorBoolean
+    public boolean colored = false;
```
Thunderblade73, line 18: We use the spelling "colour" not "color".
You also need to change it in at the places where it is used,
```suggestion
    @Expose
    @ConfigOption(name = "Coloured", desc = "Whether to make the numbers coloured.")
    @ConfigEditorBoolean
    public boolean coloured = false;
```

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,106 @@
...
+import at.hannibal2.skyhanni.utils.ItemUtils.getInternalNameOrNull
+import at.hannibal2.skyhanni.utils.ItemUtils.getItemCategoryOrNull
+import at.hannibal2.skyhanni.utils.ItemUtils.getItemRarityOrNull
+import at.hannibal2.skyhanni.utils.ItemUtils.getLore
+import at.hannibal2.skyhanni.utils.LorenzRarity
+import at.hannibal2.skyhanni.utils.LorenzUtils
+import at.hannibal2.skyhanni.utils.NEUInternalName.Companion.asInternalName
+import at.hannibal2.skyhanni.utils.RegexUtils.matches
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+import io.github.moulberry.notenoughupdates.util.stripControlCodes
```
Thunderblade73: Don't use neu's code here. We want to be as independet as possible, since we plan to remove the dependecy.
Also we simply have .removeColor(). Which does the same but better (thx nea).

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,106 @@
...
+        return when (this) {
+            LorenzRarity.COMMON, LorenzRarity.SPECIAL -> 3
+            LorenzRarity.UNCOMMON, LorenzRarity.VERY_SPECIAL -> 5
+            LorenzRarity.RARE -> 8
+            LorenzRarity.EPIC -> 12
+            LorenzRarity.LEGENDARY -> 16
+            LorenzRarity.MYTHIC -> 22
+            else -> null
+        }
+    }
```
Thunderblade73: A better syntax for this function. This way one identation is gone (And we love low identation)
```suggestion
    private fun LorenzRarity.toMP(): Int?  = when (this) {
        LorenzRarity.COMMON, LorenzRarity.SPECIAL -> 3
        LorenzRarity.UNCOMMON, LorenzRarity.VERY_SPECIAL -> 5
        LorenzRarity.RARE -> 8
        LorenzRarity.EPIC -> 12
        LorenzRarity.LEGENDARY -> 16
        LorenzRarity.MYTHIC -> 22
        else -> null
    }
```

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,106 @@
...
+            endMP *= 2
+        if (itemID == "RIFT_PRISM".asInternalName())
+            endMP = 11
+        if (item.isAbicase())
+            endMP += contactAmount / 2
+        event.stackTip = "${if (config.colored) rarity.chatColorCode else "§7"}${endMP}"
+    }
+
+    @SubscribeEvent
+    fun onInventoryOpened(event: InventoryOpenEvent) {
```
Thunderblade73: InventoryFullyOpenedEvent does the job here. No need to also have the inventory update event here. Since stuff only changes the time a inventory is opened, updating does not add anything to this kind of gui.
```suggestion
    fun onInventoryOpened(event: InventoryFullyOpenedEvent) {
```

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,106 @@
...
+        if (!isEnabled()) return
+        if (event.inventoryName.startsWith("Abiphone")) {
```
Thunderblade73: Use a repoPattern instead.

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,106 @@
...
+            val theBookLore = event.inventoryItems[51]?.getLore() ?: return
+            for (line in theBookLore) {
+                val stripped = line.stripControlCodes()
+                if (stripped.startsWith("Your contacts: ")) {
+                    contactAmount = stripped.split(" ")[2].split("/")[0].toInt()
+                    return
+                }
```
Thunderblade73: Use a repoPattern with a group here instead of this. 
Hint: matchMatcher

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,106 @@
...
+            }
+        }
+    }
+
+    private fun ItemStack.isAbicase(): Boolean {
+        val id = this.getInternalNameOrNull() ?: return false
+        return id.startsWith("ABICASE_")
+    }
```
Thunderblade73: You have the internalName. So now need to refetch the name, sure it is bascily free (fetching is cached for most item stuff. Eg: InternalName, Rarity, ItemCategory, ...), but also not needed,
```suggestion
    private fun NeuInternalName.isAbicase(): Boolean = id.startsWith("ABICASE_")
```

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,106 @@
...
+
+        val item = event.stack
+        val rarity = item.isAccessory() ?: return
+        val itemID = item.getInternalNameOrNull() ?: return
+
+        var endMP = rarity.toMP() ?: run {
+            ErrorManager.skyHanniError(
+                "Unknown rarity '$rarity' for item '${item.displayName}§7'"
+            )
+        }
```
Thunderblade73: There is no need to warp it into a run.
```suggestion
        var endMP = rarity.toMP() ?: ErrorManager.skyHanniError(
            "Unknown rarity '$rarity' for item '${item.displayName}§7'",
        )
```

Thunderblade73 (changes requested):
Code structure is good. But you missed some stuff that is mostly unqiue to our repo (eg: pattern for mostly everything). I also added some formatting stuff that makes the code look better.

src/main/java/at/hannibal2/skyhanni/config/features/inventory/MagicalPowerConfig.java
Thunderblade73: I take that back. I guess hannibal changed his mind about that. Leave it as is.

hannibal002: With the planned gb/us English setting, the words in the config and the search doesn't matter for the user-facing part. In the backend, I still prefer color over colour, for consistency:
image
image

hannibal002: Related info: https://github.com/hannibal002/SkyHanni/pull/2258

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
minhperry: Since Contact Directory is only available at XI Ultra Style and above I would need at least 5 patterns for

- 6 Ultra Style
- 7 Mega (Color)
- 8 Pro (Giga)
- Enormous
- Flip

hannibal002: you can all that put into one single regex (`Abiphone .*`)

minhperry: But then the lower ones does not have and yet still get accepted

Thunderblade73: But how does it differ from your current approach?

hannibal002: Maybe we have missed something. The existing code is filled with string operations (startsWith, endsWith, contains, ==) on data we directly get from Hypixel. For maintaining compatibility with changed formatting on Hypixel's end, we want to change all those cases with regex via repo pattern. Since it's a ton of work to change the existing code at once, we start by trying to get rid of those patterns in new PRs as much as possible.

If you find an open pr where this is still the case, feel free to comment on it under.

hannibal002: Wait are you talking about different abiphone names and not other prs?
in that case, please tell me what names are wrongly accepted and we can find a solution together

hannibal002: In either case, the regex is checking the same thing as your startswith, so the existing logic has the same problems aready and this has nothing to do with using repo patterns over string operations

minhperry: Ok nvm I can just let through Abiphone and with current approach the book is checked instead. For lower type of phone the book will be null.

src/main/java/at/hannibal2/skyhanni/config/features/misc/MiscConfig.java
```diff
@@ -288,4 +288,6 @@ public class MiscConfig {
     @ConfigEditorBoolean
     @FeatureToggle
     public boolean maintainGameVolume = false;
+
```
hannibal002: remove those empty lines

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,108 @@
...
+        LorenzRarity.COMMON, LorenzRarity.SPECIAL -> 3
+        LorenzRarity.UNCOMMON, LorenzRarity.VERY_SPECIAL -> 5
+        LorenzRarity.RARE -> 8
+        LorenzRarity.EPIC -> 12
+        LorenzRarity.LEGENDARY -> 16
+        LorenzRarity.MYTHIC -> 22
+        else -> null
+    }
+
+    private fun ItemStack.isAccessory(): LorenzRarity? {
```
hannibal002: this function name is confusing

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,108 @@
...
+            else -> if (itemID.isAbicase()) endMP += contactAmount / 2
+        }
+
+        event.stackTip = "${if (config.colored) rarity.chatColorCode else "§7"}${endMP}"
+    }
+
+    @SubscribeEvent
+    fun onInventoryOpened(event: InventoryFullyOpenedEvent) {
+        if (!isEnabled()) return
+        if (event.inventoryName.startsWith("Abiphone")) {
```
hannibal002: please use early return and repo patterns.

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
```diff
@@ -0,0 +1,108 @@
...
+import at.hannibal2.skyhanni.utils.RegexUtils.matches
+import at.hannibal2.skyhanni.utils.StringUtils.removeColor
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+import net.minecraft.item.ItemStack
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+@SkyHanniModule
+object MagicalPowerDisplay {
+    private val config get() = SkyHanniMod.feature.inventory.magicalPower
+    private var contactAmount: Int
```
hannibal002: please dont use -1, let it return null and then handle the nullable case at use.

src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.java
```diff
@@ -659,4 +659,7 @@ public static class WardrobeStorage {
 
     @Expose
     public UpgradeReminder.CommunityShopUpgrade communityShopProfileUpgrade = null;
+
+    @Expose
+    public int abiphoneContactAmount = -1;
```
hannibal002: please init as 0, not -1. we never check against -1 in code, so this adds unnecessary complexity.

hannibal002 (changes requested):
as Thunder said, please use repoPattern.

minhperry:
I will probably not be available from now until around 8th August

src/main/java/at/hannibal2/skyhanni/features/misc/MagicalPowerDisplay.kt
minhperry: couldn't i just default 0 instead?

hannibal002: There is no benefit of using 0 over null. But kotlin has string nulllability support, making it easier to use null instead of 0 or -1

minhperry: So somewhat like this?

```kt
private var contactAmount: Int?
        get() = ProfileStorageData.profileSpecific?.abiphoneContactAmount ?: null
        private set(value) {
            ProfileStorageData.profileSpecific?.abiphoneContactAmount = value ?: 0
        }
```

hannibal002: Not quite. you can remove the `?: ..` at the end for both. for get() the `?: null` at the end is just rendundant code as this already returns null if null. and for set(), we actually want to store the value, even if it is null, inside the porofile specific space.

minhperry: should be done

hannibal002 (approved):

PR #2248, opened 2024-07-21
Fix: Dicer Tracker Reset Button

hannibal002:
## What
Fixed Dicer Tracker Reset Button being misaligned.
~~Reported: https://discord.com/channels/997079228510117908/1250100311108882433~~ (This is another bug)

## Changelog Fixes
+ Fixed Dicer Tracker Reset Button misalignment. - hannibal2


Thunderblade73 (changes requested):
Fix did not work. Also see no reason why it should work.
image

Thunderblade73:
The problem is here: [SkyHanniTracker L91](https://github.com/hannibal002/SkyHanni/blob/beta/src/main/java/at/hannibal2/skyhanni/utils/tracker/SkyHanniTracker.kt#L91)
renderStringsAndItems does not work with elements that have a y size that is different from 10

hannibal002:
This pr does not fix the bug linked to the discord report, but rather fixes another bug. (see changelog text)

PR #2257, opened 2024-07-25
Fix: Jasper Gemstone Color

lunaynx:
## What
Fixed the check for Jasper gemstones using pink color instead of magenta.

<details>
<summary>Images</summary>

image
</details>

+ Fixed wrong color for Jasper gemstone in OreBlock API. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/mining/OreBlock.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/mining/OreBlock.kt
@@ -176,7 +176,7 @@ enum class OreBlock(
         checkArea = { inCrystalHollows || inGlacite },
     ),
     JASPER(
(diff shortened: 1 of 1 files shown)


Thunderblade73:
Could you validated if that is also the case in the mineshafts? Since I remember that they glass color for jasper was strange. Since it is very much possible that hypixel messed the color up and the feature got it right for the mineshafts, since that was the original intendet purpose.

lunaynx:
I'll try, but it may take me a while to get a Jasper mineshaft.

lunaynx:
Confirmed it's the same in mineshafts.

Thunderblade73 (approved):

ItsEmpa (approved):

PR #2263, opened 2024-07-26
Fix: HotM total powder being reset

lunaynx:
## What
Fixed HotmAPI not checking the inventory name causing total powder to reset when opening the SkyBlock Menu on a mining island.

## Changelog Fixes
+ Fixed total powder amounts resetting when opening the SkyBlock Menu on a mining island. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/HotmData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/HotmData.kt
@@ -673,7 +673,8 @@ enum class HotmData(
         @SubscribeEvent
         fun onInventoryFullyOpen(event: InventoryFullyOpenedEvent) {
             if (!LorenzUtils.inSkyBlock) return
-            if (!inventoryPattern.matches(event.inventoryName)) return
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/data/HotmData.kt
```diff
@@ -673,7 +673,7 @@ enum class HotmData(
         @SubscribeEvent
         fun onInventoryFullyOpen(event: InventoryFullyOpenedEvent) {
             if (!LorenzUtils.inSkyBlock) return
-            inInventory = inventoryPattern.matches(event.inventoryName)
```
Thunderblade73, line 676: The inInventory Variable is Still needed

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/HotmData.kt
Thunderblade73: The value is used in HotmFeatures, if you wonder.

Thunderblade73 (approved):

PR #2277, opened 2024-07-30
Fix: TunnelMaps Deadlock

Thunderblade73:
## What
I forgot to reset the goalReached which leads to a deadlock since 

## Changelog Fixes
+ Fixed an issue where TunnelMaps could stop functioning. - Thunderblade73


hannibal002:
Why is there no function reset() ?

PR #2281, opened 2024-07-31
Fix: Flawless Gemstones in Powder Filter

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1268045402394067114
Fixed the fact that Flawless gemstones were not included in the pattern.

## Changelog Fixes
+ Fixed Flawless Gemstones not being formatted correctly when the Powder Mining filter is enabled. - Daveed


DavidArthurCole:
The enum `FINE_ONLY` needs to get renamed to `FINE_UP` and mapped to the new value before this is ready

PR #2303, opened 2024-08-04
Fix: Clicking on weight pass chat message

not-a-cow:
## What
Fixed clicking on a passed player in chat sending a chat message instead of a command due to a missing /

## Changelog Fixes
+ Fixed command sent in chat for Farming Weight Display - not-a-cow


CalMWolfs (changes requested):
code looks fine, but you have not formatted the pr description (and therefore the changelog) properly

PR #2306, opened 2024-08-04
Feature: Corpse Tracker

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1230761536176521249
Adds a source-addressable tracker for Glacite Mineshaft Corpses.

<details>
<summary>Images</summary>

Due to not having any actual loot strings for corpses, the tracker is displaying Nucleus run loot, but 🤷

All Loot Buckets (sources), will merge loot from all buckets
image

"Bucket" (source) selectors, similar to the Display Mode toggles, that will only appear while inventory.
image

With a "Bucket" selected, only display loot that came from that 'type'
image

</details>

## Changelog New Features
+ Added Tracker for Glacite Corpses. - Daveed
    * Tracks overall loot and loot per type.

+ Added an abstract "bucketed" tracker that can store where items came from. - Daveed
    * Takes an enum in instantiation and has built-in logic for swapping display types.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/CorpseTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/CorpseTracker.kt
@@ -51,8 +51,8 @@ object CorpseTracker {
             )
         }
 
-        override fun getCoinName(item: TrackedItem) = "<no coins>"
-        override fun getCoinDescription(item: TrackedItem): List<String> = listOf("<no coins>")
+        override fun getCoinName(bucket: CorpseType?, item: TrackedItem) = "<no coins>"
+        override fun getCoinDescription(bucket: CorpseType?, item: TrackedItem): List<String> = listOf("<no coins>")
 
         @Expose
         var corpsesLooted: MutableMap<CorpseType, Long> = EnumMap(CorpseType::class.java)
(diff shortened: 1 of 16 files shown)


src/main/java/at/hannibal2/skyhanni/utils/tracker/BucketedItemTrackerData.kt
```diff
@@ -0,0 +1,78 @@
...
+
+abstract class BucketedItemTrackerData<E : Enum<E>> : TrackerData() {
+
+    private val config get() = SkyHanniMod.feature.misc.tracker
+
+    abstract fun resetItems()
+
+    abstract fun getDescription(timesGained: Long): List<String>
+
+    abstract fun getCoinName(item: TrackedItem): String
```
hannibal002: i dont think coins and coin description belongs into this bucketed data

src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/CorpseTracker.kt
```diff
@@ -0,0 +1,142 @@
...
+                tracker.modify {
+                    it.addItem(event.corpseType, item, amount)
+                }
+            }
+        }
+    }
+
+    private fun drawDisplay(bucketData: BucketData): List<Searchable> = buildList {
+        addSearchString("§b§lGlacite Corpse Profit Tracker")
+        tracker.addBucketSelectorsReified<CorpseType>(this, bucketData, "Corpse Type")
```
hannibal002: No need to include bucketData in this function. the tracker has the data already

src/main/java/at/hannibal2/skyhanni/utils/tracker/SkyHanniBucketedItemTracker.kt
```diff
@@ -0,0 +1,214 @@
...
+        }
+    }
+
+    inline fun <reified F : Enum<F>> addBucketSelectorsReified(
+        lists: MutableList<Searchable>,
+        data: BucketedData,
+        sourceString: String = "",
+    ) {
+        if (isInventoryOpen()) {
+            val sourceStringPrefix = sourceString.takeIf { it.isNotEmpty() } ?: "Item Source"
```
hannibal002: have sourceStringPrefix as a constructor parameter to the class instead. also rather throw an error in dev env when the string is empty isntead of showing a generic name

src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/CorpseTracker.kt
```diff
@@ -0,0 +1,142 @@
...
+            NEUInternalName.fromItemNameOrNull(itemName)?.let { item ->
+                tracker.modify {
+                    it.addItem(event.corpseType, item, amount)
+                }
+            }
+        }
+    }
+
+    private fun drawDisplay(bucketData: BucketData): List<Searchable> = buildList {
+        addSearchString("§b§lGlacite Corpse Profit Tracker")
```
hannibal002: can we use the name "mineshaft corpse" instead of glacite corpse"?

src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/CorpseTracker.kt
```diff
@@ -0,0 +1,142 @@
...
+        tracker.addBucketSelectorsReified<CorpseType>(this, bucketData, "Corpse Type")
+
+        var profit = tracker.drawItems(bucketData, { true }, this)
+
+        if (bucketData.getCorpseCount() > 0) {
```
hannibal002: use early return (no need to show the price button if there is nothing to show)

hannibal002 (changes requested):
works fine in game (got 1.5b profit while testing lol)
couple smaller problems, not related to corpse but bucket:
the color of the bracket is wrong:
image

there is no option to select "no ´corpse type selected, snow all instead".

please make it a "click to switch through options" selection, not a list. like in fishing profit tracker.
per default hide empty brackets from this list.

src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/CorpseTracker.kt
DavidArthurCole: All `Tracker...`s' `drawDisplay` methods take their `data: Data` object as an input - don't know why this would be different. The bucketed data does exist in the tracker, but not in an accessible manner - `drawDisplay(it)` is common across all others, don't think this needs changing.

src/main/java/at/hannibal2/skyhanni/utils/tracker/BucketedItemTrackerData.kt
DavidArthurCole: Would also challenge this one - although the `CorpseTracker` doesn't have use for `addCoins()`, if we take the example of something like the `SlayerTracker` which _could_ be morphed into `BucketedItemTracker` use, there is definitely a use case.

I will update it though to be bucket-specific, as I could see having different names per bucket.

DavidArthurCole:
> works fine in game (got 1.5b profit while testing lol) couple smaller problems, not related to corpse but bucket: the color of the bracket is wrong: image
> 
> there is no option to select "no ´corpse type selected, snow all instead".
> 
> please make it a "click to switch through options" selection, not a list. like in fishing profit tracker. per default hide empty brackets from this list.

bracket color being wrong was fixed in [ab6d21a](https://github.com/hannibal002/SkyHanni/pull/2306/commits/ab6d21a63915da45a675bf361873eca563b178b8)

None option was added in [16bcf3f](https://github.com/hannibal002/SkyHanni/pull/2306/commits/16bcf3fcbfcff10620bfb4eedf92b758e3e2b158) - but was done as a list, should be fixed now.

DavidArthurCole:
gif

PR #2311, opened 2024-08-05
Feature + Fix: Hoppity Event Summarization

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1262558938650378250
Adds a recap for Hoppity Events, added an event that gets fired when a rabbit is collected. Reworked the `HoppityEggType` enum. Tangentially related to the enum, this PR also adds functionality for Hoppity Compact to compact:
- Side Dish rabbits
- All-Time Chocolate Milestone rabbits
- Shop milestone rabbits

<details>
<summary>Images</summary>

image

image

image

</details>

## Changelog New Features
+ Added Hoppity Hunt event summary. - Daveed
    * Use /shhoppitystats for live stats.

## Changelog Fixes
+ Fixed "Side Dish" and Milestone rabbits not sending a message when Compact Hoppity is enabled. - Daveed

+ Added `HoppityAPI`, which should house event-specific features rather than `ChocolateFactoryAPI`. - Daveed
    * Added an `Always Hoppity` dev config option, which will override `isHoppityEvent`.
+ Added `RabbitFoundEvent`, which will be fired when a chocolate rabbit is found from an egg, bought from Hoppity, or claimed from a milestone. - Daveed
+ Added `SIDE_DISH`, `CHOCOLATE_SHOP_MILESTONE`, `CHOCOLATE_FACTORY_MILESTONE`, and `BOUGHT` as `HoppityEggType` options. - Daveed


DavidArthurCole:
Couple bugs I need to fix still:
- Leaderboard change needs to be removed, it's needlessly complicating the entire module
- Found eggs divisor is 0 for historical data, should default to 279 for past data

src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityAPI.kt
```diff
@@ -0,0 +1,170 @@
...
+        val nameText = (if (clickedStack.hasDisplayName()) clickedStack.displayName else clickedStack.itemName)
+
+        milestoneNamePattern.matchMatcher(nameText) {
+            val itemLore = clickedStack.getLore()
+            if (!itemLore.any { it == "§eClick to claim!" }) return
+
+            // Will never match both all time and shop patterns together
+            allTimeLorePattern.firstMatcher(clickedStack.getLore()) {
+                LorenzChatEvent(
+                    "§d§lHOPPITY'S HUNT §r§dYou claimed a §r§6§lChocolate Milestone Rabbit §r§din the Chocolate Factory§r§d!",
```
hannibal002, line 104: why do we send the chat message here? if we want to impact code in other spots, please use functions/events.

src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityRabbitRarity.kt
```diff
@@ -0,0 +1,18 @@
+package at.hannibal2.skyhanni.features.event.hoppity
+
+enum class HoppityRabbitRarity(val displayName: String, val colorCode: String) {
```
hannibal002: is there a reason for this map? LorenzRarity should get used here. the two special functions below can both move into their classes where they are used

hannibal002 (changes requested):
works fine so far, only small things

src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityAPI.kt
DavidArthurCole: This was my way to leave the Hoppity chat compacting and the other chat parsers intact while still firing the new data - I didn't see a clean way to interface with both of the existing solutions at once while also using event fires. I can look into it more, but this was the functional way I was able to get both modules interfacing with the API cleanly.

hannibal002: I see. For now, the solution is fine, but when we want to expand the Hoppity/chocolate feature in the future, we should start to rework all this. Maybe into an event system for chat, or something.

src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityRabbitRarity.kt
hannibal002: Yeah, just noticed https://github.com/hannibal002/SkyHanni/blob/f3b5eaf245aa4975ff4a3787c7ec45f6799b369f/src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityEventSummary.kt#L271

DavidArthurCole: I _might_ be able to work around using LorenzRarity instead, and using some selective querying to do what I am right now, but would take some reworking

PR #2312, opened 2024-08-06
Backend: fix pet regex

NopoTheGamer:
## What
Fixes pet regex so it doesn't display [Lvl {LVL}] pet name

exclude_from_changelog


hannibal002:
its exclude_from_changelog not remove_from_changelog

PR #2316, opened 2024-08-06
Fix: Some Dungeon Features to support changed Master Mode text changes in Artist's Adobe update

Fazfoxy:
## What
fixes Master Mode Stack Size

## Changelog Fixes
+ Fixed the Master Mode Dungeon stack size display in the party finder menu. - fazfoxy

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonFinderFeatures.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonFinderFeatures.kt
@@ -57,13 +57,14 @@ object DungeonFinderFeatures {
         "note",
         "§7§7Note: §f(?<note>.*)"
     )
-    /* 
-    * REGEX-TEST: The Catacombs
-    * REGEX-TEST: Master Mode The Catacombs
-    */
+
+    /**
+     * REGEX-TEST: The Catacombs
+     * REGEX-TEST: Master Mode The Catacombs
+     */
     private val floorTypePattern by patternGroup.pattern(
         "floor.type",
-        "(The Catacombs).*|.*(Master Mode The Catacombs).*"
+        "(The Catacombs).*|.*(Master Mode The Catacombs).*",
     )
     private val checkIfPartyPattern by patternGroup.pattern(
         "check.if.party",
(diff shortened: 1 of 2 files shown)


Thunderblade73:
Please add Regex Tests.
Example:
```Kotlin
/** REGEX-TEST: Abcs
*/
Repopattern.pattern("...","Abcs")
```

src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonFinderFeatures.kt
```diff
@@ -57,9 +57,15 @@ object DungeonFinderFeatures {
         "note",
         "§7§7Note: §f(?<note>.*)"
     )
+    /* 
+    REGEX-TESTS:
```
Thunderblade73: The line REGEX-TESTS is not needed.

src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonFinderFeatures.kt
```diff
@@ -57,9 +57,15 @@ object DungeonFinderFeatures {
...
+    * REGEX-TEST: The Catacombs
+    * REGEX-TEST: Master Mode The Catacombs
+    * REGEX-TEST: MM Catacombs
```
Thunderblade73: This test can't match that better. So either the pattern is wrong or the test is wrong

hannibal002 (changes requested):
The master star items still have no stack number.
grafik


or was this pr fixing somehting else?

Fazfoxy:
image
This isn't what I'm fixing but also, those stack sizes are working locally, this feature does sometimes fail to work though maybe re-check it?

hannibal002:
what else are you fixing then?

Thunderblade73:
@hannibal002 I think your "repo pattern force locale" is on the wrong value. If you are in dev enviroment the option must be off. At least that would be the obvious case why it wouldn't work.

Fazfoxy:
more specifically, this was to do with hypixel's changing of PF notes to be Master Mode The Catacombs unlike previously + dungeon clean end being updated for changed lines

hannibal002:
ok i still dont know what the fix actually is about :/ @Fazfoxy

Fazfoxy:
> ok i still dont know what the fix actually is about :/ @Fazfoxy

ok so, it does fix the MM highlight apparently, but also it makes party stack sizes go back to the correct floor, before PR MM parties would stacksize as the non MM variant because of hypixel changing the description + should fix clean end because hypixel changed that line as well I believe

Fazfoxy:
image
when using non changed it should show like this for MM floors

hannibal002:
Now I see, thanks. I have changed the changelog wording accordingly.

hannibal002 (approved):

PR #2326, opened 2024-08-11
Add copper dye to block visitor rewards

not-a-cow:


CalMWolfs (changes requested):
Two things, 
1. you are missing a trailing comma
2. please add it to RewardWarningConfig so that future users of the mod will automatically have this as a blocked thingy. (If you are up to it you could work out how to use config fix event to add it to all existing users configs as well)

CalMWolfs (approved):

PR #2327, opened 2024-08-11
Add: Vacuum warning

not-a-cow:
## What
Warns on screan and in chat when accidentally using a vacuum instead of a Egglocator

## Changelog New Features
+ Warns you when using vacuum instead of Egglocator - Not_a_cow


not-a-cow:
i did not mean to send this to this repo

PR #2328, opened 2024-08-11
Improvement garden keybinds

Cedric-Ab:
## What
Moved GardenCustomKeybinds to it's own config page, then add custom keybinds for each different crop

## Changelog Improvements
+ Added custom keybinds for each crop type. - Cedric-Ab
  * Moved settings into their own category.

+ Create a universal config file for crops. - Cedric-Ab


src/main/java/at/hannibal2/skyhanni/features/garden/GardenCustomKeybinds.kt
```diff
@@ -0,0 +1,202 @@
...
+
+    private val config get() = GardenAPI.config.keyBind
+    private val mcSettings get() = Minecraft.getMinecraft().gameSettings
+
+    private val map: MutableMap<KeyBinding, MutableMap<String, () -> Int>> = IdentityHashMap()
+    private var currentCropName: String = ""
+    private var lastWindowOpenTime = SimpleTimeMark.farPast()
+    private var lastDuplicateKeybindsWarnTime = SimpleTimeMark.farPast()
+
+    init {
```
hannibal002: A very long list like this was the reason we didn't want to create unique keybinds for every crop type in the first place

Cedric-Ab: I don't like that part either, isn't there a way to make it without such a long list?

hannibal002: to reduce code, we could reuse the same config object for every crop. but then we need to handle default values and the crop name in the description special.
also this would still not avoid the biggest problem: a very large config for the user, with tons to settings they might have changed accidentally

Cedric-Ab: I don't understand what the problem for the user would be exactly?

jani270:
Also the provided changelog is rather unhelpful, please change it and tell what you did

Cedric-Ab:
@jani270 
> Also the provided changelog is rather unhelpful, please change it and tell what you did

Is it better now, or what do you want exactly?

jani270:
> @jani270
> 
> > Also the provided changelog is rather unhelpful, please change it and tell what you did
> 
> Is it better now, or what do you want exactly?

I will just do it myself

jani270:
Why did you change it to something worse again 💀

hannibal002:
"Create a universal config file for crops"
please move this to the technical changelog, this is not user facing.

CalMWolfs:

PR #2340, opened 2024-08-13
Fix custom wardrobe not working with patcher inv scale

NopoTheGamer:
## What
Fixes the custom wardobe being offset with patcher inv scale on by just turning off the feature while in that menu

## Changelog New Features
+ Fix custom wardrobe not working with patcher inv scale. - nopo


NopoTheGamer:
apparently this doesnt work and my intellij just stopped working so ill come back to this a bit later I suppose

NopoTheGamer:
ok well my skyhanni dev env still doesnt work soo

PR #2341, opened 2024-08-13
Backend: Dependency Label Workflow

j10a1n15:
## What
This Pull Request adds a workflow, that automatically adds the "Waiting for Dependency PR" label, if the PR has open dependencies. 

exclude_from_changelog


.github/workflows/check_dependencies.yml
```diff
@@ -0,0 +1,26 @@
+name: Check PR Dependencies
+
+on:
+    pull_request:
```
Thunderblade73: needs to be ```pull_request_target``` without it won't have the permission it needs to add a label.
[Good Explanation](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Thunderblade73 (changes requested):

hannibal002:
lets test this in action

PR #2344, opened 2024-08-14
Fix: Visible color codes in Discord RPC

NetheriteMiner:
## What
Fixes a bug where, if the "pets" status is selected, and the user has selected a pet whose name has color codes after the initial color code that determines the pet's rarity (via a pet skin), those additional color codes are visible.

## Changelog Fixes
+ Fixed color codes in Discord RPC pet status. - NetheriteMiner

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/discordrpc/DiscordStatus.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/discordrpc/DiscordStatus.kt
@@ -56,7 +56,7 @@ var beenAfkFor = SimpleTimeMark.now()
 
 fun getPetDisplay(): String = PetAPI.currentPet?.let {
     val colorCode = it.substring(1..2).first()
-    val petName = it.substring(2).removeColor().removeColor()
+    val petName = it.substring(2).removeColor()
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/misc/discordrpc/DiscordStatus.kt
```diff
@@ -56,7 +56,7 @@ var beenAfkFor = SimpleTimeMark.now()
 
 fun getPetDisplay(): String = PetAPI.currentPet?.let {
     val colorCode = it.substring(1..2).first()
-    val petName = it.substring(2)
+    val petName = it.substring(2).removeColor().removeColor()
```
j10a1n15: why do you remove it twice

j10a1n15 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/discordrpc/DiscordStatus.kt
ItsEmpa: you can never be too sure!

Nealkitt:
Wow, annoying people in dms to fix something really does work!

PR #2345, opened 2024-08-14
Fix: ScoreboardData and splitting

j10a1n15:
## What
Its hard to explain, basically, the current System worked fine on "§a111<split>§a222" and "§a§a111<split>§a§a222", but didnt work on "§a§a111<split>§a222", which caused this error. Look I assume this fixes it, I cant test it because it's Spring 4th.

## Changelog Fixes
+ Fixed a Custom Scoreboard error during the Winter Event. - j10a1n15

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/ScoreboardData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/ScoreboardData.kt
@@ -35,8 +35,14 @@ object ScoreboardData {
 
             val lastColor = start.lastColorCode() ?: ""
 
-            if (end.startsWith(lastColor)) {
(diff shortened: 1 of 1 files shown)


hannibal002:
How bad is this bug for users and how big is the potential problem when this fix actually breaks stuff? 
Do you recommend merging it? I'm unsure now

j10a1n15:
> How bad is this bug for users and how big is the potential problem when this fix actually breaks stuff? Do you recommend merging it? I'm unsure now

Works fine in the [kotlin playground](https://pl.kotl.in/hpsb-x0-a)
(bottom line is the problematic line)

PR #2348, opened 2024-08-15
Improvement: Added parkour and teleport messages to blockable messages

not-a-cowfr:
## What
Hides **parkour** and **teleport pad** messages. I wasn't able to test, **I'M NOT SURE IF IT WORKS**

## Changelog Improvements
+ Added more options to the Chat Hider feature. - not_a_cow
    * Option to block parkour messages and teleport pad messages.


CalMWolfs (changes requested):
please use repo patterns and repo pattern lists along with regex tests

not-a-cowfr:
> please use repo patterns and repo pattern lists along with regex tests

do i need a test for every message? or will one per list do? (sorry im not very familiar with this)

hannibal002:
Please write at least one text line per unique pattern. In terms of lists this means one for every entry in the list.

PR #2358, opened 2024-08-16
Fix: Pet detection

hannibal002:
## What
I am not sure why exactly, but with newer versions of NEU, the SkyHanni pet detection fails again.
It does not look like this is related to different NEU-REPO versions. On the outdated NEU version in Dev Env it worked, in production with NEU 2.3.3 it did not. This PR fixes this.

## Changelog Fixes
+ Fixed pet detection with newer NEU versions. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestProfitTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestProfitTracker.kt
@@ -89,6 +89,7 @@ object PestProfitTracker {
             val internalName = NEUInternalName.fromItemNameOrNull(group("item")) ?: return
 
             tracker.addItem(internalName, 1)
+            addKill()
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestProfitTracker.kt
```diff
@@ -89,6 +89,7 @@ object PestProfitTracker {
             val internalName = NEUInternalName.fromItemNameOrNull(group("item")) ?: return
 
             tracker.addItem(internalName, 1)
+            addKill()
```
lunaynx, line 92: Ah yes, pet detection.

hannibal002: this is an unrelated fix

PR #2372, opened 2024-08-17
Backend: Renderable Wrapped String Internal Alignement

Thunderblade73:
## What
The Renderable Wrapped Sting now has a option to align the text inside the wrapped string box.
Also fixes the increased size if the it wrapped (since it always assumed the input width instead of the width of the longest line).

+ The Renderable Wrapped String now has an option to align text inside the wrapped string box. - Thunderblade73


j10a1n15 (approved):
works

PR #2383, opened 2024-08-19
Improvement + Fix: No Rarity on Compacted Dupes

DavidArthurCole:
## What
Follow-up to #2364 - this adds the rarity for duplicate rabbits as well.
Also adds a config message to turn the rarity-adding off.

## Changelog Improvements
+ Added config option to disable rarity in Compact Hoppity messages. - Daveed

## Changelog Fixes
+ Fixed rabbit rarity not appearing on compacted Duplicate rabbit messages. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityEggsCompactChat.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityEggsCompactChat.kt
@@ -14,6 +14,8 @@ import at.hannibal2.skyhanni.utils.TimeUtils.format
 import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.Duration.Companion.seconds
 
+typealias RarityType = HoppityEggsConfig.CompactRarityTypes
+
 object HoppityEggsCompactChat {
 
     private var hoppityEggChat = mutableListOf<String>()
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityEggsCompactChat.kt
```diff
@@ -65,17 +65,19 @@ object HoppityEggsCompactChat {
...
+        val rarityText = if (showRarity) "$lastRarity " else ""
         return if (duplicate) {
             val format = lastDuplicateAmount?.shortFormat() ?: "?"
             val timeFormatted = lastDuplicateAmount?.let {
                 ChocolateFactoryAPI.timeUntilNeed(it).format(maxUnits = 2)
             } ?: "?"
 
             val timeStr = if (config.showDuplicateTime) ", §a+§b$timeFormatted§7" else ""
-            "$mealNameFormatted! §7Duplicate $lastName §7(§6+$format Chocolate§7$timeStr)"
+            "$mealNameFormatted! §7Duplicate $rarityText$lastName §7(§6+$format Chocolate§7$timeStr)"
```
hannibal002: should never show rarity for duplicate rabbits, or having a separate option

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityEggsCompactChat.kt
DavidArthurCole: Why wouldn't we? If they're turning on the option, wouldn't it make sense for consistency? Why different options for new versus duplicate?

hannibal002: I don't care about duplicates nearly as much as I care about uniques. For dupes, the current format with just showing the rarity as color for the rabbit name is perfectly fine for me. Having a bold COMMON text next to it would annoy me, I maybe even think this is unique/special.
image

Instead, for truly unique rabbits, having the rarity blast next to the name sounds like a great idea.

This is just my personal preference about this. Therefore, I think we can add those as settings for the config.

Or simply not show rarity name for dupes always. The dupe message is intentionally designed to be less shiny/flashy as the purple big NEW text for non-dupes.

PR #2388, opened 2024-08-21
Fix: New Alpha Ores

ItsEmpa:
## What
Adds new ores added in alpha

## Changelog Fixes
+ Added new Pure Ores support. - Empa


hannibal002:
will it break on live server when we merge this now?

ItsEmpa:
no it wouldnt break

PR #2391, opened 2024-08-21
improvement: Item errors in beta only

hannibal002:
## What
item category and rarity errors will only show up in beta versions now

## Changelog Improvements
+ Item rarity or item category not found errors will only show up on beta versions now. - hannibal2

PR #2392, opened 2024-08-21
Fix: ItemCategory error spam

Thunderblade73:
## What
Makes them Beta only.
Before sending the error it checks a repo pattern which maches all entires, which updates remotly to include the new stuff.

Also replaces #2391

## Changelog Fixes
+ Fixed ItemCategory errors being sent even though we already have a fix. - Thunderblade73


src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
```diff
@@ -213,7 +225,9 @@ object ItemUtils {
                     "internal name" to getInternalName(),
                     "item name" to name,
                     "inventory name" to InventoryUtils.openInventoryName(),
+                    "pattern result" to rarity,
```
hannibal002, line 228: why pattern result? isnt rarity as label here better?
also where is the beta only tag for item rarity stack error log

Thunderblade73: Since it is the result of the pattern. And yes I forgot to add betaOnly here but added this already

hannibal002 (approved):
I really move the implementation of this fix

PR #2410, opened 2024-08-25
Fix: Patcher coordinates chat detection

hannibal002:
## What
fix error with Patcher coordinate detection from chat.
We can't use a more simple regex without breaking support for older versions.

## Changelog Fixes
+ Fixed an error with Patcher Coordinate Detection from chat. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/PatcherSendCoordinates.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/PatcherSendCoordinates.kt
@@ -30,10 +30,12 @@ object PatcherSendCoordinates {
 
     /**
      * REGEX-TEST: hannibal2: x: 2, y: 3, z: 4
+     * REGEX-TEST: hannibal2: x: 2, y: 3, z: 4broken
+     * REGEX-TEST: hannibal2: x: 2, y: 3, z: 4 extra text
      */
     private val coordinatePattern by RepoPattern.pattern(
         "misc.patchercoords.coords",
-        "(?<playerName>.*): [xX]: (?<x>[0-9.-]+),? [yY]: (?<y>[0-9.-]+),? [zZ]: (?<z>.*)"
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/misc/PatcherSendCoordinates.kt
```diff
@@ -30,10 +30,12 @@ object PatcherSendCoordinates {
 
     /**
      * REGEX-TEST: hannibal2: x: 2, y: 3, z: 4
+     * REGEX-TEST: hannibal2: x: 2, y: 3, z: 4broken
+     * REGEX-TEST: hannibal2: x: 2, y: 3, z: 4 extra text
      */
     private val coordinatePattern by RepoPattern.pattern(
         "misc.patchercoords.coords",
-        "(?<playerName>.*): [xX]: (?<x>[0-9.-]+),? [yY]: (?<y>[0-9.-]+),? [zZ]: (?<z>.*)"
+        "(?<playerName>.*): [xX]: (?<x>[0-9.-]+),? [yY]: (?<y>[0-9.-]+),? [zZ]: (?<z>[0-9.-]+(?: .*)?)"
```
j10a1n15, line 38:
```suggestion
        "(?<playerName>.*): [xX]: (?<x>[0-9.-]+),? [yY]: (?<y>[0-9.-]+),? [zZ]: (?<z>[0-9.-]+)(?:.*)"
```

why not just
image

j10a1n15 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/PatcherSendCoordinates.kt
hannibal002: The second line in the regex test will break SkyHanni code. We expect one empty line between z and extra text.
In a future pr we might want to change the SkyHanni code, but for now, we want to also fix this for older sh versions via a regex only change.

PR #2413, opened 2024-08-26
Improvement: Warning rancher chat command

hannibal002:
## What
Improved wrong rancher speed while farming warning, allows clicking on the message to open the edit sign

<details>
<summary>Images</summary>

</details>

## Changelog Improvements
+ Improved incorrect Rancher Boots speed warning. - hannibal2
    * Also allows clicking on the message to open the edit sign directly.
    * Now warns only when actually farming and wearing Rancher Boots.


CalMWolfs (approved):

lunaynx (changes requested):
It would be best to check if the player is actually wearing Rancher's Boots before offering to change the speed.

PR #2415, opened 2024-08-26
Fix: Clicked Blocks Highlight Migrator

j10a1n15:
## What
Was outdated in the pr

exclude_from_changelog


hannibal002 (approved):
good catch, thank you!

PR #2419, opened 2024-08-27
Backend: Remove more neu code from SkyHanni

CalMWolfs:
## What
Removed most remaining function calls to Neu methods by replacing them with ones in skyhanni.
Updated Neu version in dev environment

Removed the ability to use neu chroma in ghost tracker as they can just use skyhanni chroma instead and ghost tracker is not in a great state rn anyways

This allows us to be one step closer to a version of SkyHanni that doesnt require NEU and is more ready for transition to modern minecraft versions

+ Removed references to NEU code, especially code from the Utils class in NEU. - CalMWolfs
+ Banned importing `io.github.moulberry.notenoughupdates.util.Utils`. - CalMWolfs


hannibal002 (changes requested):
Looks great overall, please also add credits to neu to smaller functions that are directly copied over.

PR #2425, opened 2024-08-28
Improvement: Mining Event Display Icons nbt id

Thunderblade73:
## What
Set the id of the icons to the matching ones (Raffle and Mithril Gourment). So that it can be retextured by texture packs.
Idk which id would be correct for the 2x powder event icon. Since I could not find a example texture in furrysky (to know which id the use for powder).

## Changelog Improvements
+ Some icons in the Mining Event Display now use the texture of the item they represent. - Thunderblade73
    * E.g., Raffle Event -> Raffle Ticket.


hannibal002:
This feels more like a fix than an improvement to me. Also, please change the wording of the changelog to include the info of what was broken, not only how it should work.

PR #2429, opened 2024-08-30
Fix: Github Actions Artifact path

hannibal002:
## What
GitHub Actions are not providing the download build as an artifact. This PR fixes the changed build paths.

## Changelog Fixes
+ Fixed incorrect Build Path in GitHub actions. - nea

exclude_from_changelog


CalMWolfs (approved):
looks like it worked

PR #2432, opened 2024-08-30
Fix: Tunnel Maps commission

hannibal002:
## What
fix tunnel maps repeatedly showing the last auto-commission goal even when there is no new commission

## Changelog Fixes
+ Fixed Glacite Tunnel Maps repeatedly showing the last commission goal even when there is no new commission. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/mining/TunnelsMaps.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/mining/TunnelsMaps.kt
@@ -187,7 +187,7 @@ object TunnelsMaps {
             } ?: run {
                 if (isCommission) {
                     active = ""
-                    goal = null
+                    clearPath()
                     isCommission = false
                 }
             }
--- a/src/main/java/at/hannibal2/skyhanni/features/mining/TunnelsMaps.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/mining/TunnelsMaps.kt
@@ -152,6 +152,8 @@ object TunnelsMaps {
 
     private val ROYAL_PIGEON by lazy { "ROYAL_PIGEON".asInternalName() }
 
+    private var isCommission = false
+
     @SubscribeEvent
     fun onInventoryFullyOpened(event: InventoryFullyOpenedEvent) {
         if (!isEnabled()) return
@@ -180,7 +182,14 @@ object TunnelsMaps {
         }.toMap()
         if (config.autoCommission) {
             clickTranslate.values.firstOrNull()?.let {
+                isCommission = true
                 setActiveAndGoal(it)
+            } ?: run {
+                if (isCommission) {
+                    active = ""
+                    goal = getNext()
+                    isCommission = false
+                }
             }
         }
     }
@@ -198,6 +207,7 @@ object TunnelsMaps {
         if (!isEnabled()) return
         if (event.clickedButton != 1) return
         clickTranslate[event.slotId]?.let {
+            isCommission = false
             setActiveAndGoal(it)
         }
     }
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/mining/TunnelsMaps.kt
```diff
@@ -198,6 +207,7 @@ object TunnelsMaps {
         if (!isEnabled()) return
         if (event.clickedButton != 1) return
         clickTranslate[event.slotId]?.let {
+            isCommission = false
```
Thunderblade73: This should be true

hannibal002: Originally I wanted to remove that line but forgot.

What is the difference between this click detection and the normal open inventory detection? Since either the goal gets reset, or we need to wait for the next inventory update to see the next goal anyway.
so there should be nothing to change "tunnel maps"-wise when clicking the on the commission item. or am i missing something?

src/main/java/at/hannibal2/skyhanni/features/mining/TunnelsMaps.kt
```diff
@@ -180,7 +182,14 @@ object TunnelsMaps {
         }.toMap()
         if (config.autoCommission) {
             clickTranslate.values.firstOrNull()?.let {
+                isCommission = true
                 setActiveAndGoal(it)
+            } ?: run {
+                if (isCommission) {
+                    active = ""
+                    goal = getNext()
```
Thunderblade73: This should be null not getNext() (Since getNext() alwalys returns null in this context)

hannibal002: good catch!

src/main/java/at/hannibal2/skyhanni/features/mining/TunnelsMaps.kt
Thunderblade73: We have 3 modes of operation:
 * Manuel (using our gui to set any goal)
 * Automatic (which grabs the goal at inventory open)
 * Select (You can click a commission to chose to follow it instead)
 
 And this is option 3.

hannibal002: can you unfollow a automatic commission by clicking this way as well?

Thunderblade73: The last action done by the user has the highest priority.

src/main/java/at/hannibal2/skyhanni/features/mining/TunnelsMaps.kt
```diff
@@ -180,7 +182,14 @@ object TunnelsMaps {
         }.toMap()
         if (config.autoCommission) {
             clickTranslate.values.firstOrNull()?.let {
+                isCommission = true
                 setActiveAndGoal(it)
+            } ?: run {
+                if (isCommission) {
+                    active = ""
+                    goal = null
```
Thunderblade73: clear path would be more correct here. Since the path does not need to exist without a goal. Normally goal setting clears the old path. But when the goal is null it returns early. Meaning the path that the path should be explicitly cleared.

Thunderblade73 (approved):
Looks fine. Hope it does not create a deadlock (which we should assume it does not until proven otherwise)

hannibal002:
We will hear in support from beta testers if this causes a deadlock. When we get to the next full version without deadlock, we are safe.

PR #2437, opened 2024-08-31
Fix: Tungsten Detection on Alpha

hannibal002:
## What
Fix tungsten detection on alpha.
bug report: https://discord.com/channels/997079228510117908/1279527967739084972

## Changelog Fixes
+ Fixed Tungsten Ore and Hardstone detection after the mining update. - Empa


src/main/java/at/hannibal2/skyhanni/features/mining/OreBlock.kt
```diff
@@ -222,7 +222,7 @@ enum class OreBlock(
         checkArea = { inGlacite },
     ),
     LOW_TIER_TUNGSTEN(
-        checkBlock = { it.block == Blocks.cobblestone },
+        checkBlock = { it.block.let { it == Blocks.cobblestone || it == Blocks.monster_egg } },
```
j10a1n15: You should also check for the monster egg block variant, since this just triggers on all monster egg blocks (stone, cobble, bricks,...)

j10a1n15 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/mining/OreBlock.kt
hannibal002: done

ItsEmpa (approved):

PR #2453, opened 2024-09-03
Feature: Secret Chime

Ovi-111:
## What
This Pull Request introduces a Secret Chime feature for dungeons.
- Plays a sound when a chest, lever, or wither essence is clicked inside dungeons.
- Secret Chime pitch and sound can be changed.

I'm still new to contributing. So I'm still learning. How things work.

## Changelog New Features
+ Secret Chime. - Ovi_1
    * Plays a user-specified sound when a chest, lever, or wither essence is clicked inside dungeons.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/dungeon/SecretChimeConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/dungeon/SecretChimeConfig.java
@@ -2,6 +2,7 @@ package at.hannibal2.skyhanni.config.features.dungeon;
 
 import at.hannibal2.skyhanni.config.FeatureToggle;
 import at.hannibal2.skyhanni.features.dungeon.DungeonSecretChime;
+import at.hannibal2.skyhanni.utils.OSUtils;
 import com.google.gson.annotations.Expose;
 import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorBoolean;
 import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorButton;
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/config/features/dungeon/SecretChimeConfig.java
CalMWolfs, line 1: This should link to the list of sounds like other features in the mod that allow you to select a sound do

src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonSecretChime.kt
```diff
@@ -0,0 +1,49 @@
...
+    fun onBlockClick(event: BlockClickEvent) {
+        if (!isEnabled() || event.clickType != ClickType.RIGHT_CLICK) return
+
+        val position = event.position
+        val blockType = when (position.getBlockAt()) {
+            Blocks.chest, Blocks.trapped_chest, Blocks.lever, Blocks.skull -> true
+            else -> false
+        }
+
+        if (event.position.getBlockAt() == Blocks.skull) {
```
CalMWolfs: why doesnt this use the position variable you created above

src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonSecretChime.kt
```diff
@@ -0,0 +1,49 @@
...
+            val text = BlockUtils.getTextureFromSkull(position.toBlockPos())
```
CalMWolfs: probably rename the variable to texture so that its more descriptive

src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonSecretChime.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    @SubscribeEvent
+    fun onBlockClick(event: BlockClickEvent) {
+        if (!isEnabled() || event.clickType != ClickType.RIGHT_CLICK) return
+
+        val position = event.position
+        val blockType = when (position.getBlockAt()) {
+            Blocks.chest, Blocks.trapped_chest, Blocks.lever, Blocks.skull -> true
+            else -> false
+        }
```
CalMWolfs: instead of setting a boolean variable here, just return where you are currently setting it to false. If you want to make this code better you would also do the wither essence texture check in this when statement if the block is of skull type

CalMWolfs (changes requested):
Just a few suggestions

Also you did not follow the [pull request template](https://github.com/hannibal002/SkyHanni/blob/beta/pull_request_template.md) which means your pr will not be added to the changelogs. It will not be merged until this is fixed

hannibal002 (changes requested):
Code looks overall fine. I have not yet tested it in game.

One big problem though: this is code duplication to DungeonHighlightClickedBlocks.
You can avoid this by moving the enum ClickedBlockType into a own class, and create a "dungeon clicked block event" with contains one of those enums as well. move that check into dungeon api. then you can listen to that new event inside the clicked blocks and secret chime feature at the same time.

PR #2457, opened 2024-09-04
Fix: Check line of sight before alerting about rare sea creature

lunaynx:
## What
Fixed missing `canBeSeen` check in rare sea creature alert.

## Changelog Fixes
+ Fixed rare sea creature alert not checking if the sea creature can be seen by the player. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureFeatures.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureFeatures.kt
@@ -39,7 +39,10 @@ object SeaCreatureFeatures {
     @SubscribeEvent
     fun onMobSpawn(event: MobEvent.Spawn.SkyblockMob) {
         if (!isEnabled()) return
+
         val mob = event.mob
+        if (!mob.canBeSeen()) return
+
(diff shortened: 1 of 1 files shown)


hannibal002:
This Logic change will cause sea creatures not getting detected by the alert feature at all. Even then, when the mob spawns behind blocks and gets pulled into the player's line of sight a second later.

~~I think there is a "SkyBlock Mob first time see event" or something especially for this~~
This Event does not exist, we need a new event for this since this very behavior is needed in multiple features. see https://github.com/hannibal002/SkyHanni/commit/3cd39585d96e8e87c2e3c98e9a095a5e1c4c1fb7

ItsEmpa:
> This Logic change will cause sea creatures entirely not getting enounced by the alert feature, even when they Spawn behind blocks and get pulled into the player's line of sight a second later.
>I think there is a "skyblock Mob first time see event" or something especially for this

that event doesnt mean first time you ACTUALLY see the mob, its just the first time that mob is detected

hannibal002:
> that event doesnt mean first time you ACTUALLY see the mob, its just the first time that mob is detected

<details>
<summary>I'm talking about this</summary>

image

</details>


The mob spawn even already handles "first detection". otherwise, the names would be horrible wrong or the spawn event would fire multiple times for the same entity, right?

PR #2469, opened 2024-09-06
Improvement: Make pest middle waypoint optional

lunaynx:
## What
Added a toggle for pest middle waypoint (off by default). This was added without much thought and is not really useful in most cases, it's just visual clutter. There's better ways we could add to indicate if the waypoint is likely inaccurate, but I'm leaving in a toggle in case someone wants it for testing purposes or other reasons.

## Changelog Improvements
+ Made the waypoint to the middle of the plot for Pest Waypoint optional (off by default). - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/garden/pests/PestWaypointConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/garden/pests/PestWaypointConfig.java
@@ -35,6 +35,14 @@ public class PestWaypointConfig {
 
     @Expose
     @ConfigOption(
+        name = "Show Middle",
+        desc = "Also show a waypoint to the middle of a plot. This can help determine if the tracker is pointing to the middle instead of a pest."
+    )
+    @ConfigEditorBoolean
+    public boolean showMiddle = false;
+
+    @Expose
+    @ConfigOption(
         name = "Show For Seconds",
         desc = "The waypoint will disappear after this number of seconds."
     )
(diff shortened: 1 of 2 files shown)


hannibal002 (changes requested):
Not having an option to disable this "middle of plot" indicator was intentional, I could not imagine someone not wanting this. Since you don't want it and put the effort in to add an option, we can add it. But:
Personally, I find this visual indicator super useful when I'm looking at a glance where the waypoint is. This is no visual clutter for me. I'm sure I'm not the only one.
The additional waypoint only exists while holding the hoover item, and also only exactly in the middle. I rarely even notice it is there, only when first entering a new plot and waiting for the pests to spawn. And exactly then I do need this feature. There is no additional clutter. The gray color is very different from the default orange color of the actual pest waypoints. 
Also, this already exists and users are adapted to it. So why make the option default off?
In my opinion, it's very unlikely someone who would use the feature is finding it in the settings, but someone who actually is annoyed by it will go further lengths to disable it. I would like to see the "show middle" option being default enabled, not default disabled.

lunaynx:
If you insist, I'll make it default enabled. But to me, this screams "debug feature left in production code". In fact, it has been bugging me ever since it was added, but I was too lazy to make a PR until now. Like I said, an alternative solution such as coloring the waypoint based on whether it points to the middle of the plot would be much better for practical use cases IMO. Or based on proximity to it if it doesn't always point *exactly* to the middle of the plot when the pests are not loaded. And maybe also making it say "Plot Guess" or "Pest Guess (Inaccurate)" or something like that in that case. (Isn't that what we did before Hypixel completely changed how the vacuum's particles work? Not sure what was wrong with that.)

lunaynx:
Also, the average user probably doesn't care enough about such a small thing, they may go "what the hell is this" and then forget about it 5 minutes later, possibly not even realizing it's from SkyHanni.

A lack of complaints doesn't automatically mean it's a useful feature that's seeing widespread use, it can very well also mean that a lot of people are just indifferent or barely even notice it.

hannibal002:
Fair points!

hannibal002 (approved):
auto disabled is fine

PR #2535, opened 2024-09-18
Backend: AreaChangeEvent and Islands in HandleEvent

ItsEmpa:
## What
Adds AreaChangeEvent and uses it in MiningAPI, and changes some stuff in LocationFixData to use a map.
Also reverts the changes for the onlyOnIsland argument in the HandleEvent annotation in #2502 , and adds a separate argument to use when using multiple IslandType.
For checking a single IslandType, use `onlyOnIsland`, and for multiple use `onlyOnIslands`.

+ Added `AreaChangeEvent`. - Empa
+ Updated `HandleEvent` to allow `onlyOnIsland` usage for a single island without brackets. - Empa


CalMWolfs (approved):
👀

PR #2546, opened 2024-09-19
Feature: Highlight Active Beacon Effect

CuzImClicks:
## What
Marks the slot with the effect that is currently selected with a green background.

## Changelog New Features
+ Added highlighting for the active Beacon Effect. - Cuz_Im_Clicks

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
@@ -3,6 +3,7 @@ package at.hannibal2.skyhanni.features.inventory
 import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.data.IslandType
 import at.hannibal2.skyhanni.events.GuiContainerEvent
+import at.hannibal2.skyhanni.events.InventoryCloseEvent
 import at.hannibal2.skyhanni.events.InventoryFullyOpenedEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.ItemUtils.getLore
@@ -49,6 +50,11 @@ object ActiveBeaconEffect {
         slot = event.inventoryItems.filter { (_, stack) ->
             stack.getLore().any { slotPattern.matches(it) }
         }.firstNotNullOfOrNull { it.key }
+    }
+
+    @SubscribeEvent
+    fun onInventoryClose(event: InventoryCloseEvent) {
+        slot = null
     }
 
     @SubscribeEvent
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
@@ -28,34 +28,25 @@ object ActiveBeaconEffect {
         "§aActive stat boost!",
     )
 
-    private var slot = -1
-    private var inInventory = false
+    private var slot: Int? = null
 
     @SubscribeEvent
     fun onInventoryFullyOpened(event: InventoryFullyOpenedEvent) {
         if (!isEnabled()) return
-        inInventory = inventoryPattern.matches(event.inventoryName)
-        if (!inInventory) return
+        if (!inventoryPattern.matches(event.inventoryName)) return
 
-        for ((slot, stack) in event.inventoryItems) {
-            if (stack.getLore().any { slotPattern.matches(it) }) {
-                this.slot = slot
-                return
-            }
-        }
-        slot = -1
+        slot = event.inventoryItems.filter { (_, stack) ->
+            stack.getLore().any { slotPattern.matches(it) }
+        }.firstNotNullOfOrNull { it.key }
     }
 
     @SubscribeEvent
     fun onBackgroundDrawn(event: GuiContainerEvent.BackgroundDrawnEvent) {
         if (!isEnabled()) return
-        if (!inInventory) return
-        if (slot == -1) return
+        val slot = slot ?: return
 
-        val guiChest = event.gui
-        val chest = guiChest.inventorySlots
-        chest.getSlot(slot) highlight LorenzColor.GREEN
+        event.gui.inventorySlots.getSlot(slot) highlight LorenzColor.GREEN
     }
 
-    fun isEnabled() = config.highlightActiveBeaconEffect && IslandType.PRIVATE_ISLAND.isInIsland()
+    fun isEnabled() = IslandType.PRIVATE_ISLAND.isInIsland() && config.highlightActiveBeaconEffect
 }
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
@@ -5,7 +5,6 @@ import at.hannibal2.skyhanni.data.IslandType
 import at.hannibal2.skyhanni.events.GuiContainerEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.InventoryUtils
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    val config get() = SkyHanniMod.feature.inventory
+
+    private val patternGroup = RepoPattern.group("inventory.activebeaconeffect")
+    private val inventoryPattern by patternGroup.pattern(
+        "inventory",
+        "Profile Stat Upgrades"
+    )
+    private val slotPattern by patternGroup.pattern(
+        "slot",
```
Thunderblade73: A little improvement for the key, Since a "slot" alone may lead to a collision down the line.
```suggestion
        "slot.active",
```

src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
```diff
@@ -0,0 +1,49 @@
...
+        val chest = guiChest.inventorySlots as ContainerChest
+        for ((slot, _) in chest.getUpperItems()) {
+            if (slot.stack.getLore().any { slotPattern.matches(it) }) {
+                slot highlight LorenzColor.GREEN
+                break
+            }
+        }
+    }
+
+    fun isEnabled() = config.highlightActiveBeaconEffect
```
Thunderblade73: isn't it only available on the private island?

src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
```diff
@@ -0,0 +1,49 @@
...
+        if (!isEnabled()) return
+        if (!inventoryPattern.matches(InventoryUtils.openInventoryName())) return
+
+        val guiChest = event.gui
+        val chest = guiChest.inventorySlots as ContainerChest
+        for ((slot, _) in chest.getUpperItems()) {
+            if (slot.stack.getLore().any { slotPattern.matches(it) }) {
+                slot highlight LorenzColor.GREEN
+                break
+            }
```
Thunderblade73: This logic may be cached on InventoryFullyOpened. But up to you if you want to change it since the performance loss is minimal here anyway in the grand scheme.

Thunderblade73 (changes requested):
Some small suggestions. But overall good.

src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
CuzImClicks: so you want to save the beacon inventory instead of getting it every time a render happens? i mean that could improve but i just did what the bingo card thing does

Thunderblade73: That's usually what we do. ( Bingo Card Overlay is just a older piece of code).

CuzImClicks: uh so do you have a feature where i can copy from? ;)

Thunderblade73: I have an abstract version in SkyblockGuideHighlightFeature. But there must be other examples somewhere.

CuzImClicks: should i move to a version where the selected one is saved as a slot id? which only gets updated when InventoryFullyOpenedEvent is called?

Thunderblade73: something like that yeah.

src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val inventoryPattern by patternGroup.pattern(
+        "inventory",
+        "Profile Stat Upgrades"
+    )
+    private val slotPattern by patternGroup.pattern(
+        "slot.active",
+        "§aActive stat boost!"
+    )
+
+    private var slot = -1
```
hannibal002: make slot nullable and use null instead of -1 as default/no value

src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
```diff
@@ -0,0 +1,61 @@
...
+        if (!isEnabled()) return
+        if (!inInventory) return
+        if (slot == -1) return
+
+        val guiChest = event.gui
+        val chest = guiChest.inventorySlots
+        chest.getSlot(slot) highlight LorenzColor.GREEN
+    }
+
+    fun isEnabled() = config.highlightActiveBeaconEffect && IslandType.PRIVATE_ISLAND.isInIsland()
```
hannibal002: please always first check if in sb (by checking if on the island) then access the config

src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
```diff
@@ -0,0 +1,61 @@
...
+        "§aActive stat boost!"
+    )
+
+    private var slot = -1
+    private var inInventory = false
+
+    @SubscribeEvent
+    fun onInventoryFullyOpened(event: InventoryFullyOpenedEvent) {
+        if (!isEnabled()) return
+        inInventory = inventoryPattern.matches(event.inventoryName)
```
hannibal002: since we use `slot` in `onBackgroundDrawn` already, the variable `inInventory`  can stay locally in the function

src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
```diff
@@ -0,0 +1,61 @@
...
+        if (!inInventory) return
+
+        for ((slot, stack) in event.inventoryItems) {
+            if (stack.getLore().any { slotPattern.matches(it) }) {
+                this.slot = slot
+                return
+            }
+        }
+        slot = -1
```
hannibal002: This can be greatly improved via
```suggestion
        slot = event.inventoryItems.filter { (_, stack) ->
            stack.getLore().any { slotPattern.matches(it) }
        }.firstNotNullOfOrNull { it.key } ?: -1
```

src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @SubscribeEvent
+    fun onBackgroundDrawn(event: GuiContainerEvent.BackgroundDrawnEvent) {
+        if (!isEnabled()) return
+        if (!inInventory) return
+        if (slot == -1) return
+
+        val guiChest = event.gui
```
hannibal002: this variable can be inlined

src/main/java/at/hannibal2/skyhanni/features/inventory/ActiveBeaconEffect.kt
```diff
@@ -0,0 +1,61 @@
...
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+@SkyHanniModule
+object ActiveBeaconEffect {
+
+    val config get() = SkyHanniMod.feature.inventory
+
+    private val patternGroup = RepoPattern.group("inventory.activebeaconeffect")
+    private val inventoryPattern by patternGroup.pattern(
```
hannibal002, line 27: even though the patterns are very simple, please add regex tests for consistency.

hannibal002 (changes requested):
code looks good, some cleanup is still possible

PR #2554, opened 2024-09-20
Fix: Totem of Corruption expiry warning sometimes not working

lunaynx:
## What
Fixed Totem of Corruption expiry warning not working if the timer skipped a second by changing the check to `<=` and adding a time-limited cache of totems that have been warned about.

## Changelog Fixes
+ Fixed Totem of Corruption expiry warning not working in some cases. - Luna


hannibal002 (changes requested):
Use TimeLimitedSet instead

PR #2558, opened 2024-09-20
Backend: Add blossom and make source checker show version again

ThatGravyBoat:
## What
Adds blossom to do token replacement in code. 
Adds back the version number to the security error screen.
Changes the hardcoded version number in SkyhanniMod to the token (should reduce amount of merge conflicts)

+ Added Blossom Gradle plugin for token replacement in code. - ThatGravyBoat
    * `@MOD_VERSION@` is now a token that will be replaced.


CalMWolfs (approved):
Worked for me

settings.gradle.kts
```diff
@@ -14,6 +14,7 @@ pluginManagement {
         maven("https://repo.nea.moe/releases")
         maven("https://repo.sk1er.club/repository/maven-releases/")
         maven("https://maven.deftu.dev/releases")
+        maven("https://repo.polyfrost.org/releases")
```
hannibal002: is there no other repo to get this from?

hannibal002 (changes requested):

settings.gradle.kts
ThatGravyBoat: wdym? this is the repo in which has a fork that supports the newer kotlin version as they changed how it works in 2.x but the fix wasnt backported to 1.x

ThatGravyBoat: This is the changes that Polyfrost made https://github.com/KyoriPowered/blossom/compare/master...Polyfrost:blossom:master

ThatGravyBoat: if the issue is about uptime of polyfrost infrastructure, I can put a mirror of it on my own maven (which is already used for the RPC)

hannibal002 (approved):

PR #2540, opened 2024-09-18
Improvement: Add Pickobulus support to OreMineEvent

ItsEmpa:
## What
Adds support for using the Pickobulus ability for the OreMineEvent.

## Changelog Improvements
+ Added support for using the Pickobulus ability for the Mineshaft Pity Display. - Empa

PR #2561, opened 2024-09-21
Feature: Flowstate helper

martimavocado:
## What
adds a status gui with the current state of the flowstate enchantment bonus

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added Flowstate Helper. - martimavocado
    * Displays stats for the Flowstate enchantment on mining tools.


src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
```diff
@@ -515,4 +515,9 @@ object ItemUtils {
             true, true,
         )
     }
+
+    fun getHeldItem(): ItemStack? {
```
Thunderblade73: InventoryUtils.getItemInHand already exits

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,123 @@
...
+            config.flowstateHelperPosition.renderRenderables(
+                listOf(Renderable.verticalContainer(it, 1)),
+                posLabel = "Flowstate Helper",
+                addToGuiManager = true
+            )
+        }
+    }
+
+    private fun createDisplay() {
+        val helperDisplay: MutableList<Renderable> = mutableListOf()
```
Thunderblade73: use a listBuilder instead of a directly mutable list.

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,123 @@
...
+
+        helperDisplay.add(Renderable.string("§d§lFlowstate Helper"))
+        helperDisplay.add(createDisplayTimer())
+        helperDisplay.add(createDisplayBlock())
+        helperDisplay.add(Renderable.string("§6+${getSpeedBonus()}⸕"))
```
Thunderblade73: You can cache all values until block break (except the timer). The list creation it self is not that expensive.

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    private fun attemptClearDisplay() {
+        DelayedRun.runDelayed(10.seconds) {
+            if (timeSinceBreak.isInFuture()) return@runDelayed
+            blockBreakStreak = 0
+            createDisplay()
+        }
+    }
```
Thunderblade73: I have a bad feeling about this, Since you can break a lot of blocks in quick succession Which would launch a bunch of redundant delayed Runs. A simple onTick logic would be better in my opinion.

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,123 @@
...
+        createDisplay()
+        attemptClearDisplay()
+    }
+
+    @SubscribeEvent
+    fun onChat(event: LorenzChatEvent) {
+        if (!MiningAPI.inCustomMiningIsland()) return
+        if (hasFlowstate() == null) return
+
+        pickobulusPattern.matchMatcher(event.message) {
```
Thunderblade73: Empa has an open [PR](https://github.com/hannibal002/SkyHanni/pull/2540) that adds the pickoblus support to OreMinedEvent. So maybe you should give it a quick look and potentially add a todo here to remove it when the other Pr is merged.

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,123 @@
...
+        return Renderable.string("Streak: $textColor$blockBreakStreak")
+    }
+
+    private fun getSpeedBonus(): Int {
+        val flowstateLevel = hasFlowstate() ?: 0
+        if (blockBreakStreak >= 200) return 200 * flowstateLevel
+        return blockBreakStreak * flowstateLevel
+    }
+
+    private fun hasFlowstate(): Int? {
```
Thunderblade73: Please cache the result (as long as the item didn't change)

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,123 @@
...
+        "mining.pickobulus.blockdestroy",
+        "§7Your §r§aPickobulus §r§7destroyed §r§e(?<amount>\\d+) §r§7blocks!"
+    )
+
+    @HandleEvent(onlyOnSkyblock = true)
+    fun onBlockMined(event: OreMinedEvent) {
+        if (!MiningAPI.inCustomMiningIsland()) return
+        if (hasFlowstate() == null) return
+
+        timeSinceBreak = SimpleTimeMark.now().plus(10.seconds)
```
Thunderblade73: I don't like the variable name since it tells me the opposite of what it represents.

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
martimavocado: figured something like this would already exist, ty

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
martimavocado: can you check if the current implementation is more around what you had in mind

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,179 @@
...
+            createDisplay()
+            attemptClearDisplay()
+            displayDirty = true
+        }
+    }
+
+    @SubscribeEvent
+    fun onTick(event: LorenzTickEvent) {
+        if (!MiningAPI.inCustomMiningIsland()) return
+        if (hasFlowstate() == null) return
```
Thunderblade73: this will block the clear if you switch to another tool.

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,179 @@
...
+    @SubscribeEvent
+    fun onRenderOverlay(event: GuiRenderEvent.GuiOverlayRenderEvent) {
+        if (!MiningAPI.inCustomMiningIsland()) return
+        if (!config.flowstateHelper) return
+        if (hasFlowstate() == null) return
+
+        if (display == null || streakEndTimer.isInFuture()) createDisplay()
+        display?.let {
+            config.flowstateHelperPosition.renderRenderables(
+                listOf(Renderable.verticalContainer(it, 1)),
```
Thunderblade73: renderRendeables is equivlant to a verticalContainer. So you can just use it here and set extraSpace to 1.

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,179 @@
...
+                config.flowstateHelperAppearance.forEach {
+                    add(it.renderable)
+                }
+            }
+        } else {
+            createDisplayTimer()
+            display = buildList {
+                config.flowstateHelperAppearance.forEach {
+                    add(it.renderable)
+                }
```
Thunderblade73: this can be a single map operation instead of a buildList and a forEach.

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,179 @@
...
+            }
+        }
+    }
+
+    private fun createDisplayTitle(): Renderable {
+        FlowstateElements.TITLE.renderable = Renderable.string("§d§lFlowstate Helper")
+        return FlowstateElements.TITLE.renderable
+    }
```
Thunderblade73: This can be static. Just set it in the enum once. So there is never a need to update it.

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
Thunderblade73: Yeah it fine now. (with a minor logic error)

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,173 @@
...
+        "mining.pickobulus.blockdestroy",
+        "§7Your §r§aPickobulus §r§7destroyed §r§e(?<amount>\\d+) §r§7blocks!"
+    )
+
+    enum class FlowstateElements(val label: String, var renderable: Renderable) {
+        TITLE("§d§lFlowstate Helper", Renderable.string("§d§lFlowstate Helper")),
+        TIMER("Time Remaining: §b6.71", Renderable.string("")),
+        STREAK("Streak: §7234", Renderable.string("")),
+        SPEED("§6+600⸕", Renderable.string(""));
+
```
CalMWolfs:
```suggestion
        SPEED("§6+600⸕", Renderable.string("")),
        ;

```

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,173 @@
...
+            displayDirty = true
+        }
+    }
+
+    @SubscribeEvent
+    fun onTick(event: LorenzTickEvent) {
+        if (!MiningAPI.inCustomMiningIsland()) return
+
+        attemptClearDisplay()
+    }
```
CalMWolfs, line 64: does this need to be here now that there is a world swap event?

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,173 @@
...
+    fun onChangeItem(event: ItemInHandChangeEvent) {
+        hasChangedItem = true
+    }
+
+    @SubscribeEvent
+    fun onIslandChange(event: IslandChangeEvent) {
+        blockBreakStreak = 0
+        displayDirty = true
+        streakEndTimer = SimpleTimeMark.now()
+        createDisplay()
```
CalMWolfs: could you just set the streak end timer to now (or probably better farPast) then call attemptClearDisplay()

CalMWolfs (changes requested):
just a few things i saw (didnt test)
also could you call the hasFlowstate function on item change and then cache that value as flowstateCache instead of also having the hasChangedItem variable

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
martimavocado: well, the streak is supposed to expire after 10s without mining blocks, i don't think having only the world swap would be useful

CalMWolfs: > well, the streak is supposed to expire after 10s without mining blocks, i don't think having only the world swap would be useful

oh yeah that makes sense

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,168 @@
...
+            config.flowstateHelperPosition.renderRenderables(
+                it,
+                1,
+                "Flowstate Helper",
+                true
+            )
+        }
+    }
+
+    private fun createDisplay() {
```
hannibal002, line 96: the create functions dont use the return value. Either make use of the return function (via buildList() and also remove the renderable in the enum) or remove the return value

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,168 @@
...
+        if (displayDirty) {
+            displayDirty = false
+            createDisplayTimer()
+            createDisplayBlock()
+            createDisplaySpeed()
+        } else {
+            createDisplayTimer()
```
hannibal002: createDisplayTimer can be moved outside the if check

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,168 @@
...
+    private var displayDirty = false
+
+    private var flowstateCache: Int? = null
+
+    private val pickobulusPattern by RepoPattern.pattern(
+        "mining.pickobulus.blockdestroy",
+        "§7Your §r§aPickobulus §r§7destroyed §r§e(?<amount>\\d+) §r§7blocks!"
+    )
+
+    enum class FlowstateElements(val label: String, var renderable: Renderable) {
```
hannibal002: Rename the enum to GUIElements

hannibal002: renderable can use a default value so that timer, streak and speed dont need code duplication

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,168 @@
...
+        hasFlowstate()
+    }
+
+    @SubscribeEvent
+    fun onIslandChange(event: IslandChangeEvent) {
+        streakEndTimer = SimpleTimeMark.farPast()
+        attemptClearDisplay()
+    }
+
+    private fun hasFlowstate(): Int? {
```
hannibal002: remove return value or make us of it

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,168 @@
...
+        attemptClearDisplay()
+    }
+
+    @SubscribeEvent
+    fun onChat(event: LorenzChatEvent) { //TODO: Remove once #2540 is merged
+        if (!MiningAPI.inCustomMiningIsland()) return
+        if (flowstateCache == null) return
+
+        pickobulusPattern.matchMatcher(event.message) {
+            streakEndTimer = SimpleTimeMark.now().plus(10.seconds)
```
hannibal002: use `10.seconds.fromNow()` instead

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,168 @@
...
+            blockBreakStreak += group("amount").toInt()
+            createDisplay()
+            attemptClearDisplay()
```
hannibal002: why trying to clear the display immediately after creating it?

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,168 @@
...
+    @HandleEvent(onlyOnSkyblock = true)
+    fun onBlockMined(event: OreMinedEvent) {
+        if (!MiningAPI.inCustomMiningIsland()) return
+        if (flowstateCache == null) return
+
+        streakEndTimer = SimpleTimeMark.now().plus(10.seconds)
+        blockBreakStreak += event.extraBlocks.values.sum()
+        createDisplay()
+        displayDirty = true
+        attemptClearDisplay()
```
hannibal002: why trying to clear the display immediately after creating it?

src/main/java/at/hannibal2/skyhanni/config/features/mining/MiningConfig.java
```diff
@@ -96,4 +103,19 @@ public class MiningConfig {
...
+    @Expose
+    @ConfigOption(name = "Flowstate Helper", desc = "Shows stats for the Flowstate enchantment.")
+    @ConfigEditorBoolean
+    @FeatureToggle
+    public boolean flowstateHelper = false;
+
+    @Expose
+    @ConfigOption(name = "Flowstate Appearance", desc = "Drag text to change the appearance.")
+    @ConfigEditorDraggableList()
+    public List<FlowstateHelper.FlowstateElements> flowstateHelperAppearance = new ArrayList<>(FlowstateHelper.FlowstateElements.defaultOption);
```
hannibal002: You can remove the `new ArrayList<>(` wrap

src/main/java/at/hannibal2/skyhanni/config/features/mining/MiningConfig.java
```diff
@@ -96,4 +103,19 @@ public class MiningConfig {
     @ConfigEditorBoolean
     @FeatureToggle
     public boolean highlightYourGoldenGoblin = true;
+
+    @Expose
+    @ConfigOption(name = "Flowstate Helper", desc = "Shows stats for the Flowstate enchantment.")
+    @ConfigEditorBoolean
+    @FeatureToggle
+    public boolean flowstateHelper = false;
```
hannibal002: move flowstate into a new Accordion

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,168 @@
...
+            createDisplayBlock()
+            createDisplaySpeed()
+        } else {
+            createDisplayTimer()
+        }
+        display = config.flowstateHelperAppearance.map { it.renderable }
+    }
+
+    private fun createDisplayTimer(): Renderable {
+        var timeRemaining = streakEndTimer.minus(SimpleTimeMark.now()).inPartialSeconds
```
hannibal002: work with duration, not double. compare the time below with `< 0.seconds`. format the time for the rendererable with `timeRemaining.format()`. Optionaly define when millis should be shown or not.

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,168 @@
...
+        return FlowstateElements.TIMER.renderable
+    }
+
+    private fun createDisplayBlock(): Renderable {
+        val textColor = if (blockBreakStreak < 200) "§7" else "§f"
+        FlowstateElements.STREAK.renderable = Renderable.string("Streak: $textColor$blockBreakStreak")
+        return FlowstateElements.STREAK.renderable
+    }
+
+    private fun createDisplaySpeed(): Renderable {
```
hannibal002: those create... functions all look very similar. could maybe code cleanup them a bit by making a more abstract `FlowstateElements.create(thing)` extension function or similar.

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,168 @@
...
+        createDisplay()
+    }
+
+    @SubscribeEvent
+    fun onRenderOverlay(event: GuiRenderEvent.GuiOverlayRenderEvent) {
+        if (!MiningAPI.inCustomMiningIsland()) return
+        if (!config.flowstateHelper) return
+        if (flowstateCache == null) return
+
+        if (display == null || streakEndTimer.isInFuture()) createDisplay()
```
hannibal002: `createDisplay()` gets called every frame, even when there is nothing to show (no flowstate active). cache it so that it updates only 10 times a second at most, and ideally dont update at all when not in use (feature enabled but no enchantment on a tool)

hannibal002 (changes requested):
Not tested in game yet. 
Just a bunch of code cleanup suggestions. If you need help with anything please ask

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
martimavocado: was old code i forgot to cleanup

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
```diff
@@ -0,0 +1,180 @@
...
+    private var timeSinceHibernation = SimpleTimeMark.farPast()
+
+    private var flowstateCache: Int? = null
+
+    private val pickobulusPattern by RepoPattern.pattern(
+        "mining.pickobulus.blockdestroy",
+        "§7Your §r§aPickobulus §r§7destroyed §r§e(?<amount>\\d+) §r§7blocks!"
+    )
+
+    enum class GUIElements(val label: String, var renderable: Renderable = Renderable.string("")) {
```
CalMWolfs: This enum class name should specify that it is for flowstate helper

CalMWolfs (changes requested):
code looks fine apart from this enum name

src/main/java/at/hannibal2/skyhanni/features/mining/FlowstateHelper.kt
martimavocado: well, https://github.com/hannibal002/SkyHanni/pull/2561#discussion_r1770763611

CalMWolfs: > well, [#2561 (comment)](https://github.com/hannibal002/SkyHanni/pull/2561#discussion_r1770763611)

😭 im not a fan of typing GUIElements anywhere else in the mod and getting this random enum. We already have 2 classes named Item that it autocompletes before the minecraft one. But i guess it is up to @hannibal002

hannibal002: We have no real consistency over the project for those gui enum names.

I see your argument for avoiding using the same unspecific and abstract enum name for all those cases all over the project. But I don't think this is such a big deal.
When searching for flowstate, you will still find the enum quicky.

My argument against having the enum having the name flowstate is that the access (in the config class) looks like "FlowstateFeatures.FlowstateGUI".

This is just a visual problem, and there are different solutions to avoid it.
E.g. moving the "flowstate display element enum" into a separate class (not separate file) or using element imports in the config file, or using a abstract class name for all enums that have the same purpose (e.g. GUIElement) or something.

In the end, I'm not advocating for something specific here. If one of you thinks a different name, or generally a different naming convention for those config enum display classes is better, please use it.

I'm not overruling anything that makes sense and my suggestion earlier was not the only solution that makes sense.

CalMWolfs (approved):
code lgtm

hannibal002 (changes requested):
Works as expected. 

Couple sugestions:

<details>
<summary>Show max streak while not yet met</summary>

image

</details>

Show flowstate display even when not holding a mining tool in hand, while the time is still remaining. maybe make it an option (default enabled)

the "auto hide" option doesnt seem to work, also i dont think this is necessary as an seting. simply hide after the 10 seconds are over.

change the color of the time to blue, for consistency over the whole mod

hannibal002 (changes requested):
The display disappears after reaching 200 streaks, even when still actively mining.

martimavocado:
> The display disappears after reaching 200 streaks, even when still actively mining.

this is intended, in my opinion there isn't much of a point to keep showing it when it has no more useful info

ItsEmpa:
> > The display disappears after reaching 200 streaks, even when still actively mining.
> 
> this is intended, in my opinion there isn't much of a point to keep showing it when it has no more useful info

but i wanna see funny number go up

PR #2567, opened 2024-09-22
Fix: Flare Display negative values in chat with lag

Nealkitt:
## What
Fixes flare display spamming chat with negative with Negative numbers, now will stop at 0

<details>
<summary>Images</summary>

image

</details>

## Changelog Fixes
+ Fixed Flare Display spamming chat with negative values when the server is lagging. - Stella

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/combat/FlareDisplay.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/combat/FlareDisplay.kt
@@ -100,7 +100,7 @@ object FlareDisplay {
                     }
                 }
             }
-            if (remainingTime > 5.seconds) continue
+            if (remainingTime > 5.seconds || remainingTime < 0.seconds) continue
             val message = "$name §eexpires in: §b${remainingTime.inWholeSeconds}s"
             when (config.alertType) {
                 FlareConfig.AlertType.CHAT -> {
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/combat/FlareDisplay.kt
```diff
@@ -194,7 +194,6 @@ object FlareDisplay {
         SOS("§5SOS Flare", "+125%"),
         ALERT("§9Alert Flare", "+50%"),
         WARNING("§aWarning Flare", null),
-        ;
```
CalMWolfs, line 197: this doesnt serve much purpose now but i see no reason to remove it as its probably there due to ktlint before which is planned to be brough back in some sense in the future

CalMWolfs (changes requested):
skyhanni checks if the entity is alive before it sends this chat message, so while this fixes the problem of incorrect time being printed it does not provide any detail into why this time is incorrect. 
I'd much rather a full solution to this issue or debug logging than just removing incorrect chat messages so its harder to identify this issue in the future

src/main/java/at/hannibal2/skyhanni/features/combat/FlareDisplay.kt
hannibal002: we want trailing commas everywhere, please leave it

CalMWolfs (approved):

PR #2568, opened 2024-09-22
Backend: Add Helper Text to `/shtestmessage`

DavidArthurCole:
## What
[Here](<https://discord.com/channels/997079228510117908/1094190239532208228/1287433867657085110>).
Adds arg usage message when `/shtestmessage` is run without args.

exclude_from_changelog


CuzImClicks:
would tab completion for the (-clipboard etc.) be possible?

Thunderblade73:
> would tab completion for the (-clipboard etc.) be possible?

yeah. But I would suggest to not add this in this pr. Since I'm working on a new command tech that would make this trivial.

CalMWolfs (approved):

PR #2571, opened 2024-09-22
Backend: Carry Tracker Number format depending on Custom Scoreboard

j10a1n15:
## What
This Pull Request updates the CarryTracker to not use CustomScoreboard's .formatNum() method.

The todo attached to the method is already resolved in refactor, which causes refactor to not build anymore
image

exclude_from_changelog


hannibal002 (changes requested):
Please use `shortFormat()` instead of `addSeparators()`

PR #2577, opened 2024-09-23
Fix: Mining Commissions Blocks Color connected textures v2

NopoTheGamer:
## What
Make it so you can have your connected textures and this colouring feature

## Changelog Fixes
+ Fixed "Mining Commissions Block Color" causing OptiFine connected textures not to connect properly. - nopo


NopoTheGamer:
so the other pr made it so neu custom gemstones still worked, this one makes it so the texture actually are connected

PR #2582, opened 2024-09-23
Improvement: Add ability to use `/shedittracker` with `CorpseTracker`

DavidArthurCole:
## What
Because `CorpseTracker` uses separated logic and buckets, it didn't fit the generic `/shedititemtracker` logic. This PR adds separate logic for it.

## Changelog Improvements
+ Added ability to use `/shedittracker` on Corpse Profit Tracker. - Daveed


src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/CorpseTracker.kt
```diff
@@ -136,6 +140,20 @@ object CorpseTracker {
         }
     }
 
+    @SubscribeEvent
+    fun onItemAdd(event: ItemAddEvent) {
+        if (!isEnabled()) return
+        ProfileStorageData.profileSpecific?.mining?.mineshaft?.corpseProfitTracker?.let { trackerData ->
+            trackerData.getSelectedBucket()?.let {
+                tracker.addItem(it, event.internalName, event.amount)
+                ChatUtils.chat("Added ${event.internalName.itemName} §8x${event.amount} to Corpse Tracker §8(${it.displayName}§8)§e.")
```
hannibal002: this should be abstracted into the bucketed item tracker.
ideally the bucketed item tracker would reuse the logic from skyhann iitem tracker so that this unique chat message is not necessary in the first place.

src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/CorpseTracker.kt
```diff
@@ -136,6 +140,20 @@ object CorpseTracker {
...
+            } ?: ChatUtils.chat(
+                "§cYou do not have a Corpse Type selected, and cannot add loot to this tracker generically.\n" +
```
hannibal002: i think this will clash with the default answer messages from /shedittracker

src/main/java/at/hannibal2/skyhanni/utils/tracker/SkyHanniBucketedItemTracker.kt
```diff
@@ -36,7 +36,7 @@ class SkyHanniBucketedItemTracker<E : Enum<E>, BucketedData : BucketedItemTracke
         addItem(bucket, SKYBLOCK_COIN, coins)
     }
 
-    fun addItem(bucket: E, internalName: NEUInternalName, amount: Int) {
+    fun addItem(bucket: E, internalName: NEUInternalName, amount: Int, manuallyAdded: Boolean = false) {
```
hannibal002: as mentioned before, i recomment moving this into skyhanni item tracker, avoiding the code duplication alltogether

hannibal002 (changes requested):
not sure how easy this is to implement. but i dont like the code duplication

src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/CorpseTracker.kt
DavidArthurCole: We can't re-use the logic - they're entirely separate trackers. The item trackers just have the concept of one set of global "loot" that's added to - this is an enum driven base of buckets, the logic doesn't fit. The only way I could think to do this was using the currently-selected bucket.

Best I could do would be to abstract it back to the bucketed item tracker, however that message is going to end up having to be _very_ generic. We can't do it in `SkyhanniItemTracker` and expect it to work here - the `addItem` signatures are already overridden for `SkyhanniBucketedItemTracker`, as it takes a bucket as first param.

We also don't have the context of the `tracker` instance in the tracker itself, which would prevent actually moving this into the generic `SkyhanniBucketedItemTracker`, as we need to know the currently selected bucket;

```kotlin
ProfileStorageData.profileSpecific?.mining?.mineshaft?.corpseProfitTracker?.let {
}
```

hannibal002: The only way I could think to do this was using the currently-selected bucket.

I was not expecting anything more.

hannibal002: How about abstracting the existing item tracker class and creating a new "single bucketed tracker" for everything else?

hannibal002 (changes requested):
doesnt work for me
image
image

DavidArthurCole:
> doesnt work for me

https://youtu.be/osf3jJuzVOE

So, it is working, it just isn't updating until something else triggers the tracker. Which, seems weird, but also I have up until now been unable to fix this, and can't get it to update when `/shedittracker` is called....

hannibal002:
moving it t0 0.28 for now

DavidArthurCole:
I'm gonna close this out. I truly can't figure out what's going wrong with the `update()` not taking, and it doesn't seem to be high profile anyways.

PR #2588, opened 2024-09-25
Improvement: Option to ignore Spaceman in visitor shopping list

lunaynx:
## What
Added an option to exclude Spaceman from Visitor Shopping List.

## Changelog Improvements
+ Added an option to exclude Spaceman from the Visitor Shopping List. - Luna


hannibal002:
I would prefer if we add an option inside the visitor menu to hide that specific visitor for this time from the shopping list. then allow to unignore at the same time as well. also show an new tag "ignored" above the visitor

PR #2623, opened 2024-09-30
Improvement: Alert Clock Offsets

j10a1n15:
## What
This Pull Request adds a chat alert when a user with more than 10s time offset joins SkyBlock.
I wouldve loved if it could directly fix the time displays by doing some quirky stuff in SkyBlockTime and SimpleTimeMark, buttt uhh when I tried that it never really worked soo uhh :3 Surely this will do

<details>
<summary>Images</summary>

image

</details>

## Changelog Improvements
+ Added a warning in chat when the computer's time is inaccurate, along with a tutorial on how to fix it. - j10a1n15

+ Added Clock Offset Millis using an NTP server. - j10a1n15


src/main/java/at/hannibal2/skyhanni/utils/TimeOffset.kt
```diff
@@ -0,0 +1,56 @@
...
+    private fun getNtpOffset(ntpServer: String): Long? {
+        return try {
+            val client = NTPUDPClient()
+            val address = InetAddress.getByName(ntpServer)
+            val timeInfo = client.getTime(address)
+
+            timeInfo.computeDetails()
+            timeInfo.offset
+        } catch (e: Exception) {
+            e.printStackTrace()
```
hannibal002: please throw a skyhanni error in chat

src/main/java/at/hannibal2/skyhanni/utils/TimeOffset.kt
```diff
@@ -0,0 +1,56 @@
...
+            null
+        }
+    }
+
+    @SubscribeEvent
+    fun onProfileJoin(event: ProfileJoinEvent) {
+        if (abs(offsetMillis) > 10 * 1000) {
```
hannibal002: please use duration instead of long for math and formatting

src/main/java/at/hannibal2/skyhanni/utils/TimeOffset.kt
```diff
@@ -0,0 +1,56 @@
...
+import kotlin.time.toDuration
+
+@SkyHanniModule
+object TimeOffset {
+    var offsetMillis = 0L
+        private set
+
+    init {
+        SkyHanniMod.coroutineScope.launch {
+            offsetMillis = getNtpOffset("time.google.com") ?: 0
```
hannibal002: please use null instead of 0 to indicate a invalid result

src/main/java/at/hannibal2/skyhanni/utils/TimeOffset.kt
```diff
@@ -0,0 +1,56 @@
...
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+import org.apache.commons.net.ntp.NTPUDPClient
+import java.net.InetAddress
+import kotlin.math.abs
+import kotlin.math.absoluteValue
+import kotlin.time.DurationUnit
+import kotlin.time.toDuration
+
+@SkyHanniModule
+object TimeOffset {
```
hannibal002: please rename to ComputerTimeOffset or LocalTime or LocalComputerTime or something to be a bit more descriptive

src/main/java/at/hannibal2/skyhanni/utils/TimeOffset.kt
```diff
@@ -0,0 +1,56 @@
...
+        if (abs(offsetMillis) > 10 * 1000) {
+            val formatted = offsetMillis.toDuration(DurationUnit.MILLISECONDS).format()
+            ChatUtils.userError("Your computer's clock is off by $formatted. Please update your time settings.")
+        }
+    }
+
+    @SubscribeEvent
+    fun onDebugCollect(event: DebugDataCollectEvent) {
+        event.title("Time Offset")
+        event.addData("$offsetMillis ms")
```
hannibal002: only show the data if offset is greater than some value. e.g. 100ms?

hannibal002 (changes requested):
please add a toggle in the config to opt out

src/main/java/at/hannibal2/skyhanni/utils/TimeOffset.kt
j10a1n15: this runs during game boot, I'm pretty sure the user wouldn't even see the error

src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
```diff
@@ -0,0 +1,63 @@
...
+            null
+        }
+    }
+
+    @SubscribeEvent
+    fun onProfileJoin(event: ProfileJoinEvent) {
+        if (!SkyHanniMod.feature.misc.warnAboutPcTimeOffset) return
+        offsetMillis?.let {
+            if (it.absoluteValue > 5.seconds) {
+                ChatUtils.userError("Your computer's clock is off by ${it.format()}. Please update your time settings.")
```
CalMWolfs: We should either explain how to do this or link them to somewhere that explains it. Could detect which operating system they are on for more precise information. As I don't trust skyblock players to know what this means

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/TimeOffset.kt
CalMWolfs: > this runs during game boot, I'm pretty sure the user wouldn't even see the error

this is a correct assumption

src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
```diff
@@ -82,8 +84,8 @@ object ComputerTimeOffset {
         val expectedDuration = 1.seconds
         val deviation = timeDifference - expectedDuration
 
-        if (deviation.absoluteValue > 100.milliseconds && config) {
-            ChatUtils.chat("Offset changed from ${lastDetectedOffset.format()} to ${deviation.format()}")
+        if (deviation.absoluteValue > 1.seconds && config) {
```
hannibal002: deviation is normally around -1ms to 1ms. no need to increase the check range

src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
```diff
@@ -58,15 +59,16 @@ object ComputerTimeOffset {
...
         null
     }
 
-    var lastSystemTime = System.currentTimeMillis()
-    var lastDetectedOffset = 1.seconds
+    private var lastSystemTime = System.currentTimeMillis()
+    private var lastDetectedOffset = 1.seconds
 
     private fun startMonitoringTimeChanges() {
-        thread(start = true) {
```
hannibal002: why removing the start value? this will again never run

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
j10a1n15: image
start true if the default value

src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
j10a1n15: during joining hypixel I got 3 triggers that reached up to 0.7s

src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
```diff
@@ -0,0 +1,124 @@
...
+    private var lastSystemTime = System.currentTimeMillis()
+    private var lastDetectedOffset = 1.seconds
+
+    private fun startMonitoringTimeChanges() {
+        thread {
+            while (true) {
+                Thread.sleep(1000)
+                detectTimeChange()
+            }
+        }
```
CalMWolfs: I wouldve said to use a fixed rate timer here becauyse from testing it is sligtly more accurate but it seems they use the current time to schedule the next time it runs (so when i fixed my clock the fixed rate timer stopped triggering for a while 😆)

src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
CalMWolfs: > during joining hypixel I got 3 triggers that reached up to 0.7s

yeah this happens for me when i first click the multiplayer button, and then while im connecting and disconnecting from a server.

src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
```diff
@@ -0,0 +1,124 @@
...
+import kotlin.concurrent.thread
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
+
+@SkyHanniModule
+object ComputerTimeOffset {
+    var offsetMillis: Duration? = null
+        private set
```
CalMWolfs: i dont think this was ever used anywhere else so it can probably be a regular private variable unless there are still plans for this offset to be integrated elsewhere into the mod

src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
```diff
@@ -0,0 +1,124 @@
...
+    private fun detectTimeChange() {
+        val currentSystemTime = System.currentTimeMillis()
+        val timeDifference = (currentSystemTime - lastSystemTime).milliseconds
+        lastSystemTime = currentSystemTime
+
+        val expectedDuration = 1.seconds
+        val deviation = timeDifference - expectedDuration
+
+        if (deviation.absoluteValue > 1.seconds && config) {
+            ChatUtils.chat("Time Offset changed from ${lastDetectedOffset.format()} to ${deviation.format()}")
```
CalMWolfs: the offsetMillis variable should probably be updated here. Also maybe print the old offsetMillis and the new one found from the ntp as currently the message is kinda reversed.
[SkyHanni] Time Offset changed from 0.0s to -3m -55s
when it should really be changing from +3m 55s to 0s

src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @SubscribeEvent
+    fun onProfileJoin(event: ProfileJoinEvent) {
+        if (!config) return
+        val offsetMillis = offsetMillis ?: return
+        if (offsetMillis.absoluteValue < 5.seconds) return
+
+        ChatUtils.clickableLinkChat(
+            "Your computer's clock is off by ${offsetMillis.format()}. Please update your time settings. Click here for instructions.",
```
CalMWolfs: This here printed SkyHanni detected a time offset of -(3m 55.701s). When my clock is ahead by about 4 mins
image

While the way to find the proper time is SystemTime + offset, the value displayed to the user should either be flipped or just the absolute value

src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
```diff
@@ -0,0 +1,124 @@
...
+        )
+    }
+
+    @SubscribeEvent
+    fun onSecondPassed(event: SecondPassedEvent) {
+        if (lastCheckTime.passedSince() < 30.minutes) return
+
+        lastCheckTime = SimpleTimeMark.now()
+        checkOffset()
+    }
```
CalMWolfs: This can be replaced by just updating the value when there is a new offset detected in the detectTimeChange function

CalMWolfs (changes requested):
a few things i noticed while testing

src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
hannibal002: Therefore we are implementing the repeating async task ourselves instead of using a library 😆

src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
hannibal002: The idea was to repeatedly warn the user that their time is off. Some might miss the first message on server join

CalMWolfs: > Some might miss the first message on server join

#2630 would ensure they dont miss the message 😛 


But yeah then this event doesnt need to be removed (but the value should still be updated when a time change is detected)

hannibal002: What if their ram is off by more mb than their time by minutes?

NopoTheGamer: ?

hannibal002: > ?

Neu warns when a user is having more than 4gb of ram allocated in exactly this popup.

Sometimes my jokes are too weird, sorry 😐

j10a1n15:
Ive now updated the code to not send a chat message if a new offset is detected, and instead just refetch it

hannibal002 (approved):
LGTM

PR #2627, opened 2024-10-01
Fix: Strays Wrongly Counting as Unique

DavidArthurCole:
## What
`duplicate` was not being set at all during the stray section of the Hoppity API. This fixes that.

## Changelog Fixes
+ Fixed Fish the Rabbit and El Dorado incorrectly counting as unique rabbits during Hoppity's Hunt. - Daveed


lunaynx:
Shouldn't these be a RepoPattern?

PR #2637, opened 2024-10-01
Improvement: Add Suggestion Provider and guild support in tab complete

ThatGravyBoat:
## What
Adds suggestion provider to create suggestion trees to suggest appropriate arguments for commands.

## Changelog Improvements
+ Added support for Guilds in player-related tab completions. - ThatGravyBoat
+ Added support for all Guild and Friend commands in tab completions. - ThatGravyBoat

+ Added SuggestionProvider. - ThatGravyBoat
    * This new class simplifies building suggestion trees.


src/main/java/at/hannibal2/skyhanni/features/commands/tabcomplete/PlayerTabComplete.kt
```diff
@@ -72,6 +77,12 @@ object PlayerTabComplete {
                 }
             }
 
+            if ((config.guild || command == "guild" || command == "g") && PlayerCategory.GUILD !in ignoredCategories) {
```
hannibal002: why are there the "guild" and "g" checks? i dont see the reason to always tab complete the guild members on those two commands even when the config is disabled.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/commands/tabcomplete/PlayerTabComplete.kt
ThatGravyBoat: Tab completion is almost always wanted on the guild commands and not the other commands for guild players, all guild commands that take a player in guild require a guild member only besides 1 the invite.

Currently how the config and system is setup if you add a new category its added for every other command which is fine because some may want to use the other commands on their guild members but most wont but most will want to use it in the guild commands.

ThatGravyBoat: If you want an additional config option can be added for "Guild Commands" "Tab complete guild commands with guild members"

hannibal002 (approved):

PR #2638, opened 2024-10-02
Improvement: Slight Touch up for Hoppity Event Summary

DavidArthurCole:
## What
Fixed a formatting issue, and finally fixed the problem with the empty lines causing unnecessary spacing in the stats card.

## Changelog Improvements
+ Improved the formatting of the Hoppity Event Stats card. - Daveed


CalMWolfs (approved):
lgtm

on another note the formatting in this should probably switch over to our Text thing as that has methods to center text and create the ----- dividers.

hannibal002 (changes requested):
Use `Text.createDivider()`

DavidArthurCole:
> Use `Text.createDivider()`

I can't without refactoring the method. It only supports color or bold, and not both. Also, not sure why we'd change that here rather than a month ago when this was first put into beta.

PR #2650, opened 2024-10-03
Fix: Made it so you can shift click to disable features like it says

NopoTheGamer:
## What
Chat popups said "shift click or control click to disable" however shift clicking never worked
I do make this function always return false but I searched around on github and couldn't really find a mod that used this "Insertion" feature thingy that I'm disabling so I think its fine

<details>
<summary>Images</summary>

image

</details>

## Changelog Fixes
+ Fixed being unable to shift-click on chat messages. - nopo


CalMWolfs (approved):

hannibal002 (changes requested):
This breaks Skytils
grafik

NopoTheGamer:
> This breaks Skytils grafik

not true

hannibal002 (approved):
lieception

PR #2657, opened 2024-10-03
Backend: Detekt Fixes Part 6

DavidArthurCole:
## What
Fixes all instances of the following detektions:
- `Indentation`
- `NoMultipleSpaces`
- `SpacingAroundCurly`
- `VariableNaming`
- `NoNameShadowing`
- `ImplicitDefaultLocale`
- `UnsafeCallOnNullableType`
- `WrongEqualsTypeParameter`
- `ForbiddenVoid`
- `Wrapping` (big one, ~200 detektions)
- `UnnecessaryApply`
- `UnnecessaryFilter` 

Detektion count down to 315 from 642 at HEAD.

exclude_from_changelog


src/main/java/at/hannibal2/skyhanni/features/misc/items/enchants/EnchantsJson.kt
```diff
@@ -5,12 +5,15 @@ import com.google.gson.annotations.Expose
...
+    @Suppress("VariableNaming")
     var NORMAL: HashMap<String, Enchant.Normal> = hashMapOf()
 
     @Expose
+    @Suppress("VariableNaming")
     var ULTIMATE: HashMap<String, Enchant.Ultimate> = hashMapOf()
 
     @Expose
+    @Suppress("VariableNaming")
     var STACKING: HashMap<String, Enchant.Stacking> = hashMapOf()
```
CalMWolfs: personally I think these shouldnt be all uppercase and should follow normal naming conventions

CalMWolfs (changes requested):
just this one thing i saw that i thought could be different

src/main/java/at/hannibal2/skyhanni/features/misc/items/enchants/EnchantsJson.kt
DavidArthurCole: It's coming from the [REPO](https://github.com/DavidArthurCole/SkyHanni-REPO/blob/main/constants/Enchants.json), so if we want to change that, it's going to break on older versions of SH, so I don't think this should get changed?

hannibal002: Use `@SerializedName`

CalMWolfs (approved):

PR #2667, opened 2024-10-04
Backend: Detekt Fixes Part 7

DavidArthurCole:
## What
Fixes all instances of the following detektions:
- `MaxLineLength`

Detektion count down to 533 from 642 at HEAD.
There were likely some regressions in this, but it's really hard to tell without Part 6 merged.

exclude_from_changelog


CalMWolfs (approved):
looked good to me (and maybe felt like there was some duplication from part 6 but idk, not that that matters)

PR #2673, opened 2024-10-05
Backend: Add backup repo

CalMWolfs:
## What
Adds backup repo to the build. This is automatically included in every gradle build, (only adds .5mb)

This uses the exact same logic as neu uses

This aims to solve problems for people who play somewhere that github blocks them from downloading the repo, as currently all they can do is disable auto-update and manually download the repo. This results in them getting many new errors each time they update their mod version.
Repo patterns are also forced to be local while using the backup repo to avoid the edge case of a release being built before the github action to update the repo patterns completes.

This pr can be tested by deleteing your repo folder, setting your repo branch to something that isnt real and starting the game. Reading the logs will show it didnt download but you will have no repo errors and the backup repo will be copied into the repo folder.

## Changelog Improvements
+ Added backup Repo. - CalMWolfs
    * Added a backup repository, which provides a backup in case you are unable to download from GitHub. - CalMWolfs

+ Included backup repo in builds. - CalMWolfs


hannibal002 (changes requested):
to test his, ive renamed the host name github.com to something else inside RepoManager.kt and deleted my local repo.
This has worked, i can join skyblock and do not get error messages in chat.

But:
/shdebug does not show anything about that. It would be good if we somewhere store that the backup repo is in use, and indicate so in /shdebug.

Also, when running `/shupdaterepo` manually, i get an error message (as expected), but also the normal "updated successfully" message. maybe change the wording of that message to indicate that the internal backup repo is in use instead.
image

CalMWolfs:
> Also, when running `/shupdaterepo` manually, i get an error message (as expected), but also the normal "updated successfully" message. maybe change the wording of that message to indicate that the internal backup repo is in use instead.

This is nothing new with my pr, but i can fix it

hannibal002:
<details>
<summary>patch to break github</summary>

```
Index: src/main/java/at/hannibal2/skyhanni/data/repo/RepoManager.kt
IDEA additional info:
<+>UTF-8
diff --git a/src/main/java/at/hannibal2/skyhanni/data/repo/RepoManager.kt b/src/main/java/at/hannibal2/skyhanni/data/repo/RepoManager.kt
--- a/src/main/java/at/hannibal2/skyhanni/data/repo/RepoManager.kt(revision cb72825685b0f1670d77dc128a83aaf2025d4d53)
+++ b/src/main/java/at/hannibal2/skyhanni/data/repo/RepoManager.kt(date 1729079369334)
@@ -310,13 +310,13 @@
         val repoUser = config.location.user
         val repoName = config.location.name
         val repoBranch = config.location.branch
-        return "https://api.github.com/repos/$repoUser/$repoName/commits/$repoBranch"
+        return "https://api.notgithub.com/repos/$repoUser/$repoName/commits/$repoBranch"
     }
 
     private fun getDownloadUrl(commitId: String?): String {
         val repoUser = config.location.user
         val repoName = config.location.name
-        return "https://github.com/$repoUser/$repoName/archive/$commitId.zip"
+        return "https://notgithub.com/$repoUser/$repoName/archive/$commitId.zip"
     }
 
     @Throws(IOException::class)

```

</details>

CalMWolfs:
Fixed some issues related to the backup repo and did the requested changes. Also removed the shdebug from the RepoManager file as there was already repo information gathered in the DebugCommand file

hannibal002 (approved):

PR #2685, opened 2024-10-07
Fix: Custom Scoreboard Update time

hannibal002:
## What
Fixed custom scoreboard not updating when the hypixel scoreboard is not updating

## Changelog Fixes
+ Fixed the custom scoreboard not updating when the Hypixel scoreboard is not updating. - hannibal2


j10a1n15 (approved):
Looks fine

PR #2711, opened 2024-10-10
Fix: Remove color codes from hoppity eggs share message 

martimavocado:
## What
this was probably unintentional
`[SkyHanni] Lunch Chocolate Egg located at x: -38, y: 193, z: 35 (on the second highest rbMountain rdpeak)`
i'm not good enough at regex to know if this is fixable with a new pattern

## Changelog Fixes
+ Fixed the chocolate egg share message sometimes displaying the wrong location name. - martimavocado


CalMWolfs (approved):
i swear i fixed this 3 months ago, i guess its back?

PR #2729, opened 2024-10-12
Backend: TimeLimitedCache

ItsEmpa:
## What
Changes TimeLimitedCache to extend MutableMap instead of Iterator. Also removed some deprecated functions, and deprecated some other ones. Also removed the deprecation on .cancel() inside LorenzEvent

+ Changed `TimeLimitedCache` to extend `MutableMap` instead of `Iterator`. - Empa


hannibal002:
> Also removed the deprecation on .cancel() inside LorenzEvent

since the whole of lorenz event is deprecated, there is no need to use the function. this is just an additional marker that when using deprecated functions you want to migrate them over to skyhanni event, like every other instance of lorenz event as well

ItsEmpa:
> > Also removed the deprecation on .cancel() inside LorenzEvent
> 
> since the whole of lorenz event is deprecated, there is no need to use the function. this is just an additional marker that when using deprecated functions you want to migrate them over to skyhanni event, like every other instance of lorenz event as well

i feel like you should only actually be able to see that LorenzEvent is deprecated if you want to create a new event. if you want to cancel an already existing event (such as cancelling a message with LorenzChatEvent), you shouldnt be told that the function to cancel the event is deprecated.

src/main/java/at/hannibal2/skyhanni/utils/TimeLimitedSet.kt
```diff
@@ -26,13 +27,15 @@ class TimeLimitedSet<T : Any>(
...
+    operator fun minusAssign(element: T) {
+        remove(element)
+    }
 
     operator fun contains(element: T): Boolean = cache.containsKey(element)
 
     fun clear() = cache.clear()
 
-    fun toSet(): Set<T> = HashSet(cache.keys())
+    fun toSet(): MutableSet<T> = cache.keys
```
hannibal002: im not sure i like the idea of exposing mutable cache outside this class. this allows unintended behaviour

hannibal002 (changes requested):
lets discuss

src/main/java/at/hannibal2/skyhanni/features/fishing/FishingTimer.kt
```diff
@@ -51,7 +51,7 @@ object FishingTimer {
 
     private var mobsToFind = 0
 
-    private val recentMobs = TimeLimitedSet<Mob>(2.seconds)
+    private var recentMobs = TimeLimitedSet<Mob>(2.seconds)
```
lineargraph: Is there a reason to make those var?

hannibal002: After my commit the `object += value` requires this for the add function. Not sure if this is the correct solution though

ItsEmpa: reverted the changes in that commit that hanni made

hannibal002 (approved):

PR #2730, opened 2024-10-12
Improvement: Stop Time Tower Usage Warning spam

MTOnline69:
## What
Improves the logic of the "Time Tower Usage Warning" feature so that it only sends one message at a time.

For example, previously if you went offline with no charges and returned with three charges available, it would notify you for each charge and then warn about the timer tower being full. Now, it will _only_ give the warning.
<details>
<summary>Image</summary>

image
</details>

I had to test by changing my config manually since my time tower is empty at the moment.

## Changelog Improvements
+ Improved the Time Tower Usage Warning so it doesn't spam messages. - MTOnline


hannibal002 (approved):
code looks fine

PR #2735, opened 2024-10-14
Improvement: Added option to show the barn fishing timer anywhere.

NeoNyaa:
## What
Adds option to show the barn fishing timer anywhere, regardless of the player's current location.

The code change is simple, a new option which the user can choose to enable or disable. When `updateLocation()` is run, an if statement will check if the setting is enabled, if so it'll override the outcome of the code above it.

## Changelog Improvements
+ Added an option to always display the Barn Fishing Timer anywhere. - NeoNyaa

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/fishing/BarnTimerConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/fishing/BarnTimerConfig.java
@@ -23,6 +23,14 @@ public class BarnTimerConfig {
 
     @Expose
     @ConfigOption(
+        name = "Show Anywhere",
+        desc = "Show the Barn Fishing Timer whenever you fish up a sea creature, regardless of location."
+    )
+    @ConfigEditorBoolean
+    public Property<Boolean> showAnywhere = Property.of(true);
+
+    @Expose
+    @ConfigOption(
         name = "Worm Fishing",
         desc = "Show the Barn Fishing Timer in the Crystal Hollows."
     )
--- a/src/main/java/at/hannibal2/skyhanni/config/features/fishing/BarnTimerConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/fishing/BarnTimerConfig.java
@@ -27,7 +27,7 @@ public class BarnTimerConfig {
         desc = "Show the Barn Fishing Timer whenever you fish up a sea creature, regardless of location."
     )
     @ConfigEditorBoolean
-    public Property<Boolean> showAnywhere = Property.of(false);
+    public boolean showAnywhere = false;
 
     @Expose
     @ConfigOption(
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/config/features/fishing/BarnTimerConfig.java
```diff
@@ -21,6 +21,14 @@ public class BarnTimerConfig {
     @FeatureToggle
     public Property<Boolean> enabled = Property.of(true);
 
+    @Expose
+    @ConfigOption(
+        name = "Show Anywhere",
+        desc = "Show the Barn Fishing Timer whenever you fish up a sea creature, regardless of location."
+    )
+    @ConfigEditorBoolean
+    public Property<Boolean> showAnywhere = Property.of(true);
```
NopoTheGamer: i dont think this should be on by default

NeoNyaa: I left it as true by default due to the other entries being true by default as well. It's not an invasive option, so if a player doesn't like it, they can always disable it.

NeoNyaa (commented):
Upon further thinking, yea it should be false. Thanks for pointing it out

src/main/java/at/hannibal2/skyhanni/config/features/fishing/BarnTimerConfig.java
```diff
@@ -21,6 +21,14 @@ public class BarnTimerConfig {
     @FeatureToggle
     public Property<Boolean> enabled = Property.of(true);
 
+    @Expose
+    @ConfigOption(
+        name = "Show Anywhere",
+        desc = "Show the Barn Fishing Timer whenever you fish up a sea creature, regardless of location."
+    )
+    @ConfigEditorBoolean
+    public Property<Boolean> showAnywhere = Property.of(false);
```
martimavocado:
```suggestion
    public boolean showAnywhere = false;
```

src/main/java/at/hannibal2/skyhanni/features/fishing/FishingTimer.kt
```diff
@@ -179,6 +179,10 @@ object FishingTimer {
             IslandType.PRIVATE_ISLAND -> config.forStranded.get() && LorenzUtils.isStrandedProfile
             else -> false
         }
+        
+        if (config.showAnywhere.get() == true) {
+            rightLocation = true
+        }
```
martimavocado: adding a return and moving this to above the when block might also not be a bad idea

```suggestion
        if (config.showAnywhere == true) {
            rightLocation = true
        }
```

martimavocado (commented):
there's no need to wrap the config in a property in this case.

having a property allows you to listen for when the config option is changed, but this isn't used here

CalMWolfs (approved):
code works for me, in future you should try to make a new branch for features as if you want to make another before this is merged it will be slightly more difficult

NeoNyaa:
> code works for me, in future you should try to make a new branch for features as if you want to make another before this is merged it will be slightly more difficult

If that's favored, that should probably be stated in CONTRIBUTING.MD as so far it doesn't mention anything about making a new branch for proposed changed.

DavidArthurCole:
> If that's favored, that should probably be stated in CONTRIBUTING.MD as so far it doesn't mention anything about making a new branch for proposed changed.

It's not really that it's 'favored', it's just the logistics of version control - if you MR/PR from your `beta` branch, any other changes you propose are going to include any other changes you make...

PR #2738, opened 2024-10-14
Backend: Remove more neu dependent code !!!

NopoTheGamer:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/2725
- https://github.com/hannibal002/SkyHanni/pull/2630

## What
Remove more required places for neu

<details>
<summary>Images</summary>

image

</details>

+ Added PlatformUtils.isNeuLoaded(). - nopo
    * Moved NEU events to a wrapper.
    * Added the ability to require NEU to load a module.
    * Integrated the horse mixin from NEU. It is required for the game to launch.


CalMWolfs (approved):

hannibal002:
have not yet looked at the code changes in this pr.
just one thing i have noticed so far:

<details>
<summary>heavy pearl detectoin</summary>

image
Either this is not yet fixed in NEU 2.4.0, or neu's detection happens after SkyHanni's fix.
I will tomorrow try with the sh fix disabled, to see if neu on its own actually detects it.
(the reason I'm confused here is that SkyHanni already checks the last neu detection time, and only shows the chat message if the fix is necessary, and we detect it via sack change/item in inventory add, and this should happen very late, compared to neu's earlier solutions to detect the heavy pearl pickup.

</details>

NopoTheGamer:
> heavy pearl detection

it works i guess just not as fast

https://github.com/user-attachments/assets/1559c313-5002-48ee-ae39-c7b48f2ae012


it took so long because the box was a bit small I've made a pr to make it a bit bigger

hannibal002 (changes requested):
I can't reproduce the fix in NEU 2.4.0! Please revert the removal of the fix in SH.

I've waited 30 seconds after leaving the matriarch, and the heavy pearls are still on the to-do list.
image

hannibal002:
Update: When moving close to matriarch after picking it up, the to-do entry finally vanishes.
But the SkyHanni fix is still fixing NEU's detection not working without moving close.

CalMWolfs:
> Update: When moving close to matriarch after picking it up, the to-do entry finally vanishes. But the SkyHanni is still fixing NEU's detection not working without moving close.

I believe https://github.com/NotEnoughUpdates/NotEnoughUpdates/pull/1396 fixes it. But yeah, this should stay for now i guess. We can use the new requiresNeu annotation for this class so it shouldnt be too much of a problem, and then remove this at a later point

hannibal002 (changes requested):
As Long as the latest NEU version does not remove the TODO entry for me, the fix in SH stays. Please revert the removal for now @NopoTheGamer

CalMWolfs:
> As Long as the latest NEU version does not remove the TODO entry for me, the fix in SH stays. Please revert the removal for now @NopoTheGamer

pearls are back

hannibal002 (approved):

PR #2742, opened 2024-10-15
Fix: El Dorado HoppityAPI

DavidArthurCole:
## What
[Info here](https://discord.com/channels/997079228510117908/1295820860707311657).
Fixes an issue where finding El Dorado will not fire an event from the Hoppity API, due to not matching the 'normal' patterns that other strays follow ... god bless Hypixel.
This only should have impacted the Compact Hoppity option, as the stray tracker still takes the click event properly, as it has its own patterns.

## Changelog Fixes
+ Fixed El Dorado not receiving a compacted chat message. - Daveed


CalMWolfs (approved):
code lgtm

PR #2746, opened 2024-10-16
Fix: Fel Highlight in Edge Case

Thunderblade73:
## What
Mobs may become only later on invisble. So I added the check on each render call. Very much a detriment to performance but ok.

Also added a Invisible Info for \shcopyentities.

## Changelog Fixes
+ Fixed a rare case where invisible Fels were highlighted even though they shouldn't. - Thunderblade73


src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonMobManager.kt
```diff
@@ -83,7 +83,7 @@ object DungeonMobManager {
     fun onRenderWorld(event: LorenzRenderWorldEvent) {
         if (!fel.highlight.get()) return
         if (fel.line) {
-            felOnTheGround.filter { it.canBeSeen() }.forEach {
+            felOnTheGround.filter { it.canBeSeen() && it.distanceToPlayer() < 30 }.forEach {
```
ItsEmpa: is the distance check really needed? since mob detection already has a distance limit in which it works

Thunderblade73: The canBeSeen logic gets sketchy at high distance in complex terrain like dungeons. I had at least two instances where it would be unrealistic to see fels (1 to 2 pixels where only visible).

Thunderblade73: The issue happens when the mob gets detected in range and then you move away from the mob. Since then distances can go up to 60 blocks. The 30 is just a value I find appropriate since it is the usable view distance. Sure there are some cases where it is oversensitive (mostly at vertical rooms). Btw. 24 is the Range for the detection.

hannibal002: please add support for distance check inside the canBeSeen function

Thunderblade73: There is 80 blocks. And it is no Problem for anything else. Just the line inside dungeons.

hannibal002: what 80 blocks do you mean? is this a reply to my message above?

Thunderblade73: Yeah.

CalMWolfs (approved):

PR #2756, opened 2024-10-18
Backend: Fix code in PlayerChatManager

CalMWolfs:
The current code checks if  SystemMessageEvent or AbstractChatEvent is cancelled and then tries to cancel LorenzChatEvent if they are cancelled. 
But none of these 3 events are cancellable and would throw an error when they are attempted to be cancelled. As far as i could see, SystemMessageEvent and AbstractChatEvent are never actually attempted to be cancelled and only are blocked (like they should be)

exclude_from_changelog


hannibal002 (approved):
Good catch, thanks!

PR #2763, opened 2024-10-19
Backend: Update dev auth

CalMWolfs:
Update dev auth which in theory should mean that we have to authenticate less due to this commit 
https://github.com/DJtheRedstoner/DevAuth/commit/dc6641fda9edad6553aaa4be972877309190fbb5

exclude_from_changelog


CalMWolfs:
Replaced by #2764

PR #2771, opened 2024-10-20
Feature: Added Primal Fear solvers

mariethepolarfox:
## What
Added Math & Public Speaking Primal Fear solvers

## Changelog New Features
+ Added chat solvers for Primal Fears. - Helium9
    * Added solvers for Math and Public Speaking Primal Fears.

## Technical Details
+ Added `generateRandomString` function. - Helium9
+ Fixed spelling in Dev Repo section. - Helium9


src/main/java/at/hannibal2/skyhanni/config/features/dev/DebugConfig.java
```diff
@@ -21,6 +21,11 @@ public class DebugConfig {
     @ConfigEditorBoolean
     public boolean commandLogs = false;
 
+    @Expose
+    @ConfigOption(name = "Always in SkyBlock", desc = "Enables SkyHanni's features everywhere, not just SkyBlock.")
```
hannibal002: change the wording. since this is in dev category, you can be more technical. something like "this sets skyhannis skylbock detection to always true"

src/main/java/at/hannibal2/skyhanni/config/features/event/GreatSpookConfig.java
```diff
@@ -47,4 +48,14 @@ public class GreatSpookConfig {
...
+    // @Expose
+    // @ConfigOption(name = "Primal Fear Solver", desc = "Solver for the Math Primal Fear")
+    // @ConfigEditorBoolean
+    // @FeatureToggle
+    // public boolean primalFearSolver = false;
+
+    @ConfigOption(name = "Primal Fear Solvers", desc = "Solvers for the Primal Fears")
+    @Accordion
+    @Expose
+    public PrimalFearSolverConfig primalFearSolvers = new PrimalFearSolverConfig();
```
hannibal002: solver or solvers? inconcistent spelling

src/main/java/at/hannibal2/skyhanni/config/features/event/GreatSpookConfig.java
```diff
@@ -47,4 +48,14 @@ public class GreatSpookConfig {
     @ConfigLink(owner = GreatSpookConfig.class, field = "greatSpookTimeLeft")
     public Position positionTimeLeft = new Position(40, 40, false, true);
 
+    // @Expose
+    // @ConfigOption(name = "Primal Fear Solver", desc = "Solver for the Math Primal Fear")
+    // @ConfigEditorBoolean
+    // @FeatureToggle
+    // public boolean primalFearSolver = false;
+
+    @ConfigOption(name = "Primal Fear Solvers", desc = "Solvers for the Primal Fears")
```
hannibal002: desc ends with a dot please

src/main/java/at/hannibal2/skyhanni/config/features/event/PrimalFearSolverConfig.java
```diff
@@ -0,0 +1,20 @@
...
+import com.google.gson.annotations.Expose;
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorBoolean;
+import io.github.notenoughupdates.moulconfig.annotations.ConfigOption;
+
+public class PrimalFearSolverConfig {
+    @Expose
+    @ConfigOption(name = "Math Primal Fear Solver", desc = "Solver for the Math Primal Fear")
+    @ConfigEditorBoolean
+    @FeatureToggle
+    public boolean mathPrimalFear = false;
```
hannibal002: nno reason to repeat the name "primal fear" for the variable name in the class that already contains that name

src/main/java/at/hannibal2/skyhanni/config/features/event/PrimalFearSolverConfig.java
```diff
@@ -0,0 +1,20 @@
+package at.hannibal2.skyhanni.config.features.event;
+
+import at.hannibal2.skyhanni.config.FeatureToggle;
+import com.google.gson.annotations.Expose;
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorBoolean;
+import io.github.notenoughupdates.moulconfig.annotations.ConfigOption;
+
+public class PrimalFearSolverConfig {
+    @Expose
+    @ConfigOption(name = "Math Primal Fear Solver", desc = "Solver for the Math Primal Fear")
```
hannibal002: dot at the end of desc

src/main/java/at/hannibal2/skyhanni/config/features/event/PrimalFearSolverConfig.java
```diff
@@ -0,0 +1,20 @@
...
+    @ConfigEditorBoolean
+    @FeatureToggle
+    public boolean mathPrimalFear = false;
+
+    @Expose
+    @ConfigOption(name = "Public Speaking Primal Fear Solver", desc = "Solver for the Math Primal Fear")
```
hannibal002: check how this title looks in the config. the name is maybe too long and needs shorting

mariethepolarfox:
Let me know if you need elaboration on any feature!

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
```diff
@@ -53,11 +62,80 @@ object TheGreatSpook {
         if (isTimeLeftEnabled()) config.positionTimeLeft.renderString(displayTimeLeft, posLabel = "Time Left Display")
     }
 
+    private val mathFearMessagePattern by RepoPattern.pattern(
+        "chat.math",
+        "§d§lQUICK MATHS! §r§7Solve: §r§e(?<math>.*)",
```
hannibal002, line 71: please add a regex test to this regex pattern. search for "regex-test" in the project for templates to copy from

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
```diff
@@ -53,11 +62,80 @@ object TheGreatSpook {
...
+    )
+
+    private val speakingFearMessagePattern by RepoPattern.pattern(
+        "chat.speaking",
+        "§4\\[FEAR] Public Speaking Demon§r§f: (Speak|Say something interesting) (?<name>.*)!",
+    )
+
+    private fun solveMath(math: String): String {
+        val mathFormatted = math.replace("x", "*").replace(" ", "")
```
hannibal002: maybe those two replace checks also belong in the math utils?

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
```diff
@@ -53,11 +62,80 @@ object TheGreatSpook {
...
+        return try {
+            val result = MathUtils.evaluateExpression(mathFormatted)
+            result.toString()
+        } catch (e: Exception) {
+            "§c§lError"
```
hannibal002: instead of saying error, return null and handle it custom

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
```diff
@@ -53,11 +62,80 @@ object TheGreatSpook {
...
+        }
+    }
+
+    @SubscribeEvent
+    fun onChat(event: LorenzChatEvent) {
+        if (!isSolverEnabled()) return
+
+        val message = event.message
+
+        if (isMathSolverEnabled()) {
```
hannibal002: you can inline this check.

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
```diff
@@ -53,11 +62,80 @@ object TheGreatSpook {
...
+            val result = MathUtils.evaluateExpression(mathFormatted)
+            result.toString()
+        } catch (e: Exception) {
+            "§c§lError"
+        }
+    }
+
+    @SubscribeEvent
+    fun onChat(event: LorenzChatEvent) {
+        if (!isSolverEnabled()) return
```
hannibal002: instead check for if in sb here once, then just the config elements below. less confusing functions

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
```diff
@@ -53,11 +62,80 @@ object TheGreatSpook {
...
     private fun isFearStatEnabled(): Boolean = LorenzUtils.inSkyBlock && config.fearStatDisplay
     private fun isTimeLeftEnabled(): Boolean = LorenzUtils.inSkyBlock && config.greatSpookTimeLeft
 
-    private fun isAllDisabled(): Boolean = !isTimeLeftEnabled() && !isTimerEnabled() && !isFearStatEnabled() && !isNotificationEnabled()
+    private fun isSolverEnabled(): Boolean = isMathSolverEnabled() || isSpeakingSolverEnabled()
+
+    private fun isMathSolverEnabled(): Boolean = LorenzUtils.inSkyBlock && config.primalFearSolvers.mathPrimalFear
+    private fun isSpeakingSolverEnabled(): Boolean = LorenzUtils.inSkyBlock && config.primalFearSolvers.publicSpeakingPrimalFear
+
+    private fun isAllDisabled(): Boolean = !isTimeLeftEnabled() && !isTimerEnabled() && !isFearStatEnabled() &&
```
hannibal002: very confusing

src/main/java/at/hannibal2/skyhanni/utils/MathUtils.kt
```diff
@@ -0,0 +1,76 @@
...
+                        }
+                    )
+                }
+            }
+        }
+
+        return nums[0]
+    }
+
+    private fun String.isDouble(): Boolean {
```
hannibal002: this belongs in number utils. also there might be a better approach to check this, not sure

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -334,6 +334,11 @@ object StringUtils {
 
     fun generateRandomId() = UUID.randomUUID().toString()
 
+    fun generateRandomString(length: Int): String {
+        val chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
```
hannibal002: store the char array otuside the function so it does not get created on each function call

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -334,6 +334,11 @@ object StringUtils {
...
+        return (1..length).map { chars.random() }.joinToString("")
```
hannibal002: look at kotlin's `repeat()` function

hannibal002 (commented):
just some first thoughts

src/main/java/at/hannibal2/skyhanni/utils/MathUtils.kt
CalMWolfs, line 1: This whole file is replaced by NEUCalculator.calculateOrNull()

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
CalMWolfs: To add onyo this, the existing code before you made this feature was also bad but adding another 3 options was the cherry on top. Maybe look at how many other features have a global enabled value and implemetn that instead

src/main/java/at/hannibal2/skyhanni/config/features/dev/DebugConfig.java
CalMWolfs: I do not like this option as this check is here to stop crashes and other user facing errors. Being an option in the config allows any user to enable it and then they will complain. If a feature needs to work outside of skyblock we already have a place for that. Also I feel that this is so you can use the quickMaths feature outside of skyblock in something like the pit, which id consider a cheat anyways and would not approve for being in skyhanni

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/MathUtils.kt
hannibal002: neu independency? i see this file as the beginning of a "what if no neu is detected"

CalMWolfs: That is something to deal with before next halloween, but for now does not really matter and I dont see a point in making this new calculator that has a lot less features. 
The best option would be to take the neu calculator and put it into skyhanni like we do with Skyblock time. But this is out of the scope of this pr

src/main/java/at/hannibal2/skyhanni/utils/MathUtils.kt
mariethepolarfox: You already have a isDouble() check in NumberUtils, I removed my redundant function.

src/main/java/at/hannibal2/skyhanni/utils/MathUtils.kt
mariethepolarfox: I read you want to move away from NEU dependency, which is why I kept mine. The always in Skyblock toggle was mainly for testing reasons in private worlds, if you do not like it, I can remove it again. I will also look at the seccond suggestion

CalMWolfs: > I read you want to move away from NEU dependency

You are correct in this but in this case it is more appropriate to leave it as we already use the neu calculator elsewhere so when we solve that it will also solve this

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
mariethepolarfox: I return Strings, how would I return null?

hannibal002: You change the return value from `String` to `String?`. Then you can replace `"§c§lError"` with `null`. Then you can deal with the nullable in the caller place. (the other function where you were using the function´).

src/main/java/at/hannibal2/skyhanni/utils/MathUtils.kt
mariethepolarfox: also removed the isAllDisabled check as it was redundant

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
mariethepolarfox: Removed entirely as it isnt needed

mariethepolarfox:
Lmk if you have any other requests!

Obsidianninja11:
I feel like the descriptions aren't very good. Maybe something like `Sends a clickable message in chat to solve the [Math/Public Speaking] Primal Fear.`?
May be out of scope for this pr, but having a keybind for the solvers would be good (and then it could work for the commitment phobia)

mariethepolarfox:
NEEDS TESTING!! Im not gonna be home for a bit, can test this evening though.

mariethepolarfox:
Reverting keybind now, when reverted it will be ready for review

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
```diff
@@ -53,11 +60,70 @@ object TheGreatSpook {
...
+    )
+
+    @SubscribeEvent
+    fun onChat(event: LorenzChatEvent) {
+        if (isMathSolverEnabled()) {
+            mathFearMessagePattern.matchMatcher(event.message) {
+                val math = group("math")
+                val result = NEUCalculator.calculateOrNull(math)?.toInt()
+                DelayedRun.runNextTick {
+                    if (result != null) {
```
hannibal002: use `let` instead of `!= null`

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
```diff
@@ -53,11 +60,70 @@ object TheGreatSpook {
...
+                            "The result is: $result",
+                            option = config.primalFearSolver::math,
+                            actionName = "Send the result",
+                            action =
+                            {
+                                HypixelCommands.allChat(result.toString())
+                            },
+                        )
+                    } else {
+                        ChatUtils.chat("An unexpected error occurred while solving the math problem.")
```
hannibal002: use error manager and include the math string

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
```diff
@@ -53,11 +60,70 @@ object TheGreatSpook {
...
+    }
+
     private fun isTimerEnabled(): Boolean = LorenzUtils.inSkyBlock && config.primalFearTimer
 
     private fun isNotificationEnabled(): Boolean = LorenzUtils.inSkyBlock && config.primalFearNotification
     private fun isFearStatEnabled(): Boolean = LorenzUtils.inSkyBlock && config.fearStatDisplay
     private fun isTimeLeftEnabled(): Boolean = LorenzUtils.inSkyBlock && config.greatSpookTimeLeft
 
-    private fun isAllDisabled(): Boolean = !isTimeLeftEnabled() && !isTimerEnabled() && !isFearStatEnabled() && !isNotificationEnabled()
+    private fun isMathSolverEnabled(): Boolean = LorenzUtils.inSkyBlock && config.primalFearSolver.math
```
hannibal002: remove the two functions: check if in sb only once in the event and inline the two feature checks

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
mariethepolarfox: only my 2? or every function

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
```diff
@@ -44,20 +52,76 @@ object TheGreatSpook {
...
-    private fun isNotificationEnabled(): Boolean = LorenzUtils.inSkyBlock && config.primalFearNotification
-    private fun isFearStatEnabled(): Boolean = LorenzUtils.inSkyBlock && config.fearStatDisplay
-    private fun isTimeLeftEnabled(): Boolean = LorenzUtils.inSkyBlock && config.greatSpookTimeLeft
+    @SubscribeEvent
+    fun onChat(event: LorenzChatEvent) {
+        if (!LorenzUtils.inSkyBlock) return
 
-    private fun isAllDisabled(): Boolean = !isTimeLeftEnabled() && !isTimerEnabled() && !isFearStatEnabled() && !isNotificationEnabled()
+        if (config.primalFearSolver.math) {
+            val math = mathFearMessagePattern.matchMatcher(event.message) { group("math") } ?: return
```
hannibal002: since the logic is moved into mathSolver, the explicit variable math can be removed, reverted to the state above, something like: mathsolver(group("math")))

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
```diff
@@ -44,20 +52,76 @@ object TheGreatSpook {
...
+
+    /**
+     * REGEX-TEST: §4[FEAR] Public Speaking Demon§r§f: Speak PlasticEating!
+     */
+    private val speakingFearMessagePattern by RepoPattern.pattern(
+        "chat.speaking",
+        "§4\\[FEAR] Public Speaking Demon§r§f: (Speak|Say something interesting) (?<name>.*)!",
+    )
+
+    private fun mathSolver(math: String?) {
```
hannibal002: maybe rename math to query, term, question, task, or something. math is too abstract imo

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
```diff
@@ -44,20 +52,76 @@ object TheGreatSpook {
...
+        val result = math?.let { NEUCalculator.calculateOrNull(it)?.toInt() } ?: run {
+            ChatUtils.userError("Failed to solve $math!")
+            return
+        }
+        ChatUtils.clickToActionOrDisable(
+            "The result is: $result",
```
hannibal002: maybe rename result to answer. not sure though

src/main/java/at/hannibal2/skyhanni/features/event/spook/TheGreatSpook.kt
```diff
@@ -44,20 +52,76 @@ object TheGreatSpook {
...
+            action = {
+                HypixelCommands.allChat(result.toString())
+            },
+        )
     }
 
-    private fun isTimerEnabled(): Boolean = LorenzUtils.inSkyBlock && config.primalFearTimer
+    private fun publicSpeakingSolver() {
+        ChatUtils.clickToActionOrDisable(
+            "Click to send a random string to complete the Primal Fear",
```
hannibal002: remove the phrase "to send a random string" from this line. otherwise the line gets too long also the whole chat message contains the information twice atm

hannibal002 (changes requested):

hannibal002 (approved):

PR #2775, opened 2024-10-20
Backend: Add custom import ordering detekt rule

CalMWolfs:
## What
Added a custom import ordering detekt rule. This works with pre-processed imports so that you no longer have to either supress the file or add it to the baseline.
I also made an enum PreprocessingPattern that can be used across detekt rules. As im sure we will run into many more cases that require custom detekt rules or something similar

+ Added a custom import ordering Detekt rule. - CalMWolfs
	* This rule works with preprocessing imports


detekt/src/main/kotlin/imports/CustomImportOrdering.kt
```diff
@@ -0,0 +1,120 @@
...
+        val formattedOriginal = originalImports.joinToString("\n") { it }
+
+        val expectedImports = imports.sortedWith(importOrder).map { "import ${it.importPath}" }
+        val formattedExpected = expectedImports.filter { !linesToIgnore.contains(it) }.joinToString("\n")
+
+        val isExpected = formattedOriginal == formattedExpected
+        if (!isExpected) {
+            println("Original imports:\n\"$formattedOriginal\"")
+            println("Expected imports:\n\"$formattedExpected\"")
+        }
```
DavidArthurCole: I'd remove this just so people aren't getting erroneous output in gradle if we ever use detekt logging as an output (or Report() this as a separate issue if it's truly an edge case)

DavidArthurCole (approved):
one note for the printlns, rest of this lgtm

DavidArthurCole (approved):

PR #2783, opened 2024-10-21
Fix: Custom Scoreboard Printing Class Name

DavidArthurCole:
## What
[Reported here](https://discord.com/channels/997079228510117908/1297726116655529997).
Fixes an issue where the class name of `PartyAPI` was being printed instead of the value it was meant to reference.

## Changelog Fixes
+ Fixed party leader not displaying correctly in the Custom Scoreboard. - Daveed


j10a1n15 (approved):
🙏

PR #2797, opened 2024-10-23
Remove: Feature toggle from focus mode

CalMWolfs:
## What
removed the feature toggle from focus mode because people keep turning it on by accident

Also added search tags so people can find it more easily when accidentily enabled

exclude_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/inventory/FocusModeConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/inventory/FocusModeConfig.java
@@ -1,6 +1,5 @@
 package at.hannibal2.skyhanni.config.features.inventory;
 
-import at.hannibal2.skyhanni.config.FeatureToggle;
 import com.google.gson.annotations.Expose;
 import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorBoolean;
 import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorKeybind;
@@ -12,7 +11,6 @@ public class FocusModeConfig {
     @Expose
     @ConfigOption(name = "Enabled", desc = "In focus mode you only see the name of the item instead of the whole description.")
     @ConfigEditorBoolean
-    @FeatureToggle
     public boolean enabled = false;
 
     @Expose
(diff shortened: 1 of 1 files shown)


hannibal002 (changes requested):
I'm against removing feature toggles because its accidentally enabled. following this logic we could start removing the feature toggle on every slightly "annoying feature when accidentally enabled".
users need to learn to read what the feature does when enabling everything.

The serach tag is fine

CalMWolfs:
> I'm against removing feature toggles because its accidentally enabled. following this logic we could start removing the feature toggle on every slightly "annoying feature when accidentally enabled".

While I agree with this as a general rule I feel that this feature is an exception. Imagine you have just installed this mod or are updating one full version, you want to turn on all features to see what the mod has to offer and suddenly every item doesn't display properly. I feel this im would be a very negative experience. I'd like you to reconsider this specific case

lunaynx:
Yes, I agree. This isn't just a "potentially annoying feature", it literally hides the description of every item with no obvious way out. It should absolutely not be enabled by an "enable all features" button. That's almost as bad as making it enable the "crash on death" feature.

ItsEmpa:
Feature Toggles shouldnt be used for stuff like troll/dev/intrusive features yeah
specially in cases where the user might not know how to call the feature, since that makes it really hard for them to find which mod its coming from, and which feature it is

hannibal002:
I also don't like the problems this currently causes.
Still, not having this feature with a feature toggle would break consistency and logic, since this IS a standalone feature!

Instead, I provide an alternative solution:
have the toggle state default disabled:
https://github.com/hannibal002/SkyHanni/blob/7d3e61c27663d7e78ff211224f6bac416465cd54/src/main/java/at/hannibal2/skyhanni/features/inventory/FocusMode.kt#L18

Then add a new second line directly below the item name (both, when lore is hidden or not)
"press x to show/hide the item's lore!"

This way, the feature will no longer cause confusion, since:
1. You need to select a keybind to use the feature
2. Even when you select a key, and press it on accident, you instantly see how to disable it again.

Then add a third toggle in the config that hides the added second line in the lore when enabled, and this toggle is default disabled and has no feature toggle annotation

The only downside i see with this change is that the feature seems to be broken when you previously just enabled it to be “always enabled” and don't want to use a keybind. But since you should actually use a keybind, forcing users to select one to use the feature is a great way to avoid having the feature unintentionally active.

What do you think? @CalMWolfs @qtlunya @ItsEmpa

CalMWolfs:
> Instead, I provide an alternative solution:


I mean I still prefer just removing the toggle, but if thats not a possibility then your proposed solution is fine. But if this new logic is not implemented by the time we are doing a full release the toggle should be removed

hannibal002:
Replaced by 
- #2844

PR #2808, opened 2024-10-24
Fix: Skill Overflow

superhize:
## What
Fix skill overflow calculation

## Changelog Fixes
+ Fixed Overflow Level/XP calculation being incorrect. - HiZe
    * Hypixel now shows the correct amount of overflow XP when max level is reached.


hannibal002 (changes requested):
please use neu repo instead of duplicating the data into sh repo: https://github.com/hannibal002/SkyHanni-REPO/pull/275#issuecomment-2433914893

also please change the changelog wording to clearly state what was wrong before/how this is now better

hannibal002 (approved):

PR #2811, opened 2024-10-24
Backend: getFormattedTextCompat method

lineargraph:
This method allows getting 1.8.9 like formattedtext on 1.21

+ Fixed getFormattedTextCompat. - !nea
+ Added buildpaths.txt to limit which files are attempted to be built on 1.21. - !nea
    * This file should be populated over time with known good files to enable compilation on 1.21.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -25,7 +25,7 @@ import kotlin.io.path.outputStream
 plugins {
     idea
     java
-    id("com.github.johnrengelman.shadow") version "7.1.2"
+    id("com.gradleup.shadow") version "8.3.4"
     id("gg.essential.loom")
     id("dev.deftu.gradle.preprocess")
     kotlin("jvm")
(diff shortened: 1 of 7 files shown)


build.gradle.kts
```diff
@@ -275,6 +275,21 @@ if (target == ProjectTarget.MAIN) {
...
+                .map { it.substringBefore("#").trim() }
+                .filter { it.isNotBlank() }
+                .toSet()
+            kotlin.include(buildPaths)
+            java.include(buildPaths)
+        }
+    }
+}
+includeBuildPaths(file("buildpaths.txt"), sourceSets.main)
+includeBuildPaths(file("buildpaths-test.txt"), sourceSets.test)
```
CalMWolfs, line 293: I'm probably doing something wrong but this causes me to get the following error when building with
`skyhanni.multi-version=compile` in my .gradle/private.properties
I've tried with both java 8 and java 21

Caused by: java.lang.IllegalArgumentException: Unsupported class file major version 65

https://hst.sh/qebuhixavu.properties

CalMWolfs (approved):

PR #2815, opened 2024-10-25
Backend: Bits Event

ItsEmpa:
## What
(Will be needed for [#2813](https://github.com/hannibal002/SkyHanni/pull/2813))
This pr changes the bits spent event to have difference, and separates the bits available gained event into a different event.
It also moves famerank to playerstorage instead of globally, and makes it not get stored by its display name

+ Added difference tracking to Bits Spent event and separated Bits Available Gained event into its own event. - Empa


hannibal002 (changes requested):
image

PR #2819, opened 2024-10-26
Fix: Catch JSON Parse Errors in APIUtils

ptlthg:
## What
This fixes getting errors from syncing contests with elitebot.dev because SkyHanni tries reading a string response as JSON. [See error here](https://discord.com/channels/997079228510117908/1000669238035497022/1299559940251324527).

The "fix" here is to just catch the parse error and return an empty JsonObject when reading the response of a POST request in APIUtils. Because the previous behavior was to error out and the string responses from my API don't need to be looked at, I don't expect this to cause any problems.

**Untested** - This is small change and I honestly don't want to figure out how to get SkyHanni to build since the last time I did a PR, so yeah.

## Changelog Fixes
+ Fixed API error when sending Jacob Contests. - Ke5o

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/APIUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/APIUtils.kt
@@ -181,9 +181,16 @@ object APIUtils {
 
     private fun readResponse(entity: HttpEntity): JsonObject {
         val retSrc = EntityUtils.toString(entity) ?: return JsonObject()
-        val parsed = parser.parse(retSrc)
-        if (parsed.isJsonNull) return JsonObject()
-        return parsed as JsonObject
+
+        try {
+            val parsed = parser.parse(retSrc)
+            if (parsed.isJsonNull) return JsonObject()
+
+            return parsed as JsonObject
+        } catch (_: Throwable) {
+            // This causes content types that aren't JSON to be ignored
+            return JsonObject()
+        }
     }
 
     fun postJSONIsSuccessful(url: String, body: String, silentError: Boolean = false): Boolean {


CalMWolfs (approved):

hannibal002 (changes requested):
I don't like this workaround, this will show a wrong success message further up the stack:

https://github.com/hannibal002/SkyHanni/blob/3a7762517230fa716d49c7efafc3d4f8482c567c/src/main/java/at/hannibal2/skyhanni/utils/APIUtils.kt#L155

Waybe we should instead make `readResponse` return null and handle it in `postJSON`, specifying that the format is no valid JSON?

ptlthg:
> I don't like this workaround, this will show a wrong success message further up the stack:

Right now this method is only used in one place, when a 200-299 response code is detected. I agree it's not a great workaround, but the request is still a success, no matter if reading the response works correctly.

hannibal002:
An empty response is not the same as an invalid JSON response, IMO. Especially since we return a JSON object, if we can't parse it, it should not silently assume its correct.

ptlthg:
The issue here is that this is just for reading the response of a POST call. We are already determining that the call was successful from the status code, but SkyHanni is erroring out and treating the HTTP request like it failed because it's not parsing the response properly.

There can definitely be better handling for parsing the response body, but the information of "did this post request succeed" doesn't depend on the response body being parsed properly.

I just changed the Elite API so the original error likely won't happen again though https://github.com/EliteFarmers/API/commit/7e521d9e8e9333c26f8ce45a6686588e00e0c6a1

hannibal002 (approved):

PR #2825, opened 2024-10-26
Fix: update dungeon potion regex

catgirlseraid:
## What
Update the dungeon potion regex as hypixel adds a x1 onto the end when in a menu where the potion is able to be sold.

bug reported in discord: https://discord.com/channels/997079228510117908/1296714423301312532

## Changelog Fixes
+ Fixed "Dungeon Potion level as stack size" not working in shop menus. - phoebe


CalMWolfs (changes requested):
please add a regex test

CalMWolfs (approved):

PR #2827, opened 2024-10-26
Fix: Ore Event Pure Ores

ItsEmpa:
## What
This PR fixes most pure ores not being detected when mined.
test

## Changelog Fixes
+ Fixed not detecting mining most "pure ores". - Empa


src/main/java/at/hannibal2/skyhanni/features/mining/OreBlock.kt
```diff
@@ -255,7 +250,11 @@ enum class OreBlock(
     ),
     ;
 
+    fun hasInitSound() = this !in oresWithoutInitSound
```
hannibal002: this could be a val or lazy

hannibal002 (changes requested):
works for me in game

hannibal002 (approved):

PR #2828, opened 2024-10-26
Backend: CurrencyChangeEvent

ItsEmpa:
## Dependencies
- #2765 

## What
Adds a CurrencyChangeEvent for copper, motes and north stars (didnt add purse/bits because those already have their own events). Also stores in profile storage all currencies.

These events will be used for adding these currencies to Item Pickup Log and for the custom scoreboard performance improvement pr im working on.

+ Added CurrencyChangeEvent. - Empa


CalMWolfs:

PR #2829, opened 2024-10-26
Improvement: Carry Tracker Remove

ItsEmpa:
## What
Makes it so that you can actually remove people from the carry tracker

## Changelog Improvements
+ Added the ability to remove people from the Carry Tracker. - Empa


hannibal002 (changes requested):
image
change the changelog to improvement, and indicate that this now additionally to the gui also works via command

hannibal002 (approved):

PR #2836, opened 2024-10-27
Fix: Downloading forge multiple times

CalMWolfs:
Fix the shots causing forge to be downloaded each time build is ran

image

exclude_from_changelog


lineargraph:
Hi just wanted to point out: while i did tell you to override hashcode and equals (as well as serializable), a data class already generates those for you, wo one of those wouldve worked also.

PR #2850, opened 2024-10-28
Backend: MaxwellAPI

ItsEmpa:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/277

## What
This PR refactors the MaxwellAPI to make it use MaxwellPower data classes instead of just storing the name, to allow for adding more data to powers in the future. Also adds Update events for power, mp and tunings, which will be used for custom scoreboard (soon™️).

+ Refactored MaxwellAPI to use MaxwellPower data classes instead of strings. - Empa


src/main/java/at/hannibal2/skyhanni/data/maxwell/MaxwellAPI.kt
```diff
@@ -31,35 +35,55 @@ import java.util.regex.Pattern
...
 
+    fun getPowerByInternalNameOrNull(name: String): MaxwellPower? = powers[name]
+
+    val NO_POWER = MaxwellPower("No Power", "NO_POWER")
+
+    private var powers = mapOf<String, MaxwellPower>()
+    private const val THAUMATURGY_TUNINGS_SLOT = 51
+    private const val THAUMATURGY_MP_SLOT = 48
+
+    //region Patterns
```
DavidArthurCole: Not saying `//region` is explicitly wrong, but parts of the code base are already using `// <editor-fold>` & `// </editor-fold>`, which can be utilized with the current rules intact, and this way we wouldn't have to add extra edge-case detection to the CustomCommentSpacing rule.

CalMWolfs:

PR #2853, opened 2024-10-28
Backend Improvement: Companionize Detekt Rule Regex

DavidArthurCole:
## What
We all say "thank you Empa 🥺"

exclude_from_changelog


ItsEmpa:
🥺

PR #2855, opened 2024-10-28
Backend: Kotlin Position

ItsEmpa:
## What
This PR translates the Position class into kotlin, and changes it to use more kotlin getters/setters, and cleaner code overall.
Also makes use of overloaded constructors to reduce the amount of them that have to be declared in code

+ Translated Position Class to Kotlin. - Empa


src/main/java/at/hannibal2/skyhanni/config/core/config/Position.kt
```diff
@@ -0,0 +1,185 @@
...
+import com.google.gson.annotations.Expose
+import io.github.notenoughupdates.moulconfig.annotations.ConfigLink
+import io.github.notenoughupdates.moulconfig.gui.GuiScreenElementWrapper
+import net.minecraft.client.Minecraft
+import net.minecraft.client.gui.ScaledResolution
+import java.lang.reflect.Field
+
+class Position @JvmOverloads constructor(
+    x: Int,
+    y: Int,
```
CalMWolfs, line 38:
```suggestion
    x: Int = 0,
    y: Int = 0,
```

in the java one these were also defaulted to 0

src/main/java/at/hannibal2/skyhanni/config/core/config/PositionList.kt
```diff
@@ -9,10 +9,9 @@ class PositionList() : ArrayList<Position>(), MutableList<Position> {
     }
 
     constructor(size: Int) : this() {
-        this.addAll((0..<size).map { Position() })
+        this.addAll(List(size) { Position(0, 0) })
```
CalMWolfs:
```suggestion
        this.addAll(List(size) { Position() })
```

src/main/java/at/hannibal2/skyhanni/features/gui/TabWidgetDisplay.kt
```diff
@@ -72,7 +72,7 @@ enum class TabWidgetDisplay(private val configName: String?, vararg val widgets:
                     "Positions" to config.displayPositions,
                 )
             } else {
-                config.displayPositions.addAll(generateSequence { Position() }.take(sizeDiff))
+                config.displayPositions.addAll(List(sizeDiff) { Position(0, 0) })
```
CalMWolfs:
```suggestion
                config.displayPositions.addAll(List(sizeDiff) { Position() })
```

src/main/java/at/hannibal2/skyhanni/config/core/config/Position.kt
```diff
@@ -0,0 +1,185 @@
...
+    scale: Float = DEFAULT_SCALE,
+    centerX: Boolean = false,
+    centerY: Boolean = false,
```
CalMWolfs: while this isnt used currently, it was default true before so we should probably keep consistancy, same can be said for this value in the constructor below

CalMWolfs (changes requested):
just a few things i saw,
also changing the center value to just be whatever the centerX value is set to is fine imo, but 4 options will no longer be centered because the old config will not migrate to the new constructor on its own (CF: customReminder, fishingHookDisplay and 2 blaze slayer options). These should probably be adressed

CalMWolfs (approved):
worked for me

hannibal002 (changes requested):
When moving GUIs in position editor, some jump to the left corner of the screen and cant get moved away, but not all.

image


my guess is this is realted to the centerX value?

hannibal002:
Can confirm, its still this PR that causes it (after testing with latest beta branch where the bug was not there)

src/main/java/at/hannibal2/skyhanni/config/core/config/Position.kt
```diff
@@ -0,0 +1,202 @@
...
+        if (adjustedAxis < 0) {
+            adjustedDelta -= adjustedAxis
+            adjustedAxis = 0
+        } else if (adjustedAxis + objLength > length) {
+            adjustedDelta += length - adjustedAxis - objLength
+            adjustedAxis = length - objLength
+        }
+
+        return adjustedAxis to adjustedDelta
+    }
```
Thunderblade73: This function does not match up with the moveX/moveY functions. The wasPositive check and differentiation is lost here. Furthermore the bounds edge case is also not here.
Those missing things probably account for the edge case that hannibal encounters.

Thunderblade73 (changes requested):

hannibal002 (approved):
works for me now

PR #2856, opened 2024-10-28
Backend: `messageHistory` Length Config Option

DavidArthurCole:
## What
Adds a dev config option to allow expanding the length of chat history that is kept in `messageHistory`.

+ Added dev config option for expanding Chat History length. - Daveed
   + Careful with large values :)

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/dev/DevConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/dev/DevConfig.java
@@ -36,6 +36,15 @@ public class DevConfig {
     public int logExpiryTime = 14;
 
     @Expose
+    @ConfigOption(
+        name = "Chat History Length",
+        desc = "The number of messages to keep in chat history.\n" +
+            "§cExcessively large values may cause memory allocation issues."
+    )
+    @ConfigEditorSlider(minValue = 100, maxValue = 5000, minStep = 10)
+    public int chatHistoryLength = 100;
+
+    @Expose
     @ConfigOption(name = "Slot Number", desc = "Show slot number in inventory while pressing this key.")
     @ConfigEditorKeybind(defaultKey = Keyboard.KEY_NONE)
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/config/features/dev/DevConfig.java
```diff
@@ -35,6 +35,15 @@ public class DevConfig {
     @ConfigEditorSlider(minValue = 1, maxValue = 30, minStep = 1)
     public int logExpiryTime = 14;
 
+    @Expose
+    @ConfigOption(
+        name = "Chat History Length",
+        desc = "The number of messages to keep in chat history.\n" +
+            "§cExcessively large values may cause memory allocation issues."
```
j10a1n15, line 42: it should probably mention that its related to shchathistory somewhere

j10a1n15 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/ChatManager.kt
```diff
@@ -41,7 +43,7 @@ object ChatManager {
             override fun removeEldestEntry(
                 eldest: MutableMap.MutableEntry<IdentityCharacteristics<IChatComponent>, MessageFilteringResult>?,
             ): Boolean {
-                return size > 100
+                return size > config.chatHistoryLength
```
j10a1n15:
```suggestion
                return size > config.chatHistoryLength.coerceAtLeast(0)
```

kinda useless but since you can put any number into that field, its probably for the better? idk depends on what others think

j10a1n15 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/ChatManager.kt
j10a1n15: eh it probably doesnt matter actually

j10a1n15 (approved):

CalMWolfs (approved):

PR #2859, opened 2024-10-29
Improvement: Option for optimal speed warning when not wearing rancher's

Obsidianninja11:
## What
Made optimal speed warning warn when not wearing rancher's boots.
It will send a message that doesn't have the clickable /setmaxspeed when not wearing rancher's boots. (it still has click to disable though)

## Changelog Improvements
+ Added an option to send the optimal speed warning even when not wearing Rancher's Boots. - Obsidian


jani270:
How is this a bug fix? This was/is the intended behavior of the feature.

Obsidianninja11:
> How is this a bug fix? This was/is the intended behavior of the feature.

Seems like Hanni accidentally disabled it entirely when not wearing rancher's
see #2413

jani270:
> > How is this a bug fix? This was/is the intended behavior of the feature.
> 
> Seems like Hanni accidentally disabled it entirely when not wearing rancher's see #2413
 
 
 > accidentally 
 
 
 "Now warns only when actually farming and wearing Rancher Boots."

Obsidianninja11:
I suppose i can add a config to `Only Warn With Rancher's` tomorrow, but seems quite useless imo

lunaynx:
Yeah, I would prefer if this was configurable. It can be useful to have, but I don't want to be notified if I'm pest farming and going at, say, 329 speed instead of 328. Also on Bingo where it's  hard to get the exact speed right or you may even be using a different farm design.

Obsidianninja11:
Why's detekt failing :(

CalMWolfs (changes requested):
The default behaviour of this feature should not change

Obsidianninja11:
> The default behaviour of this feature should not change

My reason for setting the default to false is:
That was the old behaviour, and it was changed with only a minor note in patches (I thought it was just bugged for a while).
When it's set to false, it sends a message to disable it if the user wants. When it's set to true, the user may not know of it.

I can change it to default to true, but I think it's worth considering a default of false

PR #2862, opened 2024-10-29
Backend: Merge Rev and Tara Patterns

DavidArthurCole:
## What
Merges tara and rev patterns into monster patterns, and adds regex tests for each of the things we're attempting to block, for when inevitably someone does the enum mapping option we were talking about yesterday.
I don't have the patience or gear to do blaze or eman to the level I'd need for merging those :3

exclude_from_changelog


CalMWolfs (changes requested):
This feels like the perfect place for either a repo pattern list or an unum that creates repo patterns. 
As both these options are significantly more readable and also use the repo so they can be updated remotely

DavidArthurCole:
Gonna close this one out as yeah, enums for choices of what to block are probably better. (but I do not have the will to do that 😔)

PR #2869, opened 2024-10-31
Fix: NEUInternalName string comparison

lunaynx:
## What
Fixed a regression causing comparisons between `NEUInternalName` and `String` to always returning false.

## Changelog Fixes
+ Fixed some features not working correctly because of a bug in the code. - Luna
    * This includes, but is likely not limited to, the "Merge Seeds" option for Crop Profit Display.


hannibal002:
A string representation of a neu internal name is no neu internal object.

"2" is not the same as 2.
Please fix spots in code where we compare strings to neu internal name objects instead.

PR #2873, opened 2024-11-01
Fix: Itempickup log hay bale

catgirlseraid:
## What
hay bale (wheat) is HAY_BLOCK but hay bale (furniture) gets HAY_BALE

when internal name gets resolved from the sack pickup message it incorrectly resolves to the wrong block.
this fixes it for itempickup log for now until the internalname system is changed with the removal of neu.
this fix works since there is no way for furniture to end up in the sack change message

## Changelog Fixes
+ Fixed itempickuplog hay bale resolving to the wrong item when being moved from sacks. - phoebe


src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -122,7 +122,10 @@ object ItemPickupLog {
         if (!isEnabled() || !config.sack) return
 
         event.sackChanges.forEach {
-            val itemStack = (it.internalName.getItemStack())
+            var itemStack = (it.internalName.getItemStack())
+            // TODO this should not need to be done here but the whole internal name resolving needs a rework and this fixes it for now
+            if (it.internalName == "HAY_BALE".asInternalName()) itemStack = "HAY_BLOCK".asInternalName().getItemStack()
```
ItsEmpa: internal names should be initialized as properties outside functions. also make it so it only calls .getItemStack() once

ItsEmpa (changes requested):
requested changes. also the changelog should be formatted in a way that users can understand more or less what the fix is

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
catgirlseraid: fixy fix maybe me thinks

ItsEmpa (approved):

lunaynx:
This bug also affects `/gfs`, so maybe it can be done somewhere in the code where it affects sacks in general?

I had an open PR for this but I need to get around to finishing it. I don't remember if my implementation fixed it for item pickup log.

src/main/java/at/hannibal2/skyhanni/features/inventory/ItemPickupLog.kt
```diff
@@ -122,7 +124,11 @@ object ItemPickupLog {
         if (!isEnabled() || !config.sack) return
 
         event.sackChanges.forEach {
-            val itemStack = (it.internalName.getItemStack())
+
+            // TODO this should not need to be done here but the whole internal name resolving needs a rework and this fixes it for now
+            val internalName = if (it.internalName == HAY_BALE) HAY_BLOCK else it.internalName
```
hannibal002, line 129: code duplication of https://github.com/hannibal002/SkyHanni/blob/7c2d4ed566246513bc438272e436627412147d62/src/main/java/at/hannibal2/skyhanni/utils/ItemNameResolver.kt#L66

hannibal002 (changes requested):

CalMWolfs:

PR #2893, opened 2024-11-04
Backend: Remove asInternalName

j10a1n15:
## What
This Pull Request only replaces the old asInternalName() with the new toInternalName(), making the deprecated spam smaller (by a bit).
This will cause build failures on prs that don't use the new method yet, but honestly they shouldn't even use it anyway

+ Removed the outdated `asInternalName()` method. - j10a1n15


hannibal002:
the old function is deprecated and should still be deprecated for some time. there is no need for this pr to get merged now, as it will only break other prs. lets wait on this for couple more full versions, right?

j10a1n15:
> the old function is deprecated and should still be deprecated for some time. there is no need for this pr to get merged now, as it will only break other prs. lets wait on this for couple more full versions, right?

im fine with keeping the old method, but not merging it for some time because it could break other prs feels kinda weird.
The only thing that got changed was the renaming of a method call, that could break like up to 1-2 prs that removed a function that had that call inside

CalMWolfs (approved):
yay less deprecated spam, i still feel that the old method should be fully gone but 🤷, this will have to do

PR #2897, opened 2024-11-05
Fix + Backend: Livid Detection

martimavocado:
## What
reworks how the livid color detection works and adds debug, should also fix the wrong color being detected

## Changelog Fixes
+ Fixed incorrect Livid highlighting in F5/M5 sometimes. - martimavocado

+ Reworked Livid color detection. - martimavocado
    * Added performance improvements.
    * Added easier debugging for future issues.


hannibal002 (changes requested):
every livid was marked as the correct one.
also this detection even started before the right livid was revealed. so this cant work.
image

on antoher note, this ímplementaion looks like it will only mark the livid when the feature is enabled on the beginning. when you enable the feature mid livid fight, the livid willl not get detected.

martimavocado:
> every livid was marked as the correct one.
> also this detection even started before the right livid was revealed. so this cant work.
> image

this wasn't happening to me, will look into

> on antoher note, this ímplementaion looks like it will only mark the livid when the feature is enabled on the beginning. when you enable the feature mid livid fight, the livid willl not get detected.

didn't think about this, i'll take a look at what i can do

hannibal002 (approved):
works for me

PR #2900, opened 2024-11-06
Feature: Reopen last storage command

aphased:
## What

This PR adds three new sub-commands as a (configless/always-on) feature. Example: `/bp -` (the single dash). Executing it will re-open the backpack that was last opened using `/bp [int]`.  
The inspiration for this comes from Unix-style `cd -` for going to the previous current directory. Also, I'm using `/bp` and `/ec` for brevity everywhere, `/backpack` and `/enderchest` work as well of course.

The new command `/shlo` (and `/shlastopened`) will always open the last-used storage irrespective of whether it was an ender chest or backpack page.

The key/command combinations are not yet taken AFAIK.
Currently in an unmodded client, executing
- `/bp -` gives `Command Failed: Non-Numeric argument!`.
- `/ec -` opens the first ender chest page, which is also still reachable via `/ec` or `/ec 1`
- `/shlastopened` (and its short-form variant `/shlo`) are of course also still unregistered

---

Apart from typing it out as a command (or repeatedly accessing it via the up arrow), this feature also has the advantage that a **single hotkey** from any number of mods may be defined and used the same way every time you need to access _any_ storage page multiple times in a row for removing/adding some items to it.

Also potentially noteworthy regarding the implementation, I made the explicit decision against resetting the most recent backpack/ender chest slot value in case of a non-integer argument. In this case, it will just keep the previous valid integer backpack number "active".

This feature was first suggested [here](https://discord.com/channels/997079228510117908/1258925643828035744).

## Changelog New Features
+ Added commands "/bp -", "/ec -" and "/shlastopened" to reopen last opened storage pages. - aphased


lunaynx:
Cool idea, can we have the same thing for ender chest as well?

Xclipsen:
Thank you for your hard work on this feature! I really hope it gets accepted, as it would make me genuinely happy to use it actively. I believe it has a lot of potential, and I appreciate the effort you’ve put into it.

aphased:
Converting to draft since I would still like to add @lunaynx's [suggestion](https://github.com/hannibal002/SkyHanni/pull/2900#issuecomment-2460667087) (see above) first.

src/main/java/at/hannibal2/skyhanni/features/commands/OpenLastStorage.kt
```diff
@@ -0,0 +1,125 @@
...
+        event.isCanceled = true
+
+        if (lastStorageType == StorageType.ENDER_CHEST) {
+            openLastEnderChest()
+        } else if (lastStorageType == StorageType.BACKPACK) {
+            openLastBackPack()
+        }
+    }
+
+    private fun handleBackpack(event: MessageSendToServerEvent): Boolean {
```
hannibal002: the event is only ued for reading the message. instead of using the event here, you can directly give the function only  message

src/main/java/at/hannibal2/skyhanni/features/commands/OpenLastStorage.kt
```diff
@@ -0,0 +1,125 @@
...
+        if (!event.message.startsWith("/backpack", ignoreCase = true) &&
+            !event.message.startsWith("/bp", ignoreCase = true)
+        ) {
+            return false
+        }
+
+        if (event.message.startsWith("/backpack -", ignoreCase = true) ||
```
hannibal002: instead of ignore case everywhere, you can make the message to lower case once at the beginning

src/main/java/at/hannibal2/skyhanni/features/commands/OpenLastStorage.kt
```diff
@@ -0,0 +1,125 @@
...
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+
+@SkyHanniModule
+object OpenLastStorage {
+    private enum class StorageType {
+        ENDER_CHEST, BACKPACK
+    }
+
+    private var lastBackpack = -1
+    private var lastEnderChest = -1
```
hannibal002: please use kotlin's null instead of -1 as default/invalid vlaue

hannibal002 (changes requested):
as you can see in this comparison, the two functions handleBackpack and handleEnderChest are doing the same thing, just with different values. please try using only one function for both
image

src/main/java/at/hannibal2/skyhanni/features/commands/OpenLastStorage.kt
aphased: Done

src/main/java/at/hannibal2/skyhanni/features/commands/OpenLastStorage.kt
aphased: Done

src/main/java/at/hannibal2/skyhanni/features/commands/OpenLastStorage.kt
aphased: Not entirely sure, but it feels a bit unavoidable to pass it in because I need to set the event to `isCanceled`

CalMWolfs: Resolved in my commit

CalMWolfs (approved):
Worked for me

hannibal002 (changes requested):
Please add a toggle to disable this behaviour.

PR #2905, opened 2024-11-10
Backend: No labels on prs without milestones

hannibal002:
## What
To avoid adding labels on new PRs, ones that aren't yet moved to a milestone, we will not add labels until the milestone is set.

exclude_from_changelog


hannibal002:
found a better solution:
image

PR #2907, opened 2024-11-10
Improvement: CF Stray Filter

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/285

## What
Adds the ability to only be notified (pinged & flashed) when strays of a certain rarity or higher appear in the factory, rather than all or nothing.

Don't know what we want to do about what used to be `rabbitWarning` going away in favor of `rabbitWarningType`. I can try to do a migration of `true -> ALL` and `false -> NONE`, but I'm not sure if that's worth it.

## Changelog Improvements
+ Chocolate Factory's "Stray Alerts" and "Screen Flash" can now be filtered by rabbit rarity. - Daveed


hannibal002:
I can try to do a migration of true -> ALL and false -> NONE, but I'm not sure if that's worth it.

i say not worth it

hannibal002 (changes requested):
the config value `flashScreenType` crashes during startup with old config

hannibal002 (approved):

PR #1753, opened 2024-05-10
Feature: Ghost Profit Tracker

ItsEmpa:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/271/

### WIP
Removes the Ghost Counter and replaces it with a Ghost Profit Tracker, to keep it consistent with other profit trackers, and to fix various issues with it.

TODO: 

- [x] Migrate data from ghost counter to ghost profit tracker
- [x] Rename configs to ghost profit tracker
- [x] Store allowed items in repo
- [x] Remove old ghost counter code

## Changelog New Features
+ Added Ghost Profit Tracker. - Empa
    * Replaces the outdated and buggy Ghost Counter.

PR #2909, opened 2024-11-11
Feature: Remove NEU as a dependency for SkyHanni

CalMWolfs:
## Dependencies
- #1753 (so the game doesnt crash in the dwarven mines)

## What
Removed the need for having neu installed to play with skyhanni. There are many backend changes that cannot easily be split apart as it all relies on having item resolution query and the neu repo downloaded.
Now all of the neu repo is loaded by skyhanni regardless of if neu is installed or not.
We also get lowest bin prices if neu isnt installed, if it is installed we just use the data that neu has already fetched.

## Changelog New Features
+ Removed NotEnoughUpdates requirement. - CalMWolfs
    * SkyHanni no longer requires NEU to run, preparing for updates to modern Minecraft versions.
    * Using NEU is still recommended.

+ Added functionality to download NEU repo if NEU isn't installed. - CalMWolfs
+ Parsed NEU item repo for item and recipe data. - CalMWolfs
    * Ensured parsing occurs regardless of NEU installation status.
+ Retrieved lowest BIN prices if NEU isn't installed. - CalMWolfs
    * Using NEU's item price data if NEU is installed.


src/main/java/at/hannibal2/skyhanni/utils/system/PlatformUtils.kt
```diff
@@ -24,10 +27,41 @@ object PlatformUtils {
...
+        } catch (_: Throwable) {
+        }
+
+        val text = listOf(
+            "§c§lOutdated NotEnoughUpdates version detected!",
+            "§cWhile Skyhanni doesn't require NotEnoughUpdates to function anymore,",
+            "§cif you choose to still use NotEnoughUpdates, which is recommended,",
+            "§cwe require you to use a newer version of NotEnoughUpdates to ensure",
+            "§ccompatibility with some of our features.",
+            "§cPlease update NotEnoughUpdates to version 2.4.0",
```
jani270: I would not hardcode a version number here

PR #2914, opened 2024-11-12
Improvement: Attribute Overlay Options

ItsEmpa:
## What
Adds 2 new options to attribute overlay, one to highlight attributes if they are part of a good roll even if the attribute combination isnt a good roll, and an option to fully show the good roll even if some attributes in it are not in the attribute list to show.

## Changelog Improvements
+ Added customization options for Attribute Overlay. - Empa
    * Option to highlight pieces with attributes contained in a Good Roll piece.
    * Option to show Good Rolls even if the attributes aren’t in the list.


hannibal002 (changes requested):
please explain the changelog line more

PR #2916, opened 2024-11-12
Backend: Remove Deprecated Color.withAlpha

j10a1n15:
## What
Removes ~32 deprecation messages if I counted correctly
Also can this method just please stay removed ty

+ Removed deprecated `Color.withAlpha` method in favor of `Color.addAlpha`. - j10a1n15


hannibal002:
>  Also can this method just please stay removed ty

you talk about the removal of deprecated functions should not be reverted?
as long as the deprecation got added a long time ago its fine
removing too early breaks the logic behind using deprecation entirely, i could instead just remove the function in the first place.

but we dont want this, we already have enough merge conflicts in open prs.

j10a1n15:
> as long as the deprecation got added a long time ago its fine 

it got removed ~1 month ago with #2112

CalMWolfs (approved):

PR #2922, opened 2024-11-13
Fix: Barn Fishing Timer

ItsEmpa:
## What
Adds some debug info to barn fishing timer.
Edit: Managed to pinpoint what the issue was

## Changelog Fixes
+ Fixed barn fishing timer getting stuck. - Empa


ItsEmpa:
found out what the issue was with the debugging i added and fixed the issue

PR #2923, opened 2024-11-13
Fix: Widget Update

ItsEmpa:
## What
This PR fixes the Widget Update Event sometimes not being sent after world switch until another tablist update happens.

## Changelog Fixes
+ Fixed Tab Widgets sometimes not getting updating. - Empa

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/model/TabWidget.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/model/TabWidget.kt
@@ -1,5 +1,7 @@
 package at.hannibal2.skyhanni.data.model
 
+import at.hannibal2.skyhanni.data.HypixelData
+import at.hannibal2.skyhanni.data.ProfileStorageData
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/data/model/TabWidget.kt
```diff
@@ -371,10 +375,21 @@ enum class TabWidget(
...
 
+        @SubscribeEvent
+        fun onSecond(event: SecondPassedEvent) {
+            if (sentSinceWorldChange || !LorenzUtils.inSkyBlock) return
+            sentSinceWorldChange = true
+            @Suppress("DEPRECATION")
+            update(TabListData.getTabList())
+            ChatUtils.debug("Forcefully Updated Widgets")
+        }
+
```
Thunderblade73: Here you gamble that the tablist is already correct. But this might not be the case at this time. Since full world switches can take longer than 1 second (with a bit of lag).

Thunderblade73 (changes requested):

PR #2926, opened 2024-11-14
Fix: BitsAPI

ItsEmpa:
## What
Fixes bits api getting wrong amount of bits in purse when opening the cookie menu.

## Changelog Fixes
+ Fixed incorrect bits values. - Empa


src/main/java/at/hannibal2/skyhanni/data/BitsAPI.kt
```diff
@@ -2,17 +2,24 @@ package at.hannibal2.skyhanni.data
 
 import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.config.ConfigUpdaterMigrator
+import at.hannibal2.skyhanni.data.BitsAPI.bits
```
NopoTheGamer: ?

ItsEmpa: no clue how that happened

PR #2931, opened 2024-11-15
Backend: Add regex test  detekt rule

CalMWolfs:
## What
Added a detekt rule to enforce repo patterns to be accompanied by at least one regex test.

This makes sure that each repo pattern has a regex test unless it is a plain string.
Every regex test that accompanies a repo pattern must pass, 
Every regex fail test that accompanies a repo pattern must fail

For now all repo patterns that are missing a regex test are baselined, but this means that when they are updated they will require a pattern. I will also try to add some regex tests to a lot of them.

This detekt rule also has possibilities for the future e.g.
- Enforce naming conventions for the repo pattern variable as rn ~95% of them end in "Pattern"
- Disallow unnamed capture groups, stuff like `(.*)` shouldn't be allowed
- Making max line length pr ignore repo patterns so we dont need to supress it everywhere

+ Added a Detekt rule to enforce repo patterns to be accompanied by at least one regex test. - CalMWolfs

### DIFF
--- a/.live-plugins/regexr/plugin.kts
+++ b/.live-plugins/regexr/plugin.kts
@@ -21,6 +21,7 @@ val logger =
     Logger.getInstance("SkyHanni")
 
 val regexTestPrefix = "REGEX-TEST: "
+val regexTestFailPrefix = "REGEX-FAIL: "
 
 class RegexInfo(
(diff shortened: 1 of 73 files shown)


NopoTheGamer:
lgtm to me

PR #2934, opened 2024-11-15
code cleanup in bits api

hannibal002:
ignore_from_changelog

again fun stuff i want to check agianst detekt first


hannibal002:
Should have seen #2926 first oh no

PR #2939, opened 2024-11-16
Improvement: Cake Tracker Scrollable and AH Tips

DavidArthurCole:
## What
Instead of having a maximum length, make cake tracker implement scrollable, and add tips about there being entries above/below. I tried to do a scroll bar renderable, that failed, so I instead added common logic for adding these tips that's default false. (See images).
Also adds a note to the on-hover for prices that indicates that clicking them will open AH.

<details>
<summary>Images</summary>

image

image

image

</details>

## Changelog Improvements
+ Improved the UI for the Cake Tracker. - Daveed
    * Made the Cake Tracker scrollable instead of having a fixed size.


hannibal002 (changes requested):
After scrolling 2–3 times in the new year cake tracker, the size of lines visible at any time is changing by one line.

The last line "more items below" is already disappearing on the second last scroll, instead of just on the last scroll.

DavidArthurCole:
> After scrolling 2–3 times in the new year cake tracker, the size of lines visible at any time is changing by one line.
> 
> The last line "more items below" is already disappearing on the second last scroll, instead of just on the last scroll.

The first issue is probably an underlying issue with the way that `scrollList` is choosing to render height, as the tips add their height to the `renderY` correctly. I can't replicate the second issue at all...

Thunderblade73:
After changing the size in the config and then opening the cake bag
```
SkyHanni 0.28.Beta.16: Caught a ClassCastException in CakeTracker at ChestGuiOverlayRenderEvent: java.lang.Float cannot be cast to java.lang.Integer
 
Caused by java.lang.ClassCastException: java.lang.Float cannot be cast to java.lang.Integer
	at SH.features.inventory.caketracker.CakeTracker.getMaxTrackerHeight(CakeTracker.kt:55)
	at SH.features.inventory.caketracker.CakeTracker.buildCakeRenderables(CakeTracker.kt:432)
	at SH.features.inventory.caketracker.CakeTracker.drawDisplay(CakeTracker.kt:392)
	at SH.features.inventory.caketracker.CakeTracker.reRenderDisplay(CakeTracker.kt:198)
	at SH.features.inventory.caketracker.CakeTracker.onBackgroundDraw(CakeTracker.kt:171)
	at SH.data.RenderData.onBackgroundDraw(RenderData.kt:52)
```

Thunderblade73:
The changes to the scrollable list don't account for the correct size. Which is not allowed.
The green box is the bounding box of the renderable (simply do a .renderbounds()).
image
image

Also a tip how to fix hannibals concern is to give it 1-2 pixel more space.

hannibal002 (changes requested):
I cant reproduce the original two bugs reported.

But the pixel error is visible and ideally fixed before merging the pr.

use this patch to see it clearly:


<details>
<summary>Patch</summary>

```
Index: src/main/java/at/hannibal2/skyhanni/utils/renderables/Renderable.kt
IDEA additional info:
<+>UTF-8
diff --git a/src/main/java/at/hannibal2/skyhanni/utils/renderables/Renderable.kt b/src/main/java/at/hannibal2/skyhanni/utils/renderables/Renderable.kt
--- a/src/main/java/at/hannibal2/skyhanni/utils/renderables/Renderable.kt(revision 72dc1a5ed5bda9acccacda29313c7260fb81b0e6)
+++ b/src/main/java/at/hannibal2/skyhanni/utils/renderables/Renderable.kt(date 1733059457763)
@@ -998,8 +998,9 @@
             private val end get() = scroll.asInt() + height
 
             override fun render(posX: Int, posY: Int) {
+                val hovered = isHovered(posX, posY)
                 scroll.update(
-                    isHovered(posX, posY) && shouldAllowLink(true, bypassChecks),
+                    hovered && shouldAllowLink(true, bypassChecks),
                 )
 
                 var renderY = 0
@@ -1008,9 +1009,13 @@
 
                 // If showScrollableTipsInList is true, and we are scrolled 'down', display a tip indicating
                 // there are more items above
-                if (showScrollableTipsInList && scroll.asInt() > 0) {
+                if (hovered) {
                     width = maxOf(width, scrollUpTip.width)
-                    scrollUpTip.renderXAligned(posX, posY, width)
+                    if (showScrollableTipsInList && scroll.asInt() > 0) {
+                        scrollUpTip.renderXAligned(posX, posY, width)
+                    } else {
+                        string("§7§ono items above (dont scroll)").renderXAligned(posX, posY, width)
+                    }
                     GlStateManager.translate(0f, scrollUpTip.height.toFloat(), 0f)
                     renderY += scrollUpTip.height
                     virtualY += scrollUpTip.height

```

</details>

hannibal002 (approved):
those few pixels will not break the pr

PR #2945, opened 2024-11-18
Improvement: Block Opening CF From Item(s)

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1306532441330028544
Expands the scope of `/cf`-blocking (booster cookie, rabbit pet) to also apply to the item from the calendar event, as well as Skyhanni's own added item.

Also made it so that in the future it's a lot easier to drop in new code for "reasons cf can't be opened," which might be useful for the mixin that's coming in alpha.

## Changelog Improvements
+ The item used to open the Chocolate Factory will now respect Mythic Rabbit and Booster Cookie blocking. - Daveed


NopoTheGamer:
should this also be used for the block egglocator feature?

hannibal002:
> should this also be used for the block egglocator feature?

While similar, we dont need the booster cookie check for egglocator blocking.
the actual remaining duplicate logic remains the clickToActionOrDisable call

PR #2948, opened 2024-11-18
Fix: Custom Garden Keybinds

ItsEmpa:
## What
This PR fixes custom garden keybinds sometimes spamming keybinds instead of holding them, making it function like an autoclicker. Also fixes an issue where sometimes it would suddenly stop breaking blocks, and refactors it to improve performance.

## Changelog Fixes
+ Fixed Custom Garden Keybinds sometimes spamming left-click instead of holding. - Empa
+ Fixed Custom Garden Keybinds sometimes causing blocks to stop breaking. - Empa


src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCustomKeybinds.kt
```diff
@@ -24,71 +29,121 @@ object GardenCustomKeybinds {
...
+            useItem.set(KeyboardManager.RIGHT_MOUSE)
+            left.set(Keyboard.KEY_A)
+            right.set(Keyboard.KEY_D)
+            forward.set(Keyboard.KEY_W)
+            back.set(Keyboard.KEY_S)
+            jump.set(Keyboard.KEY_SPACE)
+            sneak.set(Keyboard.KEY_LSHIFT)
+            Minecraft.getMinecraft().thePlayer.closeScreen()
+        }
     }
```
CalMWolfs: I know they didnt have it before but could we have some chat feedback for these, also idk why they exit the config menu so maybe that can be removed but thats up to you. (If the exiting config is removed there doesnt need to be a chat message)

CalMWolfs (changes requested):
It worked fine for me, just this one thing

src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCustomKeybinds.kt
hannibal002: done

PR #2956, opened 2024-11-20
Fix: IllegalStateException from Pest Traps

DavidArthurCole:
## What
Fixes an issue where pests from traps may not update the tablist/scoreboard fast enough, leading to an exception being thrown.

## Changelog Fixes
+ Fixed an occasional error when killing pests shortly after releasing them from pest traps. - Daveed


DavidArthurCole:
Why yes thank you that would be because I cannot copy past correctly

PR #2966, opened 2024-11-20
Improvement: Replace same message for New Year Cake Reminder

lunaynx:
## What
Added `replaceSameMessage = true` to New Year Cake Reminder.

## Changelog Improvements
+ Previous reminders for New Year Cake are now replaced instead of spamming the chat. - Luna


lunaynx:
Oh you can't do that with this method, fun.

hannibal002:
`clickToActionOrDisable()` always uses replaceSameMessage as true internally already. what message is getting spammed anyway?

lunaynx:
Never mind.
image

PR #2971, opened 2024-11-22
Backend: Fix duplicate input class error

CalMWolfs:
Removes the 150 lines of warnings saying duplicate input class on each build
exclude_from_changelog


hannibal002:
this doesnt work for me.
using this build script, i cant log in into my mc account anymore, i get a "invalid session try restart" error on joining hypixel.
after deleting the ms acc json file (see image) with this pr, i did not get the question in chat to relogin.
after switching back to beta branch and reloading gradle and mc, i got the link in chat again and cound join hypixel again.
image

hannibal002 (changes requested):
see above

hannibal002:
(I was testing with beta merged into this pr already)
image

PR #2998, opened 2024-11-27
Fix: ComputerTimeOffset Error

j10a1n15:
## What
Fixes by bumping commons-net and using .use { }
Also adds a Config Option for a diff NTP Server url, just incase idk

## Changelog Fixes
+ Fixed computer time offset calculation errors. - j10a1n15

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
@@ -56,6 +56,7 @@ object ComputerTimeOffset {
         val client = NTPUDPClient()
         val address = InetAddress.getByName(ntpServer)
         val timeInfo = client.getTime(address)
+        client.close()
 
         timeInfo.computeDetails()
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
```diff
@@ -56,6 +56,7 @@ object ComputerTimeOffset {
         val client = NTPUDPClient()
         val address = InetAddress.getByName(ntpServer)
         val timeInfo = client.getTime(address)
+        client.close()
```
lunaynx:
```suggestion
        val timeInfo = NTPUDPClient().use { client ->
            val address = InetAddress.getByName(ntpServer)
            client.getTime(address)
        }
```
(Requires updating `commons-net` to `3.11.1` in `build.gradle.kts`.)

lunaynx (changes requested):

lunaynx (approved):

PR #3003, opened 2024-11-29
Fix: replace roman numeral v2

Mikecraft1224:
## What
Fixed the ReplaceRomanNumeral Feature replacing things it shouldn't. Small changes may follow.

## Changelog Fixes
+ Fixed Replace Roman Numerals replacing random letters. - Mikecraft1224


CalMWolfs (changes requested):
Please make some unit tests for roman numerals so that it is easier to test if it is working as intended

hannibal002 (changes requested):
as my patch in dms: i dont think using regexes is appropriate here, as well as caching is missing. lets discuss

hannibal002 (approved):

PR #3016, opened 2024-12-03
Fix: Crimson Isle Quest Parsing

lunaynx:
## What
Fixed detection of Crimson Isle daily quests after Hypixel update.

Unfortunately due to non-robust code this can't be fixed via a repo update, it will require a mod update, so I renamed the pattern.

## Changelog Fixes
+ Fixed detection of Crimson Isle daily quests after the Hypixel update. - Luna


lunaynx:
Completion detection from town board still needs fixing.

PR #3026, opened 2024-12-06
Fix: Custom Scoreboard Player Count Including Offline Players

Chissl:
## What
Fixed custom scoreboard's player count display on private islands/garden including players that were offline or on other islands. Also changed max player count on private islands from guest limit to online island owners + guest limit. Added an "Island" widget type to pull player names on solo islands and when guesting. 

## Changelog Fixes
+ Fixed Custom Scoreboard's player count to exclude offline players in co-ops and guest islands. - Chissl

+ Introduced "Island" widget type. - Chissl


src/main/java/at/hannibal2/skyhanni/data/HypixelData.kt
```diff
@@ -570,4 +570,18 @@ object HypixelData {
         val scoreboardTitle = displayName.removeColor()
         return scoreboardTitlePattern.matches(scoreboardTitle)
     }
+
+    private fun countPlayersOnIsland(event: WidgetUpdateEvent) {
+        if (event.isClear()) return
+        playerAmountOnIsland = playerAmountOnIslandPattern.allMatches(event.lines).size
+    }
+
+    private fun isPlayerIsland(): Boolean {
```
j10a1n15: you can use `inAnyIsland(IslandType.Garden,...)`

j10a1n15 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/HypixelData.kt
```diff
@@ -89,11 +91,11 @@ object HypixelData {
...
 
     /**
-     * REGEX-TEST:           §r§b§lCoop §r§f(4)
+     * REGEX-TEST: §8[§r§a§r§8] §r§bBpoth §r§6§l℻
      */
-    private val playerAmountCoopPattern by patternGroup.pattern(
+    private val playerAmountOnIslandPattern by patternGroup.pattern(
         "playeramount.coop",
-        "^\\s*(?:§.)*Coop (?:§.)*\\((?<amount>\\d+)\\)\\s*$",
+        "^§.\\[[§\\w]{6,11}] §r.*",
```
CalMWolfs: changing this pattern from using a capture group to not using a capture group will cause NPEs for people on old versions, please change the pattern key

CalMWolfs (dismissed):

hannibal002 (changes requested):
I cant reproduce the bug
image

j10a1n15:
> I cant reproduce the bug 

image
you need to be on your own island while in a coop

hannibal002:
>  you need to be on your own island while in a coop

ive tried on private island as well

j10a1n15 (approved):
image
works for me

Chissl:
> works for me
It's a bug documented since scoreboard release too https://discord.com/channels/997079228510117908/1222605332648296468/1222605332648296468

Should break when visiting people too since it always adds 1 to amount if it detects a solo island. Maybe regex broke or something?

hannibal002:
now it breaks on private island in production. in dev env was it working fine. hmm

hannibal002:
on other players islands it doesnt break, neither in dev env nor prod
image

j10a1n15:
> on other players islands it doesnt break, neither in dev env nor prod 

 the old code just sums up the numbers. so it takes the 4 from "Coop (4)" even though only 1 person is only rn
image
image

it doesnt break when visiting coop profiles because the tablist doesnt use the "Coop (4)" line here

Chissl:
> on other players islands it doesnt break, neither in dev env nor prod 

It doesn't detect coops at all when guesting, it assumes every guest island is solo 
image

Chissl:
> it doesnt break when visiting coop profiles because the tablist doesnt use the "Coop (4)" line here

Yeah visiting people's islands and solo islands has "Island" header instead of of coop, which triggers this code instead, which will always increment 1 to amount regardless of how many people own the island and if they're on the island
image
based on this regex 
image
which is this 
image

Not sure why it's coded like there will only be one instance of it in tab list

I replaced this with with a widget that contains every island owner name

Chissl:
> on other players islands it doesnt break, neither in dev env nor prod

It is breaking here, there are 6 people on island but shows 7 in scoreboard

hannibal002 (approved):

PR #3035, opened 2024-12-08
Backend: Use Repo for IslandType Data

j10a1n15:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/307

## What
Moved IslandType.displayName to repo, added max player count & api name aswell

+ Moved IslandType data to the repo. - j10a1n15


hannibal002:
I don't see the need for the debug data

src/main/java/at/hannibal2/skyhanni/data/IslandType.kt
```diff
@@ -1,40 +1,74 @@
...
-    GARDEN("Garden"),
-    GARDEN_GUEST("Garden Guest"),
-    SPIDER_DEN("Spider's Den"),
-    WINTER("Jerry's Workshop"),
-    THE_RIFT("The Rift"),
-    MINESHAFT("Mineshaft"),
-
-    NONE(""),
-    ANY(""),
-    UNKNOWN("???"),
```
CalMWolfs: I feel we should still keep these as fall backs as peoples repos love to not work

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/data/IslandType.kt
j10a1n15: Right now the fallback is the raw enum name, I could add smth that just formats the raw name, e.g. [SkyBlockAPI](https://github.com/SkyblockAPI/SkyblockAPI/blob/0ecf9b828b67225be5c5114a6f25f40b5a604826/src/client/kotlin/tech/thatgravyboat/skyblockapi/utils/extentions/StringExtensions.kt#L136)

CalMWolfs: This wouldnt work for things like spiders den or jerrys island, its probably just better to have the actual fallback in the code instead of generating them

src/main/java/at/hannibal2/skyhanni/data/IslandType.kt
```diff
@@ -1,40 +1,74 @@
...
+    WINTER,
+    THE_RIFT,
+    MINESHAFT,
+
+    NONE,
+    ANY,
+    UNKNOWN,
     ;
 
+    var islandData: IslandData? = null
```
ItsEmpa, line 51: private setter

ItsEmpa (changes requested):

src/main/java/at/hannibal2/skyhanni/data/IslandType.kt
```diff
@@ -30,11 +36,40 @@ enum class IslandType(val displayName: String) {
...
+        private set
+
+    val displayName: String get() = islandData?.name ?: nameFallback
+
+    @SkyHanniModule
     companion object {
+        var islandTypesData: IslandTypeJson? = null
```
ItsEmpa: instead of storing the entire data class as nullable, it should just store the max players and max players mega as non-nullable variables with a fallback if island types data isnt loaded

ItsEmpa (changes requested):

src/main/java/at/hannibal2/skyhanni/data/IslandType.kt
ItsEmpa: either that or make a fallback island type json

lunaynx:
Bump

hannibal002:
max players for garden/private island is wrong in custom scoreboard. without this pr, it shows 26, with this pr, it shows 24.
is this intended?

j10a1n15:
> max players for garden/private island is wrong in custom scoreboard. without this pr, it shows 26, with this pr, it shows 24. is this intended?

all islands except the hub got switched from 26 max to 24 max, related to [this comment](https://github.com/hannibal002/SkyHanni-REPO/pull/307#issuecomment-2525839991)
the private island max limit only really works if someone is guesting, since we dont have a way to get the max players from the rank + the ones you get from the community center upgrades

hannibal002 (approved):

PR #3037, opened 2024-12-08
Fix: Composter message spam

hannibal002:
## What
Fixed two small bugs totally independent of each other.

## Changelog Fixes
+ Fixed chat error spam from Garden Composter Overlay. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/mining/OreBlock.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/mining/OreBlock.kt
@@ -86,7 +86,7 @@ enum class OreBlock(
     JADE(EnumDyeColor.LIME, { inCrystalHollows || inGlacite }),
     SAPPHIRE(EnumDyeColor.LIGHT_BLUE, { inCrystalHollows || inGlacite }),
     TOPAZ(EnumDyeColor.YELLOW, { inCrystalHollows || inGlacite }),
-    JASPER(EnumDyeColor.MAGENTA, { inCrystalHollows || inGlacite }),
+    JASPER(EnumDyeColor.MAGENTA, { inCrystalHollows || inGlacite || inMineshaft }),
     OPAL(EnumDyeColor.WHITE, { inGlacite || inCrimsonIsle }),
     AQUAMARINE(EnumDyeColor.BLUE, { inGlacite }),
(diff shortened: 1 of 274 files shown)


src/main/java/at/hannibal2/skyhanni/features/mining/OreBlock.kt
```diff
@@ -86,7 +86,7 @@ enum class OreBlock(
     JADE(EnumDyeColor.LIME, { inCrystalHollows || inGlacite }),
     SAPPHIRE(EnumDyeColor.LIGHT_BLUE, { inCrystalHollows || inGlacite }),
     TOPAZ(EnumDyeColor.YELLOW, { inCrystalHollows || inGlacite }),
-    JASPER(EnumDyeColor.MAGENTA, { inCrystalHollows || inGlacite }),
+    JASPER(EnumDyeColor.MAGENTA, { inCrystalHollows || inGlacite || inMineshaft }),
```
lunaynx: `inGlacite` already includes mineshafts, this seems unnecessary.

Also, there are no commissions to mine Jasper gemstones.

hannibal002: > inGlacite already includes mineshafts, this seems unnecessary.

didnt work for me today. So I'm not sure about that

> Also, there are no commissions to mine Jasper gemstones.

you are right, there aren't. but there are still jasper gemstones, and they aren't grayed out properly.

lunaynx: Ah, does it gray out gemstones that are *not* for commissions?

I can't see from the code how it could possibly fail `inGlacite` but not `inMineshaft` though, besides a bit flip.

hannibal002: > Ah, does it gray out gemstones that are not for commissions?

this is the whole point of the feature. otherwise, having the correct blocks highlihgt in any color will just cause confusion with the natural color of other blocks

> I can't see from the code how it could possibly fail inGlacite but not inMineshaft

after looking into the code a bit more, i also cant see how this should fix anything. will investigate further when i notice it the next time

CalMWolfs (approved):

PR #3038, opened 2024-12-08
Backend: Sulphur Skitter performance updates

hannibal002:
## What
optimizing Sulphur skitter box

+ Optimized sulphur skitter performance by caching `AxisAlignedBB` calculations and using more `LorenzVec`. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/nether/SulphurSkitterBox.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/nether/SulphurSkitterBox.kt
@@ -13,6 +13,7 @@ import at.hannibal2.skyhanni.utils.BlockUtils.getBlockAt
 import at.hannibal2.skyhanni.utils.LocationUtils
 import at.hannibal2.skyhanni.utils.LocationUtils.distanceToPlayer
 import at.hannibal2.skyhanni.utils.LorenzUtils.isInIsland
+import at.hannibal2.skyhanni.utils.LorenzVec
 import at.hannibal2.skyhanni.utils.RenderUtils
 import at.hannibal2.skyhanni.utils.RenderUtils.expandBlock
 import at.hannibal2.skyhanni.utils.SpecialColor.toSpecialColor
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/nether/SulphurSkitterBox.kt
```diff
@@ -26,54 +27,63 @@ import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
...
+            val pos1 = it.add(-RADIUS, -RADIUS, -RADIUS)
+            val pos2 = it.add(RADIUS + 1, RADIUS + 1, RADIUS + 1)
+            pos1.axisAlignedTo(pos2).expandBlock()
         }
     }
 
+    private fun calculateSpongeLocations() {
+        val location = LocationUtils.playerLocation()
+        val from = location.add(-20, -20, -20).toBlockPos()
+        val to = location.add(20, 20, 20).toBlockPos()
```
CalMWolfs: if 2 lines down we check if its within 15 blocks why is it 20 blocks up here?
by changing these all to 15, we can check 39000 less blocks each time this function is ran (which is only 43% of the original blocks checked)

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/features/nether/SulphurSkitterBox.kt
hannibal002: Those numbers aren't from me.
I assume the reason is so blocks in a 15 block radius will get instantly detected and not only every second

CalMWolfs: > Those numbers aren't from me. I assume the reason is so blocks in a 15 block radius will get instantly detected and not only every second

Those numbers may not be from you but this is a performance improvement pr and i have given performance improvement suggestion

hannibal002: I agree with the math though, we can set both to 15

CalMWolfs (approved):

PR #2927, opened 2024-11-14
Backend: Active Custom Scoreboard Patterns

j10a1n15:
## What
This Pull Request improves the Custom Scoreboard to only do important regex checks. Each Element now contains their respective patterns and using the showIslands() we now filter out not for the current island important patterns. This should reduce the amount of regex checks by like 60% I assume, so yay more fps (maybe)

<details>
<summary>/shdebug custom scoreboard</summary>

```
= Debug Information for SkyHanni 0.28.Beta.12 =

search 'custom scoreboard':

== Custom Scoreboard ==
 Custom Scoreboard Lines:
    Title - island: true - show: true - [§ZSKAI-BAI-BAI~ ≧◡≦]
    Profile - island: true - show: true - [§7♲ Ironman]
    Purse - island: true - show: true - [Purse: §61.5M]
    Motes - island: false - show: true - []
    Bank - island: true - show: true - [Bank: §6250.3M]
    Bits - island: true - show: true - [Bits: §b231k§7/§b106k]
    Copper - island: false - show: true - []
    Gems - island: true - show: true - [Gems: §a160]
    Heat - island: false - show: true - []
    Cold - island: false - show: true - []
    North_stars - island: false - show: true - []
    Chunked_stats - island: true - show: true - [§61.5M §f| §6250.3M §f| §b231k§7/§b106k, §a160]
    Soulflow - island: true - show: true - [Soulflow: §3155,642]
    Island - island: true - show: true - [§7㋖ §aHub]
    Location - island: true - show: true - [§7⏣ §bVillage]
    Player_amount - island: true - show: true - [§7Players: §a21§7/§a26]
    Visiting - island: false - show: true - []
    Date - island: true - show: true - [§9Early Winter 4th]
    Time - island: true - show: true - [§73:02am §b☽]
    Lobby_code - island: true - show: true - [§711/14/2024 §8mini4AN]
    Power - island: true - show: true - [Power: §aBloody §7(§6683§7)]
    Tuning - island: true - show: true - [Tuning: §e⚔19]
    Cookie - island: true - show: false - []
    Objective - island: true - show: true - []
    Slayer - island: true - show: false - []
    Quiver - island: true - show: false - []
    Powder - island: false - show: true - [§9§lPowder,  §7- §fMithril: §2129k/7.9M,  §7- §fGemstone: §d573k/7.2M,  §7- §fGlacite: §b255k/687k]
    Events - island: true - show: true - []
[...]

PR #3046, opened 2024-12-11
Fix: New Rift CS Lines

j10a1n15:
## What
Added some new lines, probably not all though since I didnt play yet

## Changelog Fixes
+ Fixed Custom Scoreboard errors from Rift update. - j10a1n15

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/gui/customscoreboard/ScoreboardPattern.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/gui/customscoreboard/ScoreboardPattern.kt
@@ -818,6 +818,19 @@ object ScoreboardPattern {
         "Protestors handled: §b\\d+\\/\\d+",
     )
 
+    val timeSlicedPattern by riftSb.pattern(
+        "timesliced",
+        "§c§lTIME SLICED!",
+    )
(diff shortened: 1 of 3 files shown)


hannibal002 (changes requested):
this will just hide the "unknown lines" information.
in this context, i dont think this is productive. instead. the user wants to see if their time is sliced or the time until the big damage is coming in.
instead of this, add a option to include those settings. I know this is more complex than simply hiding the error messages, but we dont want to "just hide" them.

remember, the default settings of the custom scoreboard should not hide information the vanilla scoreboard provides.

j10a1n15:
What?
This pr adds the new lines to the rift lines? The rift option is default enabled on the events priority list. I don't see how this hides information

hannibal002 (approved):
Sorry!

this list keeps confusing me. we should remove the list in the future and instead iterate over all known pattern objects to build the "known pattern list" dynamically
image

j10a1n15:
> this list keeps confusing me. we should remove the list in the future and instead iterate over all known pattern objects to build the "known pattern list" dynamically

> 

This is alr an Open pr https://github.com/hannibal002/SkyHanni/pull/2927

PR #3047, opened 2024-12-11
Fix: sending mining event data when the event tracker is disabled

DerGruenkohl:
## What
Fix sending mining event data to the soopy api when the mining event tracker is disabled
https://discord.com/channels/997079228510117908/1316207755719348324

## Changelog Fixes
+ Fixed sending mining events when the tracker is disabled. - DerGruenkohl


CalMWolfs (changes requested):
I oppose this change as not sending the events when the setting is off significantly reduces the data that is available for the people that use this feature or other projects that rely on this api, is there an actual reason to disable this behaviour?

hannibal002:
replaced by 
- #3055

PR #3052, opened 2024-12-12
Fix: Stray Screen Flash V3

DavidArthurCole:
## What
For real now fixes the issue regarding the screen flash staying around longer than it should for strays. This is another case where I really don't know why we were using `onTick` instead of the inventory updating. This _might_ fix the side dish egg doing that little flashbang after you open CF next time, idrk.
Also refactored the stray ding/flash to happen in the same object, so it's easier to have the logic centralized.

<details>
<summary>Images</summary>

If you wanna see it in action:
https://discord.com/channels/997079228510117908/1234043917792579604/1316905839734558842

</details>

## Changelog Fixes
+ Fixed stray screen flash staying around longer than it should. - Daveed


src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityAPI.kt
```diff
@@ -183,15 +183,18 @@ object HoppityAPI {
         return (month % 2 == 1) == (day % 2 == 0)
     }
 
-    private fun Map<Int, ItemStack>.filterStrayProcessable() = filter { (slotNumber, stack) ->
+    fun Map<Int, ItemStack>.filterMayBeStray() = filter { (slotIndex, stack) ->
```
hannibal002: Only generic utils funcitons should  be public and extended. this is a limited scope utils function, and we dont want to offer this funciton as option to tab complete in other features of the mod.
simplest solution: dont make a extended funciton here

hannibal002 (changes requested):

hannibal002 (approved):

PR #3053, opened 2024-12-13
Fix: NeuRenderEvent

CalMWolfs:
## What
I guess you cannot post events that are just objects from Java code

## Changelog Fixes
+ Fixed intermittent NEU rendering issues. - CalMWolfs

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/GuiData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/GuiData.kt
@@ -1,7 +1,6 @@
 package at.hannibal2.skyhanni.data
 
 import at.hannibal2.skyhanni.api.event.HandleEvent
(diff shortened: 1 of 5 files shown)


DavidArthurCole (approved):
🔥

NopoTheGamer (approved):

ThatGravyBoat:
> I guess you cannot post events that are just objects from Java code

It's not that it was done in Java code it's that the object event was Cancellable so it saved the cancelled stated across posts. 

There probably needs to be a warning or style check to make sure you can't make a cancellable event an object class.

CalMWolfs:
> It's not that it was done in Java code it's that the object event was Cancellable so it saved the cancelled stated across posts.

oh that makes much more sense, ty

PR #3004, opened 2024-11-29
Backend: Move Next Jacobs Contest, Best Crop Milestone & Jyrre Timer to Renderable

j10a1n15:
## What
This Pull Request moves said method to be Renderables, which removes the deprecated addCropIcon method

## Changelog Technical Details
+ Moved Next Jacobs Contest, Best Crop Milestone & Jyrre Timer to Renderable. - j10a1n15

PR #3059, opened 2024-12-14
Backend: remove deprecated RenderUtils methods

appable0:
## What
Removes deprecated RenderUtils methods `renderSingleLineWithItems` and `addItemIcon`, as well as an unused private method in that class. Also remove associated utility method in GardenAPI.

Assuming this goes well, I would like to remove the deprecated renderStringsAndItems in a second pass (that one has much broader usage currently, though). I think generally it might be a good idea to only render Renderables and deprecate all other methods rendering lists/strings/etc at Positions, but that can be saved for quite a bit later.

+ Removed some deprecated render methods. - appable


j10a1n15:
Hello, https://github.com/hannibal002/SkyHanni/pull/3004 alr removes the addItemIcon method

appable0:
Closing as largely duplicated by above, will reopen as a dependent PR for the `renderSingleLineWithItems` removal.

PR #3061, opened 2024-12-14
Fix: Item Value Estimation Coloring

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/310

## What
Fixes an issue with items not showing their correct colors, both due to varying name, and recombobulation not being accounted for. If this was intentional let me know and I'll make it a config option.

<details>
<summary>Images</summary>

Before
image

After
image

</details>

## Changelog Fixes
+ Fixed incorrect display colors in Item Value Estimation. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/items/EstimatedItemValue.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/items/EstimatedItemValue.kt
@@ -4,6 +4,7 @@ import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.config.ConfigUpdaterMigrator
 import at.hannibal2.skyhanni.config.features.misc.EstimatedItemValueConfig
+import at.hannibal2.skyhanni.data.jsonobjects.repo.ItemValueCalculationDataJson
 import at.hannibal2.skyhanni.data.jsonobjects.repo.ItemsJson
 import at.hannibal2.skyhanni.events.ConfigLoadEvent
 import at.hannibal2.skyhanni.events.GuiRenderEvent
@@ -49,6 +50,9 @@ object EstimatedItemValue {
     var bookBundleAmount = mapOf<String, Int>()
     private var currentlyShowing = false
 
+    lateinit var itemValueCalculationData: ItemValueCalculationDataJson
+        private set
+
     fun isCurrentlyShowing() = currentlyShowing && Minecraft.getMinecraft().currentScreen != null
 
     @HandleEvent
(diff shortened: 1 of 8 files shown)


NopoTheGamer:
in the tooltip i dont think the "base item" should reflect recomb

hannibal002:
either add reforge name to the item name, then we can simply use the real item name again, or dont show recomb.
Since showing recomb but not reforge feels unintuitive

hannibal002:
unrelated: the data in the repo pr is used to move existing logic outside of live code. this is fine. but this has noting to do with this pr's title or description

src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
```diff
@@ -460,6 +466,28 @@ object ItemUtils {
...
+     */
+    val ItemStack.itemNameBaseRarityAware: String
+        get() {
+            return getRarityName(getItemBaseRarityOrNull())
+        }
+
+    /**
+     * Returns the item name with its full color, i.e. after recombobulation
+     */
+    val ItemStack.itemNameFullyRarityAware: String
```
hannibal002, line 480: what is the difference between this and the display name?

hannibal002 (changes requested):
Im not very happy with this design change. I agree, before it was missing the recomb color entirely, and this was confusing as well, but at least the design was clear.
image


how about we add the display name of the item at the first line, then a empty line, then the base item + everything else? just for the estimated chest value, not touching the actual estiamted item value logic here.


additionallly, the changelog wording is not precise. this affects only the estimated chest value, not the estimated item value.

src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
hannibal002: if the purpose is to remove the reforge name, maybe its better to have the function be called `nameWithoutReforge` instead?

src/main/java/at/hannibal2/skyhanni/features/misc/items/EstimatedItemValueCalculator.kt
```diff
@@ -700,7 +700,7 @@ object EstimatedItemValueCalculator {
             }
         }
 
-        val name = internalName.itemName
+        val name = stack.itemNameBaseRarityAware
```
hannibal002, line 703: we dont want to show the recomb color here, and since the reforge name is also not part of the clean item name, i dont see why this line change is necessary

src/main/java/at/hannibal2/skyhanni/features/misc/items/EstimatedItemValue.kt
```diff
@@ -49,6 +50,9 @@ object EstimatedItemValue {
     var bookBundleAmount = mapOf<String, Int>()
     private var currentlyShowing = false
 
+    lateinit var itemValueCalculationData: ItemValueCalculationDataJson
```
hannibal002, line 53: this should be null, not lateinit. starting with a outdated repo will cause ugly errors instead of clean repo errors

hannibal002 (changes requested):
please also mention the "item calculation stuff moved to repo" part in the changelog/pr desc, as this is arguabily the bigger part of the pr

PR #3065, opened 2024-12-14
Backend: Remove renderSingleLineWithItems

appable0:
## Dependencies
- #3004 

## What
Removes deprecated RenderUtils method `renderSingleLineWithItems`, which was only used by one class after the changes made by the dependent PR.

+ Removed deprecated method; moved raffle event overlay to a renderable. - appable0


appable0:
I think this got merged through another PR.

PR #3069, opened 2024-12-14
Backend: Migrate some methods from renderStringsAndItems

appable0:
## What
Adds container builders (Container.vertical and Container.horizontal) which directly build Renderable containers. Converts CityProjectFeatures (untested for obvious reasons), ComposterDisplay, and ComposterOverlay to use these new builders instead of deprecated method `renderStringsAndItems`. @j10a1n15 and I plan to remove uses of `renderStringsAndItems` in future PRs.

+ Migrated several features to use Renderables. - appable0


hannibal002:
vertical container builder sounds a lot like the addLine wrapper function i added recently

Thunderblade73 (changes requested):
I don't see why we need aynthing from Container.kt. Since it is literrally a one to one wrapper of wrapped methods with bunch of redefinition.
It would be better to simply add to the function extra function prototypes that account for the other style/s.
Here some example (to allow to extra syntax that may be prefered/ do the same as the Container class here)
```kotlin
<In Renderable Companion>
        fun horizontalContainer(
            spacing: Int = 0,
            horizontalAlign: HorizontalAlignment = HorizontalAlignment.LEFT,
            verticalAlign: VerticalAlignment = VerticalAlignment.TOP,
            content: MutableList<Renderable>.() -> Unit,
        ) = horizontalContainer(mutableListOf<Renderable>().apply(content), spacing, horizontalAlign, verticalAlign)

        fun horizontalContainer(
            spacing: Int = 0,
            horizontalAlign: HorizontalAlignment = HorizontalAlignment.LEFT,
            verticalAlign: VerticalAlignment = VerticalAlignment.TOP,
            vararg content: Renderable,
        ) = horizontalContainer(content.toList(), spacing, horizontalAlign, verticalAlign)
```

hannibal002 (changes requested):
I like the PR, but i dont understand why this cant be done as RenderableBuilder instead. the term container sound unrelated to rendering logic to me. can't we simply wrap this into the renderable logic directly?
im looking at existing functions in renderable, e.g.
```kt       
 fun line(builderAction: MutableList<Renderable>.() -> Unit): Renderable {
     return horizontalContainer(buildList { builderAction() })
 }

PR #3077, opened 2024-12-17
Feature: Hitman Claim All Compact

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1309254513357750386 (and in many many other various conversations spread across SH and other servers).

Adds an option to replace the many many messages from Hitman "Claim All"-ing with a succinct summary message.
Also did a lot of work in re-mapping and migrating the Hoppity config options into respective packages, as the list was getting messy as hell.

<details>
<summary>Images</summary>

image

image

</details>

## Changelog New Features
+ Added chat compact for Hitman Claim All. - Daveed
    * Requires `Chat Compact` for Hoppity enabled to work.

+ Moved ~25 Hoppity config options to more sensible groupings/separate configs. - Daveed


hannibal002 (changes requested):
If i read the screenshot correctly, this simply reuses the existing format for the existing "total collected this event" format.
but it should be not doing that in chat, right? it should only show the amount i just claimed, and not the total, imo.

like, if there are 3 to claim, and i already claimed 20 hours ago, i dont want to see a list of 23, i want a summary of 3.

DavidArthurCole:
> like, if there are 3 to claim, and i already claimed 20 hours ago, i dont want to see a list of 23, i want a summary of 3.

What? I genuinely don't understand what you mean by this. If someone hits Claim All, it waits until all hitman rabbits have rolled, and doesn't send any messages until it's done, then sends one message with the summary. It used the same formatting as the hoppity event summary rarity formatter, yeah. It worked in the GUI and it worked in chat back when Hoppity Event Stats was a chat util instead of a UI element.

If your meaning here is that it wouldn't make sense to do this for 3 rabbits versus 28, I can see that point, and to that end I'll add a threshold.

hannibal002 (changes requested):
yeah ignore my previous comment.

`HoppityLocationConfig` should be merged into `HoppityWaypointsConfig`.
I assume you intentionally kept that separated, but i see only confusion. 
Rather, have e.g. the two "duplicates features" (Mark Duplicate Locations and Mark Nearby Duplicates) under a shared subcategory "without egglocator" in `HoppityWaypointsConfig` and the load from neu pv can belong in dev or at the end of `HoppityWaypointsConfig`.

src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityEggsManager.kt
```diff
@@ -347,7 +346,7 @@ object HoppityEggsManager {
 
         val baseConfig = "event.hoppityEggs"
         massMigrationMap.forEach { (oldKey, newKey) ->
-            event.move(70, "$baseConfig.$oldKey", "$baseConfig.$newKey")
+            event.move(79, "$baseConfig.$oldKey", "$baseConfig.$newKey")
```
hannibal002, line 349: ?

hannibal002 (changes requested):

PR #2621, opened 2024-09-30
Feature + Fix: SH Track

Thunderblade73:
## What
Feature to track anything. Currently supported: Items and Powder (but more can be easily added).
This also adds a command handling system in which you only have to specify the arguments and a processing object, which makes the creation of more complex commands easier.

<details>
<summary>Images</summary>

</details>

## Changelog New Features
+ Added /shtrack. - Thunderblade73
    * a powerful command to track anything.
    * has a lot of customization options via arguments.
    * has quick access variants like: /shtrackitems or /shtrackcollections.

## Changelog Improvements
+ Changed /shtrackcollection to use the new shtrack which allows for more customization. - Thunderblade73

## Changelog Fixes
+ Fixed /shtrackcollection not accounting for all items. - Thunderblade73

## Changelog Technical Details
+ Added Renderable.verticalEditTable. - Thunderblade73

PR #3082, opened 2024-12-18
Fix: shtrackcollection not counting Youngite and Obsolite when tracking timite

martimavocado:
## What

## Changelog Fixes
+ Fixed /shtrackcollection not tracking Youngite and Obsolite. - martimavocado

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/CollectionTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/CollectionTracker.kt
@@ -176,7 +176,6 @@ object CollectionTracker {
             }
             if (internalName == "TIMITE".toInternalName() &&
                 (it.getInternalName() == "YOUNGITE".toInternalName()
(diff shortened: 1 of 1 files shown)


CalMWolfs (changes requested):
Seeing that this is the second special case for the collection tracker, maybe it could move this special logic to the repo
Something like:

"Cactus": [
"Cactus_green"
],
"Timite": [
"Youngite",
"Other ite"
]

Thunderblade73:
This can/should be a temporary fix. Since I'm currently completely replacing the logic with https://github.com/hannibal002/SkyHanni/pull/2621. (Which I hope I complete before the end of the year)

hannibal002:
I agree with Thunderblade here, using repo now will cause bigger merge conflicts. with the later pr /rewrite we also will move this into the repo with the format suggested by Cal.

hannibal002 (approved):

PR #3087, opened 2024-12-19
Fix: Revert Failures

DavidArthurCole:
## What
Fixes detekt failures that were introduced during code cleanup earlier.

exclude_from_changelog


hannibal002:
i prefer this solution over the simple revert i did. ill revert my revert if you open this pr

PR #3098, opened 2024-12-20
Backend: TickEvent repeatSeconds Deprecation

j10a1n15:
## What
~~I kinda feel like repeatSeconds should have its @Deprecation (oops, hi 👋) removed.~~ I did that, pls feedback 
In some cases, like in the screenshot below the repeatSeconds event is better to look at.
image
In some rare cases using this is 100% better

The Ordanata/Larva changes might seem weird at first, but the main method alr ran every second, and the other variables had no other use cases so this should be fine.

~~Keeps the .repeatSeconds in the Reputation Helper since this one should be resolved by "removing" TablistUtils.getTablist()~~ Done by #3207

+ Resolved deprecations in TickEvent.repeatSeconds(). - j10a1n15


src/main/java/at/hannibal2/skyhanni/features/rift/area/wyldwoods/RiftLarva.kt
```diff
@@ -26,15 +26,13 @@ object RiftLarva {
     private val LARVA_HOOK = "LARVA_HOOK".toInternalName()
 
     @HandleEvent
-    fun onTick(event: SkyHanniTickEvent) {
+    fun onSecondPassed(event: SecondPassedEvent) {
```
hannibal002, line 29: this will only check the hand once per second instead of 20 times per second

src/main/java/at/hannibal2/skyhanni/features/rift/area/wyldwoods/RiftOdonata.kt
```diff
@@ -25,15 +25,13 @@ object RiftOdonata {
     private val emptyBottle = "EMPTY_ODONATA_BOTTLE".toInternalName()
 
     @HandleEvent
-    fun onTick(event: SkyHanniTickEvent) {
+    fun onSecondPassed(event: SecondPassedEvent) {
```
hannibal002, line 28: again check hand. maybe use the hand change event ?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/rift/area/wyldwoods/RiftLarva.kt
j10a1n15: like I mentioned in the pr desc, this doesnt matter since the only method that actually uses `hasHookInHand` runs once per second aswell
so `hasHookInHand` updating 20 times instead of just once was useless to begin with

hannibal002 (approved):
reading pr descriptions is hard

PR #3101, opened 2024-12-21
Improvement: Discrite Held Time

lunaynx:
## What
Added Discrite support for time held in lore and stack size. Also refactored related code a bit, and made the stack size a single option for all Time Pocket items, just like the lore option is, since I do not see a reason why someone would want it for some but not all of them.

~~(I kept the enum name `BOTTLE_OF_JYRRE` because I didn't want to make a mess with a multi-level migration considering it also has a legacy ID.)~~

## Changelog Improvements
+ Added time-held display for Discrite in stack size and lore. - Luna


CalMWolfs (changes requested):
All the code is fine except that fixRemovedConfigElement is only called for config version 29 which means that this wont actually do the config fix that you are wanting it to for 99% of people.
You'd have to make a new function for a new config fix version. You may as well change the enum name now as well in this new function

CalMWolfs (approved):

PR #3104, opened 2024-12-21
Backend: Better formatting settings

hannibal002:
## What
Apparently, they are better now. I'd like to have feedback from others that actually know what they are doing.
This no longer breaks my auto-formatting (makes it so that IntelliJ formatting doesn't break detekt rules)

+ Improved Detekt and .editorconfig rules. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/BossbarData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/BossbarData.kt
@@ -31,6 +31,7 @@ object BossbarData {
     fun onTick(event: LorenzTickEvent) {
         val bossbarLine = BossStatus.bossName ?: return
         //#else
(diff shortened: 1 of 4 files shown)


src/main/java/at/hannibal2/skyhanni/data/BossbarData.kt
CalMWolfs, line 1: This is not correct, you can't just move the preprocessing comments

CalMWolfs (changes requested):
We have a custom rule for annotation spacing in detekt so that it works with preprocessed comments. Add whatever you are trying to do into that custom rule instead.

Unfortunately there is no way to have preprocessing comments while also having it follow typical kotlin formatting standards unless we want to duplicate lines

CalMWolfs (approved):

PR #3106, opened 2024-12-22
Backend: Storage Classes to Kotlin

DavidArthurCole:
## What
Converts the following Java classes to Kotlin classes:
- `Features`
- `SackData`
- `PlayerSpecificStorage`
- `ProfileSpecificStorage`
- `Storage`

There was a slight edit to `EnumSkippingTypeAdapterFactory` necessary to make this work, but nothing breaking.
In the process of testing this currently, however I've managed to load up my dev env with an existing config, and all data and settings appear to be intact.

+ Converted Storage and related classes to Kotlin. - Daveed


hannibal002 (changes requested):
Since we dont want to add `@JVMFields` above every Category, we will need to wait for a moulconfig update.
```
java.lang.RuntimeException: @Category on non public field private at.hannibal2.skyhanni.config.features.About at.hannibal2.skyhanni.config.Features.about in class at.hannibal2.skyhanni.config.Features
	at io.github.notenoughupdates.moulconfig.internal.Warnings.warn0(Warnings.java:60) ~[legacy-3.0.0-beta.7.jar:?]
	at io.github.notenoughupdates.moulconfig.internal.Warnings.warn(Warnings.java:64) ~[legacy-3.0.0-beta.7.jar:?]
	at io.github.notenoughupdates.moulconfig.internal.Warnings.warn(Warnings.java:68) ~[legacy-3.0.0-beta.7.jar:?]
	at io.github.notenoughupdates.moulconfig.processor.ConfigProcessorDriver.processCategoryMeta(ConfigProcessorDriver.java:153) ~[legacy-3.0.0-beta.7.jar:?]
	at io.github.notenoughupdates.moulconfig.processor.ConfigProcessorDriver.processConfig(ConfigProcessorDriver.java:180) ~[legacy-3.0.0-beta.7.jar:?]
	at at.hannibal2.skyhanni.config.ConfigManager.firstLoad(ConfigManager.kt:95) ~[main/:?]
	at at.hannibal2.skyhanni.SkyHanniMod.init(SkyHanniMod.kt:71) ~[main/:?]
```

hannibal002:
The crash in Dev Env is now fixed, but the updated MoulConfig version added a new error with selected text boxes in the config.

src/main/java/at/hannibal2/skyhanni/config/Features.kt
```diff
@@ -0,0 +1,148 @@
...
+import io.github.notenoughupdates.moulconfig.processor.ProcessedCategory
+
+class Features : Config() {
+    private val discord = MyResourceLocation("notenoughupdates", "social/discord.png")
+    private val github = MyResourceLocation("notenoughupdates", "social/github.png")
+    private val patreon = MyResourceLocation("notenoughupdates", "social/patreon.png")
+
+    override fun shouldAutoFocusSearchbar(): Boolean {
+        return false
+    }
```
martimavocado, line 38: is this not the default behavior? this should probably be deleted

hannibal002: yes, this is the default behavior. But somehow the value is toggled inside, when setting this from true to false will actually enable auto focus search bar (what we want).
When we remove the function, we lose the information that we intentionally set this to false

PR #3108, opened 2024-12-23
Fix: Class Cast Error in Stray Warning

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1320740309453504563
I'm like 90% sure this fix will work, however I'm currently on route to vacation, and don't have IntelliJ, so I'm going off vs code and prayers.

## Changelog Fixes
+ Fixed error when opening Choc Factory via NEU buttons. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryStrayWarning.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryStrayWarning.kt
@@ -95,11 +95,8 @@ object ChocolateFactoryStrayWarning {
         else event.strayHighlight()
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryStrayWarning.kt
```diff
@@ -95,8 +95,14 @@ object ChocolateFactoryStrayWarning {
         else event.strayHighlight()
     }
 
+    private fun GuiContainerEvent.getEventChest(): ContainerChest? = try {
+        gui.inventorySlots as? ContainerChest
+    } catch (e: ClassCastException) {
```
Thunderblade73: The excpetion can never happen with as? it can only happen with as

Thunderblade73 (changes requested):

PR #3116, opened 2024-12-24
Fix: Mining Event Display Sorting

penguin4life21:
## What
Changed mining island event display to sort Dwarven Mines first and then Crystal Hollows second

## Changelog Improvements
+ Sorting mining islands in mining event. - Penguin4life


penguin4life21:
closed due to issues with codebase, apologies

PR #3124, opened 2024-12-27
Fix: Force zero price for glitched Gemstone Collection item

lunaynx:
## What
Force price to 0 for glitched `GEMSTONE_COLLECTION` item, which is supposed to be a placeholder item for Gemstone collection, but it currently has a 350M lowest BIN price, because it used to be possible to glitch it out of menus, and it is causing insane prices to be shown for Gemstone Powder in Garden visitor rewards.

## Changelog Fixes
+ Fixed Gemstone Powder showing as being worth billions of coins in Garden visitor rewards. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/ItemPriceUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/ItemPriceUtils.kt
@@ -25,6 +25,7 @@ object ItemPriceUtils {
         pastRecipes: List<PrimitiveRecipe> = emptyList(),
     ): Double? {
         when (this) {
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/utils/ItemPriceUtils.kt
```diff
@@ -25,6 +25,7 @@ object ItemPriceUtils {
         pastRecipes: List<PrimitiveRecipe> = emptyList(),
     ): Double? {
         when (this) {
+            NEUInternalName.GEMSTONE_COLLECTION -> return 0.0
```
NopoTheGamer, line 28: this should probably be a repo thing at this point

PR #3131, opened 2024-12-30
Fix: Heat "null" on Custom Scoreboard

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1322382921755005009
Fixes heat being null when 0

## Changelog Fixes
+ Fixed heat showing as `null` in Custom Scoreboard. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityEggLocator.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityEggLocator.kt
@@ -171,7 +171,7 @@ object HoppityEggLocator {
         drawDynamicText(location.up(), possibleDuplicateLabel, 1.5)
     }
 
-    private fun shouldShowAllEggs() = config.showAllWaypoints && !locatorInHotbar && HoppityEggType.eggsRemaining()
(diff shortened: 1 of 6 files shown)


j10a1n15:
I feel like this will fuck this part up 

https://github.com/hannibal002/SkyHanni/blob/beta/src/main/java/at/hannibal2/skyhanni/features/gui/customscoreboard/ChunkedStatsLine.kt (I can't copy lines on mobile, should be at ~70 or smth (where heat is))

j10a1n15:
It would probably be best to add a y check similar to this

https://github.com/j10a1n15/CustomScoreboard/blob/main/src/main/kotlin/gay/j10a1n15/customscoreboard/feature/customscoreboard/elements/ElementHeat.kt

j10a1n15:
Crazy how my concern wasn't implemented yet this was still merged

hannibal002:
ive missed your comments, im sorry!

anyway, I dont see the problem the pr should cause here, we still compare the same value
image
image

hannibal002:
wait do we want to use null and and intentionally hide it when not found?

j10a1n15:
Yes

PR #3132, opened 2024-12-30
Improvement: garden keybinds

Cedric-Ab:
## What

You can now set 5 keybind layouts, each with different keybinds.
You can then for each crop select what layout you'd wish to use.
Also moved it into it's own category

<details>
<summary>Images</summary>

image

</details>

## Changelog Improvements
+ Improved garden keybinds. - Cédric Ab
    * Now set up to 5 different keybind layouts to choose from with each crop.


src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCustomKeybinds.kt
```diff
@@ -67,43 +76,96 @@ object GardenCustomKeybinds {
...
+            config.layout1.attack, config.layout1.useItem, config.layout1.left, config.layout1.right,
+            config.layout1.forward, config.layout1.back, config.layout1.jump, config.layout1.sneak,
+            config.layout2.attack, config.layout2.useItem, config.layout2.left, config.layout2.right,
+            config.layout2.forward, config.layout2.back, config.layout2.jump, config.layout2.sneak,
+            config.layout3.attack, config.layout3.useItem, config.layout3.left, config.layout3.right,
+            config.layout3.forward, config.layout3.back, config.layout3.jump, config.layout3.sneak,
+            config.layout4.attack, config.layout4.useItem, config.layout4.left, config.layout4.right,
+            config.layout4.forward, config.layout4.back, config.layout4.jump, config.layout4.sneak,
+            config.layout5.attack, config.layout5.useItem, config.layout5.left, config.layout5.right,
+            config.layout5.forward, config.layout5.back, config.layout5.jump, config.layout5.sneak
```
Thunderblade73: Small Suggestion:
Better to add a function that gets all properties as a list and then simply combine the 5 lists. Makes it better scale able.

Cedric-Ab: @Thunderblade73 
would it be better to use stuff like getting all the variables in the config to then dynamically do the stuff. so you could just adjust the number of layouts in config and then the kotlin would automatically just work with the amount of layouts in the config?
My main concern with this is the unneeded complexity this would generate

Cedric-Ab: @Thunderblade73 
like the new version?

Cedric-Ab:
@Thunderblade73
would it be better to use stuff like getting all the variables in the config to then dynamically do the stuff. so you could just adjust the number of layouts in config and then the kotlin would automatically just work with the amount of layouts in the config?
My main concern with this is the unneeded complexity this would generate

Thunderblade73:
> @Thunderblade73 would it be better to use stuff like getting all the variables in the config to then dynamically do the stuff. so you could just adjust the number of layouts in config and then the kotlin would automatically just work with the amount of layouts in the config? My main concern with this is the unneeded complexity this would generate

Yeah that would be better imo. If done correctly it should reduce complexity and not add more (as far as I can tell).
Also if you want to give the layout enum more functionallity I would suggest to move it to kotlin, since java code gets more cluttered and the config supports kotlin enums.

Btw. enums have a way to get all entries ```.values``` (for java) and ```.entries``` for kotlin (before the dot is the Enum Type)

Cedric-Ab:
nonessential TODOs:

- [x] fix detekt thing
- [x] make kotlin detect amount of layouts

Cedric-Ab:
> > @Thunderblade73 would it be better to use stuff like getting all the variables in the config to then dynamically do the stuff. so you could just adjust the number of layouts in config and then the kotlin would automatically just work with the amount of layouts in the config? My main concern with this is the unneeded complexity this would generate
> 
> Yeah that would be better imo. If done correctly it should reduce complexity and not add more (as far as I can tell). Also if you want to give the layout enum more functionallity I would suggest to move it to kotlin, since java code gets more cluttered and the config supports kotlin enums.
> 
> Btw. enums have a way to get all entries `.values` (for java) and `.entries` for kotlin (before the dot is the Enum Type)

do you think I could replace the map: layouts with the enum?
and would it need to be in kotlin or could it stay in java?

Thunderblade73:
> do you think I could replace the map: layouts with the enum? and would it need to be in kotlin or could it stay in java?

Yes, you can replace the map.
The can stay in java yes, but makeing them in kotlin is easier.

src/main/java/at/hannibal2/skyhanni/features/garden/farming/keybinds/KeyBindLayouts.kt
```diff
@@ -0,0 +1,31 @@
...
+                keyBindAttack, keyBindUseItem, keyBindLeft, keyBindRight,
+                keyBindForward, keyBindBack, keyBindJump, keyBindSneak
+            )
+
+            map.clear()
+            keyBindings.zip(
+                getAllKeybindingsFromLayout(layout)
+            ).forEach { (keyBinding, setKeyProperty) ->
+                map[keyBinding] = setKeyProperty.get() // Update map directly
+            }
```
Thunderblade73: Imo a ``var map: Map`` is better than a ``val map: MutableMap`` here.
Also:
```suggestion
            map = keyBindings.associate(
                getAllKeybindingsFromLayout(layout).get()
            )
```

src/main/java/at/hannibal2/skyhanni/features/garden/farming/keybinds/KeyBindLayouts.kt
```diff
@@ -0,0 +1,31 @@
...
+    LAYOUT_3("Layout 3", GardenAPI.config.keyBind.layout3, mutableMapOf()),
+    LAYOUT_4("Layout 4", GardenAPI.config.keyBind.layout4, mutableMapOf()),
+    LAYOUT_5("Layout 5", GardenAPI.config.keyBind.layout5, mutableMapOf());
+
+    fun buildKeybindLayoutMap() {
+        with(mcSettings) {
+            val keyBindings = listOf(
+                keyBindAttack, keyBindUseItem, keyBindLeft, keyBindRight,
+                keyBindForward, keyBindBack, keyBindJump, keyBindSneak
+            )
```
Thunderblade73: The ``with`` only applies is used for this so close it here.
Something like that.
```Kotlin
            val keyBindings = with(mcSettings) {
             listOf(
                keyBindAttack, keyBindUseItem, keyBindLeft, keyBindRight,
                keyBindForward, keyBindBack, keyBindJump, keyBindSneak
            )
            }
```

src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCustomKeybinds.kt
```diff
@@ -67,43 +79,106 @@ object GardenCustomKeybinds {
...
+//                 currentLayout = layouts[cropLayoutSelection[cropInHand]]
+//             }
+//         }
+//     }
+
+    @JvmStatic
+    fun getAllKeybindingsFromLayout(layout: KeyBindLayout) = listOf(
+        layout.attack, layout.useItem, layout.left, layout.right,
+        layout.forward, layout.back, layout.jump, layout.sneak
+    )
```
Thunderblade73: An extension function works better here. (Or make it a member function in java, with an ArrayList/Array)
```suggestion
    @JvmStatic
    fun KeyBindLayout.getAllKeybindings() = listOf(
        this.attack, this.useItem, this.left, this.right,
        this.forward, this.back, this.jump, this.sneak
    )
```

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCustomKeybinds.kt
Cedric-Ab: done

Cedric-Ab:
@Thunderblade73 
I hope it's good now

nvm, it's not working properly...

it's working now!

Cedric-Ab:
TODO: move GardenCustomKeybinds.kt into keybinds

Cedric-Ab:
@Thunderblade73 
Is cropLayoutSelection initialized on startup?
is configloadevent called on startup?

Thunderblade73:
> @Thunderblade73 Is cropLayoutSelection initialized on startup? is configloadevent called on startup?

It is loaded via Gson at ConfigLoadEvent (or what ever it is called) or if no value is present in the config json than it calls the constructor.

Cedric-Ab:
> > @Thunderblade73 Is cropLayoutSelection initialized on startup? is configloadevent called on startup?
> 
> It is loaded via Gson at ConfigLoadEvent (or what ever it is called) or if no value is present in the config json than it calls the constructor.

The cropLayoutSelection is a map inside of the kotlin, you got that like this, right?

Thunderblade73:
> The cropLayoutSelection is a map inside of the kotlin, you got that like this, right?

Nope, I saw the value in KeyBindConfig and assume you where referencing it,

Cedric-Ab:
> > The cropLayoutSelection is a map inside of the kotlin, you got that like this, right?
> 
> Nope, I saw the value in KeyBindConfig and assume you where referencing it,

So is ConfigLoadEvent called upon launch?

Thunderblade73:
> is configloadevent called on startup?

yes

> Is cropLayoutSelection initialized on startup?

Yes, but you override it every update().
Btw. it does not need to exist since you can simply add a extension function to CropType that gets the value from an when since it does not compute anything since the getLayoutByDisplayName(something.toString()) is redundant because ``something`` is already a Layout

Cedric-Ab:
> > is configloadevent called on startup?
> 
> yes
> 
> > Is cropLayoutSelection initialized on startup?
> 
> Yes, but you override it every update(). Btw. it does not need to exist since you can simply add a extension function to CropType that gets the value from an when since it does not compute anything since the getLayoutByDisplayName(something.toString()) is redundant because `something` is already a Layout

Yes, I see
This is a thing that needs to be done

### TODO

Cedric-Ab:
TODO:

- [ ] Move GardenCustomKeybinds.kt into keybinds

Cedric-Ab:
so when should this get merged?

PR #3138, opened 2024-12-31
fix heat

hannibal002:
## What
fix heat after broken by
- #3131

exclude_from_changelog


src/main/java/at/hannibal2/skyhanni/features/gui/customscoreboard/elements/ScoreboardElementHeat.kt
```diff
@@ -9,7 +9,7 @@ import at.hannibal2.skyhanni.features.gui.customscoreboard.CustomScoreboardUtils
 // scoreboard update event
 object ScoreboardElementHeat : ScoreboardElement() {
     override fun getDisplay(): String? {
-        val heat = getHeat()
+        val heat = getHeat() ?: "§c♨ 0"
```
j10a1n15:
```suggestion
        val heat = getHeat() ?: return null
```

j10a1n15 (changes requested):

j10a1n15 (approved):

PR #3142, opened 2025-01-01
Fix: HitmanAPI Calculations

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/3106

## What
Fixes a couple miscellaneous issues with HitmanAPI calculations, leading to off values when there are unclaimed eggs. Reliant on 3106 since the storage class needed a rework anyways. Didn't need migrations since these values are refreshed when needed.

## Changelog Fixes
+ Fixed Rabbit Hitman claim estimates. - Daveed


src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
```diff
@@ -192,7 +192,7 @@ class ProfileSpecificStorage {
 
         class HitmanStatsStorage {
             @Expose
-            var availableEggs: Int? = null
+            var availableHitmanEggs: Int = 0
```
hannibal002, line 195: cant the value be null for legacy users and then cause a config problem without migration fix?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
DavidArthurCole: the key changed, if it was null before, it won't map to the new name

hannibal002 (approved):

PR #3143, opened 2025-01-01
Backend: Less Lazy

ItsEmpa:
## What
This pr changes a lot of lazy vals to either be getters if they aren't called a lot, or just be initialized normally.

+ Changed some lazy vals to getters or normal vals. - Empa


hannibal002 (changes requested):
while i agree that we have too many by lazy blocks and we should check the usefulness of those, i still dont like some of the changes done.

after reviewing all changes of this PR, i think ServerBlockChangeEvent and BlockClickEvent should be reverted, as they perfectly fit the usage of lazy loading, and replacing them to `=` is worse for performance. Sure the difference is not huge, but why not use lazy when we clearly can here?

hannibal002 (approved):

PR #3144, opened 2025-01-01
Fix: Rift Transfer Logic

lunaynx:
## What
Fixed some Rift transfer related stuff.

## Changelog Fixes
+ Fixed Rift-Exportable items incorrectly marked as non-transferable with Hide Not Clickable Items. - Luna
    * Items can still be transferred inside the Rift but cannot be sold for motes.
+ Fixed "Rift-Exported" items incorrectly marked as "salable" for motes in the features "Show Motes Price" and "Hide Not Clickable Items". - Luna


lunaynx:
I forgot that this is also incorrectly applying to Rift-Transferable items, I guess I'll need to do a lore check.

hannibal002:
please include the feature name "hide not clickable items" in the changelog

lunaynx:
Done

PR #3146, opened 2025-01-02
Fix: skill overflow event not posted

appable0:
## What
Fixes SkillOverflowLevelUpEvent never being posted (not sure if this was an intentional removal; it was removed in #3068 without any mention). Also removes unneeded dependency of overflow level up message on the skill progress display.

<details>
<summary>Images</summary>
Screenshot 2025-01-01 at 11 45 26 PM

</details>

## Changelog Fixes
+ Fixed overflow level-up message not showing and removed dependency on skill progress display. - appable

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/SkillAPI.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/SkillAPI.kt
@@ -9,6 +9,7 @@ import at.hannibal2.skyhanni.events.InventoryFullyOpenedEvent
 import at.hannibal2.skyhanni.events.NeuRepositoryReloadEvent
 import at.hannibal2.skyhanni.events.SecondPassedEvent
 import at.hannibal2.skyhanni.events.SkillExpGainEvent
+import at.hannibal2.skyhanni.events.SkillOverflowLevelUpEvent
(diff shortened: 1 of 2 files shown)


hannibal002 (changes requested):
while this properly reverts the removal of the old code, the old code was flawed if i see this correctly:
~~we compare two xp values exactly, not if the xp value is exceeded, this event will almost never get triggered in practice, right?~~ (im blind sry)
also this will not trigger for overflow between 50 and 60 for skills that are max at 50 from hypixel

do you want to fix this in this pr or should we merge it as is and do this later?

appable0:
> do you want to fix this in this pr or should we merge it as is and do this later?

I think that might need to be saved for the future unless I'm missing an obvious way to do that (I don't see anywhere that we have max skill caps stored, only default skill caps?).

hannibal002:
Use https://github.com/appable0/SkyHanni/blob/41d7bf47d654431b6da3f3498ba5acf0efd0e861/src/main/java/at/hannibal2/skyhanni/api/SkillAPI.kt#L153

hannibal002 (approved):
its fine

PR #3151, opened 2025-01-03
Backend: GardenAPI - replace LorenzWorldChangeEvent with IslandChangeEvent

Cedric-Ab:
## What
Replace LorenzWorldChangeEvent with IslandChangeEvent in the GardenAPI.kt

+ Replaced `LorenzWorldChangeEvent` with `IslandChangeEvent` in GardenAPI. - Cédric Ab


hannibal002 (changes requested):
world change is not the same as island change. And at the time of writing of the garden api class, there was a meaningful distinction in it.

as a result of that, the delay can now be removed maybe.
also please test this logic chnage with all features relying on the "item in hand" information from the garden api as those two events are getting fired in different moments of the island swap logic

as reminder:
world change event gets fired when hypixel teleports the player from one world to another, this happens 2-3 timer per /warp action in hypixel for some reason.
island change event gets fired when the scoreboard/tab list information indicates the player is now on a different skyblock island

Cedric-Ab:
### Features to check

#### uses GardenToolChangeEvent
- [x] FarmingFortuneDisplay (/)
- [x] GardenOptimalSpeed
- [x] GardenYawAndPitch
- [x] CropMoneyDisplay 
- [x] DicerRngDropTracker (
- [ ] FarmingWeightDisplay
- [x] GardenCropSpeed
- [x] WrongFungiCutterWarning
- [x] FarmingLane
- [x] CaptureFarmingGear

#### uses toolInHand
- [x] FarmingFortuneDisplay (/)
- [x] GardenYawAndPitch (/)
- [x] SensitivityReducer (/)
- [x] GardenCustomKeybinds (/)

#### uses itemInHand
- [x] GardenCropMilestoneDisplay (/)

#### uses cropInHand
- [x] FarmingFortuneDisplay (/)
- [x] GardenStartLocation (/)

hannibal002:
You can create a check box list:
```
- [ ] a
- [ ] b
- [ ] c
```

Cedric-Ab:
@hannibal002 
how thoroughly do these tests have to be?
also I just looked at the implementation of the events etc and watched out for anything that would be time depending etc

hannibal002:
when you dont notice a bug by using the feature for couple seconds its fine imo

Cedric-Ab:
@hannibal002 
I tested everything, so should be fine to push or so

PR #3156, opened 2025-01-04
Backend: ModVersion changes

ItsEmpa:
## What
This PR changes some things related to the new ModVersion class and how we store the current version. It makes the ModVersion class implement the Comparable interface instead of just having the compareTo operator, and it changes the string replaced with the actual version string by blossom to be the one accessed, instead of blossom replacing the one in the @.Mod annotation and the version val being lazily accessed by getting it in the indexed mods list. Also changed the CurrentVersion implementation in UpdateManager to use ModVersion class for comparing versions.

+ Improved ModVersion class and integrated it into UpdateManager. - Empa


CalMWolfs (approved):
lgtm

PR #3163, opened 2025-01-05
Improvement: King Talisman Helper Reset Command

lunaynx:
## Changelog Improvements
+ Added `/shresetkinghelper` command to reset the King Talisman Helper. - Luna


jani270 (changes requested):
New commands should be added in the feature class and not in the Commands file. 

See > https://github.com/hannibal002/SkyHanni/blob/2fe67fcf9be36f6718219916dc2d52c9ee3d73d5/src/main/java/at/hannibal2/skyhanni/features/misc/visualwords/ModifyVisualWords.kt#L36

hannibal002:
I think we already have that command somewhere?

lunaynx:
I don't think so, `/shkingfix` is different, doesn't reset the tracker.

jani270:
> I think we already have that command somewhere?

There is `shkingfix` https://github.com/hannibal002/SkyHanni/blob/2fe67fcf9be36f6718219916dc2d52c9ee3d73d5/src/main/java/at/hannibal2/skyhanni/config/commands/Commands.kt#L456

j10a1n15:
Well I suppose we can move this command aswell while we are touching this file

CalMWolfs (approved):

PR #3168, opened 2025-01-06
Backend: Removing more Deprecated methods

j10a1n15:
## What
.asInternalName() has been deprecated for 2-3 months now I think.
And removed some other basically unused methods, rewrote some small things to not use deprecated stuff anymore.

+ Removed deprecated methods. - j10a1n15


CalMWolfs (approved):
lgtm

PR #3171, opened 2025-01-07
Fix: Define backports for autoupdater

CalMWolfs:
## What
As far as i understand it, a backport is defined as a version where the beta version (second number) is 0 and the bugfix version (third number) isnt a 0
so 
1.1.0 wouldnt be a backport but is a beta
1.0.1 would be a backport but isnt a beta
1.0.0 wouldnt be a backport or a beta

## Changelog Fixes
+ Fixed autoupdater not working with new backport updates in certain situations. - CalMWolfs


src/main/java/at/hannibal2/skyhanni/features/misc/update/CustomGithubReleaseUpdateSource.kt
```diff
@@ -11,6 +12,12 @@ import moe.nea.libautoupdate.UpdateData
  */
 class CustomGithubReleaseUpdateSource(owner: String, repository: String) : GithubReleaseUpdateSource(owner, repository) {
 
+    override fun selectUpdate(updateStream: String, releases: MutableList<GithubRelease>) = when (updateStream) {
+        "pre" -> findLatestRelease(releases.filter { !it.isDraft && !ModVersion.fromString(it.tagName).isBackport })
+        "full" -> findLatestRelease(releases.filter { !it.isDraft && !it.isPrerelease })
+        else -> null
+    }
```
lineargraph: You could in theory also just override findLatestRelease to order by the modversion:
```kt
		override fun findLatestRelease(validReleases: Iterable<GithubRelease>): UpdateData {
			return validReleases.asSequence().mapNotNull { findAsset(it) }.maxBy { ModVersion.fromString(it.tagName) }
		}
```

lineargraph: This would also allow for upgrading to backport releases if no betas have yet been released:

1.0.0 -> 1.0.1 (no 1.1.0 released yet)

lineargraph (approved):

lineargraph (changes requested):
Does not currently support upgrading to backports if there have been no betas but there have been backports

src/main/java/at/hannibal2/skyhanni/features/misc/update/CustomGithubReleaseUpdateSource.kt
CalMWolfs: Your solution does allow this but it the version comparison that we have it the mod would mean that using this would also try and update users from the stable/backport branch to the betas which is not intended. I guess I can look into our version comparison further but I feel the current solution I have is more than enough and a backport before a beta is rare enough that it is not needed

lineargraph: Im not sure how my change would make users upgrade to betas? The default selectUpdate function already filters for github prereleases, so the findLatestRelease function will not get any betas on the full update stream

hannibal002: > Im not sure how my change would make users upgrade to betas? The default selectUpdate function already filters for github prereleases, so the findLatestRelease function will not get any betas on the full update stream

iirc the problem is that when someone updates from full version to beta their update channel gets set to beta, and stays on beta, even after updating to a full version later

lineargraph: > iirc the problem is that when someone updates from full version to beta their update channel gets set to beta, and stays on beta, even after updating to a full version later

I am entirely unsure how my code would ever upgrade someone to a beta. The point is to update people who already *want* a beta version to a newer bugfix version if no betas have been released.

hannibal002: https://github.com/hannibal002/SkyHanni/blob/b50fa0505b34572bef60355263ef58b8c839aa3b/src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt#L88

lineargraph: Again: I dont update people to betas with my suggested piece of code, so the `SkyHanniMod.isBetaVersion` two lines above would not come to be true.

hannibal002: I was not talking about your suggestion here, but rather about the state of the auto updater afaik

CalMWolfs:
I think ive done what you suggested nea.
I will not change whatever hannibal is talking about as that is outside of the scope of this pr and imo not necessary to really touch at all as its worked fine for 1.5 years

hannibal002 (approved):
waiting for neas approval

src/main/java/at/hannibal2/skyhanni/features/misc/update/CustomGithubReleaseUpdateSource.kt
```diff
@@ -11,6 +12,11 @@ import moe.nea.libautoupdate.UpdateData
  */
 class CustomGithubReleaseUpdateSource(owner: String, repository: String) : GithubReleaseUpdateSource(owner, repository) {
 
+    override fun findLatestRelease(validReleases: Iterable<GithubRelease>): UpdateData {
+        return validReleases.asSequence().maxBy { ModVersion.fromString(it.tagName) }.let { findAsset(it) }
+            ?: throw IllegalStateException("No valid release found")
```
lineargraph, line 17: im pretty sure traditionally this just returns null, but that is just fine. it will practically never happen, and even if it did, it would still just be caught.

lineargraph (approved):

PR #3175, opened 2025-01-07
Fix: workaround to fix minion upgrade feature

hannibal002:
## What
Added workaround to fix minion upgrade feature showing outside minion menus 

## Changelog Fixes
+ Fixed Minion Upgrade Feature appearing outside minion menus. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/minion/MinionFeatures.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/minion/MinionFeatures.kt
@@ -22,6 +22,7 @@ import at.hannibal2.skyhanni.events.MinionStorageOpenEvent
 import at.hannibal2.skyhanni.events.SkyHanniRenderEntityEvent
 import at.hannibal2.skyhanni.events.entity.EntityClickEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
+import at.hannibal2.skyhanni.test.command.ErrorManager
 import at.hannibal2.skyhanni.utils.BlockUtils.getBlockStateAt
 import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.CollectionUtils.editCopy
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/minion/MinionFeatures.kt
```diff
@@ -177,7 +177,15 @@ object MinionFeatures {
     @SubscribeEvent
     fun onInventoryOpen(event: InventoryFullyOpenedEvent) {
         if (!enableWithHub()) return
-        if (!minionTitlePattern.find(event.inventoryName)) return
+        if (!minionTitlePattern.find(event.inventoryName)) {
+            // This should never happen, but somehow it still does. Therefore the workaround
+            if (minionInventoryOpen) {
+                minionInventoryOpen = false
+                MinionCloseEvent().post()
+                ChatUtils.debug("Detected unexpected minion menu closing. please report in discord what you have done between the last minion open and now, exactly.")
```
martimavocado: this being a debug message might mean not a lot of people see it

CalMWolfs: Yeah this should use errormanager.skyhanniError or whatever it is

hannibal002: ~~Fair, but on the other side i dont want to spam this in chat if we finally see it. maybe should do a ChatUtils.sendToBeta()~~ cals message is better

CalMWolfs: We can also log extra info here which can be used to help debug the cause of the problem if we want

martimavocado: yeah i agree with cal here, ideally we would be able to hide the error after knowing the cause with repoerrors but i don't think that's a thing right now

martimavocado (approved):

src/main/java/at/hannibal2/skyhanni/features/minion/MinionFeatures.kt
CalMWolfs: Interesting idea, right now we can change the error message but not hide it. I like the idea of being able to hide the message and then once the update is released we can tell them to update to the latest version

CalMWolfs (approved):

src/main/java/at/hannibal2/skyhanni/features/minion/MinionFeatures.kt
hannibal002: We can hide it! if replaceMessage is null ,we dont show anything in chat

hannibal002: image

PR #3177, opened 2025-01-07
Fix: Overflow levelups no longer hardcoded

mariethepolarfox:
Fix: Overflow levelups no longer hardcoded so they now appear correctly

## What
Fixes Overflow levelup messages only appearing when above Level 60, even if the max level of the skill is not 60

## Changelog Fixes
+ Fixed overflow skill level-up messages only appearing above level 60. - Helium9


hannibal002:
We decided against this, since this could confuse greatly for taming and farming.

lunaynx:
This changes nothing for Taming and Farming.

mariethepolarfox:
I will try to fix taming and farming (I mainly intended the fix for levels with a hard cap at 50)

hannibal002:
> This changes nothing for Taming and Farming.

Do we assume the player already has a max cap of 60 for taming and farming?

hannibal002 (approved):
apparently we do

PR #3178, opened 2025-01-07
Backend + Fix: Use correct type for item attributes

lunaynx:
## What
Fixed SkyBlockItemModifierUtils incorrectly treating multiple attributes as booleans, when they're actually integers. While the current way of doing it caused no issues so far except for the Book of Stats case, relying on implicit type conversion here is bad practice.

## Changelog Fixes
+ Fixed Book of Stats not counting in Estimated Item Value for items with 0 kills. - Luna

+ Fixed some item attributes treated as Boolean instead of Int. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
@@ -168,20 +168,19 @@ object SkyBlockItemModifierUtils {
         }
     }
 
-    fun ItemStack.isRecombobulated() = getAttributeBoolean("rarity_upgrades")
+    fun ItemStack.isRecombobulated() = (getAttributeInt("rarity_upgrades") ?: 0) > 0
 
-    fun ItemStack.hasJalapenoBook() = getAttributeBoolean("jalapeno_count")
+    fun ItemStack.hasJalapenoBook() = (getAttributeInt("jalapeno_count") ?: 0) > 0
 
     fun ItemStack.hasEtherwarp() = getAttributeBoolean("ethermerge")
 
-    fun ItemStack.hasWoodSingularity() = getAttributeBoolean("wood_singularity_count")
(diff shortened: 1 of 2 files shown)


hannibal002 (changes requested):
If you find yourself doing this check often, consider an extension function:

`fun Int?.isPositive(): Boolean = (this ?: 0) > 0`

src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
```diff
@@ -168,20 +169,19 @@ object SkyBlockItemModifierUtils {
...
+    fun ItemStack.hasWoodSingularity() = getAttributeInt("wood_singularity_count").isPositive()
 
     fun ItemStack.hasDivanPowderCoating() = getAttributeBoolean("divan_powder_coating")
 
-    fun ItemStack.hasArtOfWar() = getAttributeBoolean("art_of_war_count")
+    fun ItemStack.hasArtOfWar() = getAttributeInt("art_of_war_count").isPositive()
 
-    // TODO untested
-    fun ItemStack.hasBookOfStats() = getAttributeBoolean("stats_book")
+    fun ItemStack.hasBookOfStats() = getAttributeInt("stats_book") != null
```
hannibal002, line 184: why is this treated different than the others?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
lunaynx: This is intentional. For this, we only care if if the tag exists at all, which means the item has Book of Stats, even if it's zero. The value can range from 0 to integer limit, and is simply the number of kills recorded on the item.

In the other cases, a tag with the value of 0 should technically not exist, but if it does, it logically means the same thing as the absence of the tag.

hannibal002 (approved):

PR #3179, opened 2025-01-08
Fix: Rift Blood Effigies (Take 2)

lunaynx:
## What
Second attempt at fixing Rift Blood Effigy Display.

## Changelog Fixes
+ Fixed Rift Blood Effigies detection. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/ScoreboardData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/ScoreboardData.kt
@@ -130,9 +130,10 @@ object ScoreboardData {
 
     }
 
-    private fun cleanSB(scoreboard: String): String {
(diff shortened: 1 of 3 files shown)


lunaynx:
I had used an `Array` to let the compiler know that the values will never be null, because IIRC earlier in my refactor there was an excessive amount of null checks. But now there isn't much, so the map solution should be fine too.

As for `editCopy`, I was guessing it's unnecessary but I didn't want to remove it just in case I was wrong.

hannibal002:
Please do not use editCopy at all. The design is really bad and will be removed some time soon anyway (no deprecation tag yet, mb)
all access functions to the map are in the same thread. this is fine here.

about the array earlier:
grafik
(the map is not nullable anywhere here

lunaynx:
> the map is not nullable anywhere here

Yeah, it was about access by index I think, because `index` can technically be an arbitrary integer and maps don't have a fixed size. But there's only a few simple checks that don't clutter the code now, so it's fine.

PR #3181, opened 2025-01-08
Improvement: "Not Accepted" Crimson Isle quest state

Donaldino7712:
## What
This PR removes the "Not Accepted" state for Daily Quests in the Crimson Isle.
After the 0.20.9 update, quests are accepted automatically.

<details>
<summary>Images</summary>
image
</details>

## Changelog Improvements
+ Added auto-accept for Crimson Isle quests. - Donaldino7712
    * Removed unused "Not Accepted" quest state after 0.20.9 SB update.


lunaynx (changes requested):
You can't just remove the `NOT_ACCEPTED` enum value, that will very likely cause issues. Any existing quests the player has with that state should be migrated to `ACCEPTED`.

lunaynx (approved):
Not tested but looks like it should be fine now.

PR #3186, opened 2025-01-09
Fix: GardenKeybinds not working with Inventory/Chat Keybind

Thunderblade73:
## Changelog Fixes
+ Fixed Custom Keybinds not working with inventory or chat keybind. - Thunderblade73

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCustomKeybinds.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCustomKeybinds.kt
@@ -44,7 +44,7 @@ object GardenCustomKeybinds {
     fun isKeyPressed(keyBinding: KeyBinding, cir: CallbackInfoReturnable<Boolean>) {
         if (!isActive()) return
         val override = map[keyBinding] ?: run {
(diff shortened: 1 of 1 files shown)


Cedric-Ab:
so what is the code now effectively doing?

Thunderblade73:
@Cedric-Ab 
> so what is the code now effectively doing?

We did only replaced the check for the keybind with our check (so the replaced key). But other stuff that would have the same Key where not checked. Eg: Attack Keybind would be mapped to E but the Inventory Keybind would still look up for e and not now of the replacement.
The change I did was that every non mapped Keybind looks at the mapping to see if it's key is present which means it shouldn't do something.

Cedric-Ab:
@Thunderblade73 
So if you bind forward to spacebar and then you press W it doesn’t go forward, is that what is happening now?
So this doesn’t fix E still opening the inventory, right?
It seems as if the problem is not in our logic but rather in the event logic, not?

Thunderblade73:
> So if you bind forward to spacebar and then you press W it doesn’t go forward, is that what is happening now?

That was always working.
> So this doesn’t fix E still opening the inventory, right?

It does, since it blocks the inventory keybind. If any of the mapped Keybind have as input the same Key as the Inventory Keybind.

PR #3202, opened 2025-01-10
Backend: Detekt Non-Mutable Types in Config

DavidArthurCole:
## What
Adds a new Detekt rule that enforces that all collections should use mutable types.

<details>
<summary>Images</summary>

image

image

</details>

exclude_from_changelog

### DIFF
--- a/detekt/detekt.yml
+++ b/detekt/detekt.yml
@@ -23,6 +23,10 @@ ImportRules:
     CustomImportOrdering:
         active: true
 
+StorageRules:
+    EnforceConfigMutableCollections:
+        active: true
+
 style:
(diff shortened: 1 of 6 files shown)


hannibal002:
Why do we not want mutable data? Feels lilke a downgrade

DavidArthurCole:
> Why do we not want mutable data? Feels lilke a downgrade

We do ... this makes sure all collection types _are_ mutable.

hannibal002:
ok. and why do we not want immutabe types then? same problem

DavidArthurCole:
> ok. and why do we not want immutabe types then? same problem

Because it crashes the game if you try to edit a non-mutable collection from the config. It would only happen the first time someone loads the mod, as from GSON it'll be mutable, but still don't want people to crash.

PR #3206, opened 2025-01-11
Fix: faulty page scrolling invert logic

Cedric-Ab:
## Changelog Fixes
+ Fixed page scrolling with inverted bypass. - Cédric Ab


Thunderblade73 (changes requested):
The cool down is not unnecessary since it prevents spamming. Since without that you can move every second between 20 pages which is not reasonable to say that you could do it without the mod. Which would break our Interpretation of the Hypixel Rules for modding.
I'm fine with a bit lower cool down, but removal is out of the question.

Cedric-Ab:
> The cool down is not unnecessary since it prevents spamming. Since without that you can move every second between 20 pages which is not reasonable to say that you could do it without the mod. Which would break our Interpretation of the Hypixel Rules for modding. I'm fine with a bit lower cool down, but removal is out of the question.

I mean I can just spam click it, but yes it would have to update
How about 0.05s?
I‘d say max is 0.1s

Thunderblade73:
> > The cool down is not unnecessary since it prevents spamming. Since without that you can move every second between 20 pages which is not reasonable to say that you could do it without the mod. Which would break our Interpretation of the Hypixel Rules for modding. I'm fine with a bit lower cool down, but removal is out of the question.
> 
> I mean I can just spam click it, but yes it would have to update How about 0.05s? I‘d say max is 0.1s

I would not go lower than 300ms since that is already ridiculous to execute manual but I would say at least possible. Btw. 0.05s is the technical limit since that is exactly one tick which would be the case for not doing any cool down.

hannibal002:
Why do we even have a duration here? Can't we just say one click per full inventory open event? Then it's depending on the users internet connection and not a cooldown that is way too long for American users and still unrealistic short for EU users

Cedric-Ab:
> Why do we even have a duration here? Can't we just say one click per full inventory open event? Then it's depending on the users internet connection and not a cooldown that is way too long for American users and still unrealistic short for EU users

Seems good, though harder to implement

ItsEmpa:
> Seems good, though harder to implement

shouldnt be, just set a boolean to true on inventory open, and set it to false when you scroll

hannibal002 (changes requested):
do not use time

Cedric-Ab:
> do not use time

How about qeueing?

hannibal002:
> How about qeueing?

Queuing would do more than the vanilla behavior. In vanilla, when you do 10 clicks in one second you get 2–3 pages forward in 2–3 seconds (for me at least), when you add queuing, this would go forward automatic for 5–10 seconds by 10 pages. This is not ok IMO

src/main/java/at/hannibal2/skyhanni/features/inventory/PageScrolling.kt
```diff
@@ -71,7 +76,17 @@ object PageScrolling {
             patterns.matches(it.stack?.displayName)
         } ?: return
         InventoryUtils.clickSlot(slot.slotNumber)
-        cooldown = 1.0.seconds.fromNow() // 1 second is not specific it is just a reasonable cooldown
+
+        currentlyScrollable = false
+        println("set currentlyScrollable to false")
```
CalMWolfs: remove println

src/main/java/at/hannibal2/skyhanni/features/inventory/PageScrolling.kt
```diff
@@ -71,7 +76,17 @@ object PageScrolling {
...
+        cooldown = 1.0.seconds.fromNow()
+    }
+
+    @HandleEvent
+    fun onInventoryOpen(event: InventoryOpenEvent) {
+        if (!isEnabled()) return
+        currentlyScrollable = true
+        println("set currentlyScrollable to true")
```
CalMWolfs: remove println

CalMWolfs (changes requested):

hannibal002:
@Cedric-Ab please re-request reviews from us if you are ready

grafik

src/main/java/at/hannibal2/skyhanni/features/inventory/PageScrolling.kt
```diff
@@ -49,20 +51,26 @@ object PageScrolling {
...
+//     with cooldown beeing a fallback to still allow for scrolling if currentlyScrollable is stuck
+    private var currentlyScrollable = false
     private var cooldown = SimpleTimeMark.farPast()
 
     @SubscribeEvent
     fun onLorenzTick(event: LorenzTickEvent) {
         if (!isEnabled()) return
         if (InventoryUtils.inStorage() && InventoryUtils.isNeuStorageEnabled) return
-        if (cooldown.isInFuture()) return
+        if (!currentlyScrollable && !cooldown.isInFuture()) return
```
hannibal002: the check here is wrong. this currently allows me to spam click, sending multiple packets in under a second

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/PageScrolling.kt
Cedric-Ab: exactly, that's the point behind it, it's either the inventory loaded again or 1 second has passed, then you can scroll again

hannibal002: Why do we want to send multiple packets? this feels very sus. i assumed the logic should be "either allow sending after a second, or after the next inventory is loaded, whatever happens first"

hannibal002 (changes requested):
see convo above

src/main/java/at/hannibal2/skyhanni/features/inventory/PageScrolling.kt
Cedric-Ab: This is what the logic should be doing rn, is it not?

hannibal002: no, currently we send the packet many times, as said already

Cedric-Ab:
@hannibal002 is the logic correct now? if not what is the logic doing now exactly and what would you expect it to do differently?

hannibal002:
The logic is sending clicks too fast, multiple times a second.

I expect it to only produce one click per second, or allow me to  click faster when the new page is already loaded.

look at the milliseconds here
grafik

hannibal002 (changes requested):
see above

hannibal002 (approved):

PR #3207, opened 2025-01-11
Backend: ReputationHelper "Rewrite"

j10a1n15:
## What
Rewrote ReputationHelper to use renderables, changed the factionType getter to use TabwidgetUpdateEvent.
Removed clickType in favor of clickTypeEnum, renamed the clickTypeEnum variable to just clickType

+ Rewrote ReputationHelper to use Renderables & TabWidgetUpdateEvent. - j10a1n15
+ Removed deprecated clickType field. - j10a1n15

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/nether/reputationhelper/CrimsonIsleReputationHelper.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/nether/reputationhelper/CrimsonIsleReputationHelper.kt
@@ -26,7 +26,6 @@ import at.hannibal2.skyhanni.utils.KeyboardManager.isKeyHeld
 import at.hannibal2.skyhanni.utils.LorenzUtils.isInIsland
 import at.hannibal2.skyhanni.utils.LorenzVec
 import at.hannibal2.skyhanni.utils.NEUItems
-import at.hannibal2.skyhanni.utils.RegexUtils.firstMatcher
 import at.hannibal2.skyhanni.utils.RenderUtils.renderRenderables
 import at.hannibal2.skyhanni.utils.renderables.Renderable
 import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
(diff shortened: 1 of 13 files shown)


src/main/java/at/hannibal2/skyhanni/features/nether/reputationhelper/CrimsonIsleReputationHelper.kt
```diff
@@ -95,6 +98,15 @@ class CrimsonIsleReputationHelper(skyHanniMod: SkyHanniMod) {
         dirty = true
     }
 
+    @HandleEvent
+    fun onWidgetUpdate(event: WidgetUpdateEvent) {
+        if (!event.isWidget(TabWidget.REPUTATION)) return
+
+        event.widget.pattern.firstMatcher(event.lines) {
```
Thunderblade73: Can be replaced with ```TabWidget.REPUTATION.matchMatcherFirstLine(){```

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/nether/reputationhelper/CrimsonIsleReputationHelper.kt
ItsEmpa: that is not the same, the method you said only checks the first line of the widget, and .firstMatcher does the matcher for all lines until it finds one that matches

Thunderblade73: But it is the first line it tries to match xD

ItsEmpa: nvm im stupid it does want to check the title

Thunderblade73 (approved):
Changes seem to be fine. 
Only thing I want to note is that the Extension functions addQuests, addKuudra, addBosses can get annoying since the come up as suggestions for any MutableList<Renderable>. Not sure how to do it better though.

j10a1n15:
> Only thing I want to note is that the Extension functions addQuests, addKuudra, addBosses can get annoying since the come up as suggestions for any MutableList.

I dont think so since they are in classes, not objects

PR #3212, opened 2025-01-12
Backend: Delay HypixelData check

ItsEmpa:
## What
This PR delays the equals check that compares the data gotten with hypixel mod api and the old hypixel data by 1 second, which should stop errors from happening in rare cases where old hypixel data is some milliseconds faster.

+ Delayed HypixelData comparison with HypixelLocationAPI. - Empa


hannibal002:
~~I have the feeling we use the tostring == "x:y:z:" behaviour somewhere else as well and it would either break a feature or look ugly. can we keep the explicit tostring impl please?~~
wrong PR oops

PR #3214, opened 2025-01-12
Backend: LorenzVec generated methods

ItsEmpa:
## What
Removes a couple of methods in LorenzVec that should already get generated by being a data class. The only exception to this is is the .toString() method, where the only difference is that instead of using curly brackets it uses parenthesis.

exclude_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/LorenzVec.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/LorenzVec.kt
@@ -73,8 +73,6 @@ data class LorenzVec(
 
     fun add(x: Int = 0, y: Int = 0, z: Int = 0): LorenzVec = LorenzVec(this.x + x, this.y + y, this.z + z)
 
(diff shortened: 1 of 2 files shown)


hannibal002:
I have the feeling we use the tostring == "x:y:z:" behaviour somewhere else as well and it would either break a feature or look ugly. can we keep the explicit tostring impl please?

Thunderblade73:
> ~~I have the feeling we use the tostring == "x:y:z:" behaviour somewhere else as well and it would either break a feature or look ugly. can we keep the explicit tostring impl please?~~

(Ups stupid, I read what I wanted to read xD)

The asStoredString can also be used in:

In Graph in the TypeAdapter.
Also in Parkour Helper, but there the formatting is strange.

ItsEmpa:
the x:y:z is a different method, not toString()

hannibal002:
> the x:y:z is a different method, not toString()

oh compared to the weird (xyz) format?
yeah fair, then i dont have any complains about this pr

CalMWolfs (approved):

PR #3149, opened 2025-01-02
Backend: Module Restructure

lunaynx:
## What
Restructured a few outlier modules to `@SkyHanniModule` format.

## Dependencies
- #3162 
(not really required but rather wait on this due to conflicts)

## Changelog Technical Details
+ Restructured several outlier modules to the `@SkyHanniModule` format. - Luna

PR #3216, opened 2025-01-12
Why does this silly class not have skyhanni module annotation

CalMWolfs:
exclude_from_changelog


lunaynx:
Yeah, it's a weird one. I've fixed it not being a SkyHanniModule in #3149.

CalMWolfs:
> Yeah, it's a weird one. I've fixed it not being a SkyHanniModule in #3149.

draft 😔

lunaynx:
I think it's in a usable state, it was a draft originally when the goal was to fix `/shreloadlisteners` but I gave up on that for now. The related PR linked in it will conflict a bit with it however, so it will need to be fixed up.

CalMWolfs:
> I think it's in a usable state, it was a draft originally when the goal was to fix `/shreloadlisteners` but I gave up on that for now. The related PR linked in it will conflict a bit with it however, so it will need to be fixed up.

ive marked the related pr as soon and then ill make yours soon once thats merged

PR #3219, opened 2025-01-13
Fix: Missing Label on DungeonFinder Item Stack

j10a1n15:
## What
idk how the original test case succeeded, it shouldn't have 🤷 

## Changelog Fixes
+ Fixed missing label on Partyfinder ItemStack. - j10a1n15


hannibal002 (changes requested):
can we fix the current floor pattern instead of creating a new one? (if the current one is used as is in another spot, we can just make the `:` optional in the regex
image

j10a1n15:
> can we fix the current floor pattern instead of creating a new one? (if the current one is used as is in another spot, we can just make the `:` optional in the regex 

well they both have different purposes, so it wouldn't be ideal. but if that's still fine then sure okay

hannibal002:
both approaches are fine

hannibal002 (approved):

PR #3227, opened 2025-01-13
Fix: Ultimate Wither Scroll

lunaynx:
## What
Fixed Wither Blades with "Ultimate Wither Scroll" admin item not being detected as equivalent to a regular three-scroll Wither Impact.

## Changelog Fixes
+ Fixed Estimated Item Value for Wither Blades with Ultimate Wither Scroll. - Luna


lunaynx:
Fair on making it a set, at the end of the day we don't care about the order besides maybe a fixed order on display.

hannibal002:
for the order: i plan a function that will order by price for every category inside the estimated item value:
image
image

hannibal002:
Basically an extension of this 
image

PR #3232, opened 2025-01-15
Fix: FAQ formatting

ItsEmpa:
Fixes a formatting error in the faq that would make point 11 be displayed incorrectly (see image comparison)
| Before  | After |
| --- | --- |
| image | image |

and adds another point in it about current mayor mode in diana trackers

Btw, when copying these into the discord channel and editing the previous message, dont edit the message, do ctrl a and then paste the new one. Instead, do ctrl a, delete the text and *then* paste the new one. If you do it as i said first, it gets pasted like this
imagen
with the extra `>` characters in front of every line.

exclude_from_changelog


hannibal002:
Updad on discord, thanks for the hint!

PR #3235, opened 2025-01-17
Feature: Compact Jacob Claim

DavidArthurCole:
## What
Adds compacting for the rewards claim from Jacob. Does some setup for an API I'll start in a later PR.

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added Compact Jacob Claim. - Daveed
    * When collecting the rewards from many farming contests at the same time at Anita.
    * Compact the Jacob Claim message, only showing full information when hovering.


src/main/java/at/hannibal2/skyhanni/features/chat/CompactJacobClaim.kt
```diff
@@ -0,0 +1,223 @@
...
+            return event.block("compact_jacob_bulk_claim")
+        }
+
+        if (!inLoop) return
+
+        rewardsPattern.matchMatcher(message) {
+            return event.block("compact_jacob_bulk_claim")
+        }
+
+        val startingHash = rewardSet.hashCode()
```
hannibal002, line 132: why are we comparing hashes? cant we simply compare the instance of the object?

src/main/java/at/hannibal2/skyhanni/features/chat/CompactJacobClaim.kt
```diff
@@ -0,0 +1,223 @@
...
+
+        ticketPattern.matchMatcher(message) {
+            val amount = group("amount").formatInt()
+            when (group("type")) {
+                "Jacob's" -> rewardSet = rewardSet.copy(jacobTickets = amount)
```
hannibal002: why are we using copy all there and not reusing the same instance of rewards all the time?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/chat/CompactJacobClaim.kt
DavidArthurCole:
```
        val hashNow = rewardSet.hashCode()
        DelayedRun.runDelayed(eventDelay) {
            if (rewardSet.hashCode() == hashNow) {
                inLoop = false
                publishEvent()
            }
        }
 ```
 
 Either way we're storing an object for comparison in the delayed run - figured it was better to store the hashcode rather than the full set in memory?

hannibal002: i would generally prefer dirty over hashes for this, but sure

PR #3238, opened 2025-01-17
Fix: Ironman and non-ranked's /show messages not being detected

Nessiesson:
## What
The regex for the /show text did not deal with the fact that the ironman symbol is randomly put in that message. o.O Also now deals with non-ranked player's /show message having a 'rando' §7 instead of §f. :( Presumably also catches other cases but idfk.

This PR does not deal with:
* Levels not showing up in /show at all.
* Bingo/stranded player icons potentially having strange other formatting. idk.
<details>
<summary>Images</summary>

image
(sorry for being in this picture random people)
</details>

## Changelog Fixes
+ Fixed `/show` messages from ironman and non-ranked players not reformatted by Chat Player Messages. - Nessiesson


lunaynx:
Yeah, it was funny when Hypixel added this as a "fix" to the Ironman symbol not being displayed in `/show` a while back. I think they forgot that it also isn't displayed in regular chat messages and is only added by mods.

PR #2272, opened 2024-07-29
Feature: Hoppity Call Warning

DavidArthurCole:
## What
Context: https://discord.com/channels/997079228510117908/1264604838994378875
Add an option to get screen-flashed and dinged when Hoppity calls via abiphone to sell you a rabbit.
Largely reuses configuration code from stray rabbit warnings.
Also adds a config hotkey for picking up the Hoppity call.

The warning will stop either after 10 seconds, or when the player picks up the call.

<details>
<summary>Images</summary>

Config option(s)
image

Short Video
https://www.youtube.com/watch?v=F5fQ7JjEFa4

</details>

## Changelog New Features
+ Added optional warning when Hoppity calls you with a rabbit to sell. - Daveed
+ Added hotkey for picking up Abiphone calls from Hoppity. - Daveed

## Changelog Technical Details
+ Rebased Hoppity config to be in its own package. - Daveed

PR #3241, opened 2025-01-17
Backend: Disable hoppity call warning by default

Chissl:
## What
Hoppity call warning is quite intrusive and should be disabled by default

exclude_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/event/hoppity/HoppityCallWarningConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/event/hoppity/HoppityCallWarningConfig.java
@@ -19,7 +19,7 @@ public class HoppityCallWarningConfig {
     @ConfigOption(name = "Hoppity Call Warning", desc = "Warn when hoppity is calling you.")
     @ConfigEditorBoolean
     @FeatureToggle
-    public boolean enabled = true;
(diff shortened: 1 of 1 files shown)


DavidArthurCole:
https://github.com/hannibal002/SkyHanni/pull/2272#discussion_r1748942214

Chissl:
> [#2272 (comment)](https://github.com/hannibal002/SkyHanni/pull/2272#discussion_r1748942214)

I disagree with this, if this was just a chat message it would be fine, but the flash is very annoying as well as being confusing for players that don't know which mod is causing it

DavidArthurCole:
> I disagree with this, if this was just a chat message it would be fine, but the flash is very annoying as well as being confusing for players that don't know which mod is causing it

I don't entirely disagree, hence my originally making it auto-disabled.
Mainly just wanted to note it was hanni himself that objected to that.

hannibal002:
the popup title + chat + sound can surely be default enabled, since you have the chat message there is no way not to know where this is coming from and how to disable.

the "full screen flash" might be annoying, could discuss if wrapping this into an separate toggle "warn me extra annoying and repeat the ding sound like adhd" option and have this opt in?

hannibal002:
arguments

hannibal002 (approved):

PR #3257, opened 2025-01-20
Fix: Bestiary Tab Widget

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1298478804167037038
Fixes an issue where the player's IGN was incorrectly being appended to the end of the Bestiary Widget display. Also adds capability to do regex matching of what expected widget lines are in case this is not an isolated issue.

## Changelog Fixes
+ Fixed Widget Display occasionally showing player names. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/gui/TabWidgetDisplay.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/gui/TabWidgetDisplay.kt
@@ -19,7 +19,7 @@ import java.util.regex.Pattern
 enum class TabWidgetDisplay(
     private val configName: String?,
     vararg val widgets: TabWidget,
-    var expectedLinePattern: Pattern? = null,
+    var expectedLinePattern: Pattern? = null, // Todo: Remove when TabWidget data is no longer bad
 ) {
     SOULFLOW(null, TabWidget.SOULFLOW),
     COINS("Bank and Interest", TabWidget.BANK, TabWidget.INTEREST),
@@ -66,6 +66,7 @@ enum class TabWidgetDisplay(
         private val config get() = SkyHanniMod.feature.gui.tabWidget
         private fun isEnabled() = LorenzUtils.inSkyBlock && config.enabled
 
+        // Todo: Remove when TabWidget data is no longer bad
         // <editor-fold desc="Patterns">
         /**
          * REGEX-TEST: §6§lBestiary:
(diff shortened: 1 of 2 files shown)


Thunderblade73:
This is not an isolated issue that comes from the widget abstraction itself. Since the ```TabWidget.filterTabList``` does not cut off at the correct position in some edge case (I would assume it has to do with having only few widgets).
Fixing that function would be preferred but this bandage fix would work for now (but at a comment that it should be removed down the line if the other bug is fixed).

CalMWolfs (changes requested):
Im not a fan of this fix, as while we see it commonly with the bestiary widget, this is not exclusive to that widget and can happen to almost any widget (unless that widget is fixed near the start of the tablist).

Im pretty sure this is an issue with how we get the tablist and could be solved by changing the `.dropLast(1)` in `TabListData` to `.subList(0, 80)` or something similar, 

The above solution is what we use for the compact tablist and in your previous bug report in the discord, while the `/shcopytablist` was incorrect, the compact tablist still showed the correct data as it uses this solution.

hannibal002 (changes requested):
see comments above

DavidArthurCole:
> Im pretty sure this is an issue with how we get the tablist and could be solved by changing the `.dropLast(1)` in `TabListData` to `.subList(0, 80)` or something similar,

Needed a bit of fixing for this specific case, but this did work 🥳

src/main/java/at/hannibal2/skyhanni/features/gui/TabWidgetDisplay.kt
CalMWolfs, line 1: are the changes in this file still needed?

DavidArthurCole: Not still strictly needed but I don't see any harm in keeping them around?

CalMWolfs: Well if its fixed i dont see a reason to keep it, there is already the todo to remove it once fixed so it should probably be removed

hannibal002 (approved):
LGTM

NopoTheGamer:
the Changelog Fixes should probably be changed away from bestiary

CalMWolfs:
> the Changelog Fixes should probably be changed away from bestiary

fixed

PR #3258, opened 2025-01-21
Fix: Anvil Combine Helper bug

Ownwn:
## What
Fix the Anvil Combine Helper not highlighting matching items when the book is in the second anvil slot

<details>
<summary>Images</summary>

## OLD
old

## NEW
new

</details>

## Changelog Fixes
+ Fixed Anvil Combine Helper not highlighting when a book is in the second slot. - Ownwn


hannibal002 (approved):
While looking at this feature I was like "why is the code so bad", then I noticed the changes you did and i was like "ok you repaired it without changing the least amount of code, thats actually impressive" (the feature should cache the slots and should use item ids instead of lore checks)

PR #3262, opened 2025-01-22
Backend: Replace `LorenzUtils.inSkyBlock` early returns

DavidArthurCole:
## What
Replaces all reasonable instances of inSkyBlock early return checks with `onlyOnSkyblock = true` in the HandleEvent annotation. Also introduced a new detekt rule to enforce it.

exclude_from_changelog

### DIFF
--- a/detekt/detekt.yml
+++ b/detekt/detekt.yml
@@ -25,6 +25,10 @@ ImportRules:
     CustomImportOrdering:
         active: true
 
+SkyHanniStyle:
+    InSkyBlockEarlyReturn:
+        active: true
+
 style:
     MagicNumber: # I, Linnea Gräf, of sound mind and body, disagree with disabling this rule
         active: false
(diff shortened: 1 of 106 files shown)


DavidArthurCole:
Sorry for +200/-300 hanni 🥺

hannibal002:
not as bad as it could have been
image

hannibal002 (approved):
Before:
image
After:
image

jani270:
LMAO

DavidArthurCole:
The fact that there were that many added in 7 commits is further enforcing why we need this ... fixing now.

hannibal002:
this will be fire after merging this into all 100+ open prs and see detekt exploding

DavidArthurCole:
It's also gonna be great as more things are changed from LorenzEvent to SkyHanniEvent, and even more event types become target-able by this rule !1!1!!!

CalMWolfs:
Noooooo now the final like 400 uses of Lorenz event are going to be more pain to change to skyhanni event 😭

DavidArthurCole:
image

Also this is even fun-ner, because we're removing `LorenzUtils.` from a lot of these files, the fixes are import related as well

hannibal002:
> Noooooo now the final like 400 uses of Lorenz event are going to be more pain to change to skyhanni event 😭

for the sake of small prs, simply baseline the new  detekt errors and let david fix it later :P

PR #3270, opened 2025-01-23
Backend: Chat event to skyhanni event

CalMWolfs:
## What
Chat chat chat, detekt detekt detekt

+ Renamed `LorenzChatEvent` to `SkyHanniChatEvent`. - CalMWolfs


hannibal002:
originally i wanted with this transformation so that we also only use clean strings in the new chat event, but ig we can transition to this later anway, especially with all the new detekt stuff happening

PR #3272, opened 2025-01-23
Backend: more detekt rules

hannibal002:
## What
public function needs explicit return type, vars that could be vals

+ Added detekt rules for public function needs explicit return type and for vars that could be vals. - hannibal2


hannibal002:
ok this is not good

PR #3204, opened 2025-01-11
Backend: Config to Kotlin #5

DavidArthurCole:
## What
Ports the `config/features/dungeon` package to Kotlin.
Also replaces a bunch of repetitive instances of "open the sound list" with a static method.

exclude_from_changelog

PR #3285, opened 2025-01-24
Backend + Fix: Warn for immutable types in the config

lineargraph:
## Dependencies

- ~~https://github.com/hannibal002/SkyHanni/pull/3204~~ (cherry-picked not to include)

Note that the actual important commit at https://github.com/nea89o/SkyHanni/commit/ef30d3dcde75818dd3009c5b5197b26e8247b645 does not depend on that PR; it should be relatively easily cherry-picked.

## Changelog Fixes
+ Fixed crash when editing the order of powder mining and chat formatting features. - !nea

+ Added extra detekt lint warning against using immutable container types in config. - !nea
    * MoulConfig modifies `List`s directly, requiring `MutableList`, a constraint not enforceable in Java's type system. We manually enforce this since moulconfig does not.


detekt/src/main/kotlin/potentialbugs/ImmutableTypesWithExpectedInteriorMutabilityInConfig.kt
```diff
@@ -0,0 +1,86 @@
...
+	override val issue: Issue
+		get() = Issue(
+			"ImmutableTypesWithExpectedInteriorMutabilityInConfig",
+			Severity.Defect,
+			"Disallow using types which may disallow MoulConfig from mutating them, due to the ",
+			Debt.FIVE_MINS,
+		)
+
+	companion object {
+		val configPackage = FqName("at.hannibal2.skyhanni.config")
```
hannibal002, line 31: We want to use immutable types in the storage part of the config.
Do you need to change this to `at.hannibal2.skyhanni.config.features`?

hannibal002 (changes requested):
see above

detekt/src/main/kotlin/potentialbugs/ImmutableTypesWithExpectedInteriorMutabilityInConfig.kt
lineargraph: The config package here is just a short circuit. Each individual property will also be checked for an `@ConfigOption` property:

https://github.com/nea89o/SkyHanni/blob/e270bdadc933dee713e5acd32046a49561b763df/detekt/src/main/kotlin/potentialbugs/ImmutableTypesWithExpectedInteriorMutabilityInConfig.kt#L71-L73

CalMWolfs (approved):

PR #3288, opened 2025-01-24
Fix: Museum Milestone not affecting bitsPerCookie()

j10a1n15:
## What
Fixed bitsPerCookie() not using the "newly" added bits multiplier you get from museum milestones, causing bitsAvailable (used by custom scoreboard) and thunders feature which changes smth in the cookie lore (I forgot its name) not using the correct numbers 

## Changelog Fixes
+ Fixed museum milestone not affecting available bits. - j10a1n15


j10a1n15:
> The function handleFameRankGui appears to be too complex based on Cyclomatic Complexity (complexity: 34). Defined complexity threshold for methods is set to '25'

image
Okay so why do we default ignore this rule on like 80% of skyhanni code
At that point we can just not use the rule

Same issue with LongMethod
image

hannibal002:
>  Okay so why do we default ignore this rule on like 80% of skyhanni code

we ignore the rule only on old code that needs rewrites

j10a1n15:
> we ignore the rule only on old code that needs rewrites

Thats not true, ignoring all methods with @HandleEvent get ignored, even the new ones.
To ignore errors the baseline exists

hannibal002:
> > we ignore the rule only on old code that needs rewrites
> 
> Thats not true, ignoring all methods with @HandleEvent get ignored, even the new ones. To ignore errors the baseline exists

I used to use @Supress before i knew baseline existed. any examples where we supress long or complicated detekt warnings? want to trasnform them now

j10a1n15:
> I used to use @Supress before i knew baseline existed. any examples where we supress long or complicated detekt warnings? want to trasnform them now

image
only in the commands object

hannibal002:
https://github.com/hannibal002/SkyHanni/commit/33766e764f5fa09693b7588dbfc3aa947c61f901

hannibal002:

(please remove the whole "what" text, or write sometiing more technically explainatory than in the changelog part

j10a1n15:
Oh shit I forgot to change that

src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
```diff
@@ -272,6 +272,9 @@ class ProfileSpecificStorage {
 
         @Expose
         var boosterCookieExpiryTime: SimpleTimeMark? = null
+
+        @Expose
+        var museumMilestone: Int = -1
```
hannibal002: can we use null instead of -1?

src/main/java/at/hannibal2/skyhanni/data/BitsApi.kt
```diff
@@ -55,6 +58,12 @@ object BitsApi {
             profileStorage?.boosterCookieExpiryTime = value
         }
 
+    var museumMilestone: Int
```
hannibal002: i dont see the need for this. we can use the profile storage directly

src/main/java/at/hannibal2/skyhanni/data/BitsApi.kt
```diff
@@ -245,138 +272,142 @@ object BitsApi {
         }
     }
 
-    fun bitsPerCookie(): Int = (defaultCookieBits * (currentFameRank?.bitsMultiplier ?: 1.0)).toInt()
+    fun bitsPerCookie(): Int {
+        val museumBonus = 1 + (museumMilestone * 0.01) // Adds 1% per level
```
hannibal002: instead of assuming milestone is 0 if config is not loaded, rather dont inlcude that part of the formular

src/main/java/at/hannibal2/skyhanni/data/BitsApi.kt
```diff
@@ -245,138 +272,142 @@ object BitsApi {
...
-                        }
-                    }
-                    continue@items
-                }
-            }
+    private fun handleMuseumGui(stacks: List<ItemStack>) {
+        val stack = stacks.firstOrNull { museumRewardStackPattern.matches(it.displayName) } ?: return
+
+        museumMilestonePattern.firstMatcher(stack.getLore()) {
+            museumMilestone = group("milestone").toInt()
```
hannibal002: please use formatInt over toInt. if the parsing fails bc hypixel changed in their format, we see the string in the error message directly

src/main/java/at/hannibal2/skyhanni/data/BitsApi.kt
```diff
@@ -245,138 +272,142 @@ object BitsApi {
...
-            bitsAvailableMenuPattern.firstMatcher(lore) {
+    private fun handleFameRankGui(stacks: List<ItemStack>) {
+        processFameRankStacks(stacks)
+        processBitsStacks(stacks)
+        processCookieStacks(stacks)
+    }
+
+    private fun processFameRankStacks(stacks: List<ItemStack>) {
+        val stack = stacks.firstOrNull { fameRankGuiStackPattern.matches(it.displayName) } ?: return
+        for (line in stack.getLore()) {
```
hannibal002, line 338: fix code duplication in this for loop

src/main/java/at/hannibal2/skyhanni/data/BitsApi.kt
```diff
@@ -245,138 +272,142 @@ object BitsApi {
...
+        val museumBonus = 1 + (museumMilestone * 0.01) // Adds 1% per level
+        return (defaultCookieBits * museumBonus * (currentFameRank?.bitsMultiplier ?: 1.0)).toInt()
+    }
 
     @HandleEvent
     fun onInventoryFullyOpened(event: InventoryFullyOpenedEvent) {
         if (!isEnabled()) return
 
-        val stacks = event.inventoryItems
+        val stacks = event.inventoryItems.values.toList()
```
hannibal002: remove the last "toList", and then change all the function parameters below from list to collection

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/BitsApi.kt
j10a1n15: `* 1` is easier to understand, im so confused what you want from me here

hannibal002: Ok yeah my suggestion is confusing. I'll look at this later

hannibal002: No longer necessary after your changes

hannibal002 (approved):

PR #3305, opened 2025-01-25
Backend: More Early Return Nuking 🥳

DavidArthurCole:
## What
Expands the detekt rules around `onlyOnIsland` to cover more Api usage, and account for `onlyOnSkyblock` unnecessary redundancy checking (i.e., not needing both `onlyOnSkyblock = true` and `onlyOnIsland = ...`).

<details>
<summary>Images</summary>

image

image

</details>

exclude_from_changelog


src/main/java/at/hannibal2/skyhanni/features/garden/contest/FarmingContestApi.kt
```diff
@@ -59,16 +58,13 @@ object FarmingContestApi {
...
+    @HandleEvent(onlyOnIsland = IslandType.GARDEN)
     fun onSecondPassed(event: SecondPassedEvent) {
-
         if (internalContest && startTime.passedSince() > 20.minutes) {
             FarmingContestEvent(contestCrop!!, FarmingContestPhase.STOP).post()
             internalContest = false
         }
 
-        if (!GardenApi.inGarden()) return
-
```
CalMWolfs: you cant do this change as then the farming contest end event will only be sent upon joining the garden again instead of after 20 minutes. This should be suppressed instead

CalMWolfs (changes requested):
just this 1 issue

CalMWolfs (approved):

PR #3314, opened 2025-01-26
Fix: non mutable lists in config

NopoTheGamer:
## What
2 things were listof

<details>
<summary>Images</summary>

image

</details>

## Changelog Fixes
+ Fixed crash with powder mining and chat formatting options. - nopo


CalMWolfs:
#3285

PR #3315, opened 2025-01-26
Backend: StoargeNeedsExpose Ignore `private` Properties

DavidArthurCole:
## What
There are valid use cases to have a private not exposed property in a config class, so this is necessary now.

exclude_from_changelog


DavidArthurCole:
Nevermind lmao. Because `detekt` doesn't run with typing analysis, it can't differentiate between private and public in a visitor, so this works for detektMain, but not detekt. Shame. `@Supress` it is.

PR #3317, opened 2025-01-26
Backend: Cleanup Command Registration

j10a1n15:
## What
Made commandsList private -> Passing the commands list over the event class
Moved CommandRegistration post into new CommandRegistry file
and other small misc things

+ Cleaned up Command Registration. - j10a1n15, ThatGravyBoat

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/commands/CommandRegistrationEvent.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/commands/CommandRegistrationEvent.kt
@@ -1,16 +1,9 @@
 package at.hannibal2.skyhanni.config.commands
 
 import at.hannibal2.skyhanni.api.event.SkyHanniEvent
-import at.hannibal2.skyhanni.config.commands.Commands.commandList
+import at.hannibal2.skyhanni.config.commands.CommandsRegistry.commandList
 import net.minecraftforge.client.ClientCommandHandler
 
 object CommandRegistrationEvent : SkyHanniEvent() {
-    fun register(name: String, block: CommandBuilder.() -> Unit) {
-        val info = CommandBuilder(name).apply(block)
-        if (commandList.any { it.name == name }) {
-            error("The command '$name is already registered!'")
-        }
-        ClientCommandHandler.instance.registerCommand(info.toSimpleCommand())
-        commandList.add(info)
-    }
+    fun register(name: String, block: CommandBuilder.() -> Unit) = CommandsRegistry.registry(name, block)
 }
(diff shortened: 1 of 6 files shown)


Thunderblade73:
I don't understand that change. It just adds a extra layer of indirection for no apparent reason. Also it parses a global value via the an argument.
The only benefit I would see is this:
```Kotlin
    @HandleEvent
    fun onPreInitFinished(event: PreInitFinishedEvent) {
        CommandRegistrationEvent.post()
    }
```

So would want to have some explanation for this change (since I need to port the complex command (https://github.com/hannibal002/SkyHanni/pull/2621) system again with that change).

j10a1n15:
We have changed it to not be as conflicting with your pr as before.
It might also be a good idea to move your complex command system outside your already 1.7k lines long pr.

src/main/java/at/hannibal2/skyhanni/config/commands/CommandRegistrationEvent.kt
```diff
@@ -1,16 +1,21 @@
 package at.hannibal2.skyhanni.config.commands
 
 import at.hannibal2.skyhanni.api.event.SkyHanniEvent
-import at.hannibal2.skyhanni.config.commands.Commands.commandList
 import net.minecraftforge.client.ClientCommandHandler
 
-object CommandRegistrationEvent : SkyHanniEvent() {
+class CommandRegistrationEvent(private val builders: MutableList<CommandBuilder>) : SkyHanniEvent() {
+
+    val commands: List<CommandBuilder> get() = builders
```
Thunderblade73, line 8: Why not have the get in the Command Registry? Since the current way needs this to be an class instead of an object,

Thunderblade73 (commented):
> We have changed it to not be as conflicting with your pr as before.
It might also be a good idea to move your complex command system outside your already 1.7k lines long pr.

Thanks for the change. But I just wanted to know the reasoning (and I may also read the help command wrong).
And yes, I probably have to move the complex command system out (But I will discuss this with hannibal, when I'm done).

src/main/java/at/hannibal2/skyhanni/config/commands/CommandRegistrationEvent.kt
j10a1n15: We dont want the commands/builder list to have a public set
So we either need to pass the list in the event itself, or move the register method outside like before.

Gravy also said:
> an event shouldnt be where everything is handled
> its so that all actual registration and checks are handled the same spot where they are stored

Thunderblade73: Yeah, I see the problem now.

hannibal002:
> I probably have to move the complex command system out

im always a fan of smaller, encapsulated prs

CalMWolfs (approved):

PR #3320, opened 2025-01-27
Fix: Fake Objective Objectives in CS

j10a1n15:
## What
exluded bc no im not writing a changelog for this stupid thing, how would one even explain this issue to an user?
Fixed fake objectives triggering as unknown lines :DDD

exclude_from_changelog


j10a1n15:
image wrong test string

hannibal002:
how would one even explain this issue to an user?
feel free to write technical detauls changelog

PR #3322, opened 2025-01-28
Fix: Stray Timer

DavidArthurCole:
## What
The resetting of the timer was inaccurate and pretty unpredictable as to which reason would or would not do it correctly.

## Changelog Fixes
+ Fixed stray timer inaccurately resetting on inventory change. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryStrayTimer.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryStrayTimer.kt
@@ -66,7 +66,7 @@ object ChocolateFactoryStrayTimer {
 
     @HandleEvent
     fun onTick(event: SkyHanniTickEvent) {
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryStrayTimer.kt
```diff
@@ -29,26 +30,43 @@ object ChocolateFactoryStrayTimer {
...
-        if (timer > Duration.ZERO) timer = 30.seconds
+        if (!isEnabled()) return
+        // Reset the timer when the inventory is closed prematurely
+        timer = 30.seconds
     }
 
     @HandleEvent
     fun onTick(event: SkyHanniTickEvent) {
-        if (!ChocolateFactoryApi.inChocolateFactory || timer <= Duration.ZERO) return
+        if (!isEnabled() && ChocolateFactoryApi.inChocolateFactory) return
```
NopoTheGamer: i think this new if statement is incorrect but im not entirely sure

PR #3323, opened 2025-01-28
Backend: Add JsonReader `.path` to GSON failure message

DavidArthurCole:
## What
Addresses a todo. One of NEU's dependencies was hard-locking us to gson 2.2 something from 2012. Needed to update to get access to `.path`.

+ Added path logging when config options fail to read/initialize. - Daveed


src/main/java/at/hannibal2/skyhanni/utils/json/SkippingTypeAdapterFactory.kt
```diff
@@ -29,8 +29,9 @@ object SkippingTypeAdapterFactory : TypeAdapterFactory {
             return try {
                 parent.read(reader)
             } catch (e: Exception) {
-                // TODO include path and value found (as string)
-                SkyHanniMod.logger.warn("Failed to read value from JSON, skipping", e)
+                val foundValue = reader.peek().toString()
```
NopoTheGamer, line 32: what does peek return if !reader.hasnext()

build.gradle.kts
```diff
@@ -44,6 +44,12 @@ java {
     // IntelliJ run configuration.
     toolchain.vendor.set(JvmVendorSpec.ADOPTIUM)
 }
+// NEU's `error_prone_annotation` override gson to 2.2 which makes us lose some features.
```
NopoTheGamer: also what is error_prone_annotation lol, i cant find it in the neu project anywhere

src/main/java/at/hannibal2/skyhanni/utils/json/SkippingTypeAdapterFactory.kt
DavidArthurCole: image

Nothing as in EOF? Seems like there's a case for that, unless I'm misunderstanding your question.

build.gradle.kts
DavidArthurCole: your guess is as good as mine, i'm not taking the 15 minutes to let the gradle cache rebuild the dependency map 🙏

src/main/java/at/hannibal2/skyhanni/utils/json/SkippingTypeAdapterFactory.kt
NopoTheGamer: Skipping reading JSON from failed path '$.players..profiles..sackContents.', found value of 'END_OBJECT'.

DavidArthurCole:
I do not have the mental capacity to wage war on whatever dependency is still injecting gson 2.2.4 into the veins of the build process

hannibal002:
Can't we just build our own version of gson, change its path (the same way we do with MoulConfig already) and just use this then everywhere?

DavidArthurCole:
> Can't we just build our own version of gson, change its path (the same way we do with MoulConfig already) and just use this then everywhere?

image

perhaps

hannibal002:
the screenshot tells me we dont work with newer gson, or what

DavidArthurCole:
> the screenshot tells me we dont work with newer gson, or what

Oh yeah. I'm realizing there's about 4 hours of context missing from my original closure. Pre-re-location of gson, the game would instantly crash claiming that it couldn't find the method `.getPath()` in gson 2.2.4, which makes sense given that version doesn't have it. But post re-location seems to be working, so.

hannibal002:
as long as we update all those functions in the same pr we also correctly bump gson version, we should be fine then?

DavidArthurCole:
The easiest way I found to test this is to just change one of the objects near the top of the config to an empty string, and you should rightly get the error in console complaining about type (with the path).

DavidArthurCole:
image

NopoTheGamer:
is there a way to make it show the file

DavidArthurCole:
> is there a way to make it show the file

we'd need to make each Config class have a calculated property that defines its position. i can do it if we want, but i don't really see the benefit? apart from dev environment, this doesn't come up too often, and most of us can traverse the config ladder imo

hannibal002:
> we'd need to make each Config class have a calculated property that defines its position. i can do it if we want, but i don't really see the benefit? apart from dev environment, this doesn't come up too often, and most of us can traverse the config ladder imo

cant we throw the error up the stack where we know the config path and only print it then?

DavidArthurCole:
> cant we throw the error up the stack where we know the config path and only print it then?

[we try to](https://github.com/hannibal002/SkyHanni/blob/8e6c5282ab2ab4dd8fb1acde0df2f0a77fca2e29/src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt#L135)

hannibal002 (approved):
```
[01:55:48] [main/WARN] (SkyHanni) Skipping reading JSON from failed path '$.about.autoUpdates', found value of 'NUMBER'.
 java.lang.IllegalStateException: Expected BOOLEAN but was NUMBER at path $.about.autoUpdates
```

PR #3328, opened 2025-01-29
Backend: Small nitpick of #3327

Thunderblade73:
## What
#3327 got merged a few seconds to early. (Or to put it another way: Hannibal distracted me with a random cleanup commit 861a72ac60c73887303d274ade1713484e400134)
exclude_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/ChocolateFactoryApi.kt
@@ -137,7 +137,7 @@ object ChocolateFactoryApi {
             }
             return
         }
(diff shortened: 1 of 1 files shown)


hannibal002:
in the long run i wanted to deprecate the "is in inv" booleans, since they describe more clearly what the actual check is doing imo

Thunderblade73:
> in the long run i wanted to deprecate the "is in inv" booleans, since they describe more clearly what the actual check is doing imo

Then why do it both ways? Either use the value or always call mainInventory.isInside() both ways is not the right call.

hannibal002:
i didnt want to change all files that touch this, since there is more clenaup to do overall (only one read access per file), and the scope of the pr was just to impl this, not repair everthing around it. so a deprecated annotation would have been better. but i forgot abut that

PR #3329, opened 2025-01-29
Fix: Experimentation Table profit tracker gray

hannibal002:
## What
Fixed Experimentation Table profit tracker being gray while in superpair inventory. the same fix as of in https://github.com/hannibal002/SkyHanni/pull/3327

## Changelog Fixes
+ Fixed Experimentation Table profit tracker being gray while in superpair inventory. - hannibal2


Thunderblade73 (approved):
Changes look fine (under the assumption to keep changes to a minimum).
Only thing I don't like is that we need to process the regex twice for the exact same match but in to different functions, but there is no easy way to do that (without a bit of restructuring).

PR #3330, opened 2025-01-29
Feature: Option to disable Skyhanni chat prefix

Chissl:
## What
Add an option to config to remove the [Skyhanni] chat prefix, as I find it annoying on some messages. Option defaulted to off, has a large red warning, and is buried at the bottom of a config menu, so people probably won't unintentionally enable it. 

## Changelog New Features
+ Added option to disable the [Skyhanni] chat prefix. - Chissl

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/chat/ChatConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/chat/ChatConfig.kt
@@ -174,6 +174,5 @@ class ChatConfig {
             "§cWarning: It's not recommended to enable this unless you know what you're doing!"
     )
     @ConfigEditorBoolean
-    @FeatureToggle
     var removeChatPrefix: Boolean = false
 }
--- a/src/main/java/at/hannibal2/skyhanni/config/features/chat/ChatConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/chat/ChatConfig.kt
@@ -170,7 +170,7 @@ class ChatConfig {
     @Expose
     @ConfigOption(
         name = "Remove [Skyhanni] Chat Prefix",
-        desc = "Removes the [Skyhanni] prefix before messages sent from the mod."
+        desc = "Remove the [Skyhanni] prefix before messages sent from the mod."
     )
     @ConfigEditorBoolean
     @FeatureToggle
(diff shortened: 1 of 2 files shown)


NopoTheGamer:
couldnt you just use shwords

src/main/java/at/hannibal2/skyhanni/config/features/chat/ChatConfig.kt
```diff
@@ -166,4 +166,14 @@ class ChatConfig {
...
     var shortenCoinAmounts: Boolean = false
+
+    @Expose
+    @ConfigOption(
+        name = "Remove [Skyhanni] Chat Prefix",
+        desc = "Remove the [Skyhanni] prefix before messages sent by the mod.\n" +
+            "§cWarning: It's not recommended to enable this unless you know what you're doing!"
+    )
+    @ConfigEditorBoolean
+    @FeatureToggle
```
CalMWolfs: should either not have a @FeatureToggle or the option should be inverted: "Show Skyhanni Chat Prefix"

Chissl:
> couldnt you just use shwords

Causes somewhat weird formatting in multi-line messages so this was easier

CalMWolfs (changes requested):
idk how i feel about this pr but if it were to be accepted this would need to be changed

src/main/java/at/hannibal2/skyhanni/utils/ChatUtils.kt
```diff
@@ -27,14 +28,15 @@ import kotlin.time.times
...
+    private val config get() = SkyHanniMod.feature.chat
 
     // TODO log based on chat category (error, warning, debug, user error, normal)
     private val log = LorenzLogger("chat/mod_sent")
     var lastButtonClicked = 0L
 
     private const val DEBUG_PREFIX = "[SkyHanni Debug] §7"
     private const val USER_ERROR_PREFIX = "§c[SkyHanni] "
-    private const val CHAT_PREFIX = "[SkyHanni] "
+    private val CHAT_PREFIX = if (config.removeChatPrefix) "" else "[SkyHanni] "
```
j10a1n15, line 39:
```suggestion
    private val CHAT_PREFIX get() = if (config.removeChatPrefix) "" else "[SkyHanni] "
```

j10a1n15 (changes requested):

src/main/java/at/hannibal2/skyhanni/config/features/chat/ChatConfig.kt
hannibal002: I'd say both

hannibal002:
I'm generally against this. Even when only few users find this message, the ones that do might get confused where a single chat message is coming from. That's the whole point behind the prefix. 
The proper solution is to fix the/shwords bug in chat. Wasn't this fixed already?

lunaynx:
I guess it might be possible to add a "Sent by SkyHanni" to the tooltip or something. But yeah, fixing `/shwords` is probably a better solution. Did #3110 not fix it?

PR #3331, opened 2025-01-29
Improvement: Filter Legacy Items Warning

Chissl:
## What
Add a filter for the legacy item warning to the "others" toggle

## Changelog Improvements
+ Added legacy item warning to the "Others" chat filter. - Chissl


CalMWolfs (changes requested):
should be a repo pattern

hannibal002:
Please use a different group, dont add to "other"

DavidArthurCole (approved):

hannibal002 (changes requested):
After thinking a bit about what this PR does, we should probably not hide this message. Since no one else noticed this, I assume It's not as big of an issue, still not convinced we should merge it without a brief discussion about this.

hypixel adds this message so you get the new items back they "stole" from you with the item update. when we hide this messge (or a user enables all features) then they dont notice this, and might dont know where the items are.

Chissl:
I understand the sentiment, but it's been nearly half a year at this point, pretty much any player that has wanted to convert their items over has. Due to hypixel's extremely bad conversion system, some players have decided to not waste a day converting hundreds of millions of coins worth of enchanted hay bales or would prefer to keep them for sentimental reasons. This message is spammed every time a player swaps worlds, which is pretty annoying, and I think there should be an option to hide it. 

Also in regards to players accidentally enabling the feature then not knowing about their legacy items, they will still occasionally get the message because of the way the chat filter system works. 

I think there would be grounds to remove this feature if hypixel adds more legacy items, but for now, I think this feature would do more good than harm.

PR #3338, opened 2025-01-30
Improvement: SackDisplay Searchables

j10a1n15:
## What
Remove deprecation from buildSearchTable (we really need it or else it wont be a table), it cant use the Searchable class unless we use List<Renderables> instead of just Renderable.
We do now have sack specific searchables though!

## Changelog Improvements
+ Made searches in Sack display GUIs sack-specific. - j10a1n15


hannibal002 (changes requested):
You confused horizontal and vertical
image

j10a1n15:
image
oh

j10a1n15:
yeah so I dont think we can remove the deprecated method

PR #3346, opened 2025-01-31
Backend: Auto Config Bump

Thunderblade73:
ignore_from_changelog

the pr got reverted. see:
- #3593
- #3594

## What
We now have a "CONFIG_MOVE_VERSION" which will automatically bump the config version up by one.
On merge with it will be substituted with the version and the config version will be incremented.

In dev enviroment it always performs a config fix for ConfigVersion + 1 to process newly moves.

(Also need some feedback where the "config_version" file should go.)

<details>
<summary>Images</summary>

Sample commit:
image

</details>

+ Added automatic config bumps. - Thunderblade73
    * Use "CONFIG_MOVE_VERSION" to get a version for newly defined config moves.


CalMWolfs (approved):
code looks fine to me and i think the file location is fine

hannibal002 (changes requested):
please include a short text about this in contributing md

PR #3345, opened 2025-01-31
Improvement: Gui Button hover

hannibal002:
## What
Gui Buttons show now all variants on hover.

<details>
<summary>Images</summary>

image

</details>

## Changelog Improvements
+ Updated UI buttons to display all variants on hover, with scroll and back options. - hannibal2 + Daveed

## Changelog Technical Details
+ Merged addButton and addSelector into addButton for hoverable items in Renderable lists. - hannibal2

PR #3349, opened 2025-02-01
Feature + Fix: GUI Button Scroll + Null Fix

DavidArthurCole:
## What
Fix for a couple issues that exist in #3345.
Also adds the ability to go through options using the scroll wheel.
@hannibal002 this was easier than a patch file :D

exclude_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt
@@ -1,9 +1,11 @@
 package at.hannibal2.skyhanni.utils.renderables
 
+import at.hannibal2.skyhanni.utils.SimpleTimeMark
 import org.lwjgl.input.Mouse
+import kotlin.time.Duration.Companion.milliseconds
 
 abstract class ScrollInput(
-    protected val scrollValue: ScrollValue,
+    private val scrollValue: ScrollValue,
     protected val minValue: Int,
     protected val maxValue: Int,
     protected val velocity: Double,
@@ -33,6 +35,8 @@ abstract class ScrollInput(
         }
 
     protected fun isMouseEventValid(): Boolean = scrollValue.isMouseEventValid()
+    protected fun isPureScrollEvent() = scrollValue.isPureScrollEvent()
+    protected fun hasRecentScrollEvent() = scrollValue.hasRecentScrollEvent()
 
     abstract fun update(isValid: Boolean)
 
@@ -69,9 +73,9 @@ abstract class ScrollInput(
                 // For pure events, we don't care about tracking state
                 // and only care about tracking a 1/-1 for the scroll status.
                 // We reset to 0 to avoid repeatedly applying the same scroll value.
-                val isBad = Mouse.getEventDWheel() == 0 || !scrollValue.isPureScrollEvent() || !scrollValue.hasRecentScrollEvent()
-                if (isBad) return dispose()
+                if (!isPureScrollEvent() || !hasRecentScrollEvent()) return dispose()
 
+                // Otherwise we let the parent class handle the rest.
                 super.update(isValid)
             }
 
@@ -88,7 +92,7 @@ class ScrollValue {
     private var mouseEventTime = 0L
     private var lastMouseX = 0
     private var lastMouseY = 0
-    private var lastScrollTime = 0L
+    private var lastScrollTime: SimpleTimeMark = SimpleTimeMark.farPast()
 
     fun getValue(): Double =
         field ?: throw IllegalStateException("ScrollValue should be initialized before get.")
@@ -112,23 +116,20 @@ class ScrollValue {
     fun isPureScrollEvent(): Boolean {
         val mouseX = Mouse.getEventX()
         val mouseY = Mouse.getEventY()
-        val deltaWheel = Mouse.getEventDWheel()
-
-        val isScrollEvent = deltaWheel != 0
+        val isScrollEvent = Mouse.getEventDWheel() != 0
         val hasMouseMoved = mouseX != lastMouseX || mouseY != lastMouseY
 
         // Update last mouse position
         lastMouseX = mouseX
         lastMouseY = mouseY
 
-        if (isScrollEvent) {
-            lastScrollTime = System.currentTimeMillis() // Store last valid scroll time
-        }
+        // Update last scroll time if it's a scroll event
+        if (isScrollEvent) lastScrollTime = SimpleTimeMark.now()
 
         return isScrollEvent && !hasMouseMoved
     }
 
     fun hasRecentScrollEvent(): Boolean {
-        return System.currentTimeMillis() - lastScrollTime < 50L
+        return lastScrollTime.passedSince() < 50.milliseconds
     }
 }
--- a/src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt
@@ -1,8 +1,6 @@
 package at.hannibal2.skyhanni.utils.renderables
 
-import at.hannibal2.skyhanni.utils.SimpleTimeMark
 import org.lwjgl.input.Mouse
-import kotlin.time.Duration.Companion.milliseconds
 
 abstract class ScrollInput(
     private val scrollValue: ScrollValue,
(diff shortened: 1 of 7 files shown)


DavidArthurCole:
Although I'm 99% sure that the edits I made are non-breaking, and I did some basic verification testing, I did have to edit functionality of `ScrollInput` to get this to work, so we should make sure that other parts of the base that are utilizing it still function as expected.

Thunderblade73:
> Although I'm 99% sure that the edits I made are non-breaking, and I did some basic verification testing, I did have to edit functionality of `ScrollInput` to get this to work, so we should make sure that other parts of the base that are utilizing it still function as expected.

You did no edit here. You simply added new behavior. So there can't be a regression error.

DavidArthurCole:
> You did no edit here. You simply added new behavior. So there can't be a regression error.

Thanks - been looking at too much Kotlin today...

src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt
```diff
@@ -56,15 +56,39 @@ abstract class ScrollInput(
...
 }
 
 class ScrollValue {
-    var field: Double? = null
-
+    private var field: Double? = null
     private var mouseEventTime = 0L
+    private var lastMouseX = 0
+    private var lastMouseY = 0
+    private var lastScrollTime = 0L
```
Thunderblade73: Use SimpleTimeMark here.

src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt
```diff
@@ -85,4 +109,26 @@ class ScrollValue {
...
+
+        val isScrollEvent = deltaWheel != 0
+        val hasMouseMoved = mouseX != lastMouseX || mouseY != lastMouseY
+
+        // Update last mouse position
+        lastMouseX = mouseX
+        lastMouseY = mouseY
+
+        if (isScrollEvent) {
+            lastScrollTime = System.currentTimeMillis() // Store last valid scroll time
```
Thunderblade73: Isn't that the same as ```isMouseEventValid()``` ?

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt
```diff
@@ -56,15 +56,39 @@ abstract class ScrollInput(
...
+         * Should not be used to hold actual data, but rather as a tracker for pure scroll events.
+         */
+        class PureVertical(
+            scrollValue: ScrollValue = ScrollValue(),
+        ) : Vertical(scrollValue, -1, 1, 1.0, null, 0.0) {
+            override fun update(isValid: Boolean) {
+                // For pure events, we don't care about tracking state
+                // and only care about tracking a 1/-1 for the scroll status.
+                // We reset to 0 to avoid repeatedly applying the same scroll value.
+                val isBad = Mouse.getEventDWheel() == 0 || !scrollValue.isPureScrollEvent() || !scrollValue.hasRecentScrollEvent()
```
Thunderblade73: ```Mouse.getEventDWheel() == 0``` is already checked inside ```scrollValue.isPureScrollEvent()```.
```hasRecentScrollEvent()``` can be packeged into ```isPureScrollEvent`` than you could add a wrapper function insided ScrollInput and keep scrollValue private (since it's scope should be limited to the ScrollInput class to have all ways of mutation at one place.)

**Review by Thunderblade73 — CHANGES_REQUESTED** [review]

File: src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt (hunk 2 above)
Comment (DavidArthurCole) [reply src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt]: Not sure what you mean here, but let me know if it's still an issue on the changed code

Comment (Thunderblade73) [reply src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt]: You check for the time change, which is the same as isMouseEventValid

Comment (Thunderblade73) [reply src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt]: It is not directly here, but it's the usage of lastScrollTime value.

Comment (Thunderblade73) [reply src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt]: Sure there is a difference between a 20ns and 50ms but the effect should be similar.

Comment (DavidArthurCole) [reply src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt]: They're set under different conditions - mouse events get fired for everything including moving the mouse, the timestamp I need is specifically the last scroll, which is why we check against `Mouse.getEventDWheel()` before setting `lastScrollTime`

Comment (Thunderblade73) [reply src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt]: But it doesn't matter in your code, since when the event is not an Dwheel only then it returns false and your closure resets the scroll value. So the only diff here is the timeframe where events are accepted.

Comment (DavidArthurCole) [reply src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt]: I'm following, but I don't think the code paths completely line up there. If you can get a functional version without the use of that mark, I'm always willing to merge working code, but the removal of that and reliance on `isMouseEventValid()` completely bricked the ability to scroll when I tried just now.

Comment (Thunderblade73) [reply src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt]: Calling isMouseEventValid() is always false in the second instance that's why it always failed. I did not account for the super call. 
But the simple solution is to just remove the check as a whole. To be technically correct we should still always call dispose, since it should be only valid for one call. Btw the object get created repeatedly which resets the scollValue. But not here to fix that.
Patch:
```diff
Index: src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt b/src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt
--- a/src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt	(revision 83207fe4ab034bb9ad376758c2f29dfb16de83bb)
+++ b/src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt	(date 1738374929514)
@@ -1,8 +1,6 @@
 package at.hannibal2.skyhanni.utils.renderables
 
-import at.hannibal2.skyhanni.utils.SimpleTimeMark
 import org.lwjgl.input.Mouse
-import kotlin.time.Duration.Companion.milliseconds
 
 abstract class ScrollInput(
     private val scrollValue: ScrollValue,
@@ -36,7 +34,6 @@
 
     protected fun isMouseEventValid(): Boolean = scrollValue.isMouseEventValid()
     protected fun isPureScrollEvent() = scrollValue.isPureScrollEvent()
-    protected fun hasRecentScrollEvent() = scrollValue.hasRecentScrollEvent()
 
     abstract fun update(isValid: Boolean)
 
@@ -73,7 +70,8 @@
                 // For pure events, we don't care about tracking state
                 // and only care about tracking a 1/-1 for the scroll status.
                 // We reset to 0 to avoid repeatedly applying the same scroll value.
-                if (!isPureScrollEvent() || !hasRecentScrollEvent()) return dispose()
+                dispose()
+                if (!isPureScrollEvent()) return
 
                 // Otherwise we let the parent class handle the rest.
                 super.update(isValid)
@@ -92,7 +90,6 @@
     private var mouseEventTime = 0L
     private var lastMouseX = 0
     private var lastMouseY = 0
-    private var lastScrollTime: SimpleTimeMark = SimpleTimeMark.farPast()
 
     fun getValue(): Double =
         field ?: throw IllegalStateException("ScrollValue should be initialized before get.")
@@ -123,13 +120,6 @@
         lastMouseX = mouseX
         lastMouseY = mouseY
 
-        // Update last scroll time if it's a scroll event
-        if (isScrollEvent) lastScrollTime = SimpleTimeMark.now()
-
         return isScrollEvent && !hasMouseMoved
     }
-
-    fun hasRecentScrollEvent(): Boolean {
-        return lastScrollTime.passedSince() < 50.milliseconds
-    }
 }
```

PR #3361, opened 2025-02-02
Fix: Neu PV not loading

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1335606268219363468

## Changelog Fixes
+ Fixed NEU's `/pv` command not functioning for your profile. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/enoughupdates/NeuEventWrappers.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/enoughupdates/NeuEventWrappers.kt
@@ -58,6 +58,7 @@ object NeuEventWrappers {
 
     @SubscribeEvent
     fun onProfileDataLoaded(event: ProfileDataLoadedEvent) {
+        // Because of varying Gson dependencies, we can't directly use .data from NEU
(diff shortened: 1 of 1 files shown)


lunaynx:
Should probably add a comment to the code explaining why this hack is required.

src/main/java/at/hannibal2/skyhanni/api/enoughupdates/NeuEventWrappers.kt
```diff
@@ -57,7 +58,8 @@ object NeuEventWrappers {
 
     @SubscribeEvent
     fun onProfileDataLoaded(event: ProfileDataLoadedEvent) {
-        val apiData = event.data ?: return
+        // Because of varying Gson dependencies, we can't directly use .data from NEU
+        val apiData = event::class.java.getDeclaredField("data").get(event) as? JsonObject ?: return
```
lineargraph, line 62: This is entirely pointless, no? That entire method can be gotten rid off, since the JsonObject is still relocated. Java does not have structural typing, so the cast here will always fail. I really don't see the major benefit of more modern GSON either.

PR #3362, opened 2025-02-02
Feature + Fix: 400Anniversary Daily Task HUD

Thunderblade73:
## What
I made a simple hud for the daily task.
But as always I made it unnecessarily complicated. Since added an bottom layer to quickly make such huds. But in the current version it doesn't really do a whole lot and just adds an layer of abstraction, maybe it could be expanded on later on. Why I thought it would be a good idea to add that layer of abstraction is that we literally have a similar feature the carnival goals, which could use this layer.

Since this feature has a max time to live of ~3 days I won't do a whole lot to make it better. But I would need the call if this is even going to be merged, if not I will separated the 3 starting commits out into 2 Prs and throw away the feature part.

I also fixed weird behavior of the scrolling list that where introduced a while ago.

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added an HUD for the Daily Tasks for the 400 year Anniversary. - Thunderblade73

## Changelog Fixes
+ Fixed inconsistent scrolling behavior in some sh guis. - Thunderblade73


Thunderblade73:
I did a full test for both chat and inventory trigger and it worked. 
Excluding the following (since I did not complete them):
image
Only thing that might be good to test is if the enter rift task works via chat since this one didn't work with my run of the tasks but I did put a fix in place that should make it work (either way the inventory detection for it worked so not to big of a deal).

src/main/java/at/hannibal2/skyhanni/features/event/anniversary/Year400DailyHUD.kt
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        if (current != last) {
+            reset()
+        }
+
+    }
+
+    // Reset Time is 0:00 EST
+    private fun getDate(): LocalDate = ZonedDateTime.now().toLocalDate()
```
NopoTheGamer: im not sure how using the users timezone could be correct

src/main/java/at/hannibal2/skyhanni/utils/TimeUtils.kt
```diff
@@ -178,6 +179,8 @@ object TimeUtils {
     val Int.ticks get() = (this * 50).milliseconds
 
     val Float.minutes get() = toDouble().minutes
+
+    val ZoneOffset_MDT: ZoneOffset = ZoneOffset.of("-6")
```
NopoTheGamer: is this used?

Thunderblade73: No, since Hypixel changed the reset time.yesterday.

src/main/java/at/hannibal2/skyhanni/features/event/anniversary/Year400DailyHUD.kt
Thunderblade73: Thx, there should be UTC as argument of now.

hannibal002:
please extract the fix part of this pr into a new pr

PR #3363, opened 2025-02-02
Improvement: Better onomatopoeia recognition and new sounds for Sound Responses

owocato:
## What

Made the regex used to recognize sounds less rigid and catch more onomatopoeia, and added Cat Purr, Purreow, Hiss, and Dog Growl and Howl sound responses.

Also added display names to the enum constants and auto-enabled the additional noises if the user already has the existing ones enabled.

## Changelog Improvements
+ Added new sounds to Sound Responses. - cato
+ Improved Sound Responses onomatopoeia recognition. - cato


hannibal002:
What is a onomatopoeia

fanibus1:
meow

Fevix:
> What is a onomatopoeia

It's a sound used to describe a noise, like how the noise of one of those clocks with a second hand is called a "tick" and the sound a big bell makes is a "bong", or how a keychain with lots of keys "jingles" when you shake it, or how a happy cat makes a "purr"

CalMWolfs (approved):
better late than ever right?

fanibus1:
arf

PR #3364, opened 2025-02-03
Fix: Crouching or sleeping sometimes made lines render strangely.

Nessiesson:
## What
Call getEyeHeight() instead of directly reading the eyeHeight field. And no it is not just a "simple" getter as it does more stuff like correct your eye height when crouching or sleeping.

Also: dunno if the 'unrelated' change at line 1040 is appropriate but that was the only other instance of `eyeHeight` being read directly that I could find. 🤔

<details>
<summary>Images</summary>

No crouch:

Crouch:

</details>

## Changelog Fixes
+ Fixed occasional offsets in in-world text render positions. - Nessiesson
+ Fixed minor error in line drawing for Golden Goblin Highlight while sneaking. - Nessiesson


hannibal002:
I remembered there were bugs with Patcher enabled that broke eyes position, and we had a custom fix for this.
The proper solution here is to not to use draw3DLineFromPlayer() at all.

PR #3366, opened 2025-02-04
Backend: Dispatcher.Minecraft for Coroutines

cyoung06:
## What
Properly Implement Dispatchers.Minecraft and Dispatchers.Minecraft.immediate (MainCoroutineDispatcher) with delay support.

This implementation is based on implementation for [kotlin-coroutines-swing](https://github.com/Kotlin/kotlinx.coroutines/blob/master/ui/kotlinx-coroutines-swing/src/SwingDispatcher.kt)

This enables things like
```kotlin
suspend fun doAthenwaitanddoB() {
  withContext(Dispatchers.Minecraft.immediate) {
    doStuffThatAbsolutelyNeedsToRunOnMinecraftThread()
    delay(1000)
    doOtherStuffThatAbsolutelyNeedsToRunOnMinecraftThread()
  }
}
```
without blocking main thread

+ Added Dispatchers.Minecraft and Dispatchers.Minecraft.immediate for coroutines.  - syeyoung

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/MinecraftDispatcher.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/MinecraftDispatcher.kt
@@ -4,16 +4,76 @@ import kotlinx.coroutines.MainCoroutineDispatcher
 import net.minecraft.client.Minecraft
 import kotlin.coroutines.CoroutineContext
 
-object MinecraftDispatcher : MainCoroutineDispatcher() {
+import kotlinx.coroutines.*
+import kotlinx.coroutines.Runnable
+import kotlinx.coroutines.internal.*
+import kotlin.time.Duration.Companion.milliseconds
 
-    override val immediate: MainCoroutineDispatcher
-        get() = this
+@Suppress("unused")
+public val Dispatchers.Minecraft : MinecraftDispatcher
+    get() = at.hannibal2.skyhanni.utils.Minecraft
 
-    override fun isDispatchNeeded(context: CoroutineContext): Boolean {
-        return !Minecraft.getMinecraft().isCallingFromMinecraftThread
-    }
-
+@OptIn(InternalCoroutinesApi::class)
+public sealed class MinecraftDispatcher : MainCoroutineDispatcher(), Delay {
     override fun dispatch(context: CoroutineContext, block: Runnable) {
         Minecraft.getMinecraft().addScheduledTask(block)
     }
+
+    override fun scheduleResumeAfterDelay(timeMillis: Long, continuation: CancellableContinuation<Unit>) {
+        val cancellableRun = schedule(timeMillis) {
+            with(continuation) { resumeUndispatched(Unit) }
+        }
+        continuation.invokeOnCancellation { cancellableRun.cancel() }
+    }
+
+    override fun invokeOnTimeout(timeMillis: Long, block: Runnable, context: CoroutineContext): DisposableHandle {
+        val cancellableRun = schedule(timeMillis) {
+            block.run()
+        }
+        return DisposableHandle { cancellableRun.cancel() }
+    }
+
+    private fun schedule(timeMillis: Long, action: () -> Unit): CancellableRun {
+        val run = CancellableRun(action);
+        DelayedRun.runDelayed(timeMillis.milliseconds, run);
+        return run;
+    }
+
+    class CancellableRun(val run: () -> Unit, @Transient var isCancelled: Boolean = false): () -> Unit {
+        override fun invoke() {
+            if (isCancelled) return
+            this.run();
+        }
+
+        fun cancel() {
+            isCancelled = true
+        }
+    }
 }
+
+@OptIn(InternalCoroutinesApi::class)
+internal class MinecraftDispatcherFactory : MainDispatcherFactory {
+    override val loadPriority: Int
+        get() = 0
+
+    override fun createDispatcher(allFactories: List<MainDispatcherFactory>): MainCoroutineDispatcher = Minecraft
+}
+
+private object ImmediateMinecraftDispatcher : MinecraftDispatcher() {
+    override val immediate: MainCoroutineDispatcher
+        get() = this
+
+    override fun isDispatchNeeded(context: CoroutineContext): Boolean = !Minecraft.getMinecraft().isCallingFromMinecraftThread
+
+    @OptIn(InternalCoroutinesApi::class)
+    override fun toString() = toStringInternalImpl() ?: "Minecraft.immediate"
+}
+
+internal object Minecraft : MinecraftDispatcher() {
+    override val immediate: MainCoroutineDispatcher
+        get() = ImmediateMinecraftDispatcher
+
+    @OptIn(InternalCoroutinesApi::class)
+    override fun toString() = toStringInternalImpl() ?: "Minecraft"
+}
+
--- a/src/main/java/at/hannibal2/skyhanni/utils/MinecraftDispatcher.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/MinecraftDispatcher.kt
@@ -1,16 +1,19 @@
 package at.hannibal2.skyhanni.utils
 
+import kotlinx.coroutines.CancellableContinuation
+import kotlinx.coroutines.Delay
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.DisposableHandle
+import kotlinx.coroutines.InternalCoroutinesApi
 import kotlinx.coroutines.MainCoroutineDispatcher
+import kotlinx.coroutines.Runnable
+import kotlinx.coroutines.internal.MainDispatcherFactory
 import net.minecraft.client.Minecraft
 import kotlin.coroutines.CoroutineContext
-
-import kotlinx.coroutines.*
-import kotlinx.coroutines.Runnable
-import kotlinx.coroutines.internal.*
 import kotlin.time.Duration.Companion.milliseconds
 
 @Suppress("unused")
-public val Dispatchers.Minecraft : MinecraftDispatcher
+public val Dispatchers.Minecraft: MinecraftDispatcher
     get() = at.hannibal2.skyhanni.utils.Minecraft
 
 @OptIn(InternalCoroutinesApi::class)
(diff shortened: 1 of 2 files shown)


hannibal002:
This looks interesting.
Not sure I see the big of a need for us as we already have run delayed or tick events.
E.g. this example could be done via
RunNextTick doStuffThatAbsolutelyNeedsToRunOnMinecraftThread
RunDelayed 1. Second doOtherStuffThatAbsolutelyNeedsToRunOnMinecraftThread

But since this is proper coroutine and kotlin gets praised for it, maybe we should switch to it.

Unrelated to this: please remove the wild card inport line and instead import every single class separately

hannibal002:
Just noticed: a big turn off for me here is the need for the function to be written as suspend. Our events don't support it, so idk where this should get used if ever?

cyoung06:
I thought about implementing suspendible event listeners, but quickly realized that it has problems with cancelling events.

You still need to launch coroutines manually in event listeners in order to use coroutines.

One way coroutines could be very helpful is in conjunction with something like
`suspend fun <T:SkyHanniEvent, R> awaitEvent(clazz: Class<T>, eventHandler: (T) -> R): R`

where it could be used like
```kotlin
suspend fun awaitChatMessageAndCancel(pattern: Pattern, blockreason: String): Matcher = withTimeout<Matcher>(10_000) {
        while(true) {
            val result: Matcher? = awaitEvent(SkyHanniChatEvent::class.java) {
                val matcher = pattern.matcher(it.message)
                if (matcher.matches()) {
                    it.blockedReason = blockreason
                    return@awaitEvent matcher;
                }
                return@awaitEvent null;
            }
            if (result != null) return@withTimeout result;
        }
        error("")
    }

suspend fun constructStateFromChatMessage(): State {
    val res1 = awaitChatMessageAndCancel(pattern1, "~").group("a")
    val res2 = awaitChatMessageAndCancel(pattern2, "~").group("b")
    val res3 = awaitChatMessageAndCancel(pattern3, "~").group("c")
    val res4 = awaitChatMessageAndCancel(pattern4, "~").group("d")
    return State(res1, res2, res3, res4)
}
```

or

```kotlin
suspend fun addClickedBlockIntoList() {
  sendchat("please click block you want to add")
  val vec3 = try {
    withTimeout(10_000) {
      awaitEvent(BlockLeftClickEvent::class.java) { it.coords }
    }
  } catch(TimeoutCancellationException: e) {
    sendchat("operation cancelled")
    throw e;
  }
  sendchat("Successfully added "+vec3+" into the list")
}
```

But nea was against it on #coding , as "it advocates for more convoluted control flow" 


or other way it could be used is
```kotlin
suspend fun dungeonSolverButItDoesLongSearch() = withContext(Dispatchers.Default) {
        val blocks = async(Dispatchers.Minecraft) {
            val world = Minecraft.getMinecraft().theWorld
            loadBlocks(world)
        }.await()

        findPath(blocks)
    }

fun onTick(event: SkyHanniTickEvent) {
    if (solver != null) return;
    solver = CoroutineScope(CoroutineName("Calculator")).launch {
        dungeonSolverButItDoesLongSearch()
    }
}

fun onRenderWorld(event: ~) {
    if (solver != null && solver.isComplete) renderSolution(solver)
}
```

Basically anywhere you need to jump between dedicated thread and main thread, it could be used
Or where you need a simple state machine, coroutines could be used.

hannibal002:
I don't see those examples as useful, sadly. Only reason for us to do something outside main thread is if the event itself is coming from async already (block change, sound, particle, anything else that directly comes from a packet event), or Io/blocking tasks.

For the events there is no need to use this async wrapper as they are already async themselves, and we can less complicated run logic on main thread from those by using run delayed class and shift the logic to the next tick event.

For Io/blocking functions we already have dedicated async behaviour with consumers or just run the function in suspend/with context functions. I also don't see a difference here.

While this coroutine stuff looks cool, I only see how it over complicates the question what is sync and what is async. (To main thread, I'm not familiar with the term dedicated here)

Especially as we ideally want everything in main thread and nothing async, except Io stuff..

But please proof me wrong

hannibal002:
> I thought about implementing suspendible event listeners, but quickly realized that it has problems with cancelling events.

Maybe partially outside of the scope of the pr, but I think a proper solution to async events would be to duplicate them into async/sync variants of the event, with the sync one always gets moved to next tick event, but no option to cancel.

cyoung06:
Yes, this pr only adds Dispatchers.Main and nothing async event related.
Anyways implementing async events would require dynamically registering events at runtime which SkyHanniEvents.kt does not allow currently.
If it would get implemented it would be another pr

I'm looking for good examples in which coroutines would make code easier to read in sh features

hannibal002:
> Anyways implementing async events would require dynamically registering events at runtime which SkyHanniEvents.kt does not allow currently.

I don't understand

cyoung06:
in SkyHanniEvents.kt, EventHandler is built with snapshot of List<EventListeners.Listener>, and sorts it according to priority
When EventListeners.addListener() gets called EventHandler is not aware of such addition of event listener

`suspend fun <T:SkyHanniEvent, R> awaitEvent(clazz: Class<T>, eventHandler: (T) -> R): R`

would require the function to either
- dynamically register itself to eventlistener, and unregister itself once event is fired
- dynamically register itself to some kind of AsyncEventManager, which in turn registers itself to EventListener. When event fires, `awaitEvent` would remove itself from AsyncEventManager. AsyncEventManager is in charge of getting event then firing events again to continuations.
- know all classes of AsyncEvents beforehand and AsyncEventManager preregisters them when the mod loads
- or it could be implemented with channels but I don't think having two kinds of event that behave differently is good

hannibal002:
I don't know all those fancy terms, so sorry for breaking this down in my words:

you say our current impl of event system cant handle async? 
this cant be, since we can perfectly fine deal with events getting fired from async enviroments, e.g. packet event.

you say we cant mark functions that register a event cant have the suspend modifier?
idk what more in depth requires the function call to allow adding suspend, but cant we simply check at the spot where we fire the event if the current thread is the main thread, and if not include a withcontext call before fireing the event?
If this doesn work as easily as i think, maybe we can add some bytecode/after compile logic similar to mixins that forcefully marks all async events as suspend?

we are only talking about the suspend modifier since we require this suspend context to run proper sync/async magic via coroutines, right?

Does this mean we would call every coroutine logic in a non-main MC thread? Why would we do this? We want as much as code as possible to run on the main thread because:
ease of mind, no need to think about what is sync/async.
no need to use mutex/concurrent logic as often
all rendering and mc world access functions already require mc main thread
there is nothing heavy running in the mod that would benefit of being lifted off the main thread
if there is a bug that causes something to be way less optimized than it should/could be, if it's not on main thread, its harder to detect via profilers.

so in summary, i dont fully understand the benefits of coroutines. what i understand so far, this doesnt work together well with a minecraft mod in general, and as nea said in coding channel, this adds complexity.
Please give me counter arguments, if there are any. Im happy to learn

cyoung06:
Current SkyHanni event system can of course handle asynchronous events that fire outside of main thread.
There of course is not requirement to call suspend event listeners outside of mc thread
Although allowing event listeners to have suspend modifier have effects such that it would not be possible to cancel events. 
Suspend event listeners should only be used for events that are not cancellable. Events that does not depend on the result of firing.


I'm just saying that in order to implement 
`suspend fun <T:SkyHanniEvent, R> awaitEvent(clazz: Class<T>, eventHandler: (T) -> R): R`

which would

1. Suspends current execution
2. Wait for new event to be received
3. Run eventHandler in sync context (so that event can be cancelled)
4. Resume the continuation with result of eventHandler

But current EventDispatcher is not aware of this new event listener that needs to be added in step 2 , and there isn't really a good way to notify EventHandler that new EventListener has been added

You can't register a new EventListener after EventHandler has formed in SkyHanni currently

hannibal002:
In what context would you use this new awaitEvent function, and why would we want this to be suspend?
the event register logic should be unaffected from async-ness of single listeners, or even functins, dont they?
the event call logic in the event mamanger should only require a simple check wether to run a sync or async context, or not?
the actuall check wether to fire a sync or async variant of the event should be done in the caller site anyway, or not?

i dont see the need for this, maybe im thinking way too primitive here? what am i misssing?

cyoung06:
Here is sample implementation of awaitEvent.

```kotlin

suspend fun <T: SkyHanniEvent, R> awaitEvent(clazz: Class<T>, eventHandler: (T) -> R): R = suspendCancellableCoroutine { continuation ->
    SkyHanniEvents.registerEventListener(CoroutineAwaitedEventListener(continuation, eventHandler))
    // this method does not exist currently, and adding a new eventlistener is not possible in current SkyHanni events architecture.
}
class CoroutineAwaitedEventListener<T: SkyHanniEvent, R>(
    val continuation: CancellableContinuation<R>,
    val eventHandler: (T) -> R
) : Consumer<T> {
    @OptIn(InternalCoroutinesApi::class)
    override fun accept(t: T) {
        val result = eventHandler.invoke(t)
        SkyHanniEvents.unregisterEventListener(this)
        // same for here. Unregistering event listener is not possible in current Skyhanni events arch
        continuation.tryResume(result)
    }
}
```

and bringing back the example I talked about earlier

```kotlin
suspend fun addClickedBlockIntoList() {
  sendchat("please click block you want to add")
  val vec3 = try {
    withTimeout(10_000) {
      awaitEvent(BlockLeftClickEvent::class.java) { it.coords }
    }
  } catch(TimeoutCancellationException: e) {
    sendchat("operation cancelled")
    throw e;
  }
  sendchat("Successfully added "+vec3+" into the list")
}
```
This code when called, waits 10 seconds for user to click a block and does some operation with it, or returns if user didn't click a block within 10 seconds 


Implementing it without coroutines
```kotlin
private var isFunctionActive = false
private var functionValidUntil = 0L

@HandleEvent
fun onTick(event: SkyHanniTickEvent) {
  if (!isFunctionActive) return
  if (functionValidUntil < System.currentTimeMilles()) {
    sendchat("operation cancelled")
    isFunctionActive = false
  }
}

@HandleEvent
fun onBlockClick(event: BlockLeftClickEvent) {
  if (!isFunctionActive) return;
  sendchat("Successfully added ${coords} into the list")
}

fun addClickedBlockIntoList() {
  sendchat("please click block you want to add")
  functionValidUntil = System.currentTimeMillies() + 10000;
  isFunctionActive = true;
}
```
1. Code is split over 3 functions and 2 event listeners
2. There are two state variables that are needed to track if function is still active or not
3. Business logic is all over the place

awaitEvent() makes implementation of such simple state machines easier

CalMWolfs:
closing as this class doesnt exist in skyhanni anymore

PR #3371, opened 2025-02-04
Improvement: better Totem of Corruption expiry sound

MTOnline69:
## What
Changes Totem of Corruption expiry sound so it's not the same as the reel in sound. Was annoying me.
https://discord.com/channels/997079228510117908/1245020820733034569
Also switches SoundUtils.playBeepSound() to take in a pitch argument (default 1f).

## Changelog Improvements
+ Changed Totem of Corruption expiry sound to differ from the reel-in sound. - MTOnline


src/main/java/at/hannibal2/skyhanni/features/fishing/TotemOfCorruption.kt
```diff
@@ -159,7 +160,7 @@ object TotemOfCorruption {
 
             val timeToWarn = config.warnWhenAboutToExpire.seconds
             if (timeToWarn > 0.seconds && timeRemaining <= timeToWarn && totem.uniqueID !in warnedTotems) {
-                playPlingSound()
+                SoundUtils.createSound("random.orb", 0.5f).playSound()
```
j10a1n15: Don't create the sound here, instead create it in soundutils with a similar method as the playPlingSound()

j10a1n15 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/fishing/TotemOfCorruption.kt
MTOnline69: i didn't do that cause playBeepSound already uses random.orb, just at a different pitch.
Am I allowed to switch that function to take in a pitch value?

```
    fun playBeepSound(pitch: Float = 1f) {
        val beepSound by lazy { createSound("random.orb", pitch) }
        beepSound.playSound()
    }
```

j10a1n15: This would create a new sound every single time aswell,
In Theory you could use a cache with getOrPut() if you don't wanna create a second playBeepSound method

j10a1n15 (approved):
lgtm

mariethepolarfox:
Little Suggestion: Add the Option to let the user chose the Sound like the Inquisitor Option but might be out of this prs scope

DavidArthurCole (approved):

PR #3374, opened 2025-02-06
Fix: Motes per Session when leaving Hypixel

hannibal002:
## What
sending island change events when leaving hypixel/skyblock.
Related bug report: https://discord.com/channels/997079228510117908/1336742301162147840

## Changelog Fixes
+ Fixed Motes per Session when leaving Hypixel. - hannibal2

+ Added `HypixelLeaveEvent` and `SkyBlockLeaveEvent`. - hannibal2


ItsEmpa:
im not entirely sure if this should be implemented in the old hypixeldata or if we should use the new one powered by modapi

hannibal002:
> im not entirely sure if this should be implemented in the old hypixeldata or if we should use the new one powered by modapi

How does modapi handle those actions? do we get an packet when moving from sb to hub or to limbo? do we get an packet when leaving skyblock? how consisthen are those?

i still get reproduceable errors when leaving limbo. as long as those errors arent fixed, i dont think we need to discuss implementing this anywhere else.

PR #3375, opened 2025-02-06
Fix: Scrolling List

Thunderblade73:
## What
Fixed weird behavior of the scrolling list that where introduced a while ago.

Extra:
I fixed the Renderable.scrollableList for the use case in #3362 and those fixed apply to any use of the scrollableList that use the showScrollableTipsInList flag which is currently only the one in the Cake tracker.

## Changelog Fixes
+ Fixed inconsistent scrolling in some SH GUIs. - Thunderblade73

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/renderables/Renderable.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/renderables/Renderable.kt
@@ -1055,7 +1055,7 @@ interface Renderable {
             private val scroll = ScrollInput.Companion.Vertical(
                 scrollValue,
                 0,
-                virtualHeight - height + if (showScrollableTipsInList) scrollUpTip.height else 0,
+                virtualHeight - height + if (showScrollableTipsInList && virtualHeight > height) scrollUpTip.height else 0,
                 velocity,
                 button,
             )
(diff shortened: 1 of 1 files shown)


hannibal002 (changes requested):
I can still reproduce the bug in the Chest Value display. In other displays it was not a problem prior.
I assume this happens because the chest value display gets recreated very frequently

Thunderblade73:
> I can still reproduce the bug in the Chest Value display. In other displays it was not a problem prior. I assume this happens because the chest value display gets recreated very frequently

They cake tracker is the only one that uses the showScrollableTipsInList flag as far as I know. And this manly fixed that one. But also fixed a wrong Conduite to a break, which just gives a bit more performance.

hannibal002:
Well, reading is hard.
Scrollable list is not the same as scroll in inventory buttons.

hannibal002 (changes requested):
The tip consistency breaks after scrolling, and after scrolling down ince the bottom is also wrong
grafik

hannibal002 (changes requested):
now the list is too long
grafik
even though the gui is smaller
grafik

also there is still a offset
grafik

hannibal002:
<details>
<summary>maybe those images help</summary>

very bottom
grafik

one scroll above the very bottom
grafik


</details>

Thunderblade73:
Will look at that tomorrow.

DavidArthurCole:
image

I was trying to work on a better version of this, but Renderable was kicking my ass, and I pretty much gave up on making progress. [/CakeTrackerFix](https://github.com/DavidArthurCole/SkyHanni/tree/CakeTrackerFix)

hannibal002 (approved):

PR #3380, opened 2025-02-06
Docs: we love feather

j10a1n15:
## What
we love feather
mentions feather as a mod that adds their own scoreboard
we love feather

exclude_from_changelog

### DIFF
--- a/docs/DISCORD_FAQ.md
+++ b/docs/DISCORD_FAQ.md
@@ -59,7 +59,7 @@ _Frequently Asked Questions_
 
 > **12: Why can I still see the normal Scoreboard when using Custom Scoreboard?**
 > Most of the time, this is a mod conflict.
-> If you are using [Feather Client](https://feathermc.com/), please disable their own Scoreboard Feature (or just uninstall Feather it).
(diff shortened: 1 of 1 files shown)


docs/DISCORD_FAQ.md
```diff
@@ -59,6 +59,7 @@ _Frequently Asked Questions_
 
 > **12: Why can I still see the normal Scoreboard when using Custom Scoreboard?**
 > Most of the time, this is a mod conflict.
+> If you are using [Feather Client](https://feathermc.com/), please disable their own Scoreboard Feature (or just uninstall Feather it).
```
DavidArthurCole: > (or just uninstall Feather it)

Wording change?

DavidArthurCole (changes requested):
> Running SkyHanni on Badlion, Lunar, Feather, or newer versions of Minecraft is not supported.

I don't know why we're adding this to the docs tbh - feather isn't supported at all

hannibal002:
/shdebug is supposed to warn when a users is using feather already

j10a1n15:
Yet people still use feather and we cant do anything about people using feather (unless we do the skytils approach and warn them every launch)
Its imo better to have this

PR #3388, opened 2025-02-07
Fix + Feature: Modernized Jerry Feature

Thunderblade73:
## What
The Jerry Feature now uses the Mob Detection instead of guessing. In the process I also added line to jerry feature.

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added Line to Jerry. - Thunderblade73
    * Shows a line to your spawned Jerry from Jerrypocalypse.

## Changelog Fixes
+ Fixed Jerry highlight being too aggressive. - Thunderblade73
    * Now only highlights your Jerrys.
    * No longer shows false positives during the Great Spook Festival.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/ConditionalUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/ConditionalUtils.kt
@@ -30,7 +30,7 @@ object ConditionalUtils {
         whenChanged { _, new -> observer(new) }
     }
 
-    fun Property<out Boolean>.onTurnOn(observer: Runnable) {
+    fun Property<out Boolean>.onEnable(observer: Runnable) {
         whenChanged { _, _ ->
             if (this.get()) {
                 observer.run()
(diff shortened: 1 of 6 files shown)


src/main/java/at/hannibal2/skyhanni/utils/ConditionalUtils.kt
```diff
@@ -30,6 +30,14 @@ object ConditionalUtils {
         whenChanged { _, new -> observer(new) }
     }
 
+    fun Property<out Boolean>.onTurnOn(observer: Runnable) {
```
j10a1n15: I feel like onEnable sounds better

j10a1n15 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/ConditionalUtils.kt
ItsEmpa: onEnabled sounds better to me than onEnable

PR #3397, opened 2025-02-08
Fix: LineToMob ignoring condition

Thunderblade73:
We do not talk about that.

## Changelog Fixes
+ Fixed non-functional "Line to Jerry" disable option. - Thunderblade73


hannibal002:
this should not ignore the changelog. there is a bug reported by users "disabling the line feature does still show the line"

hannibal002 (approved):

PR #3398, opened 2025-02-08
Backend: Readded the drawLineToEye

Thunderblade73:
## What
Added back the drawLineToEye, which was wrongly removed. Also replace most of the instances where the other ways of doing the same where used.

Since that method was more accurate than what we did since it literally draws the point at the camera spot, since it does remove the translation we do to render in world frame (though our translation is not 100% correct to begin with, but that's beside the point)

+ Added LineDrawer.drawLineToEye. - Thunderblade73


Thunderblade73:
I'm the only one bothered with that. And since I quit Skyblock it does not matter,

PR #3407, opened 2025-02-09
Backend: add button with boolean

hannibal002:
## What
Added support for adding buttons to renderable/searchable lists via boolean config.
Using button utils in cake tracker.

+ Added support for buttons in renderable/searchable lists via boolean config. - hannibal2


DavidArthurCole:
Why are we duplicating things that have already been written? #3423

PR #3409, opened 2025-02-09
Backend: Complex command

Thunderblade73:
## Dependency
* https://github.com/hannibal002/SkyHanni-REPO/pull/390
## What
This adds a command handling system in which you only have to specify the arguments and a processing object, which makes the creation of more complex commands easier.
Also adds a generalized way to handle an Item as command input with full tabcomplete support. (Currently takes all items into account but a simple filter can be used to only use domain specific inputs eg; gfs tabcomplete)

For testing use: #2621 
Pls ignore unused functions warnings since this PR only specifics the API but nothing uses it, therefore we have a bunch of unused warnings in this pr.

Note: This is split up from the main feature because of this comment: https://github.com/hannibal002/SkyHanni/pull/3317#issuecomment-2614466478

+ Added argument-definition-driven command-handling system. - Thunderblade73


hannibal002:
bump @Thunderblade73

src/main/java/at/hannibal2/skyhanni/utils/NeuItems.kt
```diff
@@ -48,6 +59,8 @@ object NeuItems {
     fun onRepoReload(event: RepositoryReloadEvent) {
         val ignoredItems = event.getConstant<MultiFilterJson>("IgnoredItems")
         ignoreItemsFilter.load(ignoredItems)
+        val aliases = event.getConstant<ItemAliases>("ItemAliases")
```
hannibal002, line 62: where is the repo pr?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/NeuItems.kt
Thunderblade73: Good question. xD
I never did do anything with that file yet, I just implemented it.
Also I sad that it is broke as hell. I try to fix it tomorrow.

PR #3411, opened 2025-02-10
Backend: Improve speed of event posting.

Nessiesson:
## What
When posting an event, the code does a check for stuff like whether it should only run when in skyblock or on certain islands etc. This information is currently queried through the annotation data of the event handler method, and this is apparently somewhat expensive. It goes through some proxy object and spends time doing linked hashmap idk things.

Instead of that expensive operation being performed every single time the event is posted, this PR queries that annotation data once when registered and then uses those values from then on. All of this assumes that the annotation values never change which I don't think they can anyways (not without gross hax).

Below are pictures of profiling data showing the before and after. These were taken from a 2 minute run where I stood still and stared on my island. Ideally more data should gathered as I do not have a ton of render stuff on my screen/etc. Also I have PolyPatcher's hud caching enabled+FPS limit of 180FPS. And things. idk. However event posting went from ~5400ms to ~3900ms which is about 30% faster.

<details>
<summary>Images</summary>

Before:
event-post-before

After:
event-post-after

</details>

## Changelog Improvements
+ Improved performance overall slightly. - Nessiesson

+ Code now queries `HandleEventCached` instead of `HandleEvent` directly. - Nessiesson

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/event/EventListeners.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/event/EventListeners.kt
@@ -69,30 +69,28 @@ class EventListeners private constructor(val name: String, private val isGeneric
         val name: String,
         val invoker: Consumer<Any>,
         val options: HandleEventCached,
-        val generic: Class<*>?,
-    ) {
-        val onlyOnIslandTypes: Set<IslandType> = getIslands(options)
-
-        companion object {
-            private fun getIslands(options: HandleEventCached): Set<IslandType> =
-                if (options.onlyOnIslands.isEmpty()) setOf(options.onlyOnIsland)
-                else options.onlyOnIslands.toSet()
-        }
-    }
+        val generic: Class<*>?
+    )
 
     class HandleEventCached(
         val onlyOnSkyblock: Boolean,
         val onlyOnIsland: IslandType,
-        val onlyOnIslands: Array<out IslandType>,
+        val onlyOnIslandTypes: Set<IslandType>,
         val priority: Int,
         val receiveCancelled: Boolean
     ) {
         constructor(event: HandleEvent) : this(
             event.onlyOnSkyblock,
             event.onlyOnIsland,
-            event.onlyOnIslands,
+            getIslands(event.onlyOnIsland, event.onlyOnIslands),
             event.priority,
             event.receiveCancelled
         )
+
+        companion object {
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/api/event/EventListeners.kt
```diff
@@ -68,15 +68,31 @@ class EventListeners private constructor(val name: String, private val isGeneric
...
 
         companion object {
-            private fun getIslands(options: HandleEvent): Set<IslandType> =
+            private fun getIslands(options: HandleEventCached): Set<IslandType> =
                 if (options.onlyOnIslands.isEmpty()) setOf(options.onlyOnIsland)
                 else options.onlyOnIslands.toSet()
         }
     }
+
+    data class HandleEventCached(
```
ItsEmpa: this doesnt need to be a data class right?

hannibal002:
Feel free to also add a user facing fix changelog alogn the lines of "improved performance overall slightly"

src/main/java/at/hannibal2/skyhanni/api/event/EventListeners.kt
Nessiesson: probably not

src/main/java/at/hannibal2/skyhanni/api/event/EventListeners.kt
```diff
@@ -68,15 +68,31 @@ class EventListeners private constructor(val name: String, private val isGeneric
...
         val invoker: Consumer<Any>,
-        val options: HandleEvent,
+        val options: HandleEventCached,
         val generic: Class<*>?,
     ) {
         val onlyOnIslandTypes: Set<IslandType> = getIslands(options)
 
         companion object {
-            private fun getIslands(options: HandleEvent): Set<IslandType> =
+            private fun getIslands(options: HandleEventCached): Set<IslandType> =
```
ThatGravyBoat: Since the data is now being stored in its own class instead of a proxy you can move this Set<> logic in there directly

Nessiesson: kinda like that?

ThatGravyBoat: yep like that

Nessiesson:
Closed in favour of https://github.com/hannibal002/SkyHanni/pull/3417.

PR #3413, opened 2025-02-10
Fix: Hide Other Players in Dance Room Helper not updating if disabled

lunaynx:
## What
Fixed the "Hide Other Players" option in Dance Room Helper not being properly respected if the main "Enabled" switch is off. (`inRoom` was not being updated, so it could also result in players being hidden outside the Mirrorverse.)

Also some small code refactors, using `onlyInIsland` as well as adding a `RealPlayer` check just in case, though it is not strictly necessary with the main bugfix, as we don't need to show any entities in the Dance Room. (I've confirmed that the mirrored players are also detected as `RealPlayer`.)

## Changelog Fixes
+ Fixed "Hide Other Players" in Dance Room Helper not working when the helper is disabled. - Luna
+ Fixed players and some mobs being hidden outside the Mirrorverse when "Hide Other Players" is enabled in Dance Room Helper. - Luna


hannibal002 (changes requested):
how about we also check if the player is in the mirror verse everywhre, not only if in the dance room? so that we dont need to check if in dance room everytime while in rift?

hannibal002:
Also the isEnabled() functio can now be removed as its only containing one check

hannibal002 (approved):

PR #3416, opened 2025-02-10
Backend: reorder storage

rueblimaster:
## What
reordered ProfileSpecificStorage,kt and added comments to specify which feature goes where
also rearranged PlayerSpecificStorage.kt
as I found it somewhat of a mess

+ Reordered internal storage. - rueblimaster

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
@@ -61,31 +61,117 @@ import java.time.LocalDate
 import kotlin.time.Duration
 
 class ProfileSpecificStorage {
+    // api
     @Expose
-    var currentPet: String = ""
+    var skillData: MutableMap<SkillType, SkillApi.SkillInfo> = enumMapOf()
 
     @Expose
-    var experimentation: ExperimentationStorage = ExperimentationStorage()
+    var totalSkyBlockXP: Int? = null
 
-    class ExperimentationStorage {
+    // features
+    // - combat
+    @Expose
+    var ghostStorage: GhostStorage = GhostStorage()
+
+    class GhostStorage {
         @Expose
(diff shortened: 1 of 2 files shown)


rueblimaster:
I'm right with no changelog right?

Thunderblade73:
> I'm right with no changelog right?

This should get a entry in Technical.
exclude_from_changelog idealy shouldn't be used.

rueblimaster:
also, we could go the route of each feature etc gets its own class, but that would need a lot more work as we'd have to fix all the config links etc

rueblimaster:
> > I'm right with no changelog right?
> 
> This should get a entry in Technical. exclude_from_changelog idealy shouldn't be used.

ok, ty, I guess like this?

src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
```diff
@@ -345,7 +385,7 @@ class ProfileSpecificStorage {
         @Expose
         var latestTrueFarmingFortune: MutableMap<CropType, Double> = enumMapOf()
 
-        // TODO use in /ff guide
+        // - TODO use in /ff guide
```
hannibal002: this is unnecessary

hannibal002 (changes requested):
looks good overall.
in the end we want to use classes isntead of comments, though this can happen later

src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
rueblimaster: you're right, I fixed it and checked all other comments for the same mistake

rueblimaster:
> looks good overall. in the end we want to use classes instead of comments, though this can happen later

yes classes are the way to go. The switch itself will just be quite some work as we'd have to fix all the config links

hannibal002 (approved):

PR #3417, opened 2025-02-10
Backend: Event Predicates

ItsEmpa:
## What
This PR changes the way in which conditions are checked when invoking events, only looking at which conditions will have to be checked when the listener is first initialized, and storing the conditions as lists of lambdas. it also makes it so that conditions such as onlyOnSkyblock and onlyOnIsland(s) are only calculated once every tick, which should help with performance a bit on events that get called a lot (such as render events getting called every frame).

This PR also indirectly fixes the issue found and fixed in #3411, because it only accesses the values of the HandleEvent object once.

Did some testing while in hub 1, this is the difference i found when profiling the previous and the new shouldInvoke function was of more or less ~66%

<details>
<summary>Profiling Images</summary>

before
imagen
after
imagen

</details>

## Changelog Improvements
+ Improved overall performance slightly. - Empa, Nessiesson

+ Changed event preconditions to be cached per tick and checked only when necessary. - Empa, ThatGravyBoat

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
@@ -86,9 +86,7 @@ object ConfigUtils {
         if (tryJumpToEditor(ConfigGuiManager.getEditorInstance())) return
 
         // TODO create utils function "crashIfInDevEnv"
-        if (EventHandler.isInEventHandler) {
-            throw Error("can not jump to editor $name")
-        }
+        ErrorManager.crashInDevEnv("Can not open the config")
         ErrorManager.logErrorStateWithData(
             "Can not open the config",
             "error while trying to jump to an editor element",
--- a/src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
@@ -1,7 +1,6 @@
 package at.hannibal2.skyhanni.utils
 
 import at.hannibal2.skyhanni.SkyHanniMod
-import at.hannibal2.skyhanni.api.event.EventHandler
 import at.hannibal2.skyhanni.config.ConfigGuiManager
 import at.hannibal2.skyhanni.config.HasLegacyId
 import at.hannibal2.skyhanni.test.command.ErrorManager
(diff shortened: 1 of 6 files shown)


src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
```diff
@@ -86,9 +86,7 @@ object ConfigUtils {
         if (tryJumpToEditor(ConfigGuiManager.getEditorInstance())) return
 
         // TODO create utils function "crashIfInDevEnv"
```
NopoTheGamer: ?

NopoTheGamer (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
ItsEmpa: i forgor to remove the todo

hannibal002 (changes requested):
waiting on changes above

src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
```diff
@@ -85,10 +84,7 @@ object ConfigUtils {
     fun KMutableProperty0<*>.jumpToEditor() {
         if (tryJumpToEditor(ConfigGuiManager.getEditorInstance())) return
 
-        // TODO create utils function "crashIfInDevEnv"
-        if (EventHandler.isInEventHandler) {
-            throw Error("can not jump to editor $name")
-        }
+        ErrorManager.crashInDevEnv("Can not open the config")
```
NopoTheGamer: why does this remove information? (name)

hannibal002: the crash log includes the stack trace
the stack trace points to the line in code
in the line in code you can see the filed name that we try to jump to

but i agree there is no harm in including the name in the error message directly (or show via extra data?)

hannibal002 (approved):
works fine for me

src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
ItsEmpa: > why does this remove information? (name)

ill be honest that was not intentional

PR #3418, opened 2025-02-10
Fix: ChestValue with non en_Us Languages

j10a1n15:
## What
Replaced String comparisons by comparing against the I18N value of container.chest & container.chestDouble to support both different languages & texturepacks changing the chest name.
I suppose this will break/trigger badly if some texturepack names container.chest to like "SkyBlock Menu" but who does that!

<details>
<summary>Images</summary>

image
image

</details>

## Changelog Fixes
+ Fixed ChestValue not showing when using different Minecraft languages. - j10a1n15


hannibal002 (changes requested):
please write abstract inNormalChest() fun that can also be used here. maybe in InventoryUtils?
https://github.com/hannibal002/SkyHanni/blob/33d615037e74625d2b7b819b5cbe9c90974d8b49/src/main/java/at/hannibal2/skyhanni/features/inventory/HideNotClickableItems.kt#L342

hannibal002 (approved):
Meine große Truhe ist wertvoll!
image

PR #3419, opened 2025-02-10
Improvement: Cache render 

hannibal002:
## What
caching the render action to only happen once per tick, then fire the cached result every frame until next tick

## does not yet work, causes flickers and doesnt show the actual display info.
idk why. this is just my current state of where i left off.
i heavily used AI to help me with the GL state calls as i have no experience with gl at all, so this might be the reason why it doesnt work.
anyway, here it is. i appreciate if others might look at it and find the obvious error im too blind to see rn.

## Changelog Fixes
+ Increased performance while rendering GUI's

+ Caching render calls

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/RenderCache.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/RenderCache.kt
@@ -177,6 +177,11 @@ object RenderCache {
     }
 
     fun onFrame() {
+        if (!useCache) {
+            heavyRenderCall()
+            return
+        }
+
         // Trigger debug logging for one second when the player is sneaking.
         val player = Minecraft.getMinecraft().thePlayer
         if (player != null && player.isSneaking && debugGLLoggingStartTime < 0L) {
@@ -191,10 +196,6 @@ object RenderCache {
         // Update the FBO if display size changed.
         updateFBOIfNeeded()
 
-        if (!useCache) {
-            heavyRenderCall()
-            return
-        }
 
         if (updateCache) {
             renderToFBO(frameBufferId, width, height)
--- a/src/main/java/at/hannibal2/skyhanni/utils/RenderCache.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/RenderCache.kt
@@ -1,0 +1,225 @@
+package at.hannibal2.skyhanni.utils
+
+import at.hannibal2.skyhanni.SkyHanniMod
+import at.hannibal2.skyhanni.api.event.HandleEvent
+import at.hannibal2.skyhanni.events.GuiRenderEvent
+import at.hannibal2.skyhanni.events.minecraft.SkyHanniTickEvent
+import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
+import at.hannibal2.skyhanni.utils.ReflectionUtils.makeAccessible
+import net.minecraft.client.Minecraft
+import net.minecraft.client.renderer.GlStateManager
+import org.lwjgl.BufferUtils
+import org.lwjgl.opengl.Display
+import org.lwjgl.opengl.GL11
+import org.lwjgl.opengl.GL30
+import java.nio.ByteBuffer
+import kotlin.math.abs
+
+@SkyHanniModule
+object RenderCache {
+
+    // -- FBO and caching variables --
+    private var frameBufferId: Int = 0
+    private var textureId: Int = 0
+    private var fboWidth: Int = Display.getWidth()
+    private var fboHeight: Int = Display.getHeight()
+    private val width get() = Display.getWidth()
+    private val height get() = Display.getHeight()
+    private val useCache get() = SkyHanniMod.feature.misc.cacheRender
+    private var updateCache = false
+
+    init {
+        // Initial FBO creation.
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/utils/RenderCache.kt
```diff
@@ -0,0 +1,226 @@
...
+        // Trigger debug logging for one second when the player is sneaking.
+        val player = Minecraft.getMinecraft().thePlayer
+        if (player != null && player.isSneaking && debugGLLoggingStartTime < 0L) {
+            startGLLogging()
+        }
+
+        // Frame guard using renderPartialTicks.
+        val currentPartialTicks = getRenderPartialTicks()
+        if (abs(currentPartialTicks - lastPartialTicks) < 0.001f) return
+        lastPartialTicks = currentPartialTicks
```
Thunderblade73: This one may cause some flickering, since it randomly drops frames if you have less than 1ms between to frames (which always happens at 1000fps+, but can happen way before since the frametime is usually not uniform.

src/main/java/at/hannibal2/skyhanni/utils/RenderCache.kt
```diff
@@ -0,0 +1,226 @@
...
+
+
+        if (updateCache) {
+            renderToFBO(frameBufferId, width, height)
+            updateCache = false
+        }
+        GL11.glPushAttrib(GL11.GL_ALL_ATTRIB_BITS)
+        GL11.glDisable(GL11.GL_DEPTH_TEST)
+        drawCachedTexture(textureId, 0f, 0f, width.toFloat(), height.toFloat())
+        GL11.glPopAttrib()
```
Thunderblade73: I think I remember that there is some major issue with the push and pop for attributes. Since the GlStateManager gets messed up see: https://github.com/MinecraftForge/MinecraftForge/issues/1637

src/main/java/at/hannibal2/skyhanni/utils/RenderCache.kt
```diff
@@ -0,0 +1,226 @@
...
+        GL11.glClear(GL11.GL_COLOR_BUFFER_BIT or GL11.GL_DEPTH_BUFFER_BIT)
+
+        // Set up orthographic projection.
+        GL11.glMatrixMode(GL11.GL_PROJECTION)
+        GL11.glPushMatrix()
+        GL11.glLoadIdentity()
+        GL11.glOrtho(0.0, width.toDouble(), height.toDouble(), 0.0, 0.0, 1.0)
+        GL11.glMatrixMode(GL11.GL_MODELVIEW)
+        GL11.glPushMatrix()
+        GL11.glLoadIdentity()
```
Thunderblade73: Why do we set up a new projection here? We should be able to use the one we are in already.

src/main/java/at/hannibal2/skyhanni/utils/RenderCache.kt
```diff
@@ -0,0 +1,226 @@
...
+        GL11.glPopMatrix() // MODELVIEW
+        GL11.glMatrixMode(GL11.GL_PROJECTION)
+        GL11.glPopMatrix()
+        GL11.glMatrixMode(GL11.GL_MODELVIEW)
+
+        // Unbind the FBO.
+        GL30.glBindFramebuffer(GL30.GL_FRAMEBUFFER, 0)
+        // Restore the previous viewport.
+        GL11.glViewport(viewport[0], viewport[1], viewport[2], viewport[3])
+        GL11.glFlush() // Ensure rendering completes
```
Thunderblade73: We should never call glFlush. Since that way we mess up frame generation of minecraft.

Thunderblade73 (changes requested):
The idea is not far off. But there are two major issues:
* Every interaction we have in a gui now runs at 20fps, which introduces a noticeable delay.
* The second one is, that we can't garantier that the client PC has Gl 3.0 support, since minecraft only requries GL 1.1 to run. We already had a few issues with the shaders since those need GL 2.1 to work. It is definitely an edge case but it still can happen. So we need to disable the whole logic if GL 3.0 is not avaialbale.

I have no experience with framebuffers (yet) so I can't tell if that code is even correct at all. But the GL 11 calls already don't make sense at some places, since it is handled as if we are the one with full control and responsibility and that there is no one else rendering in the gl context (which is not true, since at least minecraft does). Meaning there are full on setup calls and closure calls for gl. Which does not make any sense since we already wrapped inside those similar calls via minecraft.

hannibal002:
All Code lines you marked in your review are added as ways to fix the flicker. Originally none of them existed and the flicker was still there. This also includes the setup / closure calls.
Maybe I should comment out all lines that got added as fixes, and could be removed now without any change, but this is already more than 50% of the code. I left them in the pr so it's clear what we (my ai and I) have already tested.

Good point about the hover interaction, this would need extra checking/higher frame rate, or simply entirely disabling the cache while hovering. Possible to Workaround but requires more work

The missing gl3.0 support can als be checked against and the caching disabled then.

CalMWolfs:
@hannibal002 are you planning on updating this pr or can I close it

DavidArthurCole:
image

surely...

CalMWolfs:
closing as this is stale and already isnt a super great implementation for 1.8 let along 1.21

PR #3423, opened 2025-02-11
Improvement + Backend: Cake Tracker Renderable Buttons

DavidArthurCole:
## What
Removes the custom solution for type and order changing, preferring to use the enums and Renderable Buttons instead.
Ported the CakeTracker config to kotlin while I was 'here'.

Added the ability to change which year is being searched for in the auction hover + click.
There were some unnecessary methods that could get condensed down. Essentially all existing `..click..` methods within Renderable were changed to be based in `clickable()`, which now takes optional params for things like `onHover` and `tips`, then delegates appropriately. This should in theory make the construction of these much more universal and common.

<details>
<summary>Images</summary>

image

image

</details>

## Changelog Improvements
+ Standardized buttons and display in Cake Tracker. - Daveed
    * Arrow keys can now be used to change Auction Search year.

+ Added support for keyboard keys in `Renderable.multiClickable()` alongside mouse clicks. - Daveed
+ Refactored many functions in `Renderable` to delegate to `clickable()` and `multiClickable()`. - Daveed


src/main/java/at/hannibal2/skyhanni/utils/renderables/Renderable.kt
```diff
@@ -124,55 +126,64 @@ interface Renderable {
...
-                bypassChecks = bypassChecks,
-            )
-        }
+        ) = clickable(string(text), onClick, bypassChecks, condition, tips, onHover)
 
-        fun multiClickAndHover(
-            text: Any,
-            tips: List<Any>,
+        fun clickable(
+            render: Renderable, onClick: () -> Unit,
```
Thunderblade73: You miss a line break here.

src/main/java/at/hannibal2/skyhanni/utils/renderables/Renderable.kt
```diff
@@ -124,55 +126,64 @@ interface Renderable {
             )
         }
 
-        fun clickAndHover(
-            text: Any,
-            tips: List<Any>,
+        fun clickable(
+            text: String, onClick: () -> Unit,
```
Thunderblade73: You miss a line break here.

src/main/java/at/hannibal2/skyhanni/utils/renderables/Renderable.kt
```diff
@@ -124,55 +126,64 @@ interface Renderable {
...
-        fun leftAndRightClickable(
+        private fun multiClickable(
             render: Renderable,
-            onClick: (Direction) -> Unit,
+            click: Map<Int, () -> Unit>,
             bypassChecks: Boolean = false,
             condition: () -> Boolean = { true },
+            /**
+             * This function is called if no keys within [click] handle the click
+             */
```
Thunderblade73: The doc does not state what it does correctly. Since it fires on hover with none of the clicks triggering.

src/main/java/at/hannibal2/skyhanni/utils/renderables/Renderable.kt
```diff
@@ -181,64 +192,43 @@ interface Renderable {
...
-            }
 
-            override fun render(posX: Int, posY: Int) {
-                if (isHovered(posX, posY) && condition() && shouldAllowLink(true, bypassChecks)) {
-                    if ((-100).isKeyClicked()) onClick(Direction.LEFT)
-                    else if ((-99).isKeyClicked()) onClick(Direction.RIGHT)
-                    else pureScrollInput.tryUpdateScroll()
+            private fun handleClickChecks() {
+                for ((button, onClick) in click) {
+                    if (button.isKeyClicked()) return onClick()
```
Thunderblade73: idk about that return here. Since the we would need 2 calls to process two "clicks" that happend at the same time.

src/main/java/at/hannibal2/skyhanni/utils/renderables/Renderable.kt
```diff
@@ -124,55 +126,64 @@ interface Renderable {
...
+             */
+            click: Map<Int, () -> Unit>,
             bypassChecks: Boolean = false,
             condition: () -> Boolean = { true },
-        ) = leftAndRightClickable(
-            render,
-            onClick = { if (it == Direction.LEFT) onClick() },
-            bypassChecks, condition,
+            tips: List<Any>? = null,
+            onHover: () -> Unit = {},
```
Thunderblade73, line 171: The way you implemented onHover it is only called if an tip is present. But we have code that needs hover feedback but don't have a hoverTips.

src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt
```diff
@@ -22,8 +22,14 @@ abstract class ScrollInput(
...
         get() = scrollValue.getValue()
 
+    enum class ScrollDirection(val value: Int) {
+        UP(-1),
+        DOWN(1),
+    }
+
     fun asInt() = scroll.toInt()
     fun asDouble() = scroll
+    fun asDirection() = ScrollDirection.entries.firstOrNull { it.value == scroll.toInt() }
```
Thunderblade73: This is not really clean. A x >0 / x < 0 checks makes more sense. Since this one is callable by all scroll inputs and only makes sense for the pure variant the way you implemented it.
So make it a three state enum and use:
```suggestion
    fun asDirection() = if(scroll == 0.0) NONE else if(scroll > 0.0) DOWN else UP
```
Also btw why is this inverted? Why is -1 up and 1 down?

Thunderblade73 (changes requested):
Just looked at the Renderable change.

src/main/java/at/hannibal2/skyhanni/utils/renderables/Renderable.kt
DavidArthurCole: > idk about that return here. Since the we would need 2 calls to process two "clicks" that happend at the same time.

There's nothing currently that relies on two clicks at the same time, nor is there any logic in any of the clickables that relies on multi-detection. If it's not this, it's duplicating the code for the scrollable, pick your poison.

Thunderblade73: You know that the return does 99.9999999% of the calls nothing. Only in the one frame someone clicked.
I know that nothing needs it, but it simply process it one frame later. Which is not huge but with the render caching what Hannibal plans it will make a difference.

src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt
DavidArthurCole:
```kotlin
            private fun ScrollInput.Companion.PureVertical.tryUpdateScroll() {
                pureScrollInput.update(true)
                when (asInt()) {
                    -1 -> onClick(Direction.RIGHT) // Scroll Up -> Right Click
                    1 -> onClick(Direction.LEFT) // Scroll Down -> Left Click
                    else -> return
                }
                dispose()
            }
 ```
 
 That's existing code, -1 indicates up, 1 indicates down 🤷

File: src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorFeatures.kt [hunk 7]
```diff
@@ -219,17 +219,17 @@ object GardenVisitorFeatures {
...
+                    tips = internalName.createBuyTip(),
+                    onLeftClick = {
+                        if (!GardenApi.inGarden() || NeuItems.neuHasFocus()) return@clickable
                         if (Minecraft.getMinecraft().currentScreen is GuiEditSign) {
                             LorenzUtils.setTextIntoSign("$amount")
                         } else {
                             internalName.buy(amount)
                         }
                     },
-                    tips = internalName.createBuyTip(),
```
hannibal002, line 232: looks like unintended

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorFeatures.kt
DavidArthurCole: As I noted in chat, this was a standardization change as all other references to `clickable()` have tips listed before the `onLeftClick`

rueblimaster:
@DavidArthurCole so why are we changing from onClick to onLeftClick?

PR #3425, opened 2025-02-12
Fix: Mineshaft Room Id again

j10a1n15:
## What
this time surely
the lines probably get set before the island type is known, causing the room to not be in added lines after its known

## Changelog Fixes
+ Fixed Custom Scoreboard not displaying Mineshaft Room ID. - j10a1n15

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/events/ScoreboardUpdateEvent.kt
+++ b/src/main/java/at/hannibal2/skyhanni/events/ScoreboardUpdateEvent.kt
@@ -3,9 +3,11 @@ package at.hannibal2.skyhanni.events
 import at.hannibal2.skyhanni.api.event.SkyHanniEvent
 
 class ScoreboardUpdateEvent(
+    @Deprecated("Use new instead", ReplaceWith("new"))
     val full: List<String>,
     val old: List<String>,
 ) : SkyHanniEvent() {
+    val new = full
 
     val added: List<String> = full - old.toSet()
     val removed: List<String> = old - full.toSet()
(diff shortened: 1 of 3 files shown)


hannibal002:
i was about to ask why it doesnt show :(

hannibal002 (changes requested):
After joining the mineshaft, the room id was visible.
Half a second later, the room id got replaced by the server id
image

src/main/java/at/hannibal2/skyhanni/events/ScoreboardUpdateEvent.kt
```diff
@@ -3,9 +3,11 @@ package at.hannibal2.skyhanni.events
 import at.hannibal2.skyhanni.api.event.SkyHanniEvent
 
 class ScoreboardUpdateEvent(
+    @Deprecated("Use new instead", ReplaceWith("new"))
```
hannibal002, line 6: there are cases where we want to access the full scoreboard, and this only after anything changes. the scoreboard does not have an abstraction level like the tab list, so i don't think the depreciation of the `full` field is good.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/events/ScoreboardUpdateEvent.kt
hannibal002: nvm

hannibal002:
https://github.com/hannibal002/SkyHanni/commit/5cfd8adbea035054987808a46c52ec44478c65ea
funny

src/main/java/at/hannibal2/skyhanni/events/ScoreboardUpdateEvent.kt
j10a1n15: `new` just sounds better if we have `old` thats basically all that did, but yeah :3

hannibal002 (approved):

PR #3350, opened 2025-02-01
Feature: NEU Soul Path Find

hannibal002:
## What
Added NEU Soul Path Find

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added NEU-Souls Pathfinder. - hannibal2
    * When using `/neusouls`, displays a pathfinding line to the nearest missing Fairy Soul.
    * Shows the progress of Souls found of the current island in chat.

PR #3427, opened 2025-02-13
Improvement: Even better fairy soul logic

hannibal002:
## Dependencies
- #3350

## What
Using algorithms to find one path for all fairy souls in one go.

## Changelog Improvements
+ Added Fast Fairy Soul Finder. - hannibal2
    * Uses a fast pathfinder route to reach all Fairy Souls on the current island.
    * Does not require NEU.


src/main/java/at/hannibal2/skyhanni/features/misc/FastFairySoulsPathfinder.kt
```diff
@@ -0,0 +1,346 @@
...
+
+    /**
+     * REGEX-TEST: §7Fairy Souls: §e11§7/§d11
+     */
+    private val loreSoulPattern by patternGroup.pattern(
+        "new",
+        "§7Fairy Souls: §e(?<have>.*)§7\\/§d(?<total>.*)",
+    )
+
+    class Data(
```
CalMWolfs: this can be private i think

src/main/java/at/hannibal2/skyhanni/features/misc/FastFairySoulsPathfinder.kt
```diff
@@ -0,0 +1,346 @@
...
+        for (stack in event.inventoryItems.values) {
+            val island = IslandType.getByNameOrNull(stack.displayName.removeColor()) ?: continue
+            val have = stack.getLore().firstOrNull()?.let {
+                loreSoulPattern.matchMatcher(it) {
+                    group("have").toIntOrNull()
+                }
+            } ?: continue
+
+
+            if (LorenzUtils.skyBlockIsland == island) {
```
CalMWolfs: could possibly be
`if (island.isInIsland()) {`

CalMWolfs (changes requested):
Works alright,

1 complaint. The final node isnt on the fairy soul, which while this makes sense because you cant usually stand where the soul is, i found myself getting to the final node and not knowing where exactly i need to look for the soul, maybe a line to it would be helpful
image


2. I think this routing is meant to be using travelling salesman to find the best route to get all souls on the island, but when doing souls on the hub i have fouind my self being sent back and forth acrosss the whole island instead of getting all in an area together

hannibal002:
> 2. I think this routing is meant to be using travelling salesman to find the best route to get all souls on the island, but when doing souls on the hub i have fouind my self being sent back and forth acrosss the whole island instead of getting all in an area together

i find this solution is way better than the previous simple greedy solution. a full/perfect travelings salesman solution is impossible, but feel free to update/change the navigation utils class to make it even better. im happy with the current state already, and on other non hub islands the pathing is perfect in my testing. in hub there are only few loops.

hannibal002:
> 1 complaint. The final node isnt on the fairy soul, which while this makes sense because you cant usually stand where the soul is, i found myself getting to the final node and not knowing where exactly i need to look for the soul, maybe a line to it would be helpful

we dont have the exact soul data atm. we maybe also read the neu repo data and show that line, but idk if that is actually necessary as most of the times the soul is easily findable or you just use neu souls (not the setting in sh, the setting in neu)

NopoTheGamer:
i think it should auto turn on if you do /neusouls on like the other feature

CalMWolfs:
> > 1 complaint. The final node isnt on the fairy soul, which while this makes sense because you cant usually stand where the soul is, i found myself getting to the final node and not knowing where exactly i need to look for the soul, maybe a line to it would be helpful
> 
> we dont have the exact soul data atm. we maybe also read the neu repo data and show that line, but idk if that is actually necessary as most of the times the soul is easily findable or you just use neu souls (not the setting in sh, the setting in neu)

Yeah its not necessary, but the data could easily be found through the neu repo data and then finding the closest soul to that node in the island

hannibal002:
auto turn on idk, since the new feature of this pr is a competitor to the existing "neu souls" feature in sh.
but we can surely add a chat prompt "hey you wanna a faster route? click here to activate"
what do you think?

NopoTheGamer:
> auto turn on idk, since the new feature of this pr is a competitor to the existing "neu souls" feature in sh. but we can surely add a chat prompt "hey you wanna a faster route? click here to activate" what do you think?

ok

CalMWolfs:
> > 2. I think this routing is meant to be using travelling salesman to find the best route to get all souls on the island, but when doing souls on the hub i have fouind my self being sent back and forth acrosss the whole island instead of getting all in an area together
> 
> i find this solution is way better than the previous simple greedy solution. a full/perfect travelings salesman solution is impossible, but feel free to update/change the navigation utils class to make it even better. im happy with the current state already, and on other non hub islands the pathing is perfect in my testing. in hub there are only few loops.

This is also fair if its only the hub thats broken as yes, 80 souls is probably not realistic to find the optimal path.

Well im fine with this pr regardless of these 2 things provided you adress nopos concern

CalMWolfs (approved):

PR #3437, opened 2025-02-14
Fix: upae in UserLuckBreakdown

j10a1n15:
## What
JerryItem wont always be initialised (if jerryLuck is <1)

## Changelog Fixes
+ Fixed error in UserLuckBreakdown GUI. - j10a1n15


hannibal002:
I feel like using nullable for all those items is more intuitive than using lateinit

j10a1n15:
> I feel like using nullable for all those items is more intuitive than using lateinit

changed that

PR #3446, opened 2025-02-17
Fix: Use TimeLimitedCache instead of a 'hand-rolled' solution.

Nessiesson:
I think(?) it fixes the EntityPlayerSP leak. Idk. Kinda. _Maybe_. Mods seem to leak that and WorldClient all over the place. Turns out that storing those is bad since they change on world change/login/logout/etc.

## What
For some reason the `now - entry.value.first > 500.milliseconds` check isn't properly getting rid of EntityPlayerSP. Something else is probably keeping it alive. It's weird and black magic. The TimeLimitedCache internally uses weak references to the map keys and values so it does not keep them around. idk i'm just typing words now ._.

## Changelog Fixes
+ Fixed memory leak in entity rendering code. - Nessiesson


hannibal002:
please word the user facing changelolg non technical.

Nessiesson:
okay ._.

PR #3454, opened 2025-02-18
Fix: ElectionAPI not checking every minute if no mayor is known

j10a1n15:
## What
I think this should fix it, I guess we dont even need to recheck every 20mins if we have an active mayor (hi nopo I kinda agree), but I guess its a good failsafe just incase

## Changelog Fixes
+ Fixed Mayor updating too slow if no mayor is known. - j10a1n15


jani270:
The head ref may contain hidden characters: "fix/\u2702\uFE0F"

PR #3455, opened 2025-02-18
Fix: ElectionAPI not checking every minute if no mayor is known

j10a1n15:
## What
I think this should fix it, I guess we dont even need to recheck every 20mins if we have an active mayor (hi nopo I kinda agree), but I guess its a good failsafe just incase

## Changelog Fixes
+ Fixed mayor data updating too slow after new mayor takes office. - j10a1n15

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/ElectionApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/ElectionApi.kt
@@ -210,12 +210,10 @@ object ElectionApi {
 
     private fun checkHypixelApi(forceReload: Boolean = false) {
         if (!forceReload) {
-            if (!forceReload) {
-                if (currentMayor == ElectionCandidate.UNKNOWN) {
-                    if (lastUpdate.passedSince() < 1.minutes) return
-                } else {
-                    if (lastUpdate.passedSince() < 20.minutes) return
-                }
+            if (currentMayor == ElectionCandidate.UNKNOWN) {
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/data/ElectionApi.kt
```diff
@@ -210,8 +210,13 @@ object ElectionApi {
 
     private fun checkHypixelApi(forceReload: Boolean = false) {
         if (!forceReload) {
-            if (lastUpdate.passedSince() < 20.minutes) return
-            if (currentMayor == ElectionCandidate.UNKNOWN && lastUpdate.passedSince() < 1.minutes) return
+            if (!forceReload) {
```
lunaynx: This is already checked in the outer `if` block, why duplicate it?

lunaynx (changes requested):

lunaynx (approved):

lunaynx:
~~Actually, one potential concern here is that if the mayor was already set to unknown and the Hypixel API has not updated yet (it often takes a few minutes), we might end up getting the old mayor again and caching that for 20 minutes. Or do we have some protection against that which I'm not seeing?~~

lunaynx:
Never mind, I can't read. `lastMayor` is stored separately. It's probably fine.

j10a1n15:
> Never mind, `lastMayor` is stored separately and not reset, it's probably fine.

yeah was about to send that, it should be fine. I mean that already worked before so probably good

PR #3457, opened 2025-02-18
Fix: Enchanted Book Name

ItsEmpa:
## What
This pr fixes certain issues with enchanted book name, like books combined in anvils calling item add events, and also me being stupid and using < instead of > while iterating chat lines, making it so that it immediately would exit the loop and it would always send the chat message even if there was already a message in chat.

## Changelog Fixes
+ Fixed various issues with the Enchanted Book Name feature. - Empa


lunaynx:
While we're at it, can we also add "Book" after the enchantment in the message, to match cases where Hypixel itself shows the book name?

PR #3460, opened 2025-02-19
Fix: Burrow math is not mathing

hannibal002:
## What
`Double.compare(-0, 0)` is wrong.
`-0 == 0` is  correct.

## Changelog Fixes
+ Fixed rare case where the guess point shows over the known burrow location.


hannibal002:
#3461 got merged

PR #3462, opened 2025-02-19
Feature: Add Trinity helper

Bugzilla123:
## What
Add a notification when user joins a dungeons with 5 puzzles, as this dungeons have a high chance of containing rare 1x1 rooms, like the one Trinity spawns on. There is no official info on [Trinity spawn chances](https://wiki.hypixel.net/Trinity), but [flex posts](https://hypixel.net/threads/best-way-to-get-trinitys-number-instead-of-acquiring-actual-friends.5489159/) and videos I watched, all reinforce the idea of 5 puzzles == Trinity.

## Changelog New Features
+ Add Trinity helper. - Bugzilla

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonTrinityHelper.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonTrinityHelper.kt
@@ -2,38 +2,28 @@ package at.hannibal2.skyhanni.features.dungeon
 
 import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.api.event.HandleEvent
+import at.hannibal2.skyhanni.data.model.TabWidget
 import at.hannibal2.skyhanni.events.dungeon.DungeonEnterEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.HypixelCommands
 import at.hannibal2.skyhanni.utils.LorenzUtils
-import at.hannibal2.skyhanni.utils.RegexUtils.firstMatcher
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonTrinityHelper.kt
```diff
@@ -0,0 +1,42 @@
...
+        "dungeon.puzzle.count",
+        "(?:§.)*Puzzles: (?:§.)*\\((?<amount>\\d+)\\)",
+    )
+
+    private val dragonSound by lazy { createSound("mob.enderdragon.growl", 1f) }
+
+    @HandleEvent
+    fun onDungeonEnter(event: DungeonEnterEvent) {
+        if (!config.enabled) return
+        val puzzleCount = dungeonPuzzleCountPattern.firstMatcher(TabListData.getTabList()) {
```
Thunderblade73: Don't use ```TabListData.getTabList()``` use ```Widget.smth.lines``` instead. (The smth should be replaced with the widget the data is in. With /shcopytablist you should be able to figure out which one to use.)

Thunderblade73 (changes requested):

Bugzilla123:
I got a 5 puzzle room on M5 just now and no Trinity. I still think this feature is helpful for people desperate for contact, but maybe we should make it opt-in? 

Also accepting suggestions on another sound to play, I copied the code from ultra rare books and the dragon sound is the same as dungeon start lol

src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonTrinityHelper.kt
Bugzilla123: Used the mencioned API

PR #3467, opened 2025-02-19
Feature + Backend: Runic mob highlight

SuperCl4sh:
## What
Adds highlight option for runic mobs. Also fixes Mob.isRunic

## Changelog New Features
+ Added Option to highlight runic mobs. - SuperClash

+ Fixed Mob.isRunic not working. - SuperClash


lunaynx:
> ## Changelog Fixes
> * Fixed Mob.isRunic not working. - SuperClash

Should probably be backend as it's not user-facing (no feature actually seems to be using it prior to this PR?)

src/main/java/at/hannibal2/skyhanni/data/mob/Mob.kt
```diff
@@ -114,7 +114,9 @@ class Mob(
     }
 
     val isCorrupted get() = !RiftApi.inRift() && baseEntity.isCorrupted() // Can change
-    val isRunic = !RiftApi.inRift() && baseEntity.isRunic() // Does not Change
+    val isRunic = !RiftApi.inRift() &&
+        armorStand?.getNameAsString()?.startsWith("§5") == true &&
+        mobType == Mob.Type.BASIC// Does not Change
```
Thunderblade73: What do you mean with that comment?

j10a1n15: image
the comment you wrote..?

Thunderblade73: Oh, I see. This should be better be a javadoc rather than a comment. Since it reverse to the Property not the Line it is on. I got confused because I only looked at ```mobType == Mob.Type.BASIC// Does not Change``` isolated which didn't make sense to me (also I may have ignored the - diff)

src/main/java/at/hannibal2/skyhanni/data/mob/Mob.kt
```diff
@@ -114,9 +114,12 @@ class Mob(
     }
 
     val isCorrupted get() = !RiftApi.inRift() && baseEntity.isCorrupted() // Can change
```
j10a1n15: I mean you also use javadoc here aswell then

j10a1n15 (changes requested):

hannibal002:
Why do we need the info that runic does not change? mobs spawn as runic or non runic, this doesnt change after spawning. also the variable is already defined as `val`, indicating that it cant get changed. so this information is neither useful for sb information nor coding information.

NopoTheGamer:
> Why do we need the info that runic does not change? mobs spawn as runic or non runic, this doesnt change after spawning. also the variable is already defined as `val`, indicating that it cant get changed. so this information is neither useful for sb information nor coding information.

probably because the the variable above it says it can change (which is also a val because you brought it up)

ItsEmpa:
the reason why its indicated that it cant change is because the corrupted one *can* change, but its a val with a getter so it can be mistaken that it cant change

CalMWolfs (approved):

hannibal002 (changes requested):
missing highlighting all runic mobs when enabling the feature

Thunderblade73 (approved):
Has no customization for the highlight color but the rest looks fine.

CalMWolfs (approved):
I feel the fact that it doesnt allow customisation of the color & if you toggle the setting it doesnt update the highlight is fine as its in line with all the other mob highlight feataures

PR #3469, opened 2025-02-20
Fix: Burrow location wildly inaccurate under specific circumstances

Bloxigus:
## What
https://discord.com/channels/997079228510117908/1341582719582474333/1342165087170986035
Here a particle from an old burrow is received after the player uses the ability leading to a wildly inaccurate guess.
One possible solution is to prevent the player from using the ability when hypixel is still sending lava particles.

## Changelog Fixes
+ Fixed incorrect burrow guesses under certain circumstances. - bloxigus


lunaynx:
> ## Changelog Fixes
> * Fixed incorrect guesses under certain circumstances. - bloxigus

Should specify burrow guesses, it's not gonna be obvious.

PR #3491, opened 2025-02-23
Feature: Current Chat Display

ItsEmpa:
## What
Adds a GUI element that shows what chat you are currently in (guild chat, party chat, all chat, etc), as well as a bit of info on it (party size if in party, timer until the chat expires if on private messages).

## Changelog New Features
+ Added Current Chat Display. - Empa
    * Displays a GUI element that indicates what chat you are in (e.g. Party, Guild, Coop, All).


hannibal002 (changes requested):
works perfectly fine for me.

can we please make it so that the display onlys shows while the chat is open or a short cooldown after the player sent a chat message?

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -280,6 +280,9 @@ object StringUtils {
 
     fun isEmpty(message: String): Boolean = message.removeColor().trimWhiteSpaceAndResets().isEmpty()
 
+    fun <T : Enum<T>> Enum<T>.toFormattedName(): String =
+        name.split("_").joinToString(" ") { it.lowercase().replaceFirstChar(Char::uppercase) }
```
CalMWolfs: Should be in EnumUtils

CalMWolfs (changes requested):

CalMWolfs (approved):
changes lgtm

hannibal002 (approved):

PR #3501, opened 2025-02-24
Fix: Scroll Table Rendering Last Element outside

Thunderblade73:
## What
Fixed a edge case in the search scroll table logic where the last element rendered even though it did not fit.

## Changelog Fixes
+ Fixed bug with last element rendering in scrolling tables (Trackers) within other text. - Thunderblade73


hannibal002:
ill put this into 2.0.0 directly without beta ig

PR #3504, opened 2025-02-26
Fix + Backend: Renderable Test Suite

Thunderblade73:
## What
Added a good way to add tests for renderables.
The test can be toggled with the ``/shrenderable ...`` ingame inside the dev enviroment.

This moved the to renderable test we had inside SkyHanniDebugAndTest and also added the scrollable test which I used to test the new scrollable renderable.

Also added GuiRenderEvent.GuiOnTopRenderEvent. (This event should be used instead of the GuiRender, which renders twice inside the Inventory, but that change should be done at a later point, since it is outside of the scope of this pr)
And proper docs for the Gui Events.

That PR also fixes two Renderable bugs. First the bug in DragAndDrop which was not in use therefore no one noticed. Secondly a small bug in WrappedStringRenderable which made it so the size was always the text length instead of the width of the line.

## Changelog Fixes
+ Fixed changelog viewer text sometimes being in the wrong spot. - Thunderblade73

+ Added Renderable Test Suite. - Thunderblade73
    * A way to add tests for renderables. Which can be called activated with `/shrenderable`.
+ Added `GuiOnTopRenderEvent`. - Thunderblade73
    * A GuiRenderEvent that renders as overlay outside inventory and inside an inventory as an ChestOverlay.


hannibal002:
- #3916

got merged. feel free to replace the event here now

src/main/java/at/hannibal2/skyhanni/test/SkyHanniDebugsAndTests.kt
```diff
@@ -355,7 +342,8 @@ object SkyHanniDebugsAndTests {
     @HandleEvent(GuiKeyPressEvent::class)
     fun onKeybind() {
         if (!debugConfig.copyInternalName.isKeyHeld()) return
-        val stack = stackUnderCursor() ?: return
+        val focussedSlot = slotUnderCursor() ?: return
+        val stack = focussedSlot.stack ?: return
```
CalMWolfs: Is this change intentional or is it possibly a bad merge as it looks to break the logic added in #4318

CalMWolfs (changes requested):
havent tested yet, just saw this one thing

src/main/java/at/hannibal2/skyhanni/test/SkyHanniDebugsAndTests.kt
Thunderblade73: yep, a wrong merge happend here.

CalMWolfs (approved):
Crazy feature, all worked as expected

PR #3509, opened 2025-02-26
Backend: Made SkyHanniInstallerFrame  kotlin

hannibal002:
+ Changed SkyHanniInstallerFrame to Kotlin hannibal2


lineargraph:
is there a reason sans tradition to keep the installer in the root package? It could just as well be in a regular package, no?

hannibal002:
in 1.21 we will not bundle kotlin, so this will be the onlyfile in the only java file forever

PR #3510, opened 2025-02-26
Backend: Why Kotlin

hannibal002:
## What
Why do we use Kotlin and not Java for the mod?

https://github.com/hannibal002/SkyHanni/blob/wky-kotlin/docs/why_kotlin.md

AI wrote a comparison, and i think this might be interested for some to read.

ignore_from_changelog


jani270 (changes requested):
Just No.

PR #3520, opened 2025-03-02
Backend: ExpermentationTableAPI V2

DavidArthurCole:
## What
Re-worked all modules touching Experimentation to use better event driven data rather than individual subscriptions to lower level stuff.

+ Refactored ExperimentationTableAPI and related features. - Daveed


DavidArthurCole:
Too many drafts

PR #3523, opened 2025-03-05
Fix: Show Sulphur Skitter Box without lava

lunaynx:
## What
Removed the lava check for Sulphur Skitter Box (see changelog for explanation).

<details>
<summary>Images</summary>

</details>

## Changelog Fixes
+ Fixed Sulphur Skitter Box not displaying without lava nearby. - Luna
    * The Player must be within 4 blocks, not the bobber, for the box to display.


hannibal002 (changes requested):
> Fixed Sulphur Skitter Box to also show if there is no lava within its radius. - Luna 

the wording is weird.

rather use "fixed solphur never showing when there is no lava"
or use
"improved sulphor to also show if there is no lava"

lunaynx:
I reworded the changelog, should be better now.

PR #3530, opened 2025-03-09
Improvement + Backend: Outline option for Highlight Nucleus Barriers

ksndq:
## What
Adds an option to only draw the outline around the Crystal Nucleus crystal barrier blocks.

## Changelog Improvements
+ Added Outline Mode to the Nucleus Barriers Highlighter. - ksndq

+ Changed CrystalHighlighterConfig to Kotlin. - ksndq
+ Moved NucleusBarriersBox to the mining.crystalhollows package.  - ksndq


src/main/java/at/hannibal2/skyhanni/features/mining/crystalhollows/NucleusBarriersBox.kt
```diff
@@ -63,24 +63,33 @@ object NucleusBarriersBox {
         ),
     }
 
-    @HandleEvent
-    fun onAreaChange(event: GraphAreaChangeEvent) {
-        inNucleus = event.area == "Crystal Nucleus"
-    }
```
Thunderblade73: Why do you remove the caching?

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/mining/crystalhollows/NucleusBarriersBox.kt
ksndq: isnt this exactly what IslandAreas.currentAreaName does?

Thunderblade73: yes, but that check here does only happen once, the other check will happen each render frame (as it is called in the render event).

CalMWolfs (approved):

hannibal002:
i change this back, no reason to have two more lines
image

PR #3536, opened 2025-03-11
Backend: AccessoryApi

DavidArthurCole:
## What
Standardization for storing and dealing with accessories. Will read and store accessories per-profile.
Just the base setup, nothing using it yet.

+ Added AccessoryApi. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/accessories/AccessoryApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/accessories/AccessoryApi.kt
@@ -2,8 +2,6 @@ package at.hannibal2.skyhanni.features.inventory.accessories
 
 import at.hannibal2.skyhanni.api.enoughupdates.EnoughUpdatesManager
 import at.hannibal2.skyhanni.api.event.HandleEvent
-import at.hannibal2.skyhanni.config.commands.CommandCategory
-import at.hannibal2.skyhanni.config.commands.CommandRegistrationEvent
 import at.hannibal2.skyhanni.config.storage.ProfileSpecificStorage
 import at.hannibal2.skyhanni.data.ProfileStorageData
 import at.hannibal2.skyhanni.data.jsonobjects.repo.neu.NeuMiscJson
@@ -11,7 +9,6 @@ import at.hannibal2.skyhanni.events.InventoryCloseEvent
 import at.hannibal2.skyhanni.events.InventoryUpdatedEvent
 import at.hannibal2.skyhanni.events.NeuRepositoryReloadEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
-import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.ItemUtils.getInternalNameOrNull
 import at.hannibal2.skyhanni.utils.ItemUtils.getItemRarityOrNull
 import at.hannibal2.skyhanni.utils.ItemUtils.getLore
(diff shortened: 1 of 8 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/accessories/AccessoryApi.kt
```diff
@@ -0,0 +1,247 @@
...
+
+    // Acting as a pseudo-weight for the graph
+    enum class LineageType(private val displayName: String) {
+        PARENT("Parent"), // a greater accessory
+        CHILD("Child"), // a lesser accessory
+        SIBLING("Sibling"), // an accessory of the same tier
+        ;
+
+        override fun toString(): String = displayName
+    }
```
Thunderblade73: I understand the distinction between PARENT and SIBLING but CHILD does not make sense in my mind. Because the only way you would need both CHILD and PARENT is when using a directed graph, which it isn't. Therefore I would conclude that you create two edges for each "real" edge which is nonsensical.
Tip for only creating one: Define source as always the parent and target always as the child. (Or the other way arround, does not matter).

src/main/java/at/hannibal2/skyhanni/features/inventory/accessories/AccessoryApi.kt
```diff
@@ -0,0 +1,247 @@
...
+            }
+
+            family.mapNotNull { internalNameString ->
+                this.getAccessoryOrNull(internalNameString.toInternalName())
+            }.forEach { targetAccessory ->
+                // Register the connection
+                this.addLineageConnection(accessory, targetAccessory, lineageType)
+                // Register the reverse connection
+                val reverseLineageType = if (lineageType == LineageType.CHILD) LineageType.PARENT else lineageType
+                this.addLineageConnection(targetAccessory, accessory, reverseLineageType)
```
Thunderblade73: I see you double link for no reason. For child/parent I already wrote it in the other comment. For Sibling simply sort the source and target by lexicographic order or hash of the InternalName or what ever you feel like that also is accessible.
I wanted to say I have a good example for that kind of behaviour in the grapheditor but its only meh. But the hashCode implementation should get somewhat the point across (but do it at initialization not usage) [Example](https://github.com/hannibal002/SkyHanni/blob/beta/src/main/java/at/hannibal2/skyhanni/test/graph/GraphEditor.kt#L835-L848)

Thunderblade73 (changes requested):
I do not really like the double edging you have in your graph, look at the other comments for more context.

Just a side comment, feel free to ignore: I would like a graph/edge/node interface so we can use the "cool" function with them, but more of a me project and only interesting when we have a 3th case for usage of a graph pattern.

src/main/java/at/hannibal2/skyhanni/features/inventory/accessories/AccessoryApi.kt
```diff
@@ -0,0 +1,280 @@
...
+        val internalName = getInternalNameOrNull() ?: return null
+        val rarity = getItemRarityOrNull() ?: return null
+        val enrichment = SkyblockStat.getValueOrNull(getEnrichment().orEmpty())
+        return ProfileSpecificStorage.StorageAccessory(internalName, rarity, enrichment)
+    }
+
+    fun ItemStack.isAccessory(): Boolean =
+        getInternalNameOrNull()?.let { isAccessory(it, getLore()) } ?: false
+
+    private fun isAccessory(internalName: NeuInternalName, lore: List<String>): Boolean =
```
ItsEmpa: why are you checking the lore of the item to check if its an accessory? shouldnt you use the item category for this?

you just need to check that the category is HATCESSORY or
ACCESSORY

ItsEmpa (changes requested):

DavidArthurCole:
closing this to rebase the branches more sensibly

PR #3540, opened 2025-03-11
Backend: Fix some 1.21 errors

NopoTheGamer:
## What
Fixed some errors (~300) in some places

+ Fixed 1.21 errors using remap and preprocessor. - nopo


hannibal002 (changes requested):
What is the point of adding this compatibility function to only some of the usages we have of itemstack.displayName, entity.name and inventory.name? Rather, add new extended functions so that we only need to run getFormattedTextCompat three times, for those three spots, or not at all.

Otherwise, this will just cause confusion about when to use and when not, and eventually will be used wrongly in the future, again creating more 1.21 errors.

NopoTheGamer:
4.0?

hannibal002:
> 4.0?

image

NopoTheGamer:
> What is the point of adding this compatibility function to only some of the usages we have of itemstack.displayName, entity.name and inventory.name? Rather, add new extended functions so that we only need to run getFormattedTextCompat three times, for those three spots, or not at all.

I removed the compat method from this pr as it isnt needed anymore. right now this pr mostly just changes some nullablity and mappings

hannibal002 (changes requested):
Looks fine overall.
Some of the nullability checks are redundant though and should rather be handled bis shots.txt instead of increasing verbosity.

E.g all `MC.world.getPlayer ?: return` spots. Either the player ist not null there anyway since the function only already has a "in SkyBlock" check earlier (that also includes a "is player not null check already) or we have missed such a spot and instead we wanna cause a npe instead of silently failing code that potentially impacts logic.

Also I think we are already having a Minecraft.getMinecraft wrapper that should get used everywhere instead anyway

NopoTheGamer:
> E.g all `MC.world.getPlayer ?: return` spots. Either the player ist not null there anyway since the function only already has a "in SkyBlock" check earlier (that also includes a "is player not null check already) or we have missed such a spot and instead we wanna cause a npe instead of silently failing code that potentially impacts logic.

I can remove the null stuff for now if you want but this will have to be done at some point in the future

hannibal002:
yes  please remove all the null checks from the pr. we need a different approach to this in another pr

NopoTheGamer:
> yes please remove all the null checks from the pr. we need a different approach to this in another pr

done

CalMWolfs (approved):

PR #3541, opened 2025-03-11
Improvement: Temporal Pillar Dodge

hannibal002:
## depends on
- #1928

image

## Changelog Improvements
+ Added Temporal Pillar Dodge. - hannibal2
    * Tweaked pathfinding in the Rift to avoid scary Endermen.


src/main/java/at/hannibal2/skyhanni/features/rift/RiftApi.kt
```diff
@@ -52,6 +56,30 @@ object RiftApi {
...
 
+    @HandleEvent(onlyOnIsland = IslandType.THE_RIFT)
+    fun onSecondPassed(event: SecondPassedEvent) {
+        if (!config.temporalPillarDodge) {
+            if (IslandGraphs.tempRemoveReason == "Temporal Pillar") {
+                IslandGraphs.resetTempRemove()
+            }
+            return
+        }
+        val temporalPillar = MobData.skyblockMobs.find { it.name == "Temporal Pillar" }
```
Thunderblade73: Also use the mob spawn /despawn to make a list of the temporal Pillars this way we save a good chunk of performance.

Thunderblade73 (changes requested):
I hate the design of the "temporary removal". You already had to do a hack to fix accessing the original graph.
So what I suggest: Add an isEnabled var to the graph node (should not be saved, so no edit to the json serailization should be made). Then to "ignore" those nodes we simply skip them inside of findDijkstraDistances (after visted.add). This way we do not need to recreate the graph over and over and over. Plus we will be able to introduce stuff like the atov condition later on with the same interface. For now "keep" your logic, but instead of a copy simply set the isEnabled.

hannibal002:
Conditionalised is enabled for every node has the big performance downside. I assume our hacky ways to deal with graphs would need to check that Boolean more than once per update tick of the network, not only once per findDijkstraDistances call. But I can investigate if that truly is only once the case and report back where else we currently access that

src/main/java/at/hannibal2/skyhanni/features/rift/RiftApi.kt
hannibal002: Agreed

hannibal002: done

hannibal002: not that yet

src/main/java/at/hannibal2/skyhanni/utils/GraphUtils.kt
```diff
@@ -90,6 +90,7 @@ object GraphUtils {
 
         while (queue.isNotEmpty()) {
             val current = queue.poll()
+            if (!current.enabled) continue
```
Thunderblade73: Also add them to visited, as that skips the neighbors check below for neighboring nodes.

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/rift/RiftApi.kt
hannibal002: now done

src/main/java/at/hannibal2/skyhanni/utils/GraphUtils.kt
hannibal002: done

PR #3555, opened 2025-03-13
Fix: trevor mobname i18n

j10a1n15:
## What
Okay so idk if this is caused by the minecraft language
for some it worked, for some not
its a 50-50

## Changelog Fixes
+ Fixed Trevor Mob detection for non-English languages. - j10a1n15
+ Fixed Trevor Trapper triggering on invisible mobs or while blinded. - j10a1n15


j10a1n15:
Nvm the diff lang mob names are set by hypixel

PR #3574, opened 2025-03-15
Backend: create PlotSwitchEvent

rueblimaster:
## What
I'm sorry for not splitting up this pr into more digestible parts.
but it adds a PlotSwitchEvent along with changes to how the current plot can be fetched

+ Created PlotSwitchEvent. - rueblimaster

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/events/garden/PlotChangeEvent.kt
+++ b/src/main/java/at/hannibal2/skyhanni/events/garden/PlotChangeEvent.kt
@@ -1,0 +1,6 @@
+package at.hannibal2.skyhanni.events.garden
+
+import at.hannibal2.skyhanni.api.event.SkyHanniEvent
+import at.hannibal2.skyhanni.features.garden.GardenPlotApi
+
+class PlotChangeEvent(val previousPlot: GardenPlotApi.Plot?, val newPlot: GardenPlotApi.Plot?) : SkyHanniEvent()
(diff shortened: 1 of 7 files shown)


hannibal002:
is this a event that gets called when the plot you are currently at changes? e.g. you move from one plot to another?

rueblimaster:
> is this a event that gets called when the plot you are currently at changes? e.g. you move from one plot to another?

Exactly, additionally also triggers when you leave the garden

hannibal002:
any idea how to name the event better? currently it soulds to me if the plot gets changed, but the plot stays unchagned. you move around plots rather.

CalMWolfs:
> any idea how to name the event better? currently it soulds to me if the plot gets changed, but the plot stays unchagned. you move around plots rather.

PlotSwitchEvent maybe?

rueblimaster:
> any idea how to name the event better? currently it soulds to me if the plot gets changed, but the plot stays unchagned. you move around plots rather.

Chatgpt likes PlotMoveEvent, tho I think it still sounds as if the plot itself gets moved, how about PlayerPlotMoveEvent or PlayerPlotChangeEvent?

hannibal002:
The plot itself can't switch, but the player can change the plot. I like plot switch event the most

rueblimaster:
@hannibal002 
I'll still leave the event on leaving the garden, even tho it's not really intuitive based on the name itself right?

hannibal002:
> @hannibal002 I'll still leave the event on leaving the garden, even tho it's not really intuitive based on the name itself right?
You should send a event whenever the current plot changes. If this should also trigger on join garden and land in a plot when the old plot is null or when you leave garden when the new plot is null

PR #3602, opened 2025-03-17
Backend: ItemBuyApi createBuyTip() more options

rueblimaster:
## What
Add colors etc as arguments with defaults

I'm just proposing this here, I'm not sure this is what is needed or so. So just tell me.

+ Added options to change the appearance of the BuyTip. - rueblimaster

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/ItemBuyApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/ItemBuyApi.kt
@@ -6,6 +6,7 @@ import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.HypixelCommands
 import at.hannibal2.skyhanni.utils.ItemPriceUtils.isAuctionHouseItem
 import at.hannibal2.skyhanni.utils.ItemUtils.repoItemName
+import at.hannibal2.skyhanni.utils.LorenzColor
 import at.hannibal2.skyhanni.utils.NeuInternalName
 import at.hannibal2.skyhanni.utils.StringUtils.removeColor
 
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/api/ItemBuyApi.kt
```diff
@@ -19,11 +20,19 @@ object ItemBuyApi {
...
-        },
-    )
+    fun NeuInternalName.createBuyTip(
+        colorActive: LorenzColor = LorenzColor.YELLOW,
+        colorInactive: LorenzColor = LorenzColor.RED,
+        clickType: String = "Click",
+    ): List<String> {
+        val firstPart = "§${colorActive.chatColorCode}$clickType to search for $repoItemName §${colorActive.chatColorCode}in"
+
+        return listOf(
```
j10a1n15, line 30: why is this a list

j10a1n15 (changes requested):

src/main/java/at/hannibal2/skyhanni/api/ItemBuyApi.kt
rueblimaster: 1. was already like this
2. that's because tooltips are a list, so then you can just make it tooltip = createBuyTip()

hannibal002:
I dont understand why we might wanna change the format of the strings. the whole point behind that function was (behind less code duplication) a consistent format. we dont want each feature to have a separate color palette. time has to be the same color everywhere, coins as well, numbers as well, stack sizes as well, etc. so for this case active and inactive color should always be the same. why would it be not the case?

rueblimaster:
> I dont understand why we might wanna change the format of the strings. the whole point behind that function was (behind less code duplication) a consistent format. we dont want each feature to have a separate color palette. time has to be the same color everywhere, coins as well, numbers as well, stack sizes as well, etc. so for this case active and inactive color should always be the same. why would it be not the case?

I needed it because I had other stuff in a tooltip already that wasn't formatted like that, but I ended up doing a custom thing anyways, so ig you're right, this isn't needed

PR #3606, opened 2025-03-18
Backend: Extract Renderables out of CollectionUtils

DavidArthurCole:
## What
Removes all traces of `Renderable` list stuff to a new class, and moved all collection related classes into a new package.
There's some other miscellaneous cleanup too, unused functions and such.

+ Moved all `Renderable` list utils to `RenderableCollectionUtils`. - Daveed


src/main/java/at/hannibal2/skyhanni/utils/renderables/RenderableUtils.kt
```diff
@@ -41,13 +41,13 @@ internal object RenderableUtils {
...
     /** Calculates the absolute x position of the columns in a table*/
-    fun calculateTableXOffsets(content: Collection<List<Renderable?>>, xPadding: Int) = run {
+    fun calculateTableXOffsets(content: Collection<Collection<Renderable?>>, xPadding: Int) = run {
         var buffer = 0
         var index = 0
         buildList {
             add(0)
             while (true) {
-                buffer += content.map { it.getOrNull(index) }.takeIf { it.any { it != null } }?.maxOfOrNull {
+                buffer += content.map { it.toList().getOrNull(index) }.takeIf { it.any { it != null } }?.maxOfOrNull {
```
Thunderblade73: This toList call massively hurts performance (if it does a copy, idk if the Compiler does optimize that out or not)

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/collection/RenderableCollectionUtils.kt
Thunderblade73, line 1: Wouldn't that file fit more into the renderables sub folder? (I know that it will be pain to change it, sorry)

DavidArthurCole: I don't think so, as it's more about defining functions for mutating collections _of_ renderables, imo

src/main/java/at/hannibal2/skyhanni/utils/renderables/RenderableUtils.kt
DavidArthurCole: Looking at the bytecode, you're definitely right. Changed it around.

CalMWolfs (approved):
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

PR #3618, opened 2025-03-19
Backend: TitleManagerV2 + More

DavidArthurCole:
## What
Overhauls TitleManager so that it can send titles in different places, extracting `sendTitleInInventory` from the guardian pet reminder. Does some more miscellaneous cleanup, and data handling. 
This pretty quickly grew out of proportion of 'something I could do alongside Exp. Api', so here it is.

+ Overhauled `TitleManager`. - Daveed
+ CollectionUtils' `takeIfNotEmpty()` now returns the input type. - Daveed


CalMWolfs (approved):

src/main/java/at/hannibal2/skyhanni/data/TitleManager.kt
```diff
@@ -3,66 +3,126 @@ package at.hannibal2.skyhanni.data
...
-    private var fontSizeModifier = 4f
+    private val titleLocationQueues: MutableMap<TitleLocation, CollectionUtils.OrderedQueue<TitleData>> = enumMapOf()
+    private val currentTitles: MutableMap<TitleLocation, TitleData?> = enumMapOf()
+
+    private data class TitleData(
+        var text: String = "",
+        var duration: Duration = 0.seconds,
+        var height: Double = 1.8,
+        var fontSize: Float = 4f,
+        val weight: Double = 1.0,
```
hannibal002, line 41: can we remove weight from the class entirely? its not used.

src/main/java/at/hannibal2/skyhanni/data/TitleManager.kt
```diff
@@ -3,66 +3,126 @@ package at.hannibal2.skyhanni.data
...
-    private var display = ""
-    private var endTime = SimpleTimeMark.farPast()
-    private var heightModifier = 1.8
-    private var fontSizeModifier = 4f
+    private val titleLocationQueues: MutableMap<TitleLocation, CollectionUtils.OrderedQueue<TitleData>> = enumMapOf()
+    private val currentTitles: MutableMap<TitleLocation, TitleData?> = enumMapOf()
+
+    private data class TitleData(
+        var text: String = "",
+        var duration: Duration = 0.seconds,
```
hannibal002, line 38: do we need title data to be a data class? if not, we can remove duration from class member and only have it be a parameter in constructor

src/main/java/at/hannibal2/skyhanni/data/TitleManager.kt
```diff
@@ -71,64 +132,95 @@ object TitleManager {
...
-    @HandleEvent
-    fun onTick(event: SkyHanniTickEvent) {
-        if (endTime.isInPast()) {
-            sortTitles()
-            showNextTitle()
-        }
+        GlStateManager.pushMatrix()
+        GlStateManager.translate(0f, -150f, 500f)
+        Renderable.drawInsideRoundedRect(
+            Renderable.string(text, 1.5),
```
hannibal002: why do we render `text` here and not `display`? is this intended?

hannibal002 (changes requested):
please also review my changes

mariethepolarfox:
The value was specifically used for same tick title Sorting in my old implementation so idk why it isnt used anymore here

src/main/java/at/hannibal2/skyhanni/data/TitleManager.kt
DavidArthurCole: duration needs to be persistent, so that if a title is "importantly sent", i.e., with a higher weight, and pushes the current title to the front of the queue, when it's de-queued again, it has that duration still.

src/main/java/at/hannibal2/skyhanni/data/TitleManager.kt
DavidArthurCole: It was meant to be used, but wasn't. Next commit will fix.

DavidArthurCole:
> The value was specifically used for same tick title Sorting in my old implementation so idk why it isnt used anymore here

It still is being sorted, it's just no longer an active process, and happens in the background through the class in CollectionUtils.

PR #3619, opened 2025-03-19
add more more search tags

j10a1n15:
## What
Added some more skyblock xp bar search tags, just so people can find them easier

If this needs a changelog for whatever reason tell me I guess

exclude_from_changelog


hannibal002:
moulconfig should allow a list of search tags in one annotiation instead of having multiple annotations

NopoTheGamer:
> moulconfig should allow a list of search tags in one annotiation instead of having multiple annotations

im pretty sure you just seperate by space

hannibal002:
oh that is horrible, then i cant have multi words search tags

NopoTheGamer:
> oh that is horrible, then i cant have multi words search tags

moulconfig search just works like that you can have multi word search tags

PR #3621, opened 2025-03-19
Fix: Gemstone Slot Error

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1352028522218127483
Gemstones were removed from fishing rods, errors ensued. This doesn't fix them entirely, but it eats them.

## Changelog Fixes
+ Fixed errors when viewing Legacy Fishing Rods with Gemstones. - Daveed


DavidArthurCole:
To be clear: the data we would need to truly solve this doesn't exist, as that's the whole issue, is that the new versions of the items don't have gemstones from repo data, so uh yeah, eating it _is_ the fix.

PR #3622, opened 2025-03-20
Feature: Add ordered waypoints

SuperCl4sh:
## What
Adds Coleweight's (and Soopy's?) ordered waypoints with /shordered or /sho. Should be compatible with Coleweight's routes.

## Changelog New Features
+ Added Coleweight's Ordered Waypoints (/shordered or /sho). - SuperClash

+ Added AutoService to build.gradle.kts. - SuperClash


Jordyrat:
I've been using this, works well.

Only suggestion is for `/shoskip` to default to 1 with no args

CalMWolfs:
@SuperCl4sh bump

src/main/java/at/hannibal2/skyhanni/config/features/mining/orderedwaypoints/OrderedWaypointsConfig.kt
```diff
@@ -0,0 +1,80 @@
...
+class OrderedWaypointsConfig {
+
+    @ConfigOption(name = "Credits", desc = "This feature is from Coleweight and SoopyV2, huge thanks to them!")
+    @ConfigEditorInfoText
+    var notice: String = ""
+
+    @Expose
+    @ConfigOption(name = "Enable Ordered Waypoints", desc = "Enables ordered waypoints.")
+    @ConfigEditorBoolean
+    var enabled: Boolean = true
```
CalMWolfs: should have @FeatureToggle on this

src/main/java/at/hannibal2/skyhanni/config/features/mining/orderedwaypoints/OrderedWaypointsConfig.kt
```diff
@@ -0,0 +1,80 @@
...
+
+    @Expose
+    @ConfigOption(name = "Previous Color", desc = "Color of the previous ordered waypoint.")
+    @ConfigEditorColour
+    var previousWaypointColor: String = "0:153:85:85:255"
+
+    @Expose
+    @ConfigOption(name = "Next Color", desc = "Color of the next ordered waypoint.")
+    @ConfigEditorColour
+    var nextWaypointColor: String = "0:153:255:255:88"
```
CalMWolfs: Would be good if these (and the one or two further down) were initialised as a ChromaColour instead of as a string, theres a few examples of ChromaColour elsewhere in the config and HoppityCallWarningConfig has an example with rgba

src/main/java/at/hannibal2/skyhanni/config/storage/PlayerSpecificStorage.kt
```diff
@@ -80,4 +82,7 @@ class PlayerSpecificStorage {
         @Expose
         var userLuck: Float = 0f
     }
+
+    @Expose
+    var routes: MutableMap<String, Waypoints<SkyhanniWaypoint>> = mutableMapOf()
```
CalMWolfs: i dont see a reason for these to be in player specific, id suggest making a new config file and storing them there.

CalMWolfs (changes requested):
havent looked too thoroughly through the code or tested at all yet, I should be able to do that tomorrow but I noticed these things that should all be changed

CalMWolfs (changes requested):
I do not really ever mine but is the setup route line meant to be 2 blocks above the waypoint

image

I would also like there to be an option to instead of outlining the blocks you can also have it as a filled color instead

Also right now it only shows 1 waypoint ahead of where you are, i would like a slider that allows you to change the "look ahead" amount and probably default that to 2 and have it as a range of 1-5

also not really a problem but when i load the route it starts me at position 1 so it navigates me to position 2. I think best case would be it draws the line to the closest point in the route as you load it

I've also made a commit that that slightly changes around how the commands are registered so that its a bit easier to tab complete through them (especially on 1.21 as it shows usages there as you type)

the rest of it worked pretty well

SuperCl4sh:
> I do not really ever mine but is the setup route line meant to be 2 blocks above the waypoint
> 
> image
> I would also like there to be an option to instead of outlining the blocks you can also have it as a filled color instead
> 
> Also right now it only shows 1 waypoint ahead of where you are, i would like a slider that allows you to change the "look ahead" amount and probably default that to 2 and have it as a range of 1-5
> 
> also not really a problem but when i load the route it starts me at position 1 so it navigates me to position 2. I think best case would be it draws the line to the closest point in the route as you load it
> 
> I've also made a commit that that slightly changes around how the commands are registered so that its a bit easier to tab complete through them (especially on 1.21 as it shows usages there as you type)
> 
> the rest of it worked pretty well

When using setup mode, the line is supposed to be the line of sight for the etherwarp. I've changed it to 1.62 blocks above the current waypoint if the user selects that they will not sneak (using conduit) and 1.54 if they will sneak (using aotv). I'm not sure how this works with 1.21. I've heard it's still 1.54 for sneaking height (Odin fabric uses this value), but I'm not sure.
I've added more config options (addressing the comments and added more customizability) and also changed the starting waypoint index to be whichever the user is closest to when they import the route. I've also done a bit of refactoring (namely for decideWaypoints), so it would be great if you could test it again. I've done some light testing and it worked fine.

CalMWolfs (approved):

PR #3624, opened 2025-03-20
Island Type from repo

hannibal002:
## What
changing island type from enum to list of objects.
this allows to load in new island instances from the repo when the user is having an outdated version and the repo contains newer data.
sadly this doesnt work with the HandleEvent onlyOnIsland logic.

+ Changed island type to be read from repo when unknown. - hannibal2


CalMWolfs:
Closing due to #3625

PR #3625, opened 2025-03-20
Backend: Weird enum island thing

CalMWolfs:
## What
Allowed SimpleTypeAdapters for enums to have a default (fallback) value. 
- if the enum value in an enum file is unknown it instead will use the fallback value. 
- This unknown value is then saved for later when it is being written.

The only potential problem with this approach is when there is 2 unknown values, i.e. 2 new islands release at the same time.
To test you can use the `ae0a01c8dce2b8ea52e325fa6c3cb61844541c58` commit with backwater_bayou data value deleted from the found hoppity eggs, then launch your game, collect an egg then close your game and see that the daata is correct and labelled under backwater_bayou

Replaces #3624 

+ Allowed SimpleTypeAdapters for Enums to have a default (fallback) value. - CalMWolfs


NopoTheGamer (approved):
worked for me

PR #3628, opened 2025-03-20
Fix: bitsapi randomly increasing sometimes

j10a1n15:
## What
not sure if I should remove the data logging just yet

## Changelog Fixes
+ Fixed random duplication of bits in Custom Scoreboard. - j10a1n15


hannibal002:
> not sure if I should remove the data logging just yet

id say never remove it, but just keep the last 5ish entries in the list

PR #3629, opened 2025-03-20
Backend: Block Stuff, openInventoryName()

NopoTheGamer:
## What
Made a block compat which adds blocks to lists, changed getting inventory name to use utils method instead of container, mappings, 1.21

+ Added block compat, changed inventory name getting. - nopo


DavidArthurCole (approved):

src/main/java/at/hannibal2/skyhanni/features/fishing/ThunderSparksHighlight.kt
```diff
@@ -48,8 +49,10 @@ object ThunderSparksHighlight {
             if (spark.isDead) continue
             val sparkLocation = spark.getLorenzVec()
             val block = sparkLocation.getBlockAt()
+            val lavas = mutableListOf<Block>()
```
hannibal002: please cache this, instead of recreating every frame

src/main/java/at/hannibal2/skyhanni/features/mining/OreBlock.kt
```diff
@@ -44,7 +47,7 @@ enum class OreBlock(
     GOLD_ORE(Blocks.gold_ore, { inDwarvenMines || inCrystalHollows }),
     LAPIS_ORE(Blocks.lapis_ore, { inDwarvenMines || inCrystalHollows }),
     REDSTONE_ORE(
-        { it.block.equalsOneOf(Blocks.redstone_ore, Blocks.lit_redstone_ore) },
+        { it.block.equalsOneOf(mutableListOf<Block>().addRedstoneOres()) },
```
hannibal002: please cache this list, instead of recreating it for every block we check against (its a lot, this is a huge diff)

src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
```diff
@@ -242,6 +242,12 @@ object SkyBlockItemModifierUtils {
     //$$ fun ItemStack.getMinecraftId() = Registries.ITEM.getId(item)
     //#endif
 
+    //#if MC < 1.21
+    fun String.isVanillaItem() = Item.itemRegistry.getObject(ResourceLocation(this)) != null
```
hannibal002: do not use public extended functions for generic objects. in this case, we dont want the funciton isVanillaItem as suggestion for every string in skyhanni. please use a normal function instead

hannibal002 (changes requested):

PR #3632, opened 2025-03-21
Fix: cs double slash thing

j10a1n15:
## What
sorry for making it unplayable

## Changelog Fixes
+ Fixed Custom Scoreboard Player amount containing two slashes. - j10a1n15


CalMWolfs (approved):

hannibal002 (approved):
Finally playable again

PR #3633, opened 2025-03-21
Fix: visitor shopping list in sign

rueblimaster:
## What

It didn't work due to it also being rendered with GuiRenderEvent which overwrote it.

## Changelog Fixes
+ Fixed Visitor Shopping List not working inside signs. - rueblimaster


hannibal002 (changes requested):
isnt a better fix to completely block ChestGuiOverlayRenderEvent from render while inside ScreenDrawnEvent? (also maybe dont use GuiRenderEvent in feature classes to begin with?)

rueblimaster:
> isnt a better fix to completely block ChestGuiOverlayRenderEvent from render while inside ScreenDrawnEvent? (also maybe dont use GuiRenderEvent in feature classes to begin with?)

can we do that?
Also I have no idea, I just fixed it...
also why not use GuiRenderEvent?

hannibal002:
>  also why not use GuiRenderEvent?

this is the generic event that gets fired for both cases. but when you want such generic behaviour its prob fine to just use RenderDisplayHelper

rueblimaster:
> > also why not use GuiRenderEvent?
> 
> this is the generic event that gets fired for both cases. but when you want such generic behaviour its prob fine to just use RenderDisplayHelper

I can't figure out what you want me to do now, is it fine like this or what do I need to change exactly?

hannibal002 (approved):
> I can't figure out what you want me to do now, is it fine like this or what do I need to change exactly?

i have no idea

PR #3637, opened 2025-03-22
Backend: Deprecated a lot of GuiRenderUtils

Thunderblade73:
## What
Most of these functions are not in use only in old gui code (VisualWordGui and GuiEditor) which should use Renderables instead.

+ Deprecated a lot inside GuiRenderUtils. - Thunderblade73


CalMWolfs:
@Thunderblade73 are you planning on updating this pr or can I close it

Thunderblade73:
This Pr is dead, was intended to be merged quickly so that you don't have to find solutions for old code in modern, but that part is to late

PR #3639, opened 2025-03-22
Backend + Improvement: Add twilight and shuriken to damage indicator

SuperCl4sh:
## What
Add twilight arrow poison and shuriken to damage indicator. Also fix Mob class not updating health of mobs when hurt sometimes.

<details>
<summary>Images</summary>

image
image

</details>

## Changelog Improvements
+ Added Twilight Arrow Poison and Extremely Real Shuriken name and icon indicators to Damage Indicator. - SuperClash

+ Fixed Mob class not updating mob health properly when hurt. - Thunderblade73
    * Fixed ArmorStand being retained by the Internal Mob Structure; now updated when ArmorStand respawns.


hannibal002 (changes requested):
I would not recognize those icons if I see them in a game and would therefore ask what they are. Maybe write them down per default and then have a compact option to show it as is.
also, please update the changelog to make clear that we now include an icon in the damage indicator when the mob has this status effect (Maye with cooldown? I don't know how this effect works), instead of just stating "adding x to y", what is not precise.

SuperCl4sh:
This is what the no compact mode looks like (now the default option):
image

DavidArthurCole (approved):

src/main/java/at/hannibal2/skyhanni/config/features/combat/damageindicator/DamageIndicatorConfig.kt
```diff
@@ -106,6 +106,25 @@ class DamageIndicatorConfig {
...
+    var twilightIndicator: Boolean = true
+
+    @Expose
+    @ConfigOption(
+        name = "Compact Names",
+        desc = "Use the icons of Extremely Real Shuriken and twilight arrow poison " +
+            "instead of their names for the indicator."
+    )
+    @ConfigEditorBoolean
+    var compactName: Boolean = false
```
hannibal002: this config variable needs a more descriptive name.

src/main/java/at/hannibal2/skyhanni/features/combat/damageindicator/DamageIndicatorManager.kt
```diff
@@ -226,8 +226,30 @@ object DamageIndicatorManager {
             }
             event.drawDynamicText(location, bossName, sizeBossName, -9f, smallestDistanceVew = smallestDistanceVew)
 
+            val icons: MutableList<String> = mutableListOf()
+            if (config.shurikenIndicator && entity.getNameTagWith(3, "§b✯") != null) {
```
hannibal002: any way you can use mob detection instead of those deprecated functions? if you dont find a way, at least cache it so that the getNameTagWith function doesnt get called each frame

src/main/java/at/hannibal2/skyhanni/config/features/combat/damageindicator/DamageIndicatorConfig.kt
```diff
@@ -106,6 +106,25 @@ class DamageIndicatorConfig {
...
+    var shurikenIndicator: Boolean = true
+
+    @Expose
+    @ConfigOption(name = "Twilight Indicator", desc = "Indicate if twilight arrow poison has been used.")
+    @ConfigEditorBoolean
+    var twilightIndicator: Boolean = true
+
+    @Expose
+    @ConfigOption(
+        name = "Compact Names",
```
hannibal002: maybe change the name here, e.g. to status effect info or something?

src/main/java/at/hannibal2/skyhanni/config/features/combat/damageindicator/DamageIndicatorConfig.kt
```diff
@@ -106,6 +106,25 @@ class DamageIndicatorConfig {
...
+        desc = "Use the icons of Extremely Real Shuriken and twilight arrow poison " +
```
hannibal002: twiling arrow poison misses upper case

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/combat/damageindicator/DamageIndicatorManager.kt
SuperCl4sh: I decided to not use mob detection to comply with the existing format. I'm not sure if I can use mob detection without porting the entire function to mob detection which would be a semi-large backend job. I can't just solely port the shuriken and twilight to mob detection because the text drawing for the shuriken and twilight need to be positioned properly relative to the rest of the text to ensure no overlap, which would require knowing exactly where the other text was drawn.

One issue with nametag caching is that the twilight effect only lasts for 20 seconds, so if the player finds an already twilight debuffed mob, they don't know how long it's active for. I could just cache it for 5 seconds, but I'm not sure if this is the ideal approach since it'll wrongly display mobs as being inflicted with the twilight poison for up to 5 seconds occasionally.

hannibal002:
please rerequest a review when you are done with your changes
image

hannibal002 (approved):
image

PR #3641, opened 2025-03-23
Backend: Remove infix functions in RenderUtils

CalMWolfs:
## What
We dont really use them anywhere else and they dont provide any advantage over a regular extension function. Plus makes it easier for me to make it work on 1.21 if i dont have to worry about it

+ Removed infix functions from RenderUtils. - CalMWolfs


lunaynx:
Agreed, always found these a bit strange. Infix functions would make sense if you are implementing a mathematical operator that works on two numbers, but not really for this.

PR #3650, opened 2025-03-24
Fix: Crop Icon Lag

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1094190239532208228/1353828580387459262
8 second operation per tick seems unoptimal

## Changelog Fixes
+ Fixed performance issues with Garden GUIs. - Daveed


hannibal002:
cant we just remove the copy call entirely and make sure that the enum has a copy of the orignal item instead? then we dont need the cache

hannibal002:
highlight doesnt allow that

hannibal002 (approved):

PR #3651, opened 2025-03-25
Backend: Cleanup RenderUtils

CalMWolfs:
## What
Cleaned up `RenderUtils`
- Removed functions that have been in RenderUtils but have been unused for 1-3 years
- Made all rendering functions have to be called off a `SkyHanniRenderWorldEvent`, this ensures they are called at a correct time and will assist in making rendering work on modern versions. If these functions took in `partialTicks `before now they can access that directly from the event
- There were a few duplicate functions so i removed duplicate implementations of each of these

+ Cleaned up RenderUtils. - CalMWolfs
    * Removed unused functions.
    * Rendering functions must now be called from a SkyHanniRenderWorldEvent.
    * Removed duplicate functions.


DavidArthurCole (approved):

hannibal002 (changes requested):
there is something wrong with the mirrior verse lava parkour.
note that this is a funciton that was using a nea function before this pr.
image

CalMWolfs:
> there is something wrong with the mirrior verse lava parkour. note that this is a funciton that was using a nea function before this pr. image

Fixed
image

hannibal002 (changes requested):
THe graph editor no longer renders nodes
image

hannibal002:
The navigation pathfinding no longer renders the goal as block
image

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
```diff
@@ -182,23 +179,6 @@ object RenderUtils {
...
-            val read = generateSequence(0) { it + 1 }.take(16).map { matrixBuffer.get() }.toList()
-
-            val xTranslate = read[12].toInt()
-            val yTranslate = read[13].toInt()
-            val zTranslate = read[14].toInt()
-
-            matrixBuffer.flip()
-
-            Triple(xTranslate, yTranslate, zTranslate)
-        }
```
Thunderblade73, line 200: That's debug I use now and then. So maybe not delete it (but you can simply make it empty for newer minecraft versions.)

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
```diff
@@ -1673,8 +1429,7 @@ object RenderUtils {
         this.draw3DLine(cornerFour, cornerOne, color, lineWidth, depth)
     }
 
-    // TODO nea please merge with 'draw3DLine'
-    fun SkyHanniRenderWorldEvent.draw3DLineNea(
+    fun SkyHanniRenderWorldEvent.draw3DLine(
```
Thunderblade73, line 1454: At least add a comment to use the linedrawer when rendering multiple lines (as this function should be only used when one or two lines are drawn)

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/RenderUtils.kt
hannibal002: we should do the same thing here we did with the "draw line from player eye to location" logic

Thunderblade73: @hannibal002 
What did you mean by that. They change there is the same as here where the LineDrawer was just removed (which btw. was the newer version, that's why I want to readded it later on when I have time with #3398)

hannibal002: i want the function draw 3d line to only be called from within line drawer. if that makes sense.

Thunderblade73: But the change you describe where literally the opposite xD. Also the opposite change that happens here, as the draw3DLine has used the LineDrawer but this removes that.

hannibal002: we talk about lines, but we talk in cirlces

CalMWolfs:
> THe graph editor no longer renders nodes image

Seems all is fixed now
image

hannibal002 (approved):
I dont find any more bugs so i have no more reasons to delay this pr. sad

PR #3652, opened 2025-03-25
Fix: Allow hoppity guesses to guess when close to the egg

Bloxigus:
## What
The previous PR I created was far too strict with how close the particle could be, not allowing for eggs that are too close to be guessed.
This PR changes 1 number

## Changelog Fixes
+ Fixed EggLocator particles being discarded when too close. - bloxigus


NopoTheGamer:
exclude_from_changelog ?

PR #3670, opened 2025-03-27
Backend: Disabled Detekt

hannibal002:
## What
Detekt misbehaves every second day without reason. I keep this PR up as a warning to detekt to stop being mean to us. (and eventually merge it I don't have the nerve to deal with it anymore, with is yesterday)

## Changelog Improvements
+ Make all detekt errors during compile time shut up forever. - hannibal2


detekt/detekt.yml
```diff
@@ -1,162 +1,3 @@
...
-    HasPlatformType: # false positives on config get() methods
-        active: false
-    StorageNeedsExpose:
-        active: true
-    NonStorageDoesntNeedExpose:
-        active: true
-    ImmutableTypesWithExpectedInteriorMutabilityInConfig:
-        active: true
-
+detekt {
```
DavidArthurCole, line 1: ????

hannibal002: if the file is valid yml or not, the pr works by fixing the detetk errors. since this is only bandage and no permament fix, and more a moral arguemtn and a testing stuff i can switch branches off to when gralde cant work anymore, i think its fine to leave it as is

PR #3671, opened 2025-03-27
Backend: Disable Ping packet

hannibal002:
## What
Added option to disable the ping packet api in the dev config since i think i get server kicks bc of it.

+ Added option to disable Hypixel API Ping packet. - hannibal2

### DIFF
--- a/root.gradle.kts
+++ b/root.gradle.kts
@@ -21,7 +21,7 @@ allprojects {
      * Beta version
      * Bugfix version
      */
-    version = "2.5.0"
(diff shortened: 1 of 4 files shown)


root.gradle.kts
```diff
@@ -21,7 +21,7 @@ allprojects {
      * Beta version
      * Bugfix version
      */
-    version = "2.5.0"
+    version = "2.6.0"
```
jani270: so true

PR #3672, opened 2025-03-27
Feature: Rare Sea Creature Party Announcer

Jordyrat:
## What
Adds the option to announce rare sea creatures that you catch into party chat

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added Rare Sea Creature Party Announcer. - Jordyrat
    * Sends a message to party chat when a rare sea creature is caught.


mariethepolarfox:
Maybe add an option to send coordinates in the message as well plus parsing on the other end.

hannibal002:
i like the feature. how about we make the message a bit more personal, lets say "i got a sea_creature_name"?

> Maybe add an option to send coordinates in the message as well plus parsing on the other end.

you are supposed to fish as a group, or in what scenario would this come handy?
btw there is already a generic "send coords of mob nearby of itnerest in chat" feature planned

Jordyrat:
> i like the feature. how about we make the message a bit more personal, lets say "i got a sea_creature_name"?

I was struggling with how to word it, empa wasn't very helpful.
Changed to `I caught a sea_creature_name!` and `I caught a rare sea creature!` (if they have "Creature Name" disabled)

Fazfoxy:
> you are supposed to fish as a group, or in what scenario would this come handy?

If you're fishing in a group and one of the people in your group likes a specific spot so waypoint might help.

mariethepolarfox:
> i like the feature. how about we make the message a bit more personal, lets say "i got a sea_creature_name"?
> 
> 
> 
> > Maybe add an option to send coordinates in the message as well plus parsing on the other end.
> 
> 
> 
> you are supposed to fish as a group, or in what scenario would this come handy?
> 
> btw there is already a generic "send coords of mob nearby of itnerest in chat" feature planned

True, I know the feature is planned but seeing there has been no work done yet (maybe I'll get to this) this might be a decent part time solution

hannibal002 (changes requested):
please include the information what the feature does in the changelog. (simply copy the desc from the config to the changelog sub line)

src/main/java/at/hannibal2/skyhanni/config/features/fishing/RareCatchesConfig.kt
```diff
@@ -23,6 +23,14 @@ class RareCatchesConfig {
     @ConfigEditorBoolean
     var alertOtherCatches: Boolean = false
 
+    @Expose
+    @ConfigOption(
+        name = "Announce to Party",
+        desc = "Send a message to your party when you catch a rare sea creature."
+    )
+    @ConfigEditorBoolean
+    var announceRare: Boolean = false
```
hannibal002: please include in that variable name the info that this is a announce to party chat, otherwise this variable might get confused with "alertOtherCatches"

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureFeatures.kt
```diff
@@ -81,14 +83,20 @@ object SeaCreatureFeatures {
...
-            lastRareCatch = SimpleTimeMark.now()
+            if (config.alertOwnCatches) {
+                val text = if (config.creatureName) "${event.seaCreature.displayName}!"
+                else "${event.seaCreature.rarity.chatColorCode}RARE CATCH!"
+                TitleManager.sendTitle(text, height = 2.8, fontSize = 7f)
+                if (config.playSound) SoundUtils.playBeepSound()
+                lastRareCatch = SimpleTimeMark.now()
+            }
+            if (config.announceRare && PartyApi.partyMembers.size > 0) {
+                if (config.creatureName) HypixelCommands.partyChat("I caught a ${event.seaCreature.name}!")
```
hannibal002: i dont think we should use the creatureName config option in the new feature at all. there is no point in sending a chat message telling that you found "something rare" to others without telling them what you caught as well. the question that comes to mind is "why would you say that in party chat manually?" i dont see the need for that.

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureFeatures.kt
```diff
@@ -81,14 +83,20 @@ object SeaCreatureFeatures {
...
-            if (config.playSound) SoundUtils.playBeepSound()
-            lastRareCatch = SimpleTimeMark.now()
+            if (config.alertOwnCatches) {
+                val text = if (config.creatureName) "${event.seaCreature.displayName}!"
+                else "${event.seaCreature.rarity.chatColorCode}RARE CATCH!"
+                TitleManager.sendTitle(text, height = 2.8, fontSize = 7f)
+                if (config.playSound) SoundUtils.playBeepSound()
+                lastRareCatch = SimpleTimeMark.now()
+            }
+            if (config.announceRare && PartyApi.partyMembers.size > 0) {
```
hannibal002: instead of checking the party member size manually, please create a function in party api "isInParty" and use that instead here

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureFeatures.kt
```diff
@@ -81,14 +83,20 @@ object SeaCreatureFeatures {
 
     @HandleEvent(onlyOnSkyblock = true)
     fun onSeaCreatureFish(event: SeaCreatureFishEvent) {
-        if (!config.alertOwnCatches) return
+        if (!config.alertOwnCatches && !config.announceRare) return
```
hannibal002: you can remove this whole check here entirely

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureFeatures.kt
```diff
@@ -81,14 +83,20 @@ object SeaCreatureFeatures {
...
 
         if (event.seaCreature.rare) {
```
hannibal002: please use a early return here

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureFeatures.kt
hannibal002: Also please add support for dynamic "i caught a/an" depending on the sea creature name. maybe we even have a utils function for this in string utils already not sure.

hannibal002 (changes requested):
code looks fine, not tested in game. just small design changes

PR #3674, opened 2025-03-28
Backend: DVD Logo Renderable

DavidArthurCole:
## What
Adds a DVD Logo renderable that will bounce around the screen.

+ Added DVDLogoRenderable. - Daveed


src/main/java/at/hannibal2/skyhanni/utils/renderables/DVDLogoRenderable.kt
```diff
@@ -0,0 +1,108 @@
...
+        } else velocity
+    }
+
+    private fun generateNextPosition(nextVelocity: LogoVelocity): Position = Position(
+        x = (position.x + nextVelocity.x).coerceIn(leftLimit, rightLimit),
+        y = (position.y + nextVelocity.y).coerceIn(topLimit, bottomLimit)
+    )
+
+    // We don't care about posX and posY since we are using our own position
+    override fun render(posX: Int, posY: Int) = handlerRenderCall()
```
Thunderblade73: posX and posY are for the mouse position. So you should still care. You just have to offset them by your Position before rendering the renderable inside. Also you do not translate at all, so it does nothing besides messing with the mouse detection (which won't work at all). Also you need to translate from relative to absolute coordinate space to make it work correctly, as you set your limits in absolute space not relative.

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/renderables/DVDLogoRenderable.kt
```diff
@@ -0,0 +1,121 @@
...
+
+        val posXAtEdge = posXAtLeftEdge || posXAtRightEdge
+        val posYAtEdge = posYAtTopEdge || posYAtBottomEdge
+
+        velocity = generateNextVelocity(posXAtEdge, posXAtLeftEdge, posYAtEdge, posYAtTopEdge)
+        position = generateNextPosition()
+
+        GlStateManager.pushMatrix()
+        GlStateManager.translate(position.x.toFloat(), position.y.toFloat(), 0f)
+        renderable.render(posX, posY)
```
Thunderblade73:
```suggestion
        renderable.render(posX + position.x, posY + position.y)
```

mariethepolarfox:
Like I said in contrib chat please make the text customizable if you let the user enable this feature

Thunderblade73:
> Like I said in contrib chat please make the text customizable if you let the user enable this feature

That's just the technical part. There is no feature here. And the renderable supports any kind of renderable as content, so we can render anything we want to be a DVDLogo.

hannibal002:
this is only a backend change, no user facing feature at all

mariethepolarfox:
Oh I didnt look at the actual location of the feature toggle

hannibal002:
Features toggle? I have not looked at the code closely. I just read the chnagelog

src/main/java/at/hannibal2/skyhanni/config/features/dev/DebugConfig.kt
```diff
@@ -191,6 +191,11 @@ class DebugConfig {
     @ConfigEditorBoolean
     var forceGreatSpook: Property<Boolean> = Property.of(false)
 
+    @Expose
+    @ConfigOption(name = "DVD Logo", desc = "Enable the test DVD Logo Renderable")
+    @ConfigEditorBoolean
+    var dvdLogo: Boolean = false
```
mariethepolarfox, line 197: Here. But this is under the debugConfig so nothing a normal user is concerned about

DavidArthurCole: There's no `@FeatureToggle` annotation, which is I think the implication hanni got. This isn't something a user is gonna accidentally turn on.

PR #3676, opened 2025-03-29
Fix: `KeyDownEvent` Breaking Garden Warp Keys

DavidArthurCole:
## What
Fixes the fact that after #3666, you had to mouse jiggle before garden warp keys would work.

## Changelog Fixes
+ Fixed Garden Warp keybinds. - Daveed
    * Addressed a long-standing issue where keybinds sometimes failed.
    * Bug fixed here was a more exaggerated case of keybind failure.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -30,6 +30,11 @@
 + Added Timite Tracker for the Rift. - nopo (https://github.com/hannibal002/SkyHanni/pull/3102)
     + Shows collected ores, time gained from mining, profit for raw ores, and Highlite profit.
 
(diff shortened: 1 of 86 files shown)


hannibal002 (changes requested):
please let me know how to reproduce this bug exactly.

DavidArthurCole:
> please let me know how to reproduce this bug exactly.

I feel like this is the 4th time I've stated it but for completness sake:

- Trigger any mouse input event (Left/Right/Middle click or MB4/5)
- Try to use garden keybinds
- They don't work until mouse is moved

PR #3675, opened 2025-03-29
Backend: Begin preparing for modern rendering

CalMWolfs:
## Dependencies
- #3651 

## What
Started changes for modern rendering. This pr creates 3 classes, `DrawContext`, `WorldRenderContext` and `MatrixStack` which are used for rendering stuff. Modern versions do not allow calls such as `GlStateManager.popMatrix()` or `GlStateManager.translate()` and instead must use the `MatrixStack` that is stored in the respective contexts. This pr adds these contexts to the events that will need it and the context will be used in a future pr.

Merging up to this pr gets down to
image

## Changelog Technical Details
+ Began preparing for modern rendering. - CalMWolfs
    * Introduced Contexts and MatrixStack, resembling modern rendering.

PR #3677, opened 2025-03-29
Backend: Expand the use of SkyhanniBaseScreen

CalMWolfs:
## Dependencies
- #3651 
- #3675 

## What
Made all relevant places now extend SkyHanniBaseScreen instead of GuiScreen as on modern versions there are slightly differently named methods with different values.
This also converts useage of GlStateManager to use the context from the draw screen function instead. The old functions were made to be final so that you cannot accidentaly use the wrong ones.

Merging up to this pr gets down to
image

+ Expanded the use of SkyhanniBaseScreen. - CalMWolfs


hannibal002 (approved):
i dont see any issues

PR #3680, opened 2025-03-29
Backend: Use DrawContext for Positions part 1

CalMWolfs:
## Dependencies
- #3651 
- #3675 
- #3677 
- #3678
- #3679
- #3775 

## What
I lied, there was too many lines to get to Renderables this pr.
This pr uses `DrawContext` for `Positions`

+ Used DrawContext for when rendering strings and renderables on the screen. - CalMWolfs


CalMWolfs:
closing as now that we changed how this works in #3775 this pr is now only 40 lines instead of 250 and the next one is only 140 lines from beta instead of 550, so now #3685 can have these changes and its own

PR #3683, opened 2025-03-29
Feature + Backend: Disable Profile Viewer in Kuudra

ksndq:
## What
[Thread](https://discord.com/channels/997079228510117908/1355502711214440580)

Uses the exact same logic as the Block Interacting with Visitors option (now changed to use EntityClickEvent) :https://github.com/hannibal002/SkyHanni/blob/6fb20364b5f28ead0710798cc3d3ca7b63d46ce0/src/main/java/at/hannibal2/skyhanni/features/garden/visitor/HighlightVisitorsOutsideOfGarden.kt#L92-L100

Doesn't mess up shooting the bow so personally I think it is fine.

## Changelog New Features
+ Added option to prevent player interactions during the Kuudra boss fight. - ksndq
    * Added option to disable Profile Viewer on right-click.
    
+ Changed Block Interacting with Visitors to use `EntityClickEvent`. - ksndq
+ Changed `isVisitor` to use `IslandType`. - ksndq


src/main/java/at/hannibal2/skyhanni/features/nether/kuudra/KuudraProfileViewerBlocker.kt
```diff
@@ -0,0 +1,29 @@
...
+import net.minecraft.entity.player.EntityPlayer
+import net.minecraft.network.play.client.C02PacketUseEntity
+
+@SkyHanniModule
+object KuudraProfileViewerBlocker {
+
+    private val config get() = SkyHanniMod.feature.crimsonIsle
+
+    @HandleEvent(onlyOnIsland = IslandType.KUUDRA_ARENA)
+    fun onClickEntity(event: PacketSentEvent) {
```
CalMWolfs: should use EntityClickEvent instead, feel free to update the existing related feature to also use the proper event

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/HighlightVisitorsOutsideOfGarden.kt
```diff
@@ -89,21 +89,18 @@ object HighlightVisitorsOutsideOfGarden {
...
-            if (packet.action == C02PacketUseEntity.Action.INTERACT) {
+            if (event.clickType == ClickType.RIGHT_CLICK) {
                 ChatUtils.chatAndOpenConfig(
                     "Blocked you from interacting with a visitor. Sneak to bypass or click here to change settings.",
                     GardenApi.config.visitors::blockInteracting,
                 )
             }
+            event.cancel()
         }
     }
```
ksndq: for some reason this gets called twice each click?

hannibal002: should be fixed with
https://github.com/hannibal002/SkyHanni/commit/10841ed65270a6fc1c37dbafa153cede1449259c#diff-40ca80d2404a4150c354e21581dfdbb2c6aebb1a9a2de14fa9f9091989d2d8dd

src/main/java/at/hannibal2/skyhanni/config/features/crimsonisle/CrimsonIsleConfig.kt
```diff
@@ -25,6 +25,15 @@ class CrimsonIsleConfig {
...
     var matriarchHelper: MatriarchHelperConfig = MatriarchHelperConfig()
 
+    @Expose
+    @ConfigOption(
+        name = "Disable Profile Viewer in Kuudra",
+        desc = "Prevent player interactions during the Kuudra boss fight to stop Profile Viewer from opening."
+    )
+    @ConfigEditorBoolean
+    @FeatureToggle
+    var kuudraProfileViewer: Boolean = false
```
hannibal002: give the variable a name to make sure its clear this means the pv is disabled

src/main/java/at/hannibal2/skyhanni/features/nether/kuudra/KuudraProfileViewerBlocker.kt
```diff
@@ -0,0 +1,25 @@
...
+@SkyHanniModule
+object KuudraProfileViewerBlocker {
+
+    private val config get() = SkyHanniMod.feature.crimsonIsle
+
+    @HandleEvent(onlyOnIsland = IslandType.KUUDRA_ARENA)
+    fun onClickEntity(event: EntityClickEvent) {
+        if (!config.kuudraProfileViewer) return
+
+        if (event.clickedEntity !is EntityPlayer) return
```
hannibal002: make sure they are no npc. we still wanna be able to interact with elle (or whatever other npcthere might be, idk i dont play kuudra)

hannibal002 (changes requested):

hannibal002:
@ksndq bump

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/HighlightVisitorsOutsideOfGarden.kt
ksndq: https://github.com/hannibal002/SkyHanni/blob/07dd6c1bc6e815beb13c5408e5c75d38adc6878a/src/main/java/at/hannibal2/skyhanni/data/ItemClickData.kt#L48C1-L57C14

Here for each right click both INTERACT_AT and INTERACT get called at once. INTERACT_AT is turned into INTERACT so there is no way to differentiate between them later on and the event just runs twice with INTERACT action. I don't want to break existing code but can't we just keep the INTERACT_AT as it is or just return on it since I don't think its useful at all.

ksndq: @hannibal002

hannibal002: there was a very specific difference between those two. but i forgot what it was.
feel free to play around. if we break other features we know immediately and we can easily fix

hannibal002:
@ksndq is this ready for review? 
image

ksndq:
It's not, unfortunately. The interact packet behavior is unexpected. I need more time to test some more solutions. Im pretty sure we are on the same timezone so I should be done today.

src/main/java/at/hannibal2/skyhanni/data/ItemClickData.kt
```diff
@@ -49,8 +49,10 @@ object ItemClickData {
                 val clickType = when (packet.action) {
                     C02PacketUseEntity.Action.INTERACT -> ClickType.RIGHT_CLICK
                     C02PacketUseEntity.Action.ATTACK -> ClickType.LEFT_CLICK
-                    C02PacketUseEntity.Action.INTERACT_AT -> ClickType.RIGHT_CLICK
-                    else -> return
+                    else -> {
+                        event.cancel()
+                        return
+                    }
```
CalMWolfs: This cannot be correct

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/data/ItemClickData.kt
ksndq: I think it is right since both INTERACT and INTERACT_AT packets get called for entities that use it and INTERACT_AT is useless but I think i found a better solution anyway, is it possible to just add the action type to EntityClickEvent?

hannibal002: Yeah please dont change the vanilla packet behaviour this way without reason.

what is your better solution?

hannibal002 (changes requested):
see above

src/main/java/at/hannibal2/skyhanni/data/ItemClickData.kt
ksndq: `add the action type to EntityClickEvent`

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/HighlightVisitorsOutsideOfGarden.kt
```diff
@@ -88,21 +86,18 @@ object HighlightVisitorsOutsideOfGarden {
...
-        val player = MinecraftCompat.localPlayerOrNull ?: return
-        if (player.isSneaking) return
-        val packet = event.packet as? C02PacketUseEntity ?: return
-        val entity = packet.getEntityFromWorld(world) ?: return
+        if (MinecraftCompat.localPlayer.isSneaking) return
+        val entity = event.clickedEntity ?: return
         if (isVisitor(entity) || (entity is EntityArmorStand && isVisitorNearby(entity.getLorenzVec()))) {
-            event.cancel()
-            if (packet.action == C02PacketUseEntity.Action.INTERACT) {
+            if (event.action != Action.INTERACT_AT) {
```
hannibal002, line 94: why got this changed from interact to interact at? cant we remove action from the event and use right click, just like we do in kuudra profile viewer blocker?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/HighlightVisitorsOutsideOfGarden.kt
ksndq: {913C817B-097C-411E-875A-C8347E8C1C2F}
If I don't filter out one of those packets the "Blocked you from interacting with a visitor..." message gets sent twice but I also want the message when left clicking the NPC (It didn't do that before but I think it should)

PR #3687, opened 2025-03-30
Backend: Further GuiRenderUtils modifications

CalMWolfs:
## Dependencies
- #3675 
- #3677 
- #3678
- #3679
- #3776

## What
Converts all rendering of strings and renderables to use drawContext
Create and use even more methods in GuiRenderUtils that work on modern such as drawing texture rectangles, rendering items stacks and gradient rects.
Hopefully this is the last pr related to 2d rendering

Merging up to this pr gets down to
image

+ Used `DrawContext` for rendering strings and renderables. - CalMWolfs
+ Made `Renderables` use `DrawContext`. - CalMWolfs
+ Added more methods in `GuiRenderUtils`. - CalMWolfs & nopo


src/main/java/at/hannibal2/skyhanni/data/RenderData.kt
```diff
@@ -25,7 +25,7 @@ object RenderData {
         if (GuiEditManager.isInGui() || VisualWordGui.isInGui()) return
 
         DrawContextUtils.translate(0f, 0f, -3f)
-        renderOverlay(event.context)
+        renderOverlay(DrawContextUtils.drawContext)
```
hannibal002, line 28: the `context` member of `RenderingSkyHanniEvent` is now useless. can we remove it?

hannibal002 (changes requested):
im a bit confused about the `context` parameter in the events. i thought we dont need them anymore in there, you replace them in some instances, but in other instances you add them, in the same pr. whats your goal here

can we remove those lines in SkyHanniEvent now?
    interface Rendering {
        val context: DrawContext
    }

src/main/java/at/hannibal2/skyhanni/data/RenderData.kt
CalMWolfs: While it seems to be unused, it is overriding the context val of the `Rendering` interface so therefore it is needed to stay, otherwise half the rendering events wont work on modern

CalMWolfs:
> im a bit confused about the `context` parameter in the events. i thought we dont need them anymore in there, you replace them in some instances, but in other instances you add them, in the same pr. whats your goal here
> 
> can we remove those lines in SkyHanniEvent now? interface Rendering { val context: DrawContext }

While yes we dont access the context from the events anymore and instead use the global context we still need to set the global context which is done in `SkyHanniEvent#prePost` and this still requires the context value.

The reason that ive replaced `event.context` is that its never supposed to be accessed like that and so instead I am using the proper way that we've defined to access the context

src/main/java/at/hannibal2/skyhanni/utils/GuiRenderUtils.kt
```diff
@@ -250,36 +330,25 @@ object GuiRenderUtils {
...
-                lastWarn = SimpleTimeMark.now()
-                println(" ")
-                println("item: $item")
-                println("name: ${item.displayName}")
-                println("getInternalNameOrNull: ${item.getInternalNameOrNull()}")
-                println(" ")
-                ChatUtils.debug("rendering an item has failed.")
-            }
-        }
+        Minecraft.getMinecraft().renderItem.renderItemIntoGUI(item, 0, 0)
```
hannibal002: removing this try catch check here can have horrible problems for everyone.

hannibal002: I do remember seeing this error months ago last time.
So there is a high chance it is fixed by now anyway.
Still, if this is not the case for some users, removing this will cause immense performance issues for the users.

I just wanted to let you know. If you think this is still ok to remove, then it's fine for me.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/GuiRenderUtils.kt
CalMWolfs: I personally havent seen it but ill add it to error manager instead as no one would see the debug errors that it was before. This way we can see if we can fix it or if no one reports it then we can remove the try catch

PR #3689, opened 2025-03-30
Fix: Erroring from addCrimsonPrestige

ksndq:
## Changelog Fixes
+ Fixed an issue where the Estimated Item Value was not displayed for Kuudra Armors when using an outdated repository. - ksndq

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/items/EstimatedItemValueCalculator.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/items/EstimatedItemValueCalculator.kt
@@ -507,6 +507,8 @@ object EstimatedItemValueCalculator {
                         "Could not find crimson prestige cost for ${stack.displayName}",
                         "EstimatedItemValue has no crimsonPrestigeCosts for $tierName tier",
                         "internalName" to internalName,
+                        "tierIndex" to tierIndex,
+                        "armorTier" to armorTier,
                         "tierName" to tierName,
                         "crimsonPrestigeCosts" to EstimatedItemValue.crimsonPrestigeCosts,
                     )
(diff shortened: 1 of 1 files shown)


NopoTheGamer:
why is this excluded

mariethepolarfox:
Maybe because this only happens if the repo is outdated?

jani270:
> Maybe because this only happens if the repo is outdated?

That doesnt make it less of a user facing error?

mariethepolarfox:
But I think you can include this in chnagelog

mariethepolarfox:
I also think that this should get custom error handling instead of just returning 0

hannibal002 (changes requested):
please dont use `exclude for changelog` lightly.
this is clearly a user facing error 
for more infos see this: https://github.com/hannibal002/SkyHanni/blob/43291a718945aa9038718d66f38d4ee12e86a41b/CONTRIBUTING.md#no-category

also please dont hide the error but use error manager to show it in chat to the user without interrupting estimated item value

hannibal002:
please rerequest a review when you are done

src/main/java/at/hannibal2/skyhanni/features/misc/items/EstimatedItemValueCalculator.kt
```diff
@@ -502,7 +502,7 @@ object EstimatedItemValueCalculator {
 
         val allTiersCost = (0 until tierIndex).mapNotNull { index ->
             kuudraTiers.getOrNull(index)?.let { tierName ->
-                EstimatedItemValue.crimsonPrestigeCosts[tierName]
+                EstimatedItemValue.crimsonPrestigeCosts[tierName] ?: return 0.0
```
hannibal002: Please use `ErrorManager.logErrorStateWithData()` additionally to returning 0 here

hannibal002 (changes requested):

hannibal002:
bump @ksndq

ksndq:
will work on it tomorrow been pretty busy

PR #3692, opened 2025-03-30
Fix: Title Width

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/997079229302837269/1355945531104432439
https://discord.com/channels/997079228510117908/997079229302837269/1355921425176990016

Title width was a little messed up. Also there was no reason to be using wrapped strings.
Also fixes issues with titles in inventory not stopping when they should.

<details>
<summary>Images</summary>

image
image

</details>

## Changelog Fixes
+ Fixed titles showing up on multiple lines. - Daveed
+ Fixed titles in inventories overstaying their welcome. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/hypixelapi/HypixelEventApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/hypixelapi/HypixelEventApi.kt
@@ -4,8 +4,6 @@ import at.hannibal2.skyhanni.events.hypixel.modapi.HypixelApiJoinEvent
 import at.hannibal2.skyhanni.events.hypixel.modapi.HypixelApiServerChangeEvent
 import at.hannibal2.skyhanni.features.misc.CurrentPing
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
-import at.hannibal2.skyhanni.utils.ChatUtils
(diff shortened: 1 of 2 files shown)


Thunderblade73:
Wrapped string allows for longer titles to be wrapped. So you removed a feature. Also the width is infinite now as Renderable.string has no limit on it and your scaling logic does nto scale down when it is to long.

CalMWolfs:
> Wrapped string allows for longer titles to be wrapped. So you removed a feature. Also the width is infinite now as Renderable.string has no limit on it and your scaling logic does nto scale down when it is to long.

I dont think titles should ever be wrapped, also from testing this pr longer strings definitely have their font size scaled down

DavidArthurCole:
Adding onto the "titles should never be wrapped" point; TitleManagerV2 added support for subtitles in `TitleManager.sendTitle()` - if something is long enough that it needs to get wrapped, some portion of that should probably be a subtitle, the way I did for the CF opening blockers.

CalMWolfs (approved):

hannibal002 (approved):
image
`/shsendtitle 5s 10 10 hi this is a test message`
for the future (outside this pr) we should prob still scale down if the text is longer than the screen with. (ignore the dvd logo floating around)

PR #3693, opened 2025-03-30
Fix: RenderUtils boxes again

CalMWolfs:
## What
Fixed some waypoints rendering in the wrong position.
This is why there never should have been 3 functions that have the same purpose that all had different default arguments. Not impressed

## Changelog Fixes
+ Fixed some waypoints rendering in the wrong position. - CalMWolfs
    * Waypoints for Hoppity, Diana and Clicked Dungeon Blocks.


DavidArthurCole:
> 3 functions that have the same purpose that all had different default arguments

I'm naming all my function params some combination of lowercase `L`s and uppercase `i`s moving forward, and will never provide them in the same order 🥺

PR #3707, opened 2025-04-02
Backend: exact item source

rueblimaster:
#3716

## What

Tell me what you think about this please

+ Added exact item source. - rueblimaster


src/main/java/at/hannibal2/skyhanni/api/CollectionApi.kt
```diff
@@ -101,7 +101,7 @@ object CollectionApi {
 
     @HandleEvent
     fun onItemAdd(event: ItemAddEvent) {
-        if (event.source == ItemAddManager.Source.COMMAND) return
+        if (event.source == ItemSources.Source.COMMAND) return
```
hannibal002: no i dont like the name used here. either ItemSource.COMMAND or ItemManager.Source.COMMAND.

hannibal002 (changes requested):

rueblimaster:
this probably isn't feasible without a lot of effort and also currently not necessary

PR #3708, opened 2025-04-02
Fix: Crop Icons Round Two

DavidArthurCole:
## What
Guess it doesn't help to implement a caching system when the main thing that's doing enchanting doesn't use said caching system. Anyways the old system was cringe and bad, I replaced it with something much simpler.

(This was an extracted fix from #3703)

## Changelog Fixes
+ Fixed crop icons being incorrectly enchanted in some GUIs. - Daveed


hannibal002 (changes requested):
`getItemStackCopy` is already crop specific. Currently the information crop name is redundant in the `iconId`.

But how instead of simply removing the crop name in the `iconId`, we could rather remove the `map of time limit sets caches`, and instead just use one cache for all crop types. This way, the crop name in the `iconId` has a use again, and we create less redundant instances of `time limited caches`.

hannibal002 (approved):

PR #3716, opened 2025-04-03
Backend: create InventoryChangeEvent

rueblimaster:
+ Create InventoryChangeEvent. - rueblimaster

### DIFF
--- a/.live-plugins/deffunjumpr/plugin.kts
+++ b/.live-plugins/deffunjumpr/plugin.kts
@@ -1,0 +1,106 @@
+import com.intellij.codeInspection.LocalQuickFix
+import com.intellij.codeInspection.ProblemDescriptor
+import com.intellij.codeInspection.ProblemHighlightType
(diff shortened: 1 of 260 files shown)


hannibal002 (changes requested):
OwnInventoryItemUpdateEvent already exists. please merge the logic in here.

CalMWolfs (changes requested):
As hannibal said, we already have a very similar event, also i see no point in the event containing primitive item stacks instead of normal ones, you even have a todo to change this so imo this pr shouldve been opened as a draft

rueblimaster:
> As hannibal said, we already have a very similar event, also i see no point in the event containing primitive item stacks instead of normal ones, you even have a todo to change this so imo this pr shouldve been opened as a draft

the reason for the primitive item stacks is that I didn't manage to make itemStacks a deep copy, so it got mutated concurrently
Do you have an idea as to what would fix this?

PR #3724, opened 2025-04-05
Backend: Default Function "Jump to Event Declaration"

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/3722

## What
Was hoping to have some more time to test this, but with the whoopsied default function name PR merged earlier, this adds the context option to jump to the event declaration that is associated with the function:

image

+ Added context option on default function names for navigating to event declaration. - Daveed

### DIFF
--- a/.live-plugins/deffunjumpr/plugin.kts
+++ b/.live-plugins/deffunjumpr/plugin.kts
@@ -1,0 +1,106 @@
+import com.intellij.codeInspection.LocalQuickFix
+import com.intellij.codeInspection.ProblemDescriptor
+import com.intellij.codeInspection.ProblemHighlightType
+import com.intellij.openapi.project.Project
+import com.intellij.psi.JavaPsiFacade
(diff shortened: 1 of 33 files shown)


hannibal002 (changes requested):
the Alt+Space suggestion does work as intended.
image

But the Strg+B keybind does not work
image
see this:
image

hannibal002:
image

also this could be added to work with all events
image

image

DavidArthurCole:
1) It isn't possible to change the functionality of an IntelliJ keybind with live plugins. That would require a real plugin, which I am not doing nor do I think we'd expect people to install a separate one

2) The reason we're adding the jump points to the functions is so there is some indication of what event its receiving. I see no value in taking what is already a linked function, which has its event specified:

```@HandleEvent
fun onDebug(event: DebugDataCollectEvent)
```

And trying to add a secondary link there.

PR #3607, opened 2025-03-18
Backend: Remove Deprecated Function `addAsSingletonList`

DavidArthurCole:
## Dependencies
- #3606
- #3069
- #3307

## What
Removes the long-deprecated `addAsSingletonList`, which also got rid of a lot of the `List<List<Any>>` methods we had, so 🥳

Rendering classes that were actually changed during this cleanup:
- CityProjectFeatures
- ComposterDisplay
- ComposterOverlay
- JacobContestFFNeededDisplay
- JacobContestTimeNeeded
- CropMoneyDisplay
- GardenBestCropTime
- GardenVisitorDropStatistics
- GardenVisitorFeatures
- TrevorTracker
- LivingMetalSuitProgress
- CruxTalismanDisplay
- ShowMotesNpcSellPrice
- CollectionTracker
- QuickModMenuSwitch

<details>
<summary>Images</summary>

Composter Overlay + Display
image
image

Jacob Contest (both of them)
image

Visitor Drop stats:
image

GardenVisitorFeatures
image

Crop money/hr
image

Trevor Tracker
image

Living Metal Suit
image

Crux Talisman
image

ShowMotesNpcSellPrice
image

image

CollectionTracker (for as buggy as it is...)
image

QuickModMenuSwitch
image

CityProjectFeatures
image

GardenBestCropTime
image

</details>

## Changelog Removed Features
+ Removed `/shshowcropmoneycalculation`. - appable
    * Use `/shdebug Crop Money` instead.

## Changelog Technical Details
+ Removed the `addAsSingletonList` function in favor of `Renderable`s. - Daveed

PR #3731, opened 2025-04-05
Backend: remove unused RenderUtils functions

CalMWolfs:
## What
Follow up of #3607, remove these 2 unneeded functions
excluding as this shouldve been done in that pr and matches those notes

exclude_from_changelog


DavidArthurCole (approved):
oops

PR #3733, opened 2025-04-06
Fix: area detection in glacite tunnels

hannibal002:
## What
We need to use the scoreboard information here.

## Changelog Fixes
+ Fixed Mineshaft Pity Display not appearing. - hannibal2


lunaynx:
Tunnel Maps was also broken for me, I guess this fixes that too.

PR #3734, opened 2025-04-06
Feature: Dungeon Low Health Alert

jani270:
## What
Shows a title and plays a sound when a mate has low health in dungeons (Maybe some healer finally learn how to press their drop key)

Maybe this could get a detection if the ult is ready at a later point but eh. 

Surely this code is fine and doesnt have any detekt issues 
grafik

## Changelog New Features
+ Added Dungeon Low Health Alert. - jani
    * Shows a title and plays a sound when a teammate's health is low.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/dungeon/LowHealthAlert.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/dungeon/LowHealthAlert.kt
@@ -1,0 +1,53 @@
+package at.hannibal2.skyhanni.features.dungeon
+
+import at.hannibal2.skyhanni.SkyHanniMod
+import at.hannibal2.skyhanni.api.event.HandleEvent
+import at.hannibal2.skyhanni.data.IslandType
+import at.hannibal2.skyhanni.data.TitleManager
+import at.hannibal2.skyhanni.events.ScoreboardUpdateEvent
+import at.hannibal2.skyhanni.features.gui.customscoreboard.ScoreboardPattern
+import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
+import at.hannibal2.skyhanni.utils.RegexUtils.matchMatcher
+import at.hannibal2.skyhanni.utils.SimpleTimeMark
+import at.hannibal2.skyhanni.utils.SoundUtils
+import at.hannibal2.skyhanni.utils.SoundUtils.playSound
+import kotlin.time.Duration.Companion.seconds
+
+@SkyHanniModule
+object LowHealthAlert {
+
+    private val config get() = SkyHanniMod.feature.dungeon.lowHealthAlert
+    private val soundConfig get() = config.lowHealthAlertSoundConfig
+    private var lastAlert = SimpleTimeMark.farPast()
+
+    @HandleEvent(onlyOnIsland = IslandType.CATACOMBS)
+    fun onScoreboardChange(event: ScoreboardUpdateEvent) {
+        for (line in event.added.map { it }) {
+            ScoreboardPattern.teammatesPattern.matchMatcher(line) {
+                val username = group("username")
+                val color = group("color")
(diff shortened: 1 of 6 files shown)


jani270:
> ### 3 Detekt Failures were detected:
> 
>     * [ScoreboardPattern.kt#L201](https://github.com/hannibal002/SkyHanni/blob/6a73209f955f2edbc541854803906d2bde0d1a8c/src/main/java/at/hannibal2/skyhanni/features/gui/customscoreboard/ScoreboardPattern.kt#L201): Repo pattern `teammatesPattern` must have a regex test.
> 
>     * [LowHealthAlert.kt#L38](https://github.com/hannibal002/SkyHanni/blob/6a73209f955f2edbc541854803906d2bde0d1a8c/src/main/java/at/hannibal2/skyhanni/features/dungeon/LowHealthAlert.kt#L38): Redundant curly braces
> 
>     * [LowHealthAlert.kt#L25](https://github.com/hannibal002/SkyHanni/blob/6a73209f955f2edbc541854803906d2bde0d1a8c/src/main/java/at/hannibal2/skyhanni/features/dungeon/LowHealthAlert.kt#L25): This 'map' call can be removed.

This is so sad

src/main/java/at/hannibal2/skyhanni/features/dungeon/LowHealthAlert.kt
```diff
@@ -0,0 +1,53 @@
...
+    @HandleEvent(onlyOnIsland = IslandType.CATACOMBS)
+    fun onScoreboardChange(event: ScoreboardUpdateEvent) {
+        for (line in event.added) {
+            ScoreboardPattern.teammatesPattern.matchMatcher(line) {
+                val username = group("username")
+                val color = group("color")
+                val health = group("health")
+                if (color == "c" && health != null && health != "DEAD") {
+                    val isHealer = DungeonApi.playerClass == DungeonApi.DungeonClass.HEALER
+                    val shouldAlert = isEnabled() && (!config.onlyWhileHealer || isHealer) && lastAlert.passedSince() > 1.5.seconds
```
CalMWolfs: could probably move all this to the `isEnabled` function and then at the start of this event early return if `!isEnabled`

Then when sending the title just return out of the function so it doesnt send more than 1 title/sound in the same event

src/main/java/at/hannibal2/skyhanni/features/dungeon/LowHealthAlert.kt
```diff
@@ -0,0 +1,53 @@
...
+object LowHealthAlert {
+
+    private val config get() = SkyHanniMod.feature.dungeon.lowHealthAlert
+    private val soundConfig get() = config.lowHealthAlertSoundConfig
+    private var lastAlert = SimpleTimeMark.farPast()
+
+    @HandleEvent(onlyOnIsland = IslandType.CATACOMBS)
+    fun onScoreboardChange(event: ScoreboardUpdateEvent) {
+        for (line in event.added) {
+            ScoreboardPattern.teammatesPattern.matchMatcher(line) {
```
CalMWolfs: I think this can be replaced with `ScoreboardPattern.teammatesPattern.matchAll(event.added)`

CalMWolfs (changes requested):

jani270:
@DavidArthurCole Could you check if this a correct usage of the TitleManager?

DavidArthurCole (approved):
LGTM.
The only note I'll add here:

```
            lastAlert?.stop()
            TitleManager.sendTitle("§c$username §ais low", "§c$health❤", 1.seconds)?.let {
                lastAlert = it
            }
```

Calling `.stop()` will forcibly remove the title from the queue without verification it has been shown to the user yet - so if two players became low health close together, it's possible only one would display. However that seems like an edge case, especially as these only last for 1 second.

PR #3735, opened 2025-04-06
Backend: Create SignType

rueblimaster:
## Main
There still are some missing SignTypes, but I can't go and do them all.
Also you can ignore the first 2 commits.

+ Added Sign Types. - rueblimaster


rueblimaster:
Should this be repo?

hannibal002 (changes requested):
are the first two parameters always the same? if so, remove the duplicates

rueblimaster:
> are the first two parameters always the same? if so, remove the duplicates

I don't get what you're referencing?

rueblimaster:
I asked myself why even is this, and I didn't really find a good reason except for maybe this might be useful for 1 or 2 minor things in the future, so I'm closing this

PR #3736, opened 2025-04-06
Backend: refractor some events

rueblimaster:
+ Corrected the package of some events. - rueblimaster


src/main/java/at/hannibal2/skyhanni/data/HypixelData.kt
```diff
@@ -421,6 +422,12 @@ object HypixelData {
         val nowOnHypixel = LorenzUtils.onHypixel
         when {
             !wasOnHypixel && nowOnHypixel -> {
+                /* this should never happen but is still here for the odd case of it happenening */
+                if (!skyBlock) {
+                    skyBlock = true
+                    SkyBlockJoinEvent.post()
```
Bloxigus: im so confused why is this here

rueblimaster: maybe there will be a way to log onto sb instantly in the future, and that should catch that

Bloxigus: this will never work, even if they add a way to instantly join skyblock, itll only post the event twice, and it currently will break the mod if you do anything that isnt skyblock on hypixel

rueblimaster: oh okay, ty

Thunderblade73 (changes requested):
Fix Detect. The rest looks fine.

hannibal002 (changes requested):
What ist the difference of SkyBlock join event to profile join event?

Bloxigus:
> What ist the difference of SkyBlock join event to profile join event?

`ProfileJoinEvent` is used, `SkyblockJoinEvent` is not

DavidArthurCole:
gonna add on here and say we should probably stop basing events off of HypixelData until it doesn't suck from their side - these events will fire randomly when, not if, hypixel sends bad data.

CalMWolfs (approved):

hannibal002:
I've renamed "location" to "package" in the changelog

PR #3737, opened 2025-04-06
Backend: rename OwnInventoryItemUpdateEvent to ItemCollectEvent

rueblimaster:
## What
The event only gets triggered if the packet gets sent from the server, so the event doesn't trigger if the player moved an item around in their own inventory.

+ Renamed OwnInventoryItemUpdateEvent to ItemCollectEvent. - rueblimaster

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/OwnInventoryData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/OwnInventoryData.kt
@@ -3,7 +3,7 @@ package at.hannibal2.skyhanni.data
 import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.events.GuiContainerEvent
 import at.hannibal2.skyhanni.events.InventoryCloseEvent
-import at.hannibal2.skyhanni.events.OwnInventoryItemUpdateEvent
+import at.hannibal2.skyhanni.events.ItemCollectEvent
 import at.hannibal2.skyhanni.events.chat.SkyHanniChatEvent
 import at.hannibal2.skyhanni.events.entity.ItemAddInInventoryEvent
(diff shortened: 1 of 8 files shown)


Bloxigus:
This name suggests the event only gets called when a new item gets added you the inventory. This isn't true. You should consider naming it something like `OwnInventoryItemUpdateEvent`, which describes what the event does better.

rueblimaster:
> This name suggests the event only gets called when a new item gets added you the inventory. This isn't true. You should consider naming it something like `OwnInventoryItemUpdateEvent`, which describes what the event does better.

is this because in cases like supercrafting the items get removed from the server side?

hannibal002 (changes requested):
"collect" is too vague.
this could get confused with item add event

CalMWolfs (changes requested):
This pr is not correct in any way, this event is used for, as the name says when an Item in your inventory updates. E.g. you pick up an item or the data on an item updates such as a counter like those of stacking enchants. That there is most definitely not the case of "collecting" an item. As you said the event doesnt trigger if the item moves in the inventory and that is correct as the user is not updating the item. Maybe some of your confusion come from this event being used in places where it shouldnt really be such as the EnderNodeTracker or the CrystalNucleusApi where another event such as ItemAddEvent may be more appropriate provided that gets tested.

rueblimaster:
> This pr is not correct in any way, this event is used for, as the name says when an Item in your inventory updates. E.g. you pick up an item or the data on an item updates such as a counter like those of stacking enchants. That there is most definitely not the case of "collecting" an item. As you said the event doesnt trigger if the item moves in the inventory and that is correct as the user is not updating the item. Maybe some of your confusion come from this event being used in places where it shouldnt really be such as the EnderNodeTracker or the CrystalNucleusApi where another event such as ItemAddEvent may be more appropriate provided that gets tested.

Yes, you're right

PR #3738, opened 2025-04-06
Improvement: Improve accuracy of estimated item value calculator and improve consistency of the display

SuperCl4sh:
## What
This PR does the following
* Checks the price of attributes using attribute shards and compares it with the price obtained from combining item pieces (e.g., kuudra sets)
* Changes the attribute pricing for hellfire rods to be based on magma rods (similar to how non-vanquished pieces are used to calculate the price of attributes of vanquished pieces)
* Changes displayed combo price to be the actual price of the combo instead of the lowest bin of a piece with the combo (e.g., if the lowest bin of a dh fs rod is 100m and the raw craft is 95m for the rod, the displayed combo price is 5m when it was 100m before).
* Change level 1 attributes as having 0 cost because you don't really "pay" for them (e.g., if you have a bf 1 Magma Lord Chestplate, it doesn't make sense to include the price of the bf 1)

## Changelog Improvements
+ Improved Attribute Pricing accuracy in Estimated Item Value. - SuperClash
    * Uses Attribute Shards for pricing when cheaper.
    * Uses the Magma Rod attribute price for Inferno and Hellfire Rods.
    * Displays the actual combo price instead of the lowest bin piece with the attributes.
    * Changes Level 1 attributes to have 0 cost.


hannibal002:
I've updated  the changelog

PR #3740, opened 2025-04-06
Fix: Hitman Threshold

DavidArthurCole:
## What
(Potentially?) fixes the issues hannibal reported where the hitman compacting wouldn't work - I had kinda forgotten that the UI does _not_ need to stay open, so the previous method for detecting when we "can stop" was pretty shortsighted.

## Changelog Fixes
+ Fixed Hitman claim chat compact threshold not working as expected. - Daveed


src/main/java/at/hannibal2/skyhanni/features/inventory/chocolatefactory/hitman/HitmanApi.kt
```diff
@@ -24,6 +30,18 @@ object HitmanApi {
...
+    private val storage get() = ProfileStorageData.profileSpecific
+
+    @HandleEvent
+    fun onInventoryFullyOpened(event: InventoryFullyOpenedEvent) {
+        if (event.inventoryName != "Rabbit Hitman") return
+        val summaryItem = event.inventoryItems[4].takeIf { it?.displayName == "§cRabbit Hitman" } ?: return
+        val availableEggs = hitmanAvailableEggsPattern.firstMatcher(summaryItem.getLore()) {
+            group("amount").toInt()
+        } ?: return
+        val storage = storage ?: return
```
hannibal002: move storage to first line of this function. there i no need to have the code above run at all if this crucial (and easy to check) precondition is not met

hannibal002 (changes requested):

PR #3742, opened 2025-04-07
Backend: GUI -> Gui

hannibal002:
## What
This follows our code style rules:
- Follow Kotlin conventions for acronym naming:
    - Use all-uppercase for two-letter acronyms (e.g., `XP`).
    - Treat three or more letter acronyms as regular words with only the first letter capitalized (e.g., `Api`).

+ Renamed most cases of GUI to Gui. - hannibal2


CalMWolfs (changes requested):
i think we could just do these 2 todos for changing the config variable names now instead of leaving that for a future pr

PR #3752, opened 2025-04-10
Fix: Item NBT Data not saving in config (/ff or wardrobe)

NopoTheGamer:
## What
Item NBT data is not properly getting saved in the config.
This caused features that save ItemStacks in the config to break (e.g. ff or wardrobe)
This PR fixed the issue.

## Changelog Fixes
+ Fixed saving of item data in /ff and Wardrobe. - nopo
    * Takes effect after the second startup, once the data has been reloaded.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/enoughupdates/EnoughUpdatesManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/enoughupdates/EnoughUpdatesManager.kt
@@ -163,9 +163,9 @@ object EnoughUpdatesManager {
         json.addProperty("itemid", stack.item.getIdentifierString())
         json.addProperty("displayname", stack.displayName)
         //#if MC < 1.21
-        //$$ // todo nbt tag doesnt exist on modern
-        //$$ json.addProperty("nbttag", tag.toString())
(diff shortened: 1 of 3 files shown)


hannibal002 (changes requested):
please change the chagnelog.
the current text belongs in technical details
a new text can be added that says  that some of the /ff bugs are fixed now.

NopoTheGamer:
> please change the chagnelog. the current text belongs in technical details a new text can be added that says that some of the /ff bugs are fixed now.

done

hannibal002:
this doesnt impact skulls only, this impacts all item metadata. e.g. name and more. and therefore also fixes the non loading of ff stats.
image

or is the bugfix here only for skulls? then im a bit dissapointed xD

NopoTheGamer:
> this doesnt impact skulls only, this impacts all item metadata. e.g. name and more. and therefore also fixes the non loading of ff stats.
> or is the bugfix here only for skulls? then im a bit dissapointed xD

i just didnt really know what the bug was lol

hannibal002:
updated the pr desc

hannibal002:
i have added a "remove invalid items" logic in /ff:
before:
image
after:
image

PR #3754, opened 2025-04-10
Fix: Highlight Fels Skull

hannibal002:
## What
The isInvisible check doesnt work, the mob gets removed/never added from/to the list.
instead, checking if the mob moves works fine.

the same bug, broken isInvisibility check, sows fels fully with starred highlight.
To fix that, we now handle fels special, never hightlihgt them on detect like other starred mobs, but rather hightlihgt the block the same way the fel highlgiht feature does (fel highlight overrides starred) and adds the mob highlight when the fel starts moving

Reported: https://discord.com/channels/997079228510117908/1354235350721101834

## Changelog Fixes
+ Fixed Dungeon Fels Skull highlight not working. - hannibal2
+ Fixed Starred Mob Highlight incorrectly highlighting the invisible Fel in Dungeons. - hannibal2
    * Highlights only the head as intended.


src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonMobManager.kt
```diff
@@ -71,40 +64,58 @@ object DungeonMobManager {
         handleFelDespawn(event.mob)
     }
 
+    @HandleEvent(onlyOnIsland = IslandType.CATACOMBS)
+    fun onPlayerMove(event: EntityMoveEvent<EntityLivingBase>) {
```
CalMWolfs: Shouldn't this function be onEntityMove

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonMobManager.kt
hannibal002: Good Catch 🪝

PR #2149, opened 2024-06-25
Feature: Copy Chat

ItsEmpa:
## Dependencies
- #2000 

## What
Adds support for copying chat messages when right clicking them in chat. Holding Shift will copy said chat message with Shwords applied to it, holding Control will only copy the hovered line, and holding Alt will copy the message with formatting color codes.

## Changelog New Features
+ Added Right Click to Copy Chat. - Empa
    * Hold Shift to copy with `/shwords`, Control for a single line, and Alt to format color codes.

PR #3763, opened 2025-04-11
Backend: Disable copy chat

hannibal002:
## What
This PR reverts the feature entry and the mixins from
- #2149

because those mixins broke the chatting mod.

before merging this pr, first add this line to the copy chat pr desc:

exclude_from_changelog


hannibal002:
yield to 
- #3767

PR #3765, opened 2025-04-11
Fix: cs number tracking not using the config, crash when parsing Long

j10a1n15:
## What
happens + error when parsing a string to long bc it didnt use our cool method + shdebug issue when no wardrobe slots are stored

exclude_from_changelog


DavidArthurCole (approved):
happens

PR #3767, opened 2025-04-11
Fix: Chatting Crash

ItsEmpa:
## What
This PR fixes Chatting crashing when using SkyHanni, and also adds some text in the Copy Chat description to say that it will work incorrectly with Chatting installed.

exclude_from_changelog


ItsEmpa:
detekt what the fuck

PR #3774, opened 2025-04-12
Backend: Update to 1.21.5

NopoTheGamer:
## What
Updates the modern project to 1.21.5

## Changelog Technical Details
+ Updates the modern project to 1.21.5. - nopo
+ Update contributing.md to mention Minecraft Development Plugin and Access Wideners. - nopo
+ Update loom. - nopo
+ Update Moulconfig to 3.6.0 for 1.21.5 support. - nea

PR #3776, opened 2025-04-12
Backend: Nbt Compat for 1.21.5

NopoTheGamer:
## Dependencies
#3774

## What
1.21.5 Made every nbt function return an Optional instead of the real type
It also allows NbtList's to contain any type instead of just a single type
I made NbtCompat to fix most these issues

+ Added NbtCompat. - nopo


CalMWolfs (approved):

src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
```diff
@@ -164,7 +165,7 @@ object SkyBlockItemModifierUtils {
     }
 
     fun ItemStack.getAttributes() = getExtraAttributes()
-        ?.takeIf { it.hasKey("attributes", 10) }
+        ?.takeIf { it.containsCompound("attributes") }
```
hannibal002, line 168: add a wrapper for this extension function in item utils

src/main/java/at/hannibal2/skyhanni/utils/compat/NbtCompat.kt
```diff
@@ -0,0 +1,129 @@
...
+    const val TAG_FLOAT: Int = 5
+    const val TAG_DOUBLE: Int = 6
+    const val TAG_BYTE_ARRAY: Int = 7
+    const val TAG_STRING: Int = 8
+    const val TAG_LIST: Int = 9
+    const val TAG_COMPOUND: Int = 10
+    const val TAG_INT_ARRAY: Int = 11
+    const val TAG_ANY_NUMERIC = 99
+
+    fun NBTTagCompound.containsList(key: String): Boolean {
```
hannibal002: make all functions in this class non extended functions

hannibal002 (changes requested):

PR #3792, opened 2025-04-12
Backend: Added Server ticks alive to Entity Data

mariethepolarfox:
## What
Adds server ticks alive to Entity Data, helpful for some features relying on tps.

+ Added Server Ticks Alive to Entity Data. - Helium9


CalMWolfs (approved):

src/main/java/at/hannibal2/skyhanni/data/MinecraftData.kt
```diff
@@ -43,6 +43,8 @@ object MinecraftData {
             }
 
             is S32PacketConfirmTransaction -> {
+                if (packet.actionNumber > 0) return
```
hannibal002, line 46: what is this

hannibal002 (changes requested):
code looks good, im just confused about this one line

src/main/java/at/hannibal2/skyhanni/data/MinecraftData.kt
mariethepolarfox: The server can send multiple transaction packets in the same tick with different action numbers so this makes sure it only counts 1 transaction packet every tick

nwjn: transaction packet is also used to relay window id information, though i dont remember in which cases. so if you were to increment on every single transaction packet itll always end up too fast. its kind of a pet peeve that I see so many mods that dont filter it

PR #3795, opened 2025-04-13
Feature + Fix: Hide co-op members from collections and fix Collection Api not detecting Co-op contributions

ILike2WatchMemes:
## What
Adds a Feature that hides ex co-op members from the collections. Players can either open the "Historic Members" menu or manually add/remove them using /sh<add|remove>excoopmember <name>.
Also fixed the Collection Api not correctly detecting contributions from Co-op members.

## Changelog New Features
+ Added Ex Co-op Member Hider. - ILike2WatchMemes
    * Hides Ex Co-op Members from Collections.
    
## Changelog Fixes
+ Fixed Collection API contributions detection. - ILike2WatchMemes


src/main/java/at/hannibal2/skyhanni/config/commands/Commands.kt
```diff
@@ -192,6 +193,16 @@ object Commands {
...
+        event.register("shaddexcoopmember") {
+            description = "Manually add a name of an ex co-op member."
+            category = CommandCategory.USERS_ACTIVE
+            callback { HideExCoopMembers.addExMemberCommand(it) }
+        }
+        event.register("shremoveexcoopmember") {
+            description = "Manually remove a name of an ex co-op member."
+            category = CommandCategory.USERS_ACTIVE
+            callback { HideExCoopMembers.removeExMemberCommand(it) }
+        }
```
NopoTheGamer: this should be 1 command that takes 2 arguments

j10a1n15: Also can we register the command in the class where it's used

rueblimaster: Also change the description to adding to the exclusion

rueblimaster:
Very well done with all of the regexes, I'm just believe that some of your regexes should be more general than just your usecase, but that is another talking point
Also as far as I can see, the part with the "Historic Members" menu is still missing

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+
+    private const val usage = "§c/shedithiddenexcoopmembers <add|remove> <name>"
+
+    private val patternGroup = RepoPattern.group("data.exmembers")
+
+    /**
+     * REGEX-TEST: oxsss
+     * REGEX-TEST: Gillsplash
+     */
+    private val validNamePattern by patternGroup.pattern(
```
hannibal002: move into a generic utils

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+        "name",
+        "[a-zA-Z0-9_]{2,16}",
+    )
+
+    /**
+     * REGEX-TEST: Historic Members
+     */
+    private val inventoryPattern by patternGroup.pattern(
```
hannibal002, line 41: please use InventoryDetector instad

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    @HandleEvent
+    fun onInventoryOpen(event: InventoryFullyOpenedEvent) {
+        if (!config.hideExCoopMembers || !inventoryPattern.matches(event.inventoryName)) return
+
+        event.inventoryItems.values
+            .filter { it.item == Items.skull }
+            .forEach { item ->
+                namePattern.matchMatcher(item.displayName) {
```
hannibal002: unnecessary pattern. just use `String.cleanPlayerName()`

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+    private val namePattern by patternGroup.pattern(
+        "inventory.historic.name",
+        "(?:§.\\[[^]]+(?:§\\++§b)?] |§7)(?<name>\\S{2,16})",
+    )
+
+    /**
+     * REGEX-TEST: Farming Collections
+     * REGEX-TEST: Carrot Collection
+     */
+    private val collectionInventoryPattern by patternGroup.pattern(
```
hannibal002: please use CollectionApi instead

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+    )
+
+    /**
+     * REGEX-TEST: §b[MVP§f+§b] oxsss§7: §e1.9M
+     * REGEX-TEST: §a[VIP] oxsss§7: §e0
+     * REGEX-TEST: §7oxsss§7: §e0
+     */
+    private val collectedPattern by patternGroup.pattern(
+        "inventory.collections.collected",
+        "(?:§.\\[[^]]+(?:§\\++§b)?] |§7)(?<name>[^§]{2,16})§7: §e(?<amount>.+)",
```
hannibal002: Remove the rank checks from the regex. just use `String.cleanPlayerName()`

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+        "§7(?:Progress to .+|Total Collected: .+)",
+    )
+
+    @HandleEvent
+    fun onTooltip(event: ToolTipEvent) {
+        if (!config.hideExCoopMembers || !collectionInventoryPattern.matches(InventoryUtils.openInventoryName())) return
+        val storage = storage ?: return
+        if (storage.isEmpty()) return
+
+        event.toolTip = event.toolTipRemovedPrefix().handleTooltip(storage)
```
hannibal002: please cache this transformation. e.g. via slot id

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+                    remainingPlayers++
+                }
+
+                totalCollected += group("amount").formatInt()
+            }
+        }
+
+        itemsToRemove.sortedDescending().forEach { removeAt(it) }
+
+        if (remainingPlayers <= 1) {
```
hannibal002: early return

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+
+    private fun List<String>.handleTooltip(storage: MutableSet<String>): MutableList<String> = this.toMutableList().apply {
+        val coopIndex = indexOf("§7Co-op Contributions:")
+        if (coopIndex == -1) return@apply
+
+        var remainingPlayers = 0
+        var totalCollected = 0
+        val itemsToRemove = mutableListOf<Int>()
+
+        for (i in (coopIndex + 1) until size) {
```
hannibal002: use something like `drop(1).for {` or similar

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+                    itemsToRemove.add(i)
+                } else {
+                    remainingPlayers++
+                }
+
+                totalCollected += group("amount").formatInt()
+            }
+        }
+
+        itemsToRemove.sortedDescending().forEach { removeAt(it) }
```
hannibal002: what is happening here?

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+
+        if (remainingPlayers <= 1) {
+            if (!dontDisplayMaxedPattern.anyMatches(this)) {
+                if (coopIndex + 1 < size) this[coopIndex + 1] = "§7Total collected: §e${totalCollected.addSeparators()}"
```
hannibal002: why is there so much math? cant we just use add(index, entry) instead?

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+
+    private fun removeExMember(name: String): List<String> {
+        val storage = storage ?: return listOf()
+        if (!storage.remove(name)) return listOf()
+        return storage.toList()
+    }
+
+    @HandleEvent
+    fun onCommandRegistration(event: CommandRegistrationEvent) {
+        event.register("shedithiddenexcoopmembers") {
```
hannibal002: is it now "ex coop member" or "hidden coop members"? if we just want this for ex coop members, we could rather read the info from profile manager page isntead of this command

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+            when (action) {
+                "add" -> "That username is already in the list!"
+                "remove" -> "That username wasn't in the list!"
+                else -> ""
+            },
+        )
+
+        ChatUtils.hoverableChat(
+            "${action.successString()} $name (Hover to see current list).",
+            new,
```
hannibal002: either show the parameter name here, or rename the variable, or remove the line breaks for this function call

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+            .filter { it.item == Items.skull }
+            .forEach { item ->
+                namePattern.matchMatcher(item.displayName) {
+                    addExMember(group("name"))
+                }
+            }
+    }
+
+    private fun editExCoopMembers(args: Array<String>) {
+        if (args.isEmpty()) return ChatUtils.userError(usage)
```
hannibal002: `ChatUtils.userError(usage)` could be in a function

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+
+        val action = args.firstOrNull()?.takeIf { it in setOf("add", "remove") }
```
hannibal002: move that set outside the function

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,200 @@
...
+import at.hannibal2.skyhanni.utils.RegexUtils.matchMatcher
+import at.hannibal2.skyhanni.utils.RegexUtils.matches
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+import net.minecraft.init.Items
+
+@SkyHanniModule
+object HideExCoopMembers {
+
+    private val config get() = SkyHanniMod.feature.misc
+    private val storage get() = ProfileStorageData.profileSpecific?.exCoopMembers
```
hannibal002: the name storage is wrong. we use storage for the class that holds the data we access, not the data directly. the equivalent for settings is the term `config`. we dont write if (!config) return, we write if (config.!featureName) return.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
ILike2WatchMemes: removing from bottom to top as to not remove an index out of bounds since that happened sometimes

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
ILike2WatchMemes: doesnt work for nons

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
ILike2WatchMemes: I'm replacing the lines, not adding them

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
ILike2WatchMemes: wdym?

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
ILike2WatchMemes: changed the whole thing to hidden co-op members since people might have a use for that, ex co-op members are also still included tho

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
hannibal002: Inside collection API we already check if you are in that collection page. Two spots in two features checking the same thing should be avoided

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
hannibal002: How is this for nons different?

ILike2WatchMemes: because size is > 1 it returns the amount sections instead of the name

hannibal002: Can you show me what you mean

ILike2WatchMemes:
```
private fun String.internalCleanPlayerName(): String {
        val split = trim().split(" ")
        return if (split.size > 1) {
            split[1].removeColor()
        } else {
            split[0].removeColor()
        }
    }
```
grafik

src/main/java/at/hannibal2/skyhanni/features/misc/HideExCoopMembers.kt
```diff
@@ -0,0 +1,184 @@
...
+            if (line.isBlank()) return@forEachIndexed
+
+            collectedPattern.matchMatcher(line) {
+                if (group("name") in storage) {
+                    itemsToRemove.add(coopIndex + index)
+                } else {
+                    remainingPlayers++
+                }
+
+                totalCollected += group("amount").formatInt()
```
hannibal002: i have already mentioned the same thing above:we have an collection api. please use it. if the exact data you need dont exist in the collection api, please add them to the collection api instead of adding code duplication. points for reference include: https://github.com/hannibal002/SkyHanni/blob/b60151d0f7b6b298caea8d238e4d096eab7f7907/src/main/java/at/hannibal2/skyhanni/api/CollectionApi.kt#L78

and 

https://github.com/hannibal002/SkyHanni/blob/ff837769c75b2f70b35f1a11779d594d2e1a35fb/src/main/java/at/hannibal2/skyhanni/features/inventory/ItemDisplayOverlayFeatures.kt#L212

hannibal002 (changes requested):
teseted, it works fine. code looks also good. i just dont like the code duplication around collection reading logic. this doesnt belong in the hide ex coop members class imo

PR #3797, opened 2025-04-13
Backend: format files in features

rueblimaster:
## What
Still "Reformat Code" (Ctrl + Alt + L) and looked at, but only the actually relevant changes

+ Formatted files in Features package. - rueblimaster


CalMWolfs (approved):

hannibal002 (approved):
Stop touching unrelated files in future prs. this is not as helpful as you think it is

PR #3799, opened 2025-04-13
Remove Useless Todo

j10a1n15:
## What
Since #3782 got merged i assume this todo isnt needed anymore

exclude_from_changelog


hannibal002:
please elaborate

j10a1n15:
> please elaborate

Well the todo in custom scoreboard said to remove the publicly accessible config getter. The Pr mentioned above also adds a publicly accessible config getter
So yeah since that got merged this is basically just the same

hannibal002 (changes requested):
the todo talks about all the 10 configs. 
Slayer API is different.
there is one config element that a lot of classes use, thats fine.

here we have many config elemetns that only get used in one class and can be moved into that class:
arrowConfig, mayorConfig, partyConfig, maxwellConfig, backgroundConfig

the other objects that arent "config" should be inlined.
this only obfuscates the correct path of where the config comes from.

hannibal002 (approved):
waiting on junas feedback

PR #3800, opened 2025-04-13
Backend: Add MobEvent.KilledByLocalPlayer

rueblimaster:
## What
Listens for MobHurt events, then adds the hurt mobs to a TimeLimitedSet, if the source of the MobHurt is the localPlayer.
If the source of the MobHurt can be attributed to another player, the mob is removed from the set again.
(Note: if another player hits the mob, it can't be attributed to the other player, so in that case the MobEvent.KilledByLocalPlayer is falsely triggered)
Then if a mob dies and it's on the list the event gets triggered.

+ Added an event for when a mob gets killed by the player. - rueblimaster

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/mob/MobDebug.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/mob/MobDebug.kt
@@ -90,7 +90,15 @@ object MobDebug {
     @HandleEvent
     fun onMobEvent(event: MobEvent) {
         if (!config.logEvents) return
-        val text = "Mob ${if (event is MobEvent.Spawn) "Spawn" else "Despawn"}: ${
+        val eventType = when (event) {
+            is MobEvent.Spawn -> "Spawn"
+            is MobEvent.DeSpawn -> "DeSpawn"
+            is MobEvent.FirstSeen -> "FirstSeen"
+            is MobEvent.Hurt -> "Hurt"
+            else -> "Unknown"
+        }
+
+        val text = "Mob $eventType: ${
             getMobInfo(event.mob).joinToString(", ")
         }"
         MobData.logger.log(text)
--- a/src/main/java/at/hannibal2/skyhanni/data/mob/MobDebug.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/mob/MobDebug.kt
@@ -94,6 +94,7 @@ object MobDebug {
             is MobEvent.Spawn -> "Spawn"
             is MobEvent.DeSpawn -> "DeSpawn"
             is MobEvent.FirstSeen -> "FirstSeen"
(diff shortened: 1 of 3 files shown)


hannibal002:
this cant be accurate. rather name the event "mob that got meele hit by the player died"-event

rueblimaster:
> this cant be accurate. rather name the event "mob that got meele hit by the player died"-event

You're right, I also wrote that as comments, the thing is, that also wouldn't be accurate it doesn't get triggered if another player killed the mob with an arrow or so. But I think I'll find a way

src/main/java/at/hannibal2/skyhanni/data/mob/MobDebug.kt
```diff
@@ -90,7 +90,16 @@ object MobDebug {
     @HandleEvent
     fun onMobEvent(event: MobEvent) {
         if (!config.logEvents) return
-        val text = "Mob ${if (event is MobEvent.Spawn) "Spawn" else "Despawn"}: ${
+        val eventType = when (event) {
+            is MobEvent.Spawn -> "Spawn"
+            is MobEvent.DeSpawn -> "DeSpawn"
```
NopoTheGamer, line 95: Despawn is 1 word

NopoTheGamer (changes requested):

hannibal002:
well, i didnt say in my wording "and the player is the last person who hit the mob", so this is still accurate.
but overall its useless to have.

i dont think you will find a way to detekt who killed what mob that is accurate.
with pings, with particles disabled, with so many different abilities, with so many different radiases and modifiers that change that radius, its impossible to correctly calculate something in that chaos.

feel free to open a new pr if you think you mastered all those challlenges

Thunderblade73:
> > this cant be accurate. rather name the event "mob that got meele hit by the player died"-event
> 
> You're right, I also wrote that as comments, the thing is, that also wouldn't be accurate it doesn't get triggered if another player killed the mob with an arrow or so. But I think I'll find a way

See #526. I have meele and normal arrows (no spread arrows, like terminator). It might be old as hell, but still the code works, if adapted to the new mob detection and every other change that happend in the code base the last 2 years. But I plan to finish that PR some day. I just do not have enough time and this is the bottom of my TODO list.

rueblimaster:
> well, i didnt say in my wording "and the player is the last person who hit the mob", so this is still accurate. but overall its useless to have.
> 
> i dont think you will find a way to detekt who killed what mob that is accurate. with pings, with particles disabled, with so many different abilities, with so many different radiases and modifiers that change that radius, its impossible to correctly calculate something in that chaos.
> 
> feel free to open a new pr if you think you mastered all those challlenges

If you ask me it doesn't have to be 100% accurate, btw this code is kinda what I did in the fishing tracker thingy, but I'll look more into it or just implement this stuff where ever I may need it as not to confuse anybody

PR #3801, opened 2025-04-13
Fix: migrate from KeyPressEvent to KeyDownEvent

rueblimaster:
## Main
Yes, there are more uses of this event that could be migrated, I for now only migrated those with a warp in them.
If someone complains about another case of this being a problem, we can still change it then.

## Changelog Fixes
+ Fixed holding down the warp keybind leading to multiple warps. - rueblimaster


hannibal002:
>    Makes it so you don't warp multiple times, if you hold down the warp keybind.


this is no technical detail

hannibal002 (changes requested):
The changes look fine at a glance, please change the pr to a fix

rueblimaster:
> The changes look fine at a glance, please change the pr to a fix

sure, should I change something in the changelog too?

hannibal002:
yes, remove the technical parts of the chagnelog. a fix changelog is user facing and should not include any technical details.
those are for the pr desc

feel free to look up the contributing.md file again, it should explain the changelog rules.

hannibal002 (changes requested):
can we revert those changes, remove key press event, rename key down event to key press, and add a doc over the keypress event that this only gets triggered once?

rueblimaster:
> can we revert those changes, remove key press event, rename key down event to key press, and add a doc over the keypress event that this only gets triggered once?

That would mean checking every use of the KeyPressEvent, so go ahead if you'd like

rueblimaster:
@hannibal002 

What about a util, that takes in a key, a cooldown and a condition and then based on that runs code if key is tapped?

hannibal002:
> What about a util, that takes in a key, a cooldown and a condition and then based on that runs so code if key is tapped?

i like the idea

hannibal002:
> That would mean checking every use of the KeyPressEvent, so go ahead if you'd like

what do we have beta testers for?

rueblimaster:
> > That would mean checking every use of the KeyPressEvent, so go ahead if you'd like
> 
> 
> 
> what do we have beta testers for?

No, I mean just taking a quick glance at the implementations

rueblimaster:
closing this in favor of #3871

PR #3802, opened 2025-04-13
Fix + Backend: `KeyDownEvent` (again)

rueblimaster:
## Main
After #3676 & #3691 all this stuff still isn't fixed correctly.
The current problem is: mouse keys don't get tracked, this is because we stopped looking for mouse events after finding a keyboard event, which we always do

The core of what I changed was using the proper keyboard event functionality with .next() etc. Just as god intended. Additionally don't stop looking out for a mouse or a keyboard event if something is found anywhere.
Tested that both keys and mouse buttons work as warp keybinds. Also tested for the bug in #3676.

## Changelog Fixes
+ Fixed Mod-wide Keybinds V3. - rueblimaster

+ Changed KeyboardManager to allow multiple simultaneous keys and buttons. - rueblimaster
+ Added KeyUpEvent. - rueblimaster


hannibal002 (changes requested):
@rueblimaster after reading the conversations at #3803:
can you explain why your code would work and the orignal code didnt? what did change fundamentally in the keybind logic? im not familar with this at all and i want to avoid breaking other stuff in the future by merging code i dont understand.
please break it down more, in this pr or in comments in code, however you like


moving the remainig conversations in the proper pr:

> Also i havent looked much at this PR, but in general the concept of consuming all key events in a mod specific event seems dubious to me. (And something worth testing in combination with a variety of other mods)

@nea89o what do you mean with consume here?
does this mean if we use this code, other mods cant detect the same inputs anymore?
how do we avoid this? can we somehow just "listen in" without eatling the input actions?


> I dont like this pr very much, but the repeat function exists for this use case


@Bloxigus using a arbatary number to check here is always bad either way. if we truly need to read in all input actions that way we should not use a limit at all.

rueblimaster:
@hannibal002 
The fundamental problem with the code (before any fixes) was, that we assumed Mouse.Event and Keyboard.Event would have an empty key if nothing happened. That is not the case (at least not with keyboard, and from the other fix I believe the same to be the case for the mouse). So they would block each other, either the mouse or the keyboard got priority depending on the code.
Now the biggest change is that we go and handle both of them separately. This fixes those issues.
Additionally I went and used the Keyboard Eventsystem as it was supposed to be used with Keyboard.next() and then reading the event and continuing that. The concern that @nea89o had stems from the fact, that Keyboard is globally used (at least it seems like it) and because I "consume" events by calling Keyboard.next() this could cause issues in other mods.

hannibal002:
we assume that a empty key is when nothing happened? what has this to do with the mouse actions?

when multiple keys (keyboard and mouse) are pressed at the same time, that this causes isssues i see.
how can this cause issues with other mods thoough? cant they simply do the same checks we do at the same time and be fine?
and how would us detecting this hide thesame inputs from other mods?
i dont now whats going on, i dont want sh to break keybinds in any other mod. this is important and needs looked at before we move further with this pr.
the missing mouse keybind support is not that important to fast track this possible bigger probem

lineargraph:
keyboard and mouse events are global state in lwjgl. There is (figuratively speaking) an array like this in a global variable:

```kt
object lwjgl.Keyboard {
	private val keyboardEventQueue = mutableList<LWJGLKeyboardEvent>()
	private var activeEvent: LWJGLKeyboardEvent
	fun next(): Boolean {
		activeEvent = keyboardEventQueue.removeFirst() ?: return false
		return true
	}
	fun getKeyboardKey() = activeEvent.keyboardKey
}
```

This global queue is managed by lwjgl and isnt aware of multiple mods. To allow mods to use those events anyway forge adds an event:

```kt
/// somewhere in the main loop
while (Keyboard.next()) {
	val forgeEvent = ForgeInputEvent(Keyboard.getEventKey(), Keyboard.getEventKeyState())
	forgeEvent.post()
}
```

That is also why the first time around you don't need to call `Keyboard.next()` -- Forge does that for us. We can even call `Keyboard.getEventKey()` in the loop (or the event posted inside of the loop) multiple times. But if we use `Keyboard.next()` in the loop we will advance the global queue preventing forge from posting the event that distributes the event to other mods.

*Even if* we post the forge event (so as to allow other mods to accept keyboard events as well) we would still be making behaviour like that acceptable, meaning if another mod was to do the same thing, our integration would break unless we also listen to the forge event, in which case we couldve just listened to the original event.

The reason this PR came up is unclear to me as well.

It seems like there is because there is some confusing between mouse and keyboard events. If the intra-tick inter-peripheral order of mouse and keyboard events is relevant we are sadly a bit SOL: not only does forge (to my knowledge) process the events sorted by peripheral (iirc mouse first, but i could be wrong), but LWJGL does not (to my knowledge, again, i havent done lwjgl 2 in a while) provide a facility to ascertain the inter peripheral order of input events.

If the concern is an unclean state of the globals (or even the events) provided by lwjgl / forge, then wrapping the existing forge events and sanitizing them, as well as denying access to the globals / providing your own sanitized globals that get set in accordance to the currently processed event is probably a better cause.

rueblimaster:
@hannibal002 
The problem was that we checked if keyboard event and if we found one we didn't go check the mouse events.
This is what I effectively changed with the pr as it is rn

src/main/java/at/hannibal2/skyhanni/utils/KeyboardManager.kt
```diff
@@ -62,36 +63,35 @@ object KeyboardManager {
...
-            return EventKey(key, MouseCompat.getEventButtonState())
+            val keyCode = MouseCompat.getEventButton() - 100
+            lastClickedMouseButton = keyCode
+            return EventKey(keyCode, MouseCompat.getEventButtonState())
         }
         if (lastClickedMouseButton != -1 && MouseCompat.getEventButton() == -1) {
             Mouse.poll()
             val originalButton = lastClickedMouseButton + 100
             if (Mouse.isButtonDown(originalButton)) return EventKey(lastClickedMouseButton, true)
             else lastClickedMouseButton = -1
```
hannibal002: what does the else do here? the if block here is confusing to me

hannibal002 (changes requested):
have you tested the pr together with other mods that work with key events, most notably skytils?

rueblimaster:
> ### 1 Detekt Failure was detected:
> 
>     * [KeyboardManager.kt#L3](https://github.com/hannibal002/SkyHanni/blob/faf9484301c22dc98c2e74fb4b45f47096109870/src/main/java/at/hannibal2/skyhanni/utils/KeyboardManager.kt#L3): Imports begin with a pre-processed comment, make sure to check the order of the imports. If this pre-processing comment is correct you can baseline this error. If you did accidentally change the order of the imports fix them.

naughty @hannibal002

rueblimaster:
> have you tested the pr together with other mods that work with key events, most notably skytils?

nope, so this is a thing I'll have to do, but we no longer consume the events so should be less of a possible issue

rueblimaster:
@hannibal002 nah, I don't think you fixed it

But great consistency on your end

rueblimaster:
What about a general utils, that takes in a key, a cooldown and a condition and then based on that runs so code if key is tapped?

rueblimaster:
> ### 1 Detekt Failure was detected:
> 
> - [KeyboardManager.kt#L3](https://github.com/hannibal002/SkyHanni/blob/e4f9f350e7e396544176faf324cb7fe98d7827bf/src/main/java/at/hannibal2/skyhanni/utils/KeyboardManager.kt#L3): Imports begin with a pre-processed comment, make sure to check the order of the imports. If this pre-processing comment is correct you can baseline this error. If you did accidentally change the order of the imports fix them.
> 

naughty @hannibal002 V2
(but this time I can't fix it for you ❤️)

hannibal002:
@rueblimaster is this ready?
image

rueblimaster:
I also just tested it, just to make sure, works with skytils

PR #3803, opened 2025-04-13
Backend: move KeyboardManager to data

rueblimaster:
## Dependencies
- #3802 

## What
move from utils to data as it also has data alongside with some utils

+ Corrected the package of KeyboardManager . - rueblimaster

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/KeyboardManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/KeyboardManager.kt
@@ -1,4 +1,4 @@
-package at.hannibal2.skyhanni.utils
+package at.hannibal2.skyhanni.data
 
 import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.data.model.TextInput
@@ -7,6 +7,7 @@ import at.hannibal2.skyhanni.events.minecraft.KeyPressEvent
 import at.hannibal2.skyhanni.events.minecraft.KeyUpEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.test.command.ErrorManager
+import at.hannibal2.skyhanni.utils.DelayedRun
 import at.hannibal2.skyhanni.utils.compat.MouseCompat
 import net.minecraft.client.Minecraft
 import net.minecraft.client.gui.GuiChat
--- a/src/main/java/at/hannibal2/skyhanni/data/KeyboardManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/KeyboardManager.kt
@@ -69,6 +69,7 @@ object KeyboardManager {
     private fun getKeyboardEventKeys(): List<EventKey> {
         val keys = mutableListOf<EventKey>()
         // if multiple keys are pressed at the same time, only 1 key is processed. The other ones may randomly be processed later.
+        @Suppress("unused")
         for (i in 0 until 10) {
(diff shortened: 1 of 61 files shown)


src/main/java/at/hannibal2/skyhanni/data/KeyboardManager.kt
```diff
@@ -62,36 +64,41 @@ object KeyboardManager {
...
-                0 -> EventKey(Keyboard.getEventCharacter().code + 256, Keyboard.getEventKeyState())
-                else -> EventKey(Keyboard.getEventKey(), Keyboard.getEventKeyState())
-                    .also { lastClickedMouseButton = -1 }
+    private data class EventKey(val keyCode: Int, val pressed: Boolean)
+
+    private fun getKeyboardEventKeys(): List<EventKey> {
+        val keys = mutableListOf<EventKey>()
+        // if multiple keys are pressed at the same time, only 1 key is processed. The other ones may randomly be processed later.
+        @Suppress("unused")
+        for (i in 0 until 10) {
```
Bloxigus, line 73: I dont like this pr very much, but the `repeat` function exists for this use case

rueblimaster: Yes, chatGPT told me the same.
The reason it's not a repeat is the break statement. Also that's more suited to the dependency.
Do you not agree with the move or what is the problem?

DavidArthurCole: `return@repeat`

DavidArthurCole (changes requested):

src/main/java/at/hannibal2/skyhanni/data/KeyboardManager.kt
Bloxigus: Actually this seems like a while loop wanting to be written

rueblimaster: It is, the only problem with that, is that I didn't get it to work somehow, as I need to check it once before .next()
'!firstIteration || next()
Would've been the correct solution ig

lineargraph: > `return@repeat`

that is a continue not a break. you would need an outer label to break to (which you can do thanks to `-XBreakContinueInInlineLambdas`). but that isnt great either. in the end it will probably just be a while loop:

```kt
val keys = buildList {
	do {
	        val keyCode = getSyntheticKeyboardKeyCode(Keyboard.getEventKey(), Keyboard.getEventCharacter())
            val keyState = Keyboard.getEventKeyState()
            if (keyCode != 0) {
                add(EventKey(keyCode, keyState))
            }
	} while (Keyboard.next())
}
```

lineargraph:
Also i havent looked much at this PR, but in general the concept of consuming all key events in a mod specific event seems dubious to me. (And something worth testing in combination with a variety of other mods)

rueblimaster:
Please, this PR is the move of the KeyboardManager, for anything in the KeyboardManager please comment in the dependency!

hannibal002:
I dont like touching 60 files and maybe even causing merge conflicts for the sole purpose of moving a package.
this should be done in a bigger "move all stuff out of utils" sweep once the open pr numbers are down.

for future prs:
think if the change is necessary, or if this can be done later easlier with less pptential conflicts.
if the pr touches more than 10 files, dont do it without asking us before please.
this saves us all time.

PR #3804, opened 2025-04-14
Backend: Playerpos to shdebug with graph nodes

j10a1n15:
## What
adds the player boss to node graph areas

exclude_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
@@ -81,7 +81,7 @@ object SlayerApi {
             if (!isInCorrectArea) {
                 add("currentAreaType: $currentAreaType")
                 add(" hypixel area: ${HypixelData.skyBlockArea}")
-                add("graph area: ${IslandAreas.currentAreaName}")
+                add(" graph area: ${IslandAreas.currentAreaName}")
             }
             add("isInAnyArea: $isInAnyArea")
             add("latestSlayerProgress: ${latestSlayerProgress.removeColor()}")
--- a/src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
@@ -41,7 +41,7 @@ object SlayerApi {
     var latestSlayerProgress = ""
 
     val currentAreaType by RecalculatingValue(500.milliseconds) {
-        checkSlayerTypeForCurrentArea()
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
```diff
@@ -78,6 +78,11 @@ object SlayerApi {
         event.addData {
             add("activeSlayer: $activeSlayer")
             add("isInCorrectArea: $isInCorrectArea")
+            if (!isInCorrectArea) {
+                add("currentAreaType: $currentAreaType")
+                add(" hypixel area: ${HypixelData.skyBlockArea}")
+                add("graph area: ${IslandAreas.currentAreaName}")
```
CalMWolfs: i think either the hypixel area line has an un-needed space or graph area line is missing a space

CalMWolfs (changes requested):

hannibal002 (changes requested):
Instead of this workaround, we should update the graph area at the location where the detection failed.
https://discord.com/channels/997079228510117908/1000669238035497022/1361589068697309345

also please note that the hypixel area is broken on some spots and trusting the graph area more than the hypixel scoreboard area is generally preferred.
since we wanna move all area detections over to graph area eventually, i dont think we should use this workaround introduced in this pr.

j10a1n15:
Well yeah we trust the node graph more than Hypixel in this pr
It's just that all failed detections got introduced after switching to the node graph, so I think we should absolutely use this fallback for now

hannibal002:
But this workaround does mean that when the graph editor shows that there is the end of the area, the scoreboard data gets used. i dont want that, as there are cases where we have different borders. Especially on spots where Hypixel wrongfully shows the old area still even though you are map wise already in another area.

If we get reported such area navigation problems we can simply fix this via repo

src/main/java/at/hannibal2/skyhanni/test/DebugCommand.kt
```diff
@@ -145,6 +146,9 @@ object DebugCommand {
             add("skyBlockArea:")
             add("  scoreboard: '${LorenzUtils.skyBlockArea}'")
             add("  graph network: '${IslandAreas.currentAreaName}'")
+            MinecraftCompat.localPlayer.position.let {
+                add("  player pos: ${it.x}, ${it.y}, ${it.z}")
```
hannibal002: please use this: `with(location.roundTo(1)) { "/shtestwaypoint $x $y $z pathfind" }`

hannibal002 (changes requested):

PR #3805, opened 2025-04-14
Backend: PushPop/Translated/Scaled

j10a1n15:
## What
Mainly looking for feedback from cal/nop/hanni/...
This cant replace all places were push & pop is used (e.g. Movable Hotbar bc it has its operations split)
Safer way to run these operations without forgetting push/pop

+ Added PushPop, Translated & Scaled operations. - ThatGravyBoat, j10a1n15


src/main/java/at/hannibal2/skyhanni/data/RenderData.kt
```diff
@@ -36,18 +36,17 @@ object RenderData {
...
-            DrawContextUtils.translate(0f, 0f, 3f)
-        }
+        DrawContextUtils.pushPop {
+            GlStateManager.enableDepth()
 
-        GuiRenderEvent.ChestGuiOverlayRenderEvent(event.context).post()
+            if (GuiEditManager.isInGui()) {
+                DrawContextUtils.translate(0f, 0f, -3f)
+                renderOverlay(event.context)
+                DrawContextUtils.translate(0f, 0f, 3f)
```
CalMWolfs: cant this here now also use `translated`

src/main/java/at/hannibal2/skyhanni/utils/compat/DrawContextUtils.kt
```diff
@@ -61,11 +61,40 @@ object DrawContextUtils {
...
     fun popMatrix() {
         drawContext.matrices.popMatrix()
     }
+
+    @Suppress("deprecation")
+    inline fun pushPop(action: () -> Unit) {
+        pushMatrix()
+        action()
+        popMatrix()
+    }
```
CalMWolfs, line 84: could probably add a java doc to this method as well

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/compat/DrawContextUtils.kt
```diff
@@ -61,11 +61,43 @@ object DrawContextUtils {
...
+
+    /**
+     * Run operations inside a DrawContext translation
+     */
+    inline fun translated(x: Number = 0, y: Number = 0, z: Number = 0, action: () -> Unit) {
+        pushPop {
```
Thunderblade73: You will run into gl stack errors with the push pop approach here. As we render quiet deep into the stack already. Eg: In Renderable we only have ~10 stack depth left.

Thunderblade73 (changes requested):

CalMWolfs (approved):

PR #3818, opened 2025-04-16
Fix: Composter empty warning!

hannibal002:
## Changelog Fixes
+ Fixed Garden Composter empty warnings not showing. - hannibal2


hannibal002:
detekt this is not what || means

PR #3823, opened 2025-04-18
Fix: Critical Bug with Trackers

DavidArthurCole:
## What
I'm shocked to my core, I tell you

## Changelog Fixes
+ Fixed trackers not grammaring goodly. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
@@ -201,8 +201,7 @@ object StringUtils {
         return "$allButLast$delimiterColor, and ${this[lastIndex]}"
     }
 
-    fun String.pluralize(number: Int, plural: String? = null, withNumber: Boolean = false) =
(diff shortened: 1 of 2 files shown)


hannibal002:
Good thing we include the case when the tracker sums up to exactly 1 coin. I'm sure this will help many

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -201,6 +201,9 @@ object StringUtils {
         return "$allButLast$delimiterColor, and ${this[lastIndex]}"
     }
 
+    fun String.pluralize(number: Int, plural: String? = null, withNumber: Boolean = false) =
```
hannibal002: can we remove plural from the extension function? the format cactus.plural(3, cactii) feels odd. for "non s plurals" we have two strings, and having one as extension function parameter and one as normal parameter feels odd. for that case we should still just use the non extension function

hannibal002 (changes requested):

PR #3829, opened 2025-04-18
Fix: Computer Time Offset

hannibal002:
## What
The computer time offset logic currently checks every second the internet connection.
when the user had bad internet this request might take longer than 1s.
after a while of this happneing again and again, this might cause that too many connections are open at the same time, and we dont have more threads for all those tasks.

This PR now tries to fix this by detecting if a current connection is still running, and if so, first moves the delay from 1s to 10s, and if this still happens, disables the repeating checks entirely.

## Changelog Fixes
+ Fixed Computer Time Offset lag on poor internet connections. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
@@ -6,6 +6,7 @@ import at.hannibal2.skyhanni.events.DebugDataCollectEvent
 import at.hannibal2.skyhanni.events.ProfileJoinEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.test.command.ErrorManager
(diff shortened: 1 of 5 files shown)


CalMWolfs (approved):

src/main/java/at/hannibal2/skyhanni/utils/ComputerTimeOffset.kt
```diff
@@ -33,19 +44,38 @@ object ComputerTimeOffset {
...
+        if (currentlyChecking) {
+            state = state.next() ?: error("state is already TOTALLY_OFF")
+            if (state == State.TOTALLY_OFF) {
+                ErrorManager.logErrorStateWithData(
+                    "Error when checking Computer Time Offset",
+                    "trying to check again even though the previous check is stil not done",
+                )
+            }
+            if (state == State.SLOW) {
+                ChatUtils.chat("Computer Time Offset calculation takes longer than normal. Checkign now less often.")
```
NopoTheGamer, line 69: Checking*

PR #3839, opened 2025-04-19
Backend: Value compat method + misc 1.21 changes

NopoTheGamer:
## What
Adds .value() compat extension function which should be used instead of .value for hover and click events. Im not sure if value() is a great name but i couldnt think of something better

+ Added value compat method for click and hover events. - nopo
+ Added auto-join Hypixel with quick play args in dev env for modern. - nopo


CalMWolfs (approved):

src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -535,4 +528,33 @@ object StringUtils {
     private fun isFormatSpecial(formatChar: Char): Boolean {
         return formatChar in 'k'..'o' || formatChar in 'K'..'O' || formatChar in "rR"
     }
+
+    // copied from minecraft, thx jappa
+    private val colorCodes = makeColorCodeArray()
```
hannibal002: this belongs in the color utils, not string utils

src/main/java/at/hannibal2/skyhanni/features/chat/translation/Translator.kt
```diff
@@ -87,8 +87,13 @@ object Translator {
 
     private fun createClickStyle(message: String, style: ChatStyle): ChatStyle {
         val text = messageContentRegex.find(message)!!.groupValues[1].removeColor()
+        //#if MC < 1.21
         style.setChatClickEvent(ClickEvent(ClickEvent.Action.RUN_COMMAND, "/shtranslate $text"))
         style.setChatHoverEvent(HoverEvent(HoverEvent.Action.SHOW_TEXT, "§bClick to translate!".asComponent()))
+        //#else
+        //$$ style.withClickEvent(ClickEvent.RunCommand("/shtranslate $text"))
```
hannibal002: i think we already have a chat wrapper/api that allows to create chat messages with click/hover action, we should not use compat logic in feature classes this way

src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
```diff
@@ -64,7 +67,11 @@ object GardenOptimalSpeed {
 
     @HandleEvent(onlyOnIsland = IslandType.GARDEN)
     fun onTick() {
+        //#if MC < 1.21
         currentSpeed = (MinecraftCompat.localPlayer.capabilities.walkSpeed * 1000).toInt()
+        //#else
```
hannibal002: please move the "player walk speed" loigc inside the player utils class

src/main/java/at/hannibal2/skyhanni/features/garden/GardenPlotBorders.kt
```diff
@@ -23,11 +24,11 @@ object GardenPlotBorders {
         if (!isEnabled()) return
         if (timeLastSaved.passedSince() < 250.milliseconds) return
 
-        if (event.keyCode == Keyboard.KEY_G && Keyboard.isKeyDown(Keyboard.KEY_F3)) {
+        if (event.keyCode == Keyboard.KEY_G && Keyboard.KEY_F3.isKeyHeld()) {
```
hannibal002, line 27: we intentionally had the is key down function as not a extendion function. since this would show up for all integers everywhere. and we dont want to assume every int represetns a key.

hannibal002 (changes requested):
woks fine in game

src/main/java/at/hannibal2/skyhanni/features/garden/GardenPlotBorders.kt
NopoTheGamer: ok.... well i didnt change it in this pr and thats how its been for as long as i can remember

NopoTheGamer: it was you in 2023......... https://github.com/hannibal002/SkyHanni/commit/06ef6be0fd0b7085c7121f100dff354311287e2b

PR #3841, opened 2025-04-19
Fix: `KeyDownEvent` (again) V4

rueblimaster:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/396

## What
#3802 doesn't account for the user using keys with without a native keycode.

This is because we create the keycode like this: if the key doesn't have a code so 0, we take the Char.code + 256.
The problem arises when we want to check later if the key is still pressed, because there we do Keyboard.isKeyDown(keyCode) and if the keycode is one that was constructed with the Char.code + 256 we get an IndexOutOfBoundsException.

The current fix is to not go and check if the is still pressed if the keycode is above the keyboard size but instead just say that it isn't pressed anymore. 
This will trigger the attached keybinds multiple times but the effects of this should be mitigated by the existence of cooldowns in the features.

## Changelog Fixes
+ Fixed KeyboardManager errors on certain letters. - rueblimaster

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/KeyboardManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/KeyboardManager.kt
@@ -118,7 +118,11 @@ object KeyboardManager {
             val isDown = if (keyCode < 0) {
                 Mouse.isButtonDown(keyCode + 100)
             } else {
-                Keyboard.isKeyDown(keyCode)
+                if (keyCode < Keyboard.KEYBOARD_SIZE) {
+                    Keyboard.isKeyDown(keyCode)
+                } else {
+                    false
+                }
             }
 
             if (isDown) {
--- a/src/main/java/at/hannibal2/skyhanni/utils/KeyboardManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/KeyboardManager.kt
@@ -140,10 +140,12 @@ object KeyboardManager {
         //#endif
     }
 
+    /*
+    The delay below is here to make sure the Text input features in graph editor
+    and in renderable calls have time to react first, and lock this key press event properly
+     */
+
     private fun postKeyPressEvent(keyCode: Int) {
-        // This cooldown is here to make sure the Text input features in graph editor
-        // and in renderable calls have time to react first,
-        // and lock this key press event properly
         DelayedRun.runDelayed(50.milliseconds) {
             if (TextInput.isActive()) return@runDelayed
             KeyPressEvent(keyCode).post()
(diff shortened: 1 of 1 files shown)


CalMWolfs (approved):
This does feel like a bandaid fix rather than a true fix but its better than nothing

DavidArthurCole:
Can we please not touch this file apart from porting to 1.21 now? Please?

CalMWolfs:
> Can we please not touch this file apart from porting to 1.21 now? Please?

I hope this is the last fix needed

rueblimaster:
> This does feel like a bandaid fix rather than a true fix but its better than nothing

Ye, that's really what it is

hannibal002:
> Can we please not touch this file apart from porting to 1.21 now? Please?

I still want to remove the three key events and then have only one remaining that technically does key up and has the name key event but expect for that I don't intend to touch that class ever

hannibal002:
> This will trigger the attached keybinds multiple times but the effects of this should be mitigated by the existence of cooldowns in the features.

this sounds too scary.
instead of testing every featur that uses a keybind, we will just switch over to the "only key up" logic immediately ig

src/main/java/at/hannibal2/skyhanni/utils/KeyboardManager.kt
```diff
@@ -118,46 +118,44 @@ object KeyboardManager {
...
         DelayedRun.runDelayed(50.milliseconds) {
             if (TextInput.isActive()) return@runDelayed
             KeyDownEvent(keyCode).post()
         }
     }
 
     private fun postKeyUpEvent(keyCode: Int) {
-        // This cooldown is here to make sure the Text input features in graph editor
-        // and in renderable calls have time to react first,
-        // and lock this key press event properly
```
hannibal002, line 160: if you remove the comment then we dont know why the cooldown exists and someone might remove that later since they cant replicate the reason without knowing

src/main/java/at/hannibal2/skyhanni/utils/KeyboardManager.kt
```diff
@@ -118,46 +118,44 @@ object KeyboardManager {
             val isDown = if (keyCode < 0) {
                 Mouse.isButtonDown(keyCode + 100)
             } else {
-                Keyboard.isKeyDown(keyCode)
+                if (keyCode < Keyboard.KEYBOARD_SIZE) {
```
hannibal002, line 121: this line is the only fix here?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/KeyboardManager.kt
hannibal002: didnt see the comment above

src/main/java/at/hannibal2/skyhanni/utils/KeyboardManager.kt
hannibal002: didnt see the comment above

src/main/java/at/hannibal2/skyhanni/utils/KeyboardManager.kt
CalMWolfs: that is the only fix line here

hannibal002 (approved):

PR #3842, opened 2025-04-19
Backend: EquipmentApi

ItsEmpa:
## What
This PR adds an EquipmentApi that stores what equipment you have equipped, both inside and outside rift.

+ Added EquipmentApi. - Empa


Thunderblade73 (approved):
Code looks fine

src/main/java/at/hannibal2/skyhanni/features/inventory/EquipmentApi.kt
```diff
@@ -0,0 +1,113 @@
...
+    @HandleEvent(onlyOnSkyblock = true)
+    fun onInventoryUpdate(event: InventoryOpenEvent) {
+        if (!inventory.isInside()) return
+        EquipmentSlot.entries.forEach {
+            handleInventoryItem(it, event.inventoryItems[it.slot])
+        }
+    }
+
+    @HandleEvent(onlyOnSkyblock = true)
+    fun onRightClick(event: ItemClickEvent) {
```
CalMWolfs:
```suggestion
    fun onItemClick(event: ItemClickEvent) {
```

src/main/java/at/hannibal2/skyhanni/features/inventory/EquipmentApi.kt
```diff
@@ -0,0 +1,113 @@
...
+    fun onDebug(event: DebugDataCollectEvent) {
+        event.title("Equipment")
+        event.addIrrelevant {
+            val storage = storage ?: return@addIrrelevant
+            add("§aEquipment:")
+            storage.slots.forEach { item ->
+                val name = item?.displayName.toString()
+                add(name)
+            }
+            add("§aRift Equipment:")
```
CalMWolfs: I dont think ive seen adding colour codes to debug as it sends in github, 
maybe for each equipment you want to do `add(" - $name")` or `add(" $name")`

src/main/java/at/hannibal2/skyhanni/features/inventory/EquipmentApi.kt
```diff
@@ -0,0 +1,113 @@
...
+    BELT(28, ItemCategory.BELT),
+    GLOVES(37, ItemCategory.GLOVES, ItemCategory.BRACELET),
+}
+
+@SkyHanniModule
+object EquipmentApi {
+
+    val inventory = InventoryDetector { it == "Your Equipment and Stats" }
+
+    private val stained_glass_pane = Item.getItemFromBlock(Blocks.stained_glass_pane)
```
CalMWolfs: While the underscores match forge here I feel it could just be either `stainedGlassPane` or `glassPane`

CalMWolfs (changes requested):
a few small things i saw

src/main/java/at/hannibal2/skyhanni/features/inventory/EquipmentApi.kt
ItsEmpa: i forgot to actually check that the click is a right click in the code. added it now, is it fine to keep the function name as onRightClick then?

hannibal002: i would prefer for consistency sake to change it

hannibal002: we dont use funciton names to indicate what they check, but only on what event they fire upon

src/main/java/at/hannibal2/skyhanni/features/inventory/EquipmentApi.kt
```diff
@@ -0,0 +1,115 @@
...
+            add("Rift Equipment:")
+            storage.riftSlots.forEach { item ->
+                val name = item?.displayName.toString()
+                add(" - $name")
+            }
+        }
+    }
+
+    private fun handleInventoryItem(slot: EquipmentSlot, itemStack: ItemStack?) {
+        val item = if (itemStack?.item == stainedGlassPane) null else itemStack
```
CalMWolfs: to make sure this check if the item is a glass pane works properly on modern this should instead use `fun ItemStack.isStainedGlassPane` from `ColoredBlockCompat`

CalMWolfs (changes requested):
just this 1 thing for multiversion

CalMWolfs (approved):

PR #3064, opened 2024-12-14
Backend: Allow Parameterless Handling of SkyHanniEvent(s)

DavidArthurCole:
## What
For the cases where `@HandleEvent` is needed solely as the starter for a code snippet, where the normal `event: ...` function construction would warn about an unused param.
Probably the most divisive PR I've opened :D

<details>
<summary>Images</summary>

image

Screenshot 2024-12-14 030346

</details>

## Changelog Technical Details
+ SkyHanniEvents can now be `@Handle`-d without an `event:` parameter or a HandleEvent parameter if a default function name is provided for the event. - Daveed

PR #3852, opened 2025-04-20
Backend: event execution time regression

ThatGravyBoat:
## What
Fixes a regression that happened in #3064 where execution was changed to method invoke which is significantly slower than lambda meta factory by a significant amount as lambda meta factory is almost as fast as direct method execution call while method invoke has to check security manager and and access which makes its best case like 2x slower (you will rarely ever get JIT to reach this state) and its worse in the 100s of times slower.

+ Improved event execution speed. - ThatGravyBoat
+ Fixed event-fail invocation wrapping errors. - ThatGravyBoat

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/event/EventListeners.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/event/EventListeners.kt
@@ -70,14 +70,17 @@ class EventListeners private constructor(val name: String, private val isGeneric
             }
         }
 
-        return { _: Any -> method.invoke(instance) }
+        val runnable = ReflectionUtils.createRunnableFromMethod(instance, method)
+        return { _: Any -> runnable.run() }
     }
 
     private fun createSingleParameterConsumer(method: Method, instance: Any): (Any) -> Unit {
         require(SkyHanniEvent::class.java.isAssignableFrom(method.parameterTypes[0])) {
             "Method ${method.name} parameter must be a subclass of SkyHanniEvent."
         }
-        return { event -> method.invoke(instance, event) }
+
+        val consumer = ReflectionUtils.createConsumerFromMethod(instance, method)
+        return { event -> consumer.accept(event) }
     }
 
     private fun resolveGenericType(method: Method): Class<*> =
--- a/src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvents.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvents.kt
@@ -170,6 +170,7 @@ object SkyHanniEvents {
         classes.add(clazz)
 
         var current = clazz
+        @Suppress("LoopWithTooManyJumpStatements")
         while (current.superclass != null) {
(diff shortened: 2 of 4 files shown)


jani270 (dismissed):
Please either change the changelog to be a Technical Details Change or word it differently so the user can understand it.

ThatGravyBoat:
> Please either change the changelog to be a Technical Details Change or word it differently so the user can understand it.

Changed it to backend and technical change

DavidArthurCole (approved):
🙏

DavidArthurCole:
Might be worth adding to the technical change log that this detaches the handlers again, as with the method invokers it was polluting stacktraces: (https://discord.com/channels/997079228510117908/1094190239532208228/1360290368125534308)

hannibal002:
Does this remove the error entry in the list of stack trace? I could remove some special error handling then

DavidArthurCole:
> Does this remove the error entry in the list of stack trace? I could remove some special error handling then

Kinda, it's no longer at the top of the stack like it was before.

```
@HandleEvent
fun onWorldChange() {
	throw IllegalStateException("Test forcing an exception")
}
```

```
SkyHanni 2.17.0: Caught a IllegalStateException in at.hannibal2.skyhanni.data.BossbarData.onWorldChange() at WorldChangeEvent: Test forcing an exception
 
Caused by java.lang.IllegalStateException: Test forcing an exception
	at SH.data.BossbarData.onWorldChange(BossbarData.kt:20)
	at SH.api.event.EventListeners.createZeroParameterConsumer$lambda$7(EventListeners.kt:74)
	at SH.api.event.EventListeners.addListener$lambda$2(EventListeners.kt:38)
	at SH.api.event.SkyHanniEvent.prePost(SkyHanniEvent.kt:25)
	at SH.api.minecraftevents.ClientEvents.onWorldChange(ClientEvents.kt:29)
```

hannibal002:
Is this with the edge case thingy removed already?
Also unrelated to your change ere but maybe useful to do in this pr anyway: we need to remove the lines starting with event in the stack trace again

ItsEmpa (approved):

NopoTheGamer:
For people who may arive on this pr in the future and arent in the know with lambda meta factories, here are the supposed benchmarks for this performance pr that was argued to be not included in the original pr 
image

CalMWolfs (approved):
worked for me

hannibal002 (approved):
i have added even more stuff to remove the now unnecesssary workaround that hides in the `ErrorManager` and finally added proper error manager lines detection support for the "new" `createZeroParameterConsumer` and `createSingleParameterConsumer` functions.

```
SkyHanni 2.17.0: Caught a IllegalStateException in at.hannibal2.skyhanni.data.EntityData.onWorldChange() at WorldChangeEvent: Test forcing an exception
 
Caused by java.lang.IllegalStateException: Test forcing an exception
	at SH.data.EntityData.onWorldChange(EntityData.kt:50)
<Skyhanni event post>
```

PR #3854, opened 2025-04-21
Backend: Keybindings

rueblimaster:
## Dependencies
-  #3871

## What

<details>
<summary>Images</summary>

</details>

+ Added Keybinding to simplify keybindings. - rueblimaster


hannibal002:
I'd rather see an key up event that also includes the "last key up of same key duration" as a parameter

I see you copied the approach for the inventory render helper here. But I'm not sure I like this solution for keybinds

rueblimaster:
only going to be done after
- https://github.com/hannibal002/SkyHanni/pull/3871

CalMWolfs:
Have not looked through this pr yet but I would like this to not be merged before we have a working 1.21 version. As i dont want to debug completely new code after already changing a bunch of stuff to get key events to work on 1.21. Merging this will just mean there is more uncertainty as to what broke key events and will slow down fixing it as well. 
I will take it off soon but it can probably be re-added in a day or two

rueblimaster:
> Have not looked through this pr yet but I would like this to not be merged before we have a working 1.21 version. As i dont want to debug completely new code after already changing a bunch of stuff to get key events to work on 1.21. Merging this will just mean there is more uncertainty as to what broke key events and will slow down fixing it as well. I will take it off soon but it can probably be re-added in a day or two

This doesn't change the key events in any way or form, only the dependency does

rueblimaster:
hmmm, yeah fair enough
but ig something like this may be wanted/needed in the future again, but idk

PR #3863, opened 2025-04-22
Backend: Add warning to shdebug if optifine or patcher are not installed

NopoTheGamer:
## What
Adds a thingy to shdebug if you dont have optifine or patcher installed

+ Added warning in `/shdebug` if Optifine or Patcher isn’t installed. - nopo

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/ComputerEnvDebug.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/ComputerEnvDebug.kt
@@ -11,7 +11,10 @@ import at.hannibal2.skyhanni.utils.NumberUtil.roundTo
 import at.hannibal2.skyhanni.utils.RegexUtils.matchMatcher
 import at.hannibal2.skyhanni.utils.TimeUtils.format
 import at.hannibal2.skyhanni.utils.system.PlatformUtils
+//#if MC == 1.8.9
 import net.minecraftforge.fml.client.FMLClientHandler
+import net.minecraftforge.fml.common.Loader
+//#endif
 import java.lang.management.ManagementFactory
 import kotlin.math.roundToInt
 import kotlin.time.Duration.Companion.hours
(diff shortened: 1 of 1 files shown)


hannibal002:
please do the same for patcher

src/main/java/at/hannibal2/skyhanni/utils/ComputerEnvDebug.kt
```diff
@@ -216,6 +223,31 @@ object ComputerEnvDebug {
...
+        event.title("Optifine")
+        if (hasOptifine) {
+            event.addIrrelevant("Optifine is installed")
+        } else {
+            event.addData("Optifine is not installed")
+        }
+        if (hasPatcher) {
+            event.addIrrelevant("Patcher is installed")
+        } else {
+            event.addData("Patcher is not installed")
```
hannibal002: this would still have the title "optifine".
maybe change the title to "performance mods"?

src/main/java/at/hannibal2/skyhanni/utils/ComputerEnvDebug.kt
```diff
@@ -216,6 +223,31 @@ object ComputerEnvDebug {
...
+        }
+        if (!hasOptifine || !hasPatcher) {
+            event.addData("Optifine or Patcher are not installed")
```
hannibal002: this is not the proper way to add multi lines in shdebug. this will throw errors as each line assumes it has a new title.
instead, use 
```
addData{
add("string")
add("string2")
}
```

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/ComputerEnvDebug.kt
NopoTheGamer: fixed i think

hannibal002 (changes requested):
You can include the info what mod is present when only of them is missing, and merge the "x and y is present" line into one

hannibal002:
also only link to them when they are missing

PR #3880, opened 2025-04-24
Improvement: Ember Armor Trophy Fishing

DavidArthurCole:
## What
Adds a suboption to allow for showing the trophy fishing display when you're wearing trophy armor or ember armor, cause slug tech.

## Changelog Improvements
+ Added ability to show trophy fish display while wearing Ember Armor. - Daveed


DavidArthurCole:
image

I think I finally managed to write a migration that doesn't wipe the config

hannibal002:
Do we need that option to choose? I'd say just combine ember and hunter armour in one option as it was before with just hunter.

DavidArthurCole:
> Do we need that option to choose? I'd say just combine ember and hunter armour in one option as it was before with just hunter.

I didn't wanna risk there being some other obscure ember armor tech existing that would make it pop up when it's inconvenient. Idk, I don't see any issue in giving the choice, people are getting the same experience they've had via the migration.

lunaynx:
As long as you make sure it doesn't work in Kuudra, it should be fine.

It seems to check for Crimson Isle, so it's probably fine.

DavidArthurCole:
> Do we need that option to choose? I'd say just combine ember and hunter armour in one option as it was before with just hunter.

(A month and a half later)
Yeah - reverted it back and just changed the config description.

PR #3883, opened 2025-04-24
Fix: Lane Switch notification for farms with soulsand

CalMWolfs:
## What
Minecrafts constructor for when there are doubles floors before converting to int so we should do the same.

As can be seen to match the top you must do the bottom
image

## Changelog Fixes
+ Fixed Lane Switch notification not working for farms with soulsand. - CalMWolfs

+ Floor before converting to ints for LorenzVec -> BlockPos. - CalMWolfs


hannibal002:
i feel like this can have unexpected conflicts in other features

CalMWolfs:
> i feel like this can have unexpected conflicts in other features

It was changed to use .toInt like a week or two ago, this here is simply keeping the 3 int constructor but bringing back the correct behaviour

PR #3886, opened 2025-04-25
Feature + Fix: Farming Weight Monthly LB and Fix Upcoming Players

ptlthg:
## What
This PR does the following:
- Adds monthly weight leaderboard option
- Updates fallback weight values
- Loads more upcoming players when a player is high up in the leaderboard
- Uses the locally calculated rank when fetching new upcoming players

## Changelog New Features
+ Added Monthly Farming Weight Leaderboard option. - Ke5o

## Changelog Fixes
+ Fixed farming weight leaderboard upcoming players repeating. - Ke5o
    * When elitebot.dev doesn't update your data the same upcoming players would be repeated, this fixes that.


src/main/java/at/hannibal2/skyhanni/config/features/garden/EliteFarmingWeightConfig.java
```diff
@@ -29,6 +29,27 @@ public class EliteFarmingWeightConfig {
     @ConfigEditorBoolean
     public boolean leaderboard = true;
 
+    @Expose
+    @ConfigOption(name = "Leaderboard Type", desc = "Select normal or monthly weight leaderboard!")
+    @ConfigEditorDropdown
+    public Property<EliteFarmingWeightLbType> eliteLbType = Property.of(EliteFarmingWeightLbType.DEFAULT);
+
+    public enum EliteFarmingWeightLbType {
+        DEFAULT("Normal"),
```
hannibal002: please have a more descriptive name for default. is this total lb? like, "all time" stats?

src/main/java/at/hannibal2/skyhanni/features/garden/farming/FarmingWeightDisplay.kt
```diff
@@ -239,7 +251,7 @@ object FarmingWeightDisplay {
             }
         }
 
-        val totalWeight = (localWeight + weight)
+        val totalWeight = (localWeight + weight - shWeightDiff)
```
hannibal002: this formular `localWeight + weight - shWeightDiff` gets used in three spots in the code now. please use a function

src/main/java/at/hannibal2/skyhanni/features/garden/farming/FarmingWeightDisplay.kt
```diff
@@ -441,34 +455,68 @@ object FarmingWeightDisplay {
...
+        val includeUpcoming = when {
+            !isEnabled() -> ""
+            leaderboardPosition > 10_000 -> "?upcoming=50"
+            leaderboardPosition > 5_000 -> "?upcoming=30"
+            leaderboardPosition > 1_000 -> "?upcoming=20"
+            else -> "?upcoming=10"
+        }
+        // Tell the API to get upcoming players from our local rank (for when new data isn't fetched), or fallback to the
+        // provided eta goal rank from the config
+        val atRank = when {
```
hannibal002, line 490: please move the string `&atRank=` outside the `when`. rather, use `nullable int` in the `when`, and then build the `atRank` string together after the `when`

src/main/java/at/hannibal2/skyhanni/features/garden/farming/FarmingWeightDisplay.kt
```diff
@@ -196,11 +207,12 @@ object FarmingWeightDisplay {
 
         if (rankGoal == -1) rankGoal = getRankGoal()
         val leaderboard = getLeaderboard()
+        val lbType = if (isMonthlyLb()) " Monthly" else ""
 
         val list = mutableListOf<Renderable>()
         list.add(
             Renderable.clickable(
-                "§6Farming Weight§7: $weight$leaderboard",
+                "§6Farming Weight$lbType§7: $weight$leaderboard",
```
hannibal002: when you change the wording here to be "Monthly Farming Weight" you can merge three of the four isMonthglyLB checks

src/main/java/at/hannibal2/skyhanni/features/garden/farming/FarmingWeightDisplay.kt
```diff
@@ -381,6 +394,7 @@ object FarmingWeightDisplay {
     private fun inGardenEnabled() = (LorenzUtils.inSkyBlock && GardenApi.inGarden()) || config.showOutsideGarden
 
     private fun isEtaEnabled() = config.overtakeETA
+    private fun isMonthlyLb() = config.eliteLbType.get() == EliteFarmingWeightConfig.EliteFarmingWeightLbType.MONTHLY
```
hannibal002: make the last B upper case, as Leaderboard is a acronym and we write them uppercase when 2 chars long. for 3 char or longer we only uppercase the first char.

hannibal002 (changes requested):
>     Added Monthly Farming Weight leaderboard option. - Ke5o

this is a improvement and no entirely new feature imo

src/main/java/at/hannibal2/skyhanni/features/garden/farming/FarmingWeightDisplay.kt
ptlthg: I have no idea what you mean with this, but I did remove an extra lbType assignment.

ptlthg:
> > ```
> > Added Monthly Farming Weight leaderboard option. - Ke5o
> > ```
> 
> this is a improvement and no entirely new feature imo

But being able to use the monthly leaderboard is a new thing that the mod couldn't do before. I would say it's a small feature, but still a feature rather than an improvement. If this fits better as an improvement in the changelog just because similar things have been labeled improvements, I can change it, but I do think this is a feature.

src/main/java/at/hannibal2/skyhanni/features/garden/farming/FarmingWeightDisplay.kt
hannibal002: at the time of my review, the variable isMonthglyLB was used four times in the class. three of those four times were qual, with the only change if the empty space is prefix or suffix.
if you now change this line to have the bl type show at the middle and not the end, you can move the emtpy space from prefix to suffix in this example, and you can create a function `getLbTypeDisplay()` in order to reduce three identical lines

hannibal002:
> But being able to use the monthly leaderboard is a new thing that the mod couldn't do before. I would say it's a small feature, but still a feature rather than an improvement. If this fits better as an improvement in the changelog just because similar things have been labeled improvements, I can change it, but I do think this is a feature.

fair. keep it as a feature.
normally we only use new feature for actual new feature toggles, but here this is a big upgrade to the exisitng feature

hannibal002:
image

src/main/java/at/hannibal2/skyhanni/features/garden/farming/FarmingWeightDisplay.kt
hannibal002: [`fe3524b` (#3886)](https://github.com/hannibal002/SkyHanni/pull/3886/commits/fe3524b47f05ea0fe7a51e3d1112be2b162e2511)

PR #3889, opened 2025-04-26
Fix: Burrow types not showing

CalMWolfs:
## What
There was a floating point error converting floats to doubles. So we round to 2 decimal places. Also the rounding to block location was removed at some point so thats back (probably my fault)

## Changelog Fixes
+ Fixed burrows not showing their type. - CalMWolfs


mariethepolarfox:
Lgtm

hannibal002:
Doese this mean we need to use
`event.location.toBlockPos().down().toLorenzVec()`
every time we wanna have the exact location in similar events? this surely is not the solution.

i can merge this as a workaround, but there should be a TODO to remove this and fix the actual problem.

PR #3892, opened 2025-04-26
Improvement: Detect Visitors in Proximity

DavidArthurCole:
## What
Adds fallback entity detection to the garden chat blocker so that if a visitor is clicked before they appear in tab, chats are still blocked.

## Changelog Improvements
+ Added proper blocking of visitor chat messages even if the visitor isn’t in the Tab Widget yet. - Daveed


hannibal002:
looks more like a bug fix to me

CalMWolfs (approved):
code lgtm but should probably be a fix pr

PR #3894, opened 2025-04-27
Fix: Holographic entities depending on neu

CalMWolfs:
## What
Horse mixin v2, thanks for the mixins neu
also made code better 🔥 

## Changelog Fixes
+ Fixed Craft Room Holographic Helper requiring NEU to be installed. - CalMWolfs


hannibal002:
slime size and spider looking at direction are off. (tested with neu in dev env)
image


will check if those bugs already exist without the pr

hannibal002:
also the zombie doesnt have his head
image


all those bugs are independent of this pr

PR #3898, opened 2025-04-28
Fix: Prevent burrows from being removed incorrectly

Bloxigus:
## What
This PR _should_ fix the issue, might cause a `ConcurrentModificationException` due to editing a map while iterating over that map

## Changelog Fixes
+ Fixed burrows being removed incorrectly when more than 16 blocks away. - bloxigus
+ Fixed Multi Guess not saving guess locations sometimes. - bloxigus

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/event/diana/GriffinBurrowHelper.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/event/diana/GriffinBurrowHelper.kt
@@ -147,7 +147,7 @@ object GriffinBurrowHelper {
         val newLocation = calculateNewTarget()
         if (targetLocation != newLocation) {
             targetLocation = newLocation
-            // add island graphs here some day when the hub is fully added in the graph
+            // TODO: add island graphs here some day when the hub is fully added in the graph
 //             newLocation?.let {
 //                 IslandGraphs.find(it)
 //             }
@@ -189,12 +189,8 @@ object GriffinBurrowHelper {
         if (newLocation.distance(playerLocation) < 6) return
 
         latestGuess?.let {
-            if (it.precise && config.multiGuesses) {
-                if (event.new) {
-                    if (it.getLocation() !in particleBurrows) {
-                        additionalGuesses.add(it)
-                    }
-                }
+            if (it.precise && config.multiGuesses && event.new && it.getLocation() !in particleBurrows) {
+                additionalGuesses.add(it)
             }
         }
 
@@ -214,10 +210,8 @@ object GriffinBurrowHelper {
 
     private fun removePreciseGuess(location: LorenzVec) {
         latestGuess?.let {
-            if (it.precise) {
-                if (location == it.getLocation()) {
-                    latestGuess = null
-                }
+            if (it.precise && location == it.getLocation()) {
+                latestGuess = null
             }
         }
         additionalGuesses.removeIf { it.getLocation() == location }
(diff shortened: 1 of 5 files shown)


src/main/java/at/hannibal2/skyhanni/features/event/diana/GriffinBurrowHelper.kt
```diff
@@ -192,16 +190,10 @@ object GriffinBurrowHelper {
 
         latestGuess?.let {
             if (it.precise && config.multiGuesses) {
```
DavidArthurCole: Nothing wrong with your PR, but can we clean up some of the nested `if`s in this file please 🥺

```
if (x) {
	if (y) {
		z
	}
}
```

is gross

Bloxigus: That commit should fix the
```
if (a) {
    if (b) {
        if (c) {
            d
        }
    }
}
```

DavidArthurCole (approved):
Code LGTM

src/main/java/at/hannibal2/skyhanni/features/event/diana/SoopyGuessBurrow.kt
```diff
@@ -144,113 +144,112 @@ object SoopyGuessBurrow {
...
-                        } else {
-                            LorenzVec(floor(p2.x), 255.0, floor(p2.z))
-                        }
-                        BurrowGuessEvent(finalLocation, precise = false).post()
+                    val finalLocation = if (d1 < d2) {
+                        LorenzVec(floor(p1.x), 255.0, floor(p1.z))
+                    } else {
+                        LorenzVec(floor(p2.x), 255.0, floor(p2.z))
                     }
+                    BurrowGuessEvent(finalLocation, precise = false, new = false).post()
```
hannibal002, line 236: having new always false here should break multi guess support for soopy guess logic entirely. if this is intended, please mark it in the pr desc

hannibal002 (changes requested):
is there a need for the soopy guess class change? feels unintuive, or is this bc its an entirely different approach? i assumed we wanna deprecate/delete the old guess logic as the new one works as well. why are you updating the code?

hannibal002:
Also please update the changelog/pr text to clearly indicate that this is about the guess point, not the nearby burrow points

src/main/java/at/hannibal2/skyhanni/features/event/diana/GriffinBurrowParticleFinder.kt
```diff
@@ -204,16 +204,14 @@ object GriffinBurrowParticleFinder {
         var hasEnchant: Boolean = false,
         var type: Int = -1,
         var found: Boolean = false,
-        var burrowTimeToLive: Int = 0,
+        var burrowTimeToLive: Int = 1,
```
hannibal002: is this the fix for the quickly disappearing and then reappearing of nearby burrows? if so, mention this in the changelog additonally as this fixed then two unrelated bugs

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/event/diana/SoopyGuessBurrow.kt
Bloxigus: I don't think soopy guess logic ever supported multi guess
also all the changes in this file are entirely cosmetic and shouldnt change behavior

src/main/java/at/hannibal2/skyhanni/features/event/diana/GriffinBurrowParticleFinder.kt
Bloxigus: This should probably be `ceil(ping / 50ms)` to account for the time between taking out the shovel and particles first appearing.
Does skyhanni store the user's ping somewhere?

hannibal002: `CurrentPing.averagePing` exists. but note that the ping detection logic is optionally. if the user disables the dev option, the data is invalid

src/main/java/at/hannibal2/skyhanni/features/event/diana/SoopyGuessBurrow.kt
hannibal002: fine

PR #3900, opened 2025-04-28
Backend: Ignore some types in /shtrackparticles

hannibal002:
## What
added option to ignore particle types from debug command

+ Added ignore option for particle types in `/shtrackparticles`. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/test/command/TrackParticlesCommand.kt
+++ b/src/main/java/at/hannibal2/skyhanni/test/command/TrackParticlesCommand.kt
@@ -36,7 +36,7 @@ object TrackParticlesCommand {
     private val position get() = SkyHanniMod.feature.dev.debug.trackParticlePosition
 
     private var display: List<Renderable> = emptyList()
-    private var worldParticles: Map<LorenzVec, List<ReceiveParticleEvent>> = emptyMap()
+    private var worldParticles = emptyMap<LorenzVec, List<ReceiveParticleEvent>>()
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/test/command/TrackParticlesCommand.kt
```diff
@@ -71,6 +88,9 @@ object TrackParticlesCommand {
         }
     }
 
+    fun EnumParticleTypes.getByIdOrNull(name: String): EnumParticleTypes? =
+        EnumParticleTypes.entries.firstOrNull { it.name.equals(name, ignoreCase = false) }
```
CalMWolfs:
```suggestion
    private fun getParticleTypeByName(name: String): EnumParticleTypes? =
        EnumParticleTypes.entries.firstOrNull { it.name.equals(name, ignoreCase = true) }
```

I feel this is more appropriate than an extension function that has to be called on an entry of the enum as then above you can just do
`val type = getParticleTypeByName(name)`

This also makes it easier to get this working on 1.21

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/test/command/TrackParticlesCommand.kt
hannibal002: good catch, `EnumParticleTypes.CRIT.getByIdOrNull(name)` this line above is obviously not intended to say.

CalMWolfs (approved):

PR #3901, opened 2025-04-28
Fix: No hub warps unlocked error

hannibal002:
## What
Fixed an error message when trying to warp with no hub warp point unlocked

Just a workaround, in fact, there is no chance that `/hub` gets marked as not unlocked. idk whats going on exactly

## Changelog Fixes
+ Fixed error message when warping to a burrow with no unlocked hub warp point. - hannibal2


CalMWolfs (approved):
lgtm apart from detekt

PR #3904, opened 2025-04-28
Fix: Elite Config Option

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1366455153854648380
Expose was missing - this is why we need all configs to kotlin.

## Changelog Fixes
+ Fixed issue with Elite Farming config not storing eta override. - Daveed


DavidArthurCole:
why does intellij suck so much at converting java to kotlin in a clean manner

CalMWolfs:
which option was missing an expose? i dont see it in this pr

DavidArthurCole:
Pasta fazool I am a fool

PR #3905, opened 2025-04-28
Backend: Position Construction Cleanup

DavidArthurCole:
## What
Removes a bunch of instances of needless 4th parameter `true` Position instantiations.

exclude_from_changelog


CalMWolfs (approved):

hannibal002 (changes requested):
pretty sure almost all those cases can also lose the third parameter being false

DavidArthurCole:
> pretty sure almost all those cases can also lose the third parameter being false

yep, done.

hannibal002 (changes requested):
you already have 100 lines changed, what are 30 more?
grafik

hannibal002:
at a glance, all except the otherWikiCommands line can be replaced

DavidArthurCole:
> at a glance, all except the otherWikiCommands line can be replaced

What?

image

hannibal002:
i just pushed this
6a097329e1745fc69ee25b8ad44eefa6ea3ddfee

now, every instance of `, false, true)` can be replaced to `)`

DavidArthurCole:
How in the hell is changing the wiki command at all related to `Position` constructors?

PR #3906, opened 2025-04-28
Backend: Config Vals

DavidArthurCole:
## What
Adds a detekt rule for enforcing non-mutable config types to `val` instead of `var`. Grabbed some common logic out of the `StorageNeedsExpose` rule and made a util.

+ Added detekt rule to enforce non-mutable config types to be val instead of var. - Daveed
+ Changed some of the config-referencing helper functions to use KProperty0 instead of Mutable. - Daveed


hannibal002:
merge conflicts with https://github.com/hannibal002/SkyHanni/pull/3905 are fun

DavidArthurCole:
Surely there won't be another 33 detekt failures since I last merged beta in right

DavidArthurCole:
> Surely there won't be another 33 detekt failures since I last merged beta in right

technically i was right, it was 34

PR #3909, opened 2025-04-29
Add back gl statements

CalMWolfs:
Add back the gl statements

see https://github.com/hannibal002/SkyHanni/commit/8987c4edab558ff3fa0a205bca9a40f4db9260b4

exclude_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/renderables/container/ContainerRenderable.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/renderables/container/ContainerRenderable.kt
@@ -1,0 +1,85 @@
+package at.hannibal2.skyhanni.utils.renderables.container
+
+import at.hannibal2.skyhanni.data.model.TextInput
+import at.hannibal2.skyhanni.utils.RenderUtils.HorizontalAlignment
+import at.hannibal2.skyhanni.utils.RenderUtils.VerticalAlignment
+import at.hannibal2.skyhanni.utils.renderables.Renderable
+import at.hannibal2.skyhanni.utils.renderables.Renderable.Companion.filterList
+import at.hannibal2.skyhanni.utils.renderables.RenderableUtils.renderXAligned
+import at.hannibal2.skyhanni.utils.renderables.RenderableUtils.renderYAligned
+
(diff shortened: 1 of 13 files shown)


src/main/java/at/hannibal2/skyhanni/utils/renderables/container/ContainerRenderable.kt
```diff
@@ -0,0 +1,94 @@
...
+}
+
+open class VerticalContainerRenderable(
+    renderables: Collection<Renderable>,
+    spacing: Int = 0,
+    horizontalAlign: HorizontalAlignment = HorizontalAlignment.LEFT,
+    verticalAlign: VerticalAlignment = VerticalAlignment.TOP,
+) : ContainerRenderable(renderables, spacing, horizontalAlign, verticalAlign) {
+
+    override val width: Int
```
Thunderblade73: Remove the get() = in this file for width and height. That makes so that every render frame they values get recalculated. Which is not need. Idk what is going on with the vertical search container, as that one would be wrong than, but that was always wrong to begin with so it should not matter, also we only have like 2 uses of that, so it should not hurt so much.

src/main/java/at/hannibal2/skyhanni/utils/renderables/container/ContainerRenderable.kt
```diff
@@ -0,0 +1,94 @@
...
+        get() = renderables.sumOf { it.height } + spacing * (renderables.size - 1)
+
+    override fun render(posX: Int, posY: Int) {
+        var y = posY
+        var totalOffset = 0
+        renderables.forEach {
+            it.renderXAligned(posX, y, width)
+            y += it.height + spacing
+            DrawContextUtils.translate(0f, (it.height + spacing).toFloat(), 0f)
+            totalOffset += it.height + spacing
```
Thunderblade73: The total tracking is not needed, as the end result is always (height + spacing) same for the horizontal but there it is width.

Thunderblade73 (changes requested):

Thunderblade73 (approved):

PR #3910, opened 2025-04-30
Fix: Disallowing negative scales for the config values

Thunderblade73:
## Changelog Fixes
+ Fixed GUI positions going negative. - Thunderblade73


hannibal002:
please elaborate in the changelog a bit more.
the poisition value of what?

hannibal002 (changes requested):
see above

PR #3913, opened 2025-05-01
Fix: Burrows being really, really wrong + ConcurrentModificationException

Bloxigus:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/406

## What
Three fixes:
1. ConcurrentModificationException happened when an element is removed from the map while iterating over that map, which is undesirable

2. As seen [here](https://discord.com/channels/997079228510117908/1366507304190607421/1366555543354867804) sometimes the guesser makes a very wrong guess. Rather than finding the root cause (which is difficult due to being unable to replicate it) simply not showing guesses that are outside the places the player can reach should work the same.

3. Sometimes particles from other sources interfere with guesses ([here](https://discord.com/channels/997079228510117908/1355485985344848055/1355562195420975316)) and make the solver give an invalid solution. Adding bounds checks acts as a sanity check of where the guesses can be.

For testing use the repo https://github.com/Bloxigus/SkyHanni-REPO/tree/island-bounds

## Changelog Fixes
+ Fixed burrows outside island boundaries. - bloxigus
+ Fixed error when removing burrow. - bloxigus
+ Fixed other particles interfering with hotspot guesses. - bloxigus

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/IslandType.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/IslandType.kt
@@ -2,9 +2,11 @@ package at.hannibal2.skyhanni.data
 
 import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.api.event.HandleEvent.Companion.HIGHEST
+import at.hannibal2.skyhanni.data.jsonobjects.repo.IslandBounds
 import at.hannibal2.skyhanni.data.jsonobjects.repo.IslandTypeJson
 import at.hannibal2.skyhanni.events.RepositoryReloadEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
+import at.hannibal2.skyhanni.utils.LorenzVec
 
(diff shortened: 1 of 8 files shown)


hannibal002:
Burrows only exist in hub, you added the repo bounds to other islands as well. Do you plan on using this for fishing hotspot and Hoppity too?

hannibal002:
Also please change the format of the second line of the pr desc to
- link

Bloxigus:
> Burrows only exist in hub, you added the repo bounds to other islands as well. Do you plan on using this for fishing hotspot and Hoppity too?

Yes, I thought there was no harm adding it for islands outside the hub for future and it probably is a good idea for the fishing hotspot

hannibal002 (changes requested):
looking at this in next diana or when you exapnd this for hoppity/fishing

Bloxigus:
Added for hoppity/fishing, hopefully not noticeable

hannibal002 (approved):
LGTM, waiting on confirmation of @Bloxigus to merge

PR #3916, opened 2025-05-01
Backend: Fixed graph node edtior searcahble

hannibal002:
## What
calling GUI overaly render event in a screen breaks searchable.
This is a workaround to avoid this bug for the graph node editor

+ Fixed graph node editor not being searchable. - hannibal2


Thunderblade73:
Wouldn't simply using that event fix it for good?
https://github.com/hannibal002/SkyHanni/pull/3504/files#diff-41594487f6fc662ac8f1696b668c1169083a918a2fab61b1881d70aad765ee36

(In https://github.com/hannibal002/SkyHanni/pull/3504)

hannibal002:
yes this might also a more stable workaround. 
anyway, please add that event to the changelog technical details, as this is actually relevant (compared to most of the stuff we put in technical details lol)

PR #3918, opened 2025-05-01
Improvement: Added an Ironman Option to Profit Trackers

DanielMayhan:
## What
Added a toggle in the settings for the Crystal Nucleus, Fossil Excavator and Enchanting Table.
Changing the way Profits are being calculated.

Crystal Nucleus Tracker:
- Removed the inclusion of the Jungle Key & Automaton Parts from the Total Profit

<details>
<summary>Images</summary>

Screenshot 2025-05-01 153750
Screenshot 2025-05-01 153821

</details>

Fossil Excavator:
- Removed the inclusion of Scrap from the Total Profit

<details>
<summary>Images</summary>

Screenshot 2025-05-01 154138
Screenshot 2025-05-01 154349

</details>

Enchanting Table:
- Removed the starting cost from the Total Profit

<details>
<summary>Images</summary>

Screenshot 2025-05-01 153902
Screenshot 2025-05-01 154016

</details>

## Changelog Improvements
+ Different Profit Calculations for Ironman. - Mayhan

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/inventory/experimentationtable/ExperimentsProfitTrackerConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/inventory/experimentationtable/ExperimentsProfitTrackerConfig.java
@@ -9,6 +9,7 @@ import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorDraggableLi
 import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorSlider;
 import io.github.notenoughupdates.moulconfig.annotations.ConfigLink;
 import io.github.notenoughupdates.moulconfig.annotations.ConfigOption;
+import io.github.notenoughupdates.moulconfig.observer.Property;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -32,10 +33,9 @@ public class ExperimentsProfitTrackerConfig {
     public int timeDisplayed = 30;
 
     @Expose
-    @ConfigOption(name = "Ironman Profit Calculations", desc = "Removes the Cost of Experience Bottles from Profit Calculations. (Requires Restart)")
+    @ConfigOption(name = "Ironman Profit Calculations", desc = "Removes the Cost of Experience Bottles from Profit Calculations.")
     @ConfigEditorBoolean
-    public boolean isIronman = false;
-    // Unable to make the Overlay update automatically, needs fixing!! Therefore the "Requires Restart" note
+    public Property<Boolean> isIronman = Property.of(false);
 
     @Expose
     @ConfigLink(owner = ExperimentsProfitTrackerConfig.class, field = "enabled")
--- a/src/main/java/at/hannibal2/skyhanni/config/features/inventory/experimentationtable/ExperimentsProfitTrackerConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/inventory/experimentationtable/ExperimentsProfitTrackerConfig.java
@@ -32,6 +32,12 @@ public class ExperimentsProfitTrackerConfig {
     public int timeDisplayed = 30;
 
     @Expose
+    @ConfigOption(name = "Ironman Profit Calculations", desc = "Removes the Cost of Experience Bottles from Profit Calculations. (Requires Restart)")
+    @ConfigEditorBoolean
+    public boolean isIronman = false;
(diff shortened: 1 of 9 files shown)


NopoTheGamer:
https://github.com/hannibal002/SkyHanni/pull/3917 ?

DanielMayhan:
#3917 Only changes from what source the price is being calculated
This allows to completly remove the negative Profit calculations for Ironman players
As shown in the images

hannibal002:
After my source change pr (nopo mentioned it above) those values show up as 0 or -0 in the tracker. I think this pr additional is useful.
But instead of an config option, this can be checked directly from the profile data. See LorenzUtils.isIronman

DanielMayhan:
I could also use that, but I think that giving the player the option to toggle this on or off is better than just hard setting it

DanielMayhan:
Do I have to do anything here?

hannibal002:
im in favor of more options for the player, but in this case i dont see a benefit:
if the player does not play ironman, they always pay for the item. even if they crafted it manually, id argue this should not be considered normal. ironman players instead craft it.

for ironman players this is confusing. they never want the ah/bz price here, only the npc price. and those items dont have a npc price so they dont need the option to enable it.

having this default enabled confused the one group, default disabled the other group. id say dont give the option and make everyone happy.

lunaynx:
I disagree that it "should not be considered normal" for players on non-ironman profiles to grind items themselves. People should be able to play the game however they want, they may choose to buy a few specific items but otherwise play similarly to ironman.

This mindset within the SkyBlock community that "the best way to get something is always to do the best money making method and then buy it" is so bad. You might do this in some cases because either the item is really cheap or you just find grinding it yourself unfun or too difficult, and that's perfectly valid, but it is not the only way to play.

lunaynx:
Also, technically experience bottles can be bought from the Mage Emissary NPC, though there are discounts provided depending on the player's Mage Reputation, and also Seal of the Family.

hannibal002:
Valid points!

src/main/java/at/hannibal2/skyhanni/config/features/inventory/experimentationtable/ExperimentsProfitTrackerConfig.java
```diff
@@ -31,6 +31,12 @@ public class ExperimentsProfitTrackerConfig {
     @ConfigEditorSlider(minValue = 5, maxValue = 60, minStep = 1)
     public int timeDisplayed = 30;
 
+    @Expose
+    @ConfigOption(name = "Ironman Profit Calculations", desc = "Removes the Cost of Experience Bottles from Profit Calculations. (Requires Restart)")
+    @ConfigEditorBoolean
+    public boolean isIronman = false;
+    // Unable to make the Overlay update automatically, needs fixing!! Therefore the "Requires Restart" note
```
hannibal002: you can auto update the data:
1. change the config option from boolean to `Property<Boolean> name = Propety.of(default value)`
2. create a `ConfigLoadEvent` and a `onToggle`, `afterChange` or `whenChanged` function to listen to the config value change
3. when the value changes, set `TrackerManager.dirty` to `true`

search for "property.of" in the config, and see how those values are getting used for many examples.

hannibal002 (changes requested):
we can keep the options, as well as keep them separated in different settings for different trackers.

final question:
do we want this always enabled for ironamn players (ignore the option in those cases)
counter argument: maybe ironman players wanna use the npc price, e.g. exp bottles.
if we dont want to have this always enabled for ironman, we shoudl rename the options.

DanielMayhan:
> we can keep the options, as well as keep them separated in different settings for different trackers.
> 
> final question: do we want this always enabled for ironamn players (ignore the option in those cases) counter argument: maybe ironman players wanna use the npc price, e.g. exp bottles. if we dont want to have this always enabled for ironman, we shoudl rename the options.

I think we should leave it of by default and those who want to customize their game can enable it later

hannibal002:
i dont see the point of forcing ironman players to enable the feature with the name "ironman mode".
also, this might be actually counter productive if the player has multiple profiles.

DanielMayhan:
If you look at it like that, I agree. So it should just be based on the profile the player is playing on?

DanielMayhan:
Changed it so that the feature is only available on ironman and the setting is enabled, which it now is by default

CalMWolfs:
@DanielMayhan  is this pr still getting worked on or shall i close it?

NopoTheGamer:
reopen if you want to work on it again :)

PR #3924, opened 2025-05-02
Fix: Pest Particle Waypoints

ItsEmpa:
## What
After 373 days since it was broken by Hypixel, pest waypoints are finally back! 🎉

Also im unsure whether to remove the pest particle line code, so i decided to just remove the SkyHanniModule annotation from it for the time being.

## Changelog Fixes
+ Fixed Pest Particle Waypoints. - Empa
    * This replaces the Pest Particle Line feature, which was supposed to be a temporary fix when Pest Particle Waypoints was broken.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/garden/pests/PestsConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/garden/pests/PestsConfig.kt
@@ -19,6 +19,7 @@ class PestsConfig {
     @Accordion
     var pestFinder: PestFinderConfig = PestFinderConfig()
 
+    // TODO renmae to waypoint
     @Expose
     @ConfigOption(name = "Pest Waypoint", desc = "")
     @Accordion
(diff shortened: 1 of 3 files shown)


DavidArthurCole (approved):
image

After detekt is fixed, LGTM - guess waypoints are off by maybe 0.5-1 blocks in each direction, but that's probably expected given entities move around

hannibal002 (changes requested):
when the pest is in another plot, the accuracy is sometimes off by 5-10 blocks. sometimes its super accurate. not sure if the bezierFitter logic is wrongly implemented or if the particles actually show just to the middle of the plot instead of the exact pest sometimes

ItsEmpa:
> when the pest is in another plot, the accuracy is sometimes off by 5-10 blocks. sometimes its super accurate. not sure if the bezierFitter logic is wrongly implemented or if the particles actually show just to the middle of the plot instead of the exact pest sometimes

its a known thing that when pests are not loaded, hypixel points to the center of the plot instead.

hannibal002 (approved):

PR #3929, opened 2025-05-05
Backend: Floats cleanup

hannibal002:
## What
For some reason, we started using `0.0f` in the code for floats, even though `0f` represents the same in fewer chars.
And then we kept coping over and over this bad habit.

This PR does the bold action of replacing all such cases everywhere, in the hopes to increase readability.

+ Removed unnecessary float decimals. - hannibal2


NopoTheGamer:
> Removed unnecessary commas in floats

these are objectively not commas
I think you want "Period" or "Decimal"

PR #3930, opened 2025-05-05
Backend: Brigadier Commands

ItsEmpa:
## What
This PR adds support for creating commands using brigadier, and converts the `/shignore` command to use this system.

+ Added Brigadier commands support. - Empa


CalMWolfs (approved):

hannibal002 (changes requested):
idk how much or how deep in brigadier we need to change stuff in order to fix this, but this needs fixing:

image

previously, we would send this
image

hannibal002:
also the logic is wrong when no para is defined. it should show the error message above instead 
image

PR #3903, opened 2025-04-28
Backend: Config Disabling via repo

NopoTheGamer:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/403
- https://github.com/hannibal002/SkyHanni/pull/3908

## What
Adds the ability to disable config options via the repo
nea coded this for neu and i basically copy and pasted it into skyhanni

<details>
<summary>Images</summary>

image

</details>

## Changelog Technical Details
+ Added ability to disable config options via repo. - nopo & nea
+ Updated MoulConfig to 3.9.0. - nopo

PR #3931, opened 2025-05-06
Backend: Added the ability to disable config options from showing per version

NopoTheGamer:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/3903

## What
Added two annotations, OnlyModern and OnlyLegacy, which when put onto a config option makes it not show in the config
The variable still exists with whatever the user set it to, so that if a config is swapped between versions every option saves

+ Added ability to disable config options per version. - nopo


hannibal002:
@NopoTheGamer bump

PR #3935, opened 2025-05-06
Backend: Fix Ide no longer dies in K2 mode with live plugin

Thunderblade73:
## What
Not really a fix. It just hides the error and softly errors instead of killing the whole kotlin plugin.

+ K2 mode & live plugins no longer error. - Thunderblade73
    * The errors are just silents, so the someone should fix them.


Thunderblade73:
I think I fail wrong for the events...
image

CalMWolfs (approved):

PR #3944, opened 2025-05-07
fix crash in neu overlays

CalMWolfs:
Have fun in your storage and pv 

exclude_from_changelog


hannibal002:
liar. didnt crash in pv. only storage

PR #3946, opened 2025-05-08
Backend: Deprecate LorenzUtils, Adding SkyBlockUtils

hannibal002:
## What
The class is a "core utils" since the beginning.
Over time, some other functions landed there that obviously should belong somewhere else.

This PR moves all functions into better, more precise other files.
For the remaining core utils, we now have a new `SkyBlockUtils` class.

Since it's practically impossible to move all the 450 affected lines in one PR, we rather make everything deprecated, and use the new functions over time.

+ Deprecated `LorenzUtils`. - hannibal2
+ Created `SkyBlockUtils`. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/test/SkyHanniDebugsAndTests.kt
+++ b/src/main/java/at/hannibal2/skyhanni/test/SkyHanniDebugsAndTests.kt
@@ -23,6 +23,7 @@ import at.hannibal2.skyhanni.features.garden.GardenNextJacobContest
 import at.hannibal2.skyhanni.features.garden.visitor.GardenVisitorColorNames
 import at.hannibal2.skyhanni.features.inventory.bazaar.BazaarApi.getBazaarData
 import at.hannibal2.skyhanni.features.mining.OreBlock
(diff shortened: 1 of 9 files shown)


src/main/java/at/hannibal2/skyhanni/test/SkyHanniDebugsAndTests.kt
```diff
@@ -103,6 +108,23 @@ object SkyHanniDebugsAndTests {
...
+
+    val isAprilFoolsDay: Boolean
+        get() {
+            val itsTime = LocalDate.now().let { it.month == Month.APRIL && it.dayOfMonth == 1 }
+            val always = SkyHanniMod.feature.dev.debug.alwaysFunnyTime
+            val never = SkyHanniMod.feature.dev.debug.neverFunnyTime
+            val result = (!never && (always || itsTime))
+            if (previousApril != result) {
+                ModifyVisualWords.update()
+            }
```
CalMWolfs: This can get removed fully as we dont reverse words on april fools anymore

CalMWolfs (changes requested):
Apart from this it all looks good

CalMWolfs (approved):

PR #3954, opened 2025-05-10
Backend: Modern version build action & label

CalMWolfs:
## What
We currently dont require build multiversion to pass before merging a pr but in a few weeks we should make that required. For now this pr just uploads the modern jars and adds labels to prs that fails multiversion. 
Once we have more stuff working on modern, ill make it also send a comment on the pr but that isnt necessary just yet.
Then finally a bit after that we can make it a required check.

We check that its a pull request before adding the label as this action also runs on push which shouldnt have labels.

+ Upload modern JARs to GitHub Actions and add label if modern build fails. - CalMWolfs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,7 @@ on:
             - "*"
         paths-ignore:
             - ".gitignore"
-    pull_request:
+    pull_request_target:
         branches:
             - "*"
         paths-ignore:
@@ -20,6 +20,10 @@ jobs:
         steps:
             -   name: Checkout code
                 uses: actions/checkout@v4
+                with:
+                    ref: ${{ github.event.pull_request.head.ref }}
+                    repository: ${{ github.event.pull_request.head.repo.full_name }}
+
             -   uses: ./.github/actions/setup-normal-workspace
             -   name: Build with Gradle
                 run: ./gradlew assemble -x test --stacktrace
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -6,7 +6,7 @@ on:
             - "*"
         paths-ignore:
             - ".gitignore"
-    pull_request:
+    pull_request_target:
```
DavidArthurCole, line 9: I seem to remember this being `pull_request` for a reason - maybe it was so that `detekt` would run off of the branch code instead of the repo code?

DavidArthurCole (dismissed):
Maybe pin some info in a contrib-forum thread for a TLDR on how to fix the common things people are going to flag, cause I know for me at least it's somewhat daunting not having touched 1.21 on the backend

hannibal002 (changes requested):
David's comment

hannibal002:
bump @CalMWolfs

CalMWolfs:
> bump @CalMWolfs

@hannibal002 I have testing this in my own repo and it works. Daveed is correct in saying that changing to `pull_request_target` does change what code the pr is run off. This is why I have added
```yml
 with:
     ref: ${{ github.event.pull_request.head.ref }}
     repository: ${{ github.event.pull_request.head.repo.full_name }} 
```
into the pr so that it works the same as it previously did. This is also what we do for the detekt github action

here is a pr nopo made with detekt running on the pr'd code. It also correctly built a skyhanni 1.21 jar with the new code and the 1.8.9 built jar is also correct with the code that is the contents of the pr
image

DavidArthurCole:
> David's comment

Outdated - see above

DavidArthurCole (approved):

hannibal002 (approved):

lunaynx:
It's all your fault...

PR #3955, opened 2025-05-10
Improvement: Get From Sacks for Composter Crops

DavidArthurCole:
## What
Adds a shortcut to get the required selected crops from sacks as an alternative to searching the bazaar.

<details>
<summary>Images</summary>

image
image
</details>

## Changelog Improvements
+ Added option to get necessary crops to fuel composter from sack. - Daveed
    * Alternative to using the "insert from sacks" which chooses "random" crops.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/composter/ComposterOverlay.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/composter/ComposterOverlay.kt
@@ -1,5 +1,6 @@
 package at.hannibal2.skyhanni.features.garden.composter
 
+import at.hannibal2.skyhanni.api.GetFromSackApi
 import at.hannibal2.skyhanni.api.ItemBuyApi.createBuyTipLine
 import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.config.ConfigUpdaterMigrator
@@ -535,7 +536,7 @@ object ComposterOverlay {
                     }
                 },
                 RIGHT_MOUSE to {
-                    HypixelCommands.getFromSacks(internalName.asString(), itemsNeeded)
+                    GetFromSackApi.getFromSack(internalName, itemsNeeded)
                 }
             ),
             tips = tips,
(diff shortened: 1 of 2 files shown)


DavidArthurCole:
i suppose 'crops' is more accurate than fuel

Thunderblade73 (changes requested):
Use gtfs api instead. As when we send the command via Hypixel Commands it will never be queued. Therefore run it through the gtfs api.

jani270:
good that i did not update yet, why was this removed

DavidArthurCole:
> Use gtfs api instead. As when we send the command via Hypixel Commands it will never be queued. Therefore run it through the gtfs api.

Considering this is happening from within a UI, is it really a problem that it wouldn't get queued? I can't envision any scenarios where this would apply.

Thunderblade73:
> > Use gtfs api instead. As when we send the command via Hypixel Commands it will never be queued. Therefore run it through the gtfs api.
> 
> Considering this is happening from within a UI, is it really a problem that it wouldn't get queued? I can't envision any scenarios where this would apply.

Why do you want to make it more complicated? Like you literally added a util function for something what the gfs api does out of the box. Since you can add /gfs to neu buttons there is a definite case where it can happen. Simply sad every gfs call we make should go over the gfs api. Also the Hypixel Command one also calls the gfs api (since that is like the fifth time that exactly this happend), so why don't you call it directly.

hannibal002:
I confuse myself with the wording.
The composter already uses the sacks, there is no need to change the logic, just the confusing wording.

DavidArthurCole:
> I confuse myself with the wording. The composter already uses the sacks, there is no need to change the logic, just the confusing wording.

Check the new changelog, see if it makes more sense.

hannibal002:
Again, I don't know why this technical/logical change exists at all.

Without the pr, when I click on the line in the sh display it always got the items from sacks for me. I used this feature literally since I have added it every day multiple times.
Once the sacks are empty I refill with an 300m coin order from bz, then I can use the sacks for the next weeks again. And over and over.

Idk what this pr adds of value to that feature that wasn't there before except a fix to the wording of the display that currently wrongly suggests the feature can only pick from bz/ah, not the sacks.

DavidArthurCole:
> Idk what this pr adds of value to that feature that wasn't there before except a fix to the wording of the display that currently wrongly suggests the feature can only pick from bz/ah, not the sacks.

image
image

This just, is not right, then? I quite literally did not know that this already existed, because why would I assume that something that says "click to search bazaar" actually uses sacks instead???

I'll rework this to fix the fact that this presently makes absolutely no goddamn sense.

hannibal002:
The settings are fine, the description while hovering that I added recently is wrong.

PR #3957, opened 2025-05-10
Backend: Remove HasLegacyId

DavidArthurCole:
## What
See changelog.

+ Removed HasLegacyId class, and stripped legacy IDs from enum classes. - Daveed


hannibal002:
Sure, this is a long time ago since we made the jump from ids to enum names, but why would we want to break support for upgrading old configs this way? we could also start deleting old config version changes with the same argument. Im not a fan of this.

DavidArthurCole:
> Sure, this is a long time ago since we made the jump from ids to enum names

https://discord.com/channels/997079228510117908/1094190239532208228/1370632651056087041

hannibal002:
what do you mean

DavidArthurCole:
> what do you mean

```
CalMWolfs — 5/10/2025 01:24
yeah if their config is that old its not our problem
+ we made the gson not fully wipe the config on wrong value anyways
```

We're removing 900 lines of outdated code that has no reason to stick around - if people are using a version of SH that's this ancient, there are going to be bigger problems than the config not migrating over.

hannibal002:
Fair

PR #3958, opened 2025-05-10
Fix: Hoppity Meal Order

lunaynx:
## What
Fixed Meal Order for Hoppity interleaving the alternate day eggs instead of putting them after the primary day.

## Changelog Fixes
+ Fixed incorrect order for Unclaimed Eggs when set to Meal Order. - Luna


DavidArthurCole (changes requested):
`resettingEntries` gets used in a lot of places across the codebase currently, and with `firstOrNull` calls, order can be important, so I'd prefer if there is either another value added to here for `orderedResettingEntries`, or that it's ordered where used in the display class.

image

DavidArthurCole (approved):

PR #3961, opened 2025-05-11
Remove: AI images from DiscordRPC

NetheriteMiner:
## What
Removes location keys corresponding to images generated by AI in the developer portal. A week or so after the PR is merged and a beta comes out, only then should the old images actually be deleted from the rich presence account. 

## Changelog Removed Features
+ Removed AI images from Discord RPC. - NetheriteMiner
    * Older updates will display incorrectly in approximately one week.


hannibal002:
I don't think removing without replacing is the way to go

CalMWolfs:
closing this as this isnt the best way to go about this

PR #3872, opened 2025-04-22
Fix: ST middle click

hannibal002:
## What
Fixed Skytils Middle Click feature breaking SkyHanni features.

## Changelog Fixes
+ Fixed Bazaar Cancel Buy Order compatibility issue with Skytils' Middle click. - hannibal2

PR #3964, opened 2025-05-12
Fix: issues with other mods setting middle click

NopoTheGamer:
## What
Fixes other mods changing clicks to be middle clicks breaking our slot click event (didnt test with skytils)
if someone (hannibal) wants to edit the changelog to mention an actual feature it fixes that would be nice
replaces [3872](https://github.com/hannibal002/SkyHanni/pull/3872)

## Changelog Fixes
+ Fixed conflicts with other mods blocking inventory-click detection. - nopo
    * Notably `Cancelled Buy Order Clipboard` with Enchanted Books, and `Visitor Accept` clicks when Skytils’ `Middle Click GUI Items` is enabled.


DavidArthurCole (approved):
image

LGTM
(before this PR, the 'user' list on the right would not populate with NEU enchanting solvers enabled).

hannibal002:
this breaks visitor inventory (the items stay in the shopping list gui after accepting the vistor) and bazaar copy refill cancel amount feature. maybe more but those two were the most problematic i can remember.

also there is another draft pr by me that can get yield to this pr

hannibal002 (approved):
Tested both features, both now work with skytils enabled.

PR #3974, opened 2025-05-18
Backend: Hypixel Mod Api migration option

ItsEmpa:
## Dependencies
- hannibal002/SkyHanni-REPO#415

## What
This PR adds a config option that lets you choose if skyhanni will use Hypixel Mod Api or the locraw/tablist/scoreboard based detection. It also adds a way to remotely disable this from the repo. Also changed some things to make it more similar to the current system (such as setting island to none and skyblock to false on world change)

+ Added option to use Hypixel Mod Api. - Empa
+ Added a way to safely get repo constants, or a default if the file doesnt exist/isnt formatted correctly. - Empa


hannibal002 (changes requested):
Please (re)use the existing `DisabledFeatures.json` file in repo and `DisabledFeaturesJson.kt` file in the mod.

src/main/java/at/hannibal2/skyhanni/config/features/dev/DevConfig.kt
```diff
@@ -127,11 +127,14 @@ class DevConfig {
...
-//     @ConfigEditorBoolean
-//     var hypixelModApi: Boolean = true
+    @Expose
+    @ConfigOption(
+        name = "Use Hypixel Mod API",
+        desc = "Use the Hypixel Mod API for better location data.\n" +
+            "§cNote: this setting may be disabled remotely, and will only work on beta versions."
+    )
+    @ConfigEditorBoolean
+    var hypixelModApi: Boolean = false // should this be default on or off?
```
hannibal002: first keep it off per default, so that we (the devs) can enable it for our sessions and see if we notice problems. we can enable this per default later again. (and way later remove the option entirely as we phase off from scoreboard/tab list)

PR #3986, opened 2025-05-23
Improvement: Read Non-God Pot Effects from Widget

DavidArthurCole:
## What
By no means a comprehensive fix for the disaster the EffectApi is right now, but this at least makes non god pot effects read from the widget, so people don't _have to_ open `/effects` anymore, they just need to enable the widget.

I did clean up some outdated messages though.

## Changelog Improvements
+ Made non-God-Pot effects read from Tab List widget. - Daveed


hannibal002 (changes requested):
image
image
image


Per default, the active effects widget is useless.
we should also mention what options to change in the /widget menu to make this actually fix anything

hannibal002 (changes requested):
image

after changing the default settings for potion effect widget, this still doesnt work

hannibal002:
bump @DavidArthurCole

DavidArthurCole:
> after changing the default settings for potion effect widget, this still doesnt work

image

PR #4003, opened 2025-05-27
Backend: StorageAPI

Thunderblade73:
## What
Saves the state of the users storage (enderchest/backpack).

Note: To test use ``/shdebug storage``

+ Added Storage API. - Thunderblade73


CalMWolfs (changes requested):
Instead of storing the storage in the regular player specific data I believe it should be saved in a completely new file much like the sack data is. You and others have had problems with neu saving the storage (though it is worse in neu as it saves on world swap) but by having it saved in a seperate file and only saving it when values are updated instead of when the normal config saves (every minute, on settings close and on shutdown) this will reduce potential performance impacts from adding this.

CalMWolfs:
@Thunderblade73 bump

CalMWolfs (changes requested):
Does not work for rift storage on a new profile that hasnt yet unlocked a second rift storage page. I'd also assume this happens for the regular ender chest as well but I cannot confirm that
inv name in rift: `Rift Storage`
apart from this, the rest works as expected though maybe it would be good to store it in pairs of <InternalName, displayName> instead of just the display name as I feel saving the internal name opens up possibilities for more features using this in the future

Thunderblade73:
> apart from this, the rest works as expected though maybe it would be good to store it in pairs of <InternalName, displayName> instead of just the display name as I feel saving the internal name opens up possibilities for more features using this in the future

I do not understand. I save it with InternalName to Container and the Container holds both the InternalName and the DisplayName.

CalMWolfs (approved):
worked for me, i think i was not thinking properly when i said that other thing as idk why but i just assumed whatever was printed in shdebug was what was saved 🤦

PR #3919, opened 2025-05-01
Backend: Fix Optimize Imports

Thunderblade73:
## What
Fixes the optimize Imports not account for preprocessor assignments.

Done, as I do not have the skill to fix the remaing 3 cases. But at least the action works.

TODO:
- [x] Use custom one for Action 
- [ ] Use custom one for OnSave
- [ ] Use custom one for OnCommit
- [ ] Use custom one for OnFlyBy
- [x] Fix extra empty lines bug
- [x] Fix import delete bug
- [x] Fix inversed if and else
- [x] Fix annoying scroll

## Changelog Technical Details
+ Fixed optimized Imports for the Action. - Thunderblade73
    * That fixes only ``Crtl + O`` not the OnSave, OnCommit, OnFlyBy optimize import tasks.

PR #4006, opened 2025-05-28
Backend: cleanup /gfs logic

hannibal002:
## Depend on
- #3919

## What
HypixelCommand class is for sending commands to hypixel, not for redirecting to other apis.

+ Fixed /gfs command processing inconsistency. - hannibal2


Thunderblade73:
The problem is that we need to root it through the gfs api, if we send it without that the gfs queue breaks. Because the api ignores commands send from SkyHanni, so it would not get counted if we don't root it through. We need to ignore SkyHanni send /gfs or else we get an infinite recursion. The rooting in Hypixel Command is only there so that it does not get misused since that is now the 5th time someone did not use the Gfs api, even though it must be used or else the queueing breaks.

hannibal002:
Either we should listen in on commands sent from sh to hypixel in the gfs logic, or we remove the gfs command entirely from the hypixel commands class. either way, the current system is super inconsistent and needs to go

Thunderblade73:
> Either we should listen in on commands sent from sh to hypixel in the gfs logic, or we remove the gfs command entirely from the hypixel commands class. either way, the current system is super inconsistent and needs to go

We literally can't listen in on it, as we would get an infinite recursion that way. As the gfs api takes the user input and blocks it and sends it again when it is the right time. If we allow commands send from sh the gfs logic would spin in circles as it sees a new command send and then blocks that tries to send a new one, which would get blocked and resend, which is an infinite loop.
It is Hypixel Commands only because when someone wanted to use gfs I  some way they added that function to Hypixel Commands, I removed it and then later one the same happend, the third time I just made that rerouting so that at least it didn't break it. But all in all the direct sending should not be used, but I won't be able to check every Pr that needs gfs for that fact. That is why I stopped bothering, as I had that same incident 5 times no.

hannibal002:
Fair. have added addtional comment + deprecated the function.

Thunderblade73 (approved):

hannibal002:
@NopoTheGamer when can i finally reenable auto import on commit again?

NopoTheGamer:
> @NopoTheGamer when can i finally reenable auto import on commit again?

merge my 10 prs

CalMWolfs:
> @NopoTheGamer when can i finally reenable auto import on commit again?

or when #3919 is completed as i believe it is for this exact thing

PR #4008, opened 2025-05-28
Backend: Save Private Island Chest inside StorageAPI

Thunderblade73:
## Dependencies
- #4003

## What
 Additionally saves Private Island Chests within the StorageAPI. It has a disable toggle that isn't yet binded to anything (since I was to lazy to add that.

+ Added Private Island Chest to StorageAPI. - Thunderblade73


src/main/java/at/hannibal2/skyhanni/api/StorageApi.kt
```diff
@@ -0,0 +1,220 @@
...
+    /**
+     * REGEX-TEST: Rift Storage (1/2)
+     */
+    private val riftStoragePattern by RepoPattern.pattern("storage.rift", "Rift Storage \\((?<page>\\d+)/\\d+\\)")
+
+    /**
+     * REGEX-TEST: Chest
+     * REGEX-TEST: Large Chest
+     */
+    private val chestPattern by RepoPattern.pattern("storage.chest", "(?:Large )?Chest")
```
j10a1n15: I believe Island chests use the normal minecraft name so this breaks on any minecraft language that isn't en_us

j10a1n15 (changes requested):

src/main/java/at/hannibal2/skyhanni/api/StorageApi.kt
j10a1n15: ChestValue alr uses some method to help the i18n thing

CalMWolfs (approved):

PR #4011, opened 2025-05-30
Feature: Searchable Storage

ILike2WatchMemes:
## Dependencies
- #4008

## What
Adds a GUI for searching items by their name/lore in a player's backpacks/enderchests/rift storages and island chests.

## Changelog New Features
+ Added searchable storage. - ILike2WatchMemes
    * Use /shsearchstorage to search through your backpacks/enderchests/rift storage and island chests.


hannibal002:
Please include in the changelog that this is for island chests (too)

src/main/java/at/hannibal2/skyhanni/features/inventory/SearchableStorageGui.kt
```diff
@@ -0,0 +1,270 @@
...
+
+    private fun searchStorage(): List<List<Renderable>> {
+        val storageVisualMap = StorageApi.accessStorage.mapNotNull { (_, storage) ->
+            val highlightSlots = storage.items.withIndex()
+                .filter { (_, item) -> item?.matches() == true }
+                .map { it.index }
+
+            if (highlightSlots.isEmpty()) return@mapNotNull null
+
+            val renderable = storage.toRenderable(highlightSlots = highlightSlots)
```
ILike2WatchMemes: ??? whats this then

Thunderblade73: In the funtion not in the call. You did not merge correctly, you overwrote your changes inside the StorageAPI.

ILike2WatchMemes: yeah I just realised oops

ILike2WatchMemes:
can't seem to figure out why but on 1.21.5 InventoryCloseEvent is not being sent when closing with escape

src/main/java/at/hannibal2/skyhanni/features/inventory/SearchableStorage.kt
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @HandleEvent
+    fun onCommandRegistration(event: CommandRegistrationEvent) {
+        event.register("shsearchstorage") {
+            description = "Search your storage for items with their names or lore."
+            category = CommandCategory.USERS_ACTIVE
+            callback { SkyHanniMod.screenToOpen = SearchableStorageGui() }
+        }
+    }
```
CalMWolfs, line 95: there should be a way to add the search to the command

CalMWolfs (changes requested):
The feature works pretty well for doing what you intended but I feel the feature can be improved upon and changed around a bit so that it has wider use cases.

While the idea of showing the chests themselves is pretty cool, i presonally dont find that super useful. I'd much rather I can just search an item name / lore and it tells me how many i have of that item if its stackable and otherwise it shows each of that item if its unstackable.

Hovering over the item will show its tooltip and in that tooltip there will be a line added that says how many you have of that item. This could also show as the stack size in the gui as well.

Then when clicking on the items it will either open ec/bp if you have it there or otherwise highlight the chest(s) that you had it in (perhaps this here is where you could go to a secondary gui that shows the chests that contains this item and the user can select which one they want to go to).

This above approach could then also integrate sacks data which is already saved and then open the /sacks menu and possibly highlight the sack that contains the targetted item.

Perhaps the default screen when opening the gui without specifying a search could then instead be items sorted by their estimated value or based on your amount of them.

I hope all this makes sense but feel free to ask a question if you need to or if you want some help implementing part of this.

In addition to the above, here are some issues i have with the current implementation which still mostly related to the new idea/revamp i have above:

1. If the user doesn't have the option on to save private island chests it should mention this somewhere in the gui. Probably having a button in the gui that takes you to this setting would be good.

2. It should be clearer that you can click on the search by option maybe add some sort of text saying click to cycle or whatever. Also the enum short have some sort of formatting not just the default .toString which results in uppercase

3. I would also like there to be a both option in the search by that does both lore and display name.

4. Maybe somewhere in the config there should be a button to open the gui much like there is for /ff or /shwords
image

5. for some reason it doesnt highlight every item that matches the search and sometime highlijghts the wrong item if it is between 2 of the correct items
image

6. it would be good to show the lore of the item when hovering over it. Also it doesnt show the stacksize of the items in the chest

7. it needs to have some sort of scrolling or something as rn it just goes off the top of my screen if i search 1 character
image

src/main/java/at/hannibal2/skyhanni/features/inventory/SearchableStorage.kt
ILike2WatchMemes: was thinking the  same when using it yesterday, will add

ILike2WatchMemes:
> 1. If the user doesn't have the option on to save private island chests it should mention this somewhere in the gui. Probably having a button in the gui that takes you to this setting would be good.

Maybe instead of this make a pop-up like this?: 

grafik

Don't remind me being clickable and red turning green but I forgot
(amazing paint.net skills btw)

ILike2WatchMemes:
grafik
here

ILike2WatchMemes:
> 7\. it needs to have some sort of scrolling or something as rn it just goes off the top of my screen if i search 1 character

how about arrow buttons at the bottom to flip through pages?

CalMWolfs:
> > 7. it needs to have some sort of scrolling or something as rn it just goes off the top of my screen if i search 1 character
> 
> how about arrow buttons at the bottom to flip through pages?
I think arrows would work but scrolling would probably be better, but in my original review comment I talked about shifting how the feature works as a whole which would also probably eliminate a lot of these issues that you have about the display being way too tall as if you only show each item once instead of all instances of it then it will not nearly have as much data to display at once

ILike2WatchMemes:
currently working on items,
grafik
can't seem to get them bigger yet

ILike2WatchMemes:
grafik
and
grafik
which turns into this:
grafik
(will be doing stack sizes tomorrow)
@CalMWolfs thoughts?

CalMWolfs:
>  @CalMWolfs thoughts?

Will have a proper look later, just looked at the images so far, i feel that the item names dont need to be displayed above the items as you can just hover over the items and see their name + tooltip and most people will be searching in this gui anyways so it shouldn't end up too confusing. Just this change would allow your current screenshot which has 30 items in it to hold ~80 items

ILike2WatchMemes:
ItemStackRenderables seems to be rendered above everything or something so I can't get tooltips to render correctly (they're beneath the items rn but very much there), any ideas?

CalMWolfs:
ill have a look in a little bit

ptlthg:
Are we actually keeping the "for maximum QOL" text in here because that's wild if so

ILike2WatchMemes:
> Are we actually keeping the "for maximum QOL" text in here because that's wild if so

I just used that wording since I wanted to get a screenshot out I personally don't think it's a good idea either

HacktheTime:
The way the island chests are saved with modern at least must be slightly adapted. The coords in the string for island chests are saved as doubles causing chests constantly duplicating instead of getting updated.

→ I merged the changes in my fork already and want to save the hastle having to search for it since i already know the source.

j10a1n15:
Just use skyocean item search that has existed for 8+ months now it's not that deep

ItsEmpa:
> Just use skyocean item search that has existed for 7 months now it's not that deep

and also supports way more things than this, so i dont really see why this feature thats just a direct downgrade from the skyocean one should get added here personally

HacktheTime:
> > Just use skyocean item search that has existed for 7 months now it's not that deep
> 
> and also supports way more things than this, so i dont really see why this feature thats just a direct downgrade from the skyocean one should get added here personally

Good point.


But keep in mind not everything is only for the user. I personally already use it in some features in my fork. Some features can take great advantage of the data or straight IP depend on it. 

So keep in mind that having such an API might be useful for more sh features in the future.

There is NO need from me to do any changes right now. I want to highlight that. I'll make an upgraded PR if I need it and I make other PRs and Code that needs it. But I do want to point out that just because it's in sky ocean there is is not necessarily no need for it in sh.

j10a1n15:
> But I do want to point out that just because it's in sky ocean there is is not necessarily no need for it in sh.

if you wouldve read the contrib guide, you would know it specifically states "Do not copy features from other mods" and the exceptions to the rule dont apply to skyocean aswell. The feature used to be fine ish with sh supporting 1.8, but now this is just a downgrade of skyoceans search with no real reason to exist except being too lazy to install other mods/thinking installing too many mods causes fps loss

lunaynx:
I don't disagree that having the API exist can be useful – that in itself wouldn't count as copying features, since it's just backend and could be used to make unique features. But I don't know if we really *need* item tracking in SkyHanni beyond inventory/sacks/hunting box with SkyCubed and SkyOcean already having mostly everything people may need, and it is probably easier to just build any additional features that need more than that on top of those mods or another mod using [SkyblockAPI](https://github.com/SkyblockAPI/SkyblockAPI) instead.

HacktheTime:
> > But I do want to point out that just because it's in sky ocean there is is not necessarily no need for it in sh.
> 
> if you wouldve read the contrib guide, you would know it specifically states "Do not copy features from other mods" and the exceptions to the rule dont apply to skyocean aswell. The feature used to be fine ish with sh supporting 1.8, but now this is just a downgrade of skyoceans search with no real reason to exist except being too lazy to install other mods/thinking installing too many mods causes fps loss

I just completely disagree with this.

If code needs it it literally just doesn't care about the rule. Question might be whether it's worth adding a front end etc for it but it existing is most certainly not against the dont copy other mods features. 

Also in case of too lazy fps mods argument I do want to remind you again that using more mods is also meaning you need to trust more devs and offer more attack surface. This isn't that important of an argument though. 

But the fps is way more relevant. I had multiple people complain to me about fps and modern meaning that the boosts from sodium etc get more than compensated by SB mods. I personally have the luck of good hardware and decently large but only 1 RAM chip (memory broke during RAM crsisis and I don't wanna pay 1k to replace it.) But multiple people literally have significant performance issues to a point where they might choose to play basic performance only with no SB mods for the fps to be playable. Because they don't have the hardware for it.

I suggested them to contact sh to see whether there's things that can be done since performance improvements are a field I have no experience with. The only thing I can think of would be to reduce the amount of blocking code for the rendering thread. Using a processing queue for received messages sounds interesting to me to ensure order and remove the regex searches out of the main rendering thread.

HacktheTime:
> I don't disagree that having the API exist can be useful – that in itself wouldn't count as copying features, since it's just backend and could be used to make unique features. But I don't know if we really *need* item tracking in SkyHanni beyond inventory/sacks/hunting box with SkyCubed and SkyOcean already having mostly everything people may need, and it is probably easier to just build any additional features that need more than that on top of those mods or another mod using [SkyblockAPI](https://github.com/SkyblockAPI/SkyblockAPI) instead.

I dont fully understand how the inclusion of such a storage api would be done within sh if done via other mods.

Some features in current sh could benefit of better tracking (Auto king talisman complete if you have the accessory) but right now there does not seem to be a notable need.

The features I have (and more planned) depend on some form of storage api. One of the features looks like it will make some front end adaptations worth it while it should still provide the full feature set of as described in sky ocean. The feature that needs those changes is still early in development though.

src/main/java/at/hannibal2/skyhanni/utils/renderables/container/RenderableInventory.kt
```diff
@@ -96,9 +131,18 @@ object RenderableInventory {
         val finalList = uvList.map { uvRow ->
             uvRow.map { uv ->
                 val uvArray = uv.getUvCoords()
+                if (index in highlightSlots) println(index)
```
hannibal002, line 134: remove debug

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorageReminderGui.kt
```diff
@@ -0,0 +1,114 @@
...
+                DrawContextUtils.translate(guiLeft.toFloat(), guiTop.toFloat(), 0f)
+                display.render(guiLeft, guiTop)
+            }
+        }
+    }
+
+    private fun buildContent(): Renderable {
+        val warning = string("§4WARNING")
+        val warningText1 = string("§cYou currently don't save private island chests.")
+        val warningText2 = string("§cIt is recommended to enable that option for maximum QOL.")
```
hannibal002, line 53: bad wording

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorage.kt
```diff
@@ -0,0 +1,103 @@
...
+    fun onSecondPassed() {
+        if (lastCloseTime.passedSince() <= 30.seconds) return
+        highlightSlots = listOf()
+        waypoints = listOf()
+        lastCloseTime = SimpleTimeMark.farFuture()
+    }
+
+    @HandleEvent
+    fun onInventoryClose(event: InventoryCloseEvent) {
+        if (Minecraft.getMinecraft().currentScreen is SearchableStorageGui) lastCloseTime = SimpleTimeMark.now()
```
hannibal002, line 72: broken code

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorageReminderGui.kt
```diff
@@ -0,0 +1,114 @@
...
+    override fun onDrawScreen(mouseX: Int, mouseY: Int, partialTicks: Float) {
+        drawDefaultBackground(mouseX, mouseY, partialTicks)
+
+        display = Renderable.drawInsideRoundedRect(buildContent(), Color.decode("#202020"))
+
+        display?.let { display ->
+            guiLeft = (width - display.width) / 2
+            guiTop = (height - display.height) / 2
+
+            GlStateManager.disableLighting()
```
hannibal002, line 42: this does no longer exist

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorageReminderGui.kt
```diff
@@ -0,0 +1,114 @@
...
+import at.hannibal2.skyhanni.utils.compat.GuiScreenUtils
+import at.hannibal2.skyhanni.utils.compat.SkyhanniBaseScreen
+import at.hannibal2.skyhanni.utils.renderables.Renderable
+import at.hannibal2.skyhanni.utils.renderables.container.VerticalContainerRenderable.Companion.vertical
+import at.hannibal2.skyhanni.utils.renderables.primitives.StringRenderable
+import at.hannibal2.skyhanni.utils.renderables.primitives.placeholder
+import net.minecraft.client.renderer.GlStateManager
+import java.awt.Color
+
+class SearchableStorageReminderGui(private var search: String = "") : SkyhanniBaseScreen() {
```
hannibal002, line 19: broken code: missing import

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorageReminderGui.kt
```diff
@@ -0,0 +1,114 @@
...
+    private var guiLeft = 0
+    private var guiTop = 0
+
+
+    override fun guiClosed() {
+        OtherInventoryData.close("Searchable Storage GUI")
+    }
+
+    override fun onDrawScreen(mouseX: Int, mouseY: Int, partialTicks: Float) {
+        drawDefaultBackground(mouseX, mouseY, partialTicks)
```
hannibal002, line 34: broken code

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorageGui.kt
```diff
@@ -0,0 +1,377 @@
...
+
+        fun sort(input: Map<NeuInternalName, ItemData>): Map<NeuInternalName, ItemData> =
+            input.entries.sortedWith(comparator).associate { it.toPair() }
+
+        override fun toString(): String =
+            if (sortMode == this) " §f$displayName " else " §7$displayName "
+    }
+
+    override fun guiClosed() {
+        OtherInventoryData.close("Searchable Storage GUI")
```
hannibal002, line 109: have you tested this? looks wrong to me

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorage.kt
```diff
@@ -0,0 +1,103 @@
...
+    var sortMode = SortMode.NAME_DESC
+    var highlightSlots = listOf<Int>()
+    var waypoints = listOf<LorenzVec>()
+    var inventoryName = ""
+
+    @HandleEvent(GuiContainerEvent.BackgroundDrawnEvent::class)
+    fun onBackgroundDrawn() {
+        InventoryUtils.getItemsInOpenChest().forEach { slot ->
+            val slotNumber = when (inventoryName) {
+                "Chest" -> slot.slotNumber
```
hannibal002, line 52: broken code

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorage.kt
```diff
@@ -0,0 +1,103 @@
...
+            description = "Opens a gui to search for items across storages"
+            category = CommandCategory.USERS_ACTIVE
+            simpleCallback { onCommand() }
+            argCallback("search", BrigadierArguments.greedyString()) { search -> onCommand(search) }
+        }
+    }
+
+    fun onCommand(search: String = "") {
+        if (!SkyBlockUtils.inSkyBlock) return ChatUtils.userError("You can only use this in Skyblock!")
+        SkyHanniMod.screenToOpen = if (shouldRemind()) SearchableStorageReminderGui(search) else SearchableStorageGui(search)
```
hannibal002, line 99: Assignment type mismatch: actual type is 'Any', but 'Screen?' was expected.

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorage.kt
```diff
@@ -0,0 +1,103 @@
...
+
+            if (slotNumber in highlightSlots) {
+                slot.highlight(LorenzColor.GREEN)
+            }
+        }
+    }
+
+    @HandleEvent(SecondPassedEvent::class)
+    fun onSecondPassed() {
+        if (lastCloseTime.passedSince() <= 30.seconds) return
```
hannibal002, line 64: lastCloseTime is a large number, so this means its over 30s, this means the reset logic here gets called every second.

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorageUtil.kt
```diff
@@ -0,0 +1,276 @@
...
+                SearchableStorage.inventoryName = "Backpack§r (Slot #$page)"
+            }
+
+            "Ender Chest" -> {
+                HypixelCommands.enderChest(page)
+                SearchableStorage.inventoryName = "Ender Chest ($page"
+            }
+
+            "Rift Storage" -> {
+                if (SkyBlockUtils.currentIsland == IslandType.THE_RIFT) HypixelCommands.enderChest(page)
```
hannibal002, line 108: if the player is not in the rift, nothing happens.

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorageGui.kt
```diff
@@ -0,0 +1,377 @@
...
+        var amount: Int,
+        val inventories: MutableList<SkyHanniInventoryContainer>,
+    )
+
+    data class RenderPosition(
+        val x: Int,
+        val y: Int,
+    )
+
+    enum class SearchMode(val displayName: String, val matches: (ItemStack, String) -> Boolean) {
```
hannibal002, line 83: broken code

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorageUtil.kt
```diff
@@ -0,0 +1,276 @@
...
+import at.hannibal2.skyhanni.utils.SoundUtils.playSound
+import at.hannibal2.skyhanni.utils.renderables.Renderable
+import at.hannibal2.skyhanni.utils.renderables.RenderableUtils
+import at.hannibal2.skyhanni.utils.renderables.container.HorizontalContainerRenderable.Companion.horizontal
+import at.hannibal2.skyhanni.utils.renderables.container.VerticalContainerRenderable.Companion.vertical
+import at.hannibal2.skyhanni.utils.renderables.primitives.ItemStackRenderable.Companion.item
+import at.hannibal2.skyhanni.utils.renderables.primitives.StringRenderable
+import java.awt.Color
+
+object SearchableStorageUtil {
```
hannibal002, line 25: much code duplication between this file and SearchableStorage

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorageReminderGui.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    override fun onMouseClicked(originalMouseX: Int, originalMouseY: Int, mouseButton: Int) {
+        val (relativeMouseX, relativeMouseY) = getRelativeMousePos()
+
+        content?.let { content ->
+            val display = display ?: return
+            var top = 0
+
+            content.forEachIndexed { i, renderable ->
+                if (i in listOf(4, 6, 8)) {
```
hannibal002, line 88: the list should be not initialized here, not in the loop, not in the click function.

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorageReminderGui.kt
```diff
@@ -0,0 +1,114 @@
...
+                            left = (display.width - renderable.width) / 2,
+                            top,
+                            renderable.width,
+                            renderable.height,
+                        )
+                    ) {
+                        when (i) {
+                            4 -> config.searchableStorageReminder = false
+                            6 -> config::savePrivateIslandChests.jumpToEditor()
+                            8 -> SkyHanniMod.screenToOpen = SearchableStorageGui(search)
```
hannibal002, line 101: broken code

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorageUtil.kt
```diff
@@ -0,0 +1,276 @@
...
+        SearchableStorage.waypoints = with(this.inventoryContainer) {
+            SoundUtils.createSound("random.orb", 1.0f).playSound()
+            ChatUtils.chat("Set waypoint to $displayName!")
+            SearchableStorage.inventoryName = "Chest"
+
+            listOfNotNull(primaryCords, secondaryCords)
+        }
+    }
+
+    fun Int.handlePageSelection(
```
hannibal002, line 124: too generic. change to normal function, or make private

src/main/java/at/hannibal2/skyhanni/data/model/SkyHanniInventoryContainer.kt
```diff
@@ -18,13 +18,14 @@ class SkyHanniInventoryContainer(
...
 ) {
-    fun toRenderable(scale: Double = 1.0): Renderable = with(Renderable) {
+    fun toRenderable(scale: Double = 1.0, highlightSlots: List<Int> = listOf()): Renderable = with(Renderable) {
         vertical(
             text(displayName, scale),
             fakeInventory(
                 items,
                 rowSize,
                 scale,
+                highlightSlots
```
hannibal002, line 28: missing trailing comma

src/main/java/at/hannibal2/skyhanni/features/inventory/searchablestorage/SearchableStorageGui.kt
```diff
@@ -0,0 +1,377 @@
...
+        NAME("NAME", { stack, search -> stack.displayName.removeColor().contains(search, ignoreCase = true) }),
+        LORE("LORE", { stack, search -> stack.getLore().any { it.removeColor().contains(search, ignoreCase = true) } }),
+        NAME_LORE("NAME + LORE", { stack, search -> NAME.matches(stack, search) || LORE.matches(stack, search) })
+        ;
+
+        override fun toString() = if (searchMode == this) " §f${this.displayName} " else " §7${this.displayName} "
```
hannibal002, line 89: the inf condition  in the tostring function should get extracted

hannibal002 (changes requested):
the pr desc should explain why we can copy existing features from SkyOcean etc. There should be a useful improvement over the same feature in other mods

Fazfoxy:
> > I don't disagree that having the API exist can be useful – that in itself wouldn't count as copying features, since it's just backend and could be used to make unique features. But I don't know if we really _need_ item tracking in SkyHanni beyond inventory/sacks/hunting box with SkyCubed and SkyOcean already having mostly everything people may need, and it is probably easier to just build any additional features that need more than that on top of those mods or another mod using [SkyblockAPI](https://github.com/SkyblockAPI/SkyblockAPI?rgh-link-date=2025-12-31T14%3A56%3A07Z) instead.
> 
> I dont fully understand how the inclusion of such a storage api would be done within sh if done via other mods.
> 
> Some features in current sh could benefit of better tracking (Auto king talisman complete if you have the accessory) but right now there does not seem to be a notable need.
> 
> The features I have (and more planned) depend on some form of storage api. One of the features looks like it will make some front end adaptations worth it while it should still provide the full feature set of as described in sky ocean. The feature that needs those changes is still early in development though.

I'm jumping in a few *months* late but, nobody ever said to remove storage API, the entire point of conversation is that with SH dropping 1.8 there's actually 0 reason for us to have a searchable storage feature considering skyocean can search more better?

ILike2WatchMemes:
honestly if skyocean has this i dont really see a use in "copying" it (which I didn't intend), havent used the skyocean one tho; are these two basically the same?

Fazfoxy:
> honestly if skyocean has this i dont really see a use in "copying" it (which I didn't intend), havent used the skyocean one tho; are these two basically the same?

skyocean lets you search basically any storage UI on your profile, I don't think this has anything distinct that makes it anything but a worse of that?

ILike2WatchMemes:
okay woah no need to get insulting

j10a1n15:
> honestly if skyocean has this i dont really see a use in "copying" it (which I didn't intend), havent used the skyocean one tho; are these two basically the same?

image
Skyocean supports all of these item sources

HacktheTime:
I want to point out that I'm talking for myself here to give another perspective.

I know that the storage has ueses already and in some cases a search would benefit already.

For example it would be good to add an API that allows you to take resources from enderchest into account (bingo minion helper) but there can also be cases like king talismann completion if you own it already.

I think that alone for non user facing reasons it's something that should be added. Maybe without GUI or sth but an API for code sounds useful no matter what.

hannibal002:
An API that is just a `Map<Int, List<ItemStack>>` with `int= storage number` is a no brainer, 20 lines max impl. everything else is subjective

HacktheTime:
Not for the acessory example. I think a good internal api will be worth it. User facing front end is subjective in my opinion too tho.

ILike2WatchMemes:
I'll close this tbh, SkyOcean mogs this

PR #4015, opened 2025-05-31
Improvement: Kick Timer only First Kick

Fazfoxy:
## What
Makes Kick Timer only Count from first kick (makes ability to join SB notification more accurate if Pleaders keep trying to warp after kick)

This changelog is stupidly long for a change that is smaller than this entire description

## Changelog Improvements
+ Made Last Kicked from Skyblock feature count only first kick. - Fazfoxy
    * Only First Kick Counts for Rejoin cooldown so this makes rejoin notification more accurate.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/SkyBlockKickDuration.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/SkyBlockKickDuration.kt
@@ -52,7 +52,6 @@ object SkyBlockKickDuration {
         if (!isEnabled()) return
 
         if (kickPattern.matches(event.message) && !showTime) {
-
             if (SkyBlockUtils.onHypixel && !SkyBlockUtils.inSkyBlock) {
                 kickMessage = false
                 showTime = true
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/misc/SkyBlockKickDuration.kt
```diff
@@ -51,8 +51,8 @@ object SkyBlockKickDuration {
     fun onChat(event: SkyHanniChatEvent) {
         if (!isEnabled()) return
 
-        if (kickPattern.matches(event.message)) {
-            if (LorenzUtils.onHypixel && !LorenzUtils.inSkyBlock) {
+        if (kickPattern.matches(event.message) && !showTime) {
```
CalMWolfs: if both the patterns also check showTime, it could be checked in the same line as isEnabled then early return

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/SkyBlockKickDuration.kt
```diff
@@ -49,10 +49,10 @@ object SkyBlockKickDuration {
 
     @HandleEvent
     fun onChat(event: SkyHanniChatEvent) {
-        if (!isEnabled()) return
+        if (!isEnabled() && !showTime) return
```
CalMWolfs:
```suggestion
        if (!isEnabled() || !showTime) return
```
i believe this should be an or

CalMWolfs (changes requested):

CalMWolfs (approved):

PR #4022, opened 2025-06-02
Fix: chat display with disabled guild chat

rueblimaster:
https://discord.com/channels/997079228510117908/1378031892816986264

## Changelog Fixes
+ Fixed chat display when guild chat is disabled. - rueblimaster


ItsEmpa (approved):

hannibal002 (approved):
i dont like this solution. better would be to keep the code and rather add a check for the most recent chat messge being `/gc `, anyway. fine for now

rueblimaster:
> i dont like this solution. better would be to keep the code and rather add a check for the most recent chat messge being `/gc `, anyway. fine for now

I don't get why we would want to register it telling us that we can't use the guild chat rn as us using the guild chat?
I can understand if you don't like existing part of the detection, but that I don't get

PR #4000, opened 2025-05-27
Backend: `CircularRenderable`

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/3937

## What
Adds the CircularRenderable class, as well as CircularContainerRenderable.
Maybe with this being only like 200 lines changed it'll get merged quickly 🥺

## Changelog Technical Details
+ Added CircularRenderable and CircularContainerRenderable. - Daveed + Thunderblade73

PR #4025, opened 2025-06-03
Backend: `OrbitalSystemRenderable`

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/4000

## What
As the name implies, it's a renderable that has other renderables orbiting it, configurably - another one that I "need" for the pet PR.

+ Added OrbitalSystemRenderable. - Daveed


Thunderblade73 (changes requested):
I don't like the fact that we limit what renderable is usable in renderable that is a trend that I don't support. As in my personal opinion the Renderable system is great because it works with any imaginable combination of stuff and you don't have to woory about it.
Also you do not parse the new posX and posY of the sub renderables, also also you have an small overlapping space of the main renderable and sub renderables as the bounding box is rectangular, meaning mouse input (if set correctly) would trigger on both, for a fix see double layered renderable (a lot of fun).
Btw. did not look deeply at the code changes just the most potent issue I have with this.

DavidArthurCole:
> I don't like the fact that we limit what renderable is usable in renderable that is a trend that I don't support

Yeah I don't know why I did that. All of the code can be the same with `Renderable` params

DavidArthurCole:
> Also you do not parse the new posX and posY of the sub renderables

Not sure I follow here...

> Also also you have an small overlapping space of the main renderable and sub renderables as the bounding box is rectangular

Should be fixed in latest commit

PR #3991, opened 2025-05-24
Fix: TabWidget pattern white space removal

Godwhitelight:
## What
Fixing the whitespace removal on the pattern to work correctly on patterns like the EYES_PLACED, where you have few alternatives on the regex. Making the feature work better (problem described in the images provided).

<details>
<summary>Images</summary>

image
image

</details>

## Changelog Fixes
+ Fixed TabWidget Eyes Placed so it will show dragon spawned and egg respawning. - G0dwhitelight

PR #4027, opened 2025-06-03
Fix: Eyes placed tab widget

CalMWolfs:
## What
Closes #3991 as this actually fixes the 1 widget that is broken rather than messing with every other widget for no good reason

This fixes it because the whitespace that is added to the beggining of the pattern only applies to the first part of the or in regex. Making a non capturing group fixes this
image

## Changelog Fixes
+ Fixed eyes placed tab widget not showing when dragon is spawned or respawning. - CalMWolfs

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/model/TabWidget.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/model/TabWidget.kt
@@ -207,7 +207,7 @@ enum class TabWidget(
     ),
     EYES_PLACED(
         // language=RegExp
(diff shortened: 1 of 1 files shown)


Thunderblade73:
It is better to add the group around that for every widget since then such an error won't happen again for new widgets.

lunaynx:
I agree, the original fix is fine. If it somehow breaks other tab widgets, their patterns should be fixed. This fix is very unintuitive to someone reading the code. I would expect having those parentheses or not to have the same behavior.

PR #4042, opened 2025-06-07
Improvement: Pest Guess Visuals

rueblimaster:
## Changelog Improvements
+ Differentiates Plot Middle visually. - rueblimaster
    * Distinguish Pest Guesses that point to plot middle from actual Pest locations in the Garden.


src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestParticleWaypoint.kt
```diff
@@ -118,13 +121,21 @@ object PestParticleWaypoint {
             return
         }
         val waypoint = guessPosition ?: return
-        val color = LorenzColor.RED.toColor()
+        val distance = waypoint.distance(event.exactPlayerEyeLocation())
+        val color: Color
+        val middlePoint = GardenPlotApi.closestCenterPlot(waypoint)
```
hannibal002: please do not use closestCenterPlot. rather use getCurrentPlot, and remove the unused function in GardenPlotApi.

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestParticleWaypoint.kt
```diff
@@ -118,13 +121,21 @@ object PestParticleWaypoint {
...
+        if (middlePoint != null && waypoint.ceil().equalsIgnoreY(middlePoint)) {
```
hannibal002: make this "is middle" check only once when the current guess position changes, not every render frame.

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestParticleWaypoint.kt
```diff
@@ -118,13 +121,21 @@ object PestParticleWaypoint {
...
+            color = LorenzColor.YELLOW.toColor()
+            event.drawDynamicText(waypoint.add(y = -0.1 - distance / (12 * 1.7)), " §r§e(plot middle)", 1.0)
```
hannibal002: instead of the math here, use the yOff prameter of drawDynamicText

hannibal002 (changes requested):
We already had the very same fature for the old logic of pest finder (pest partile line class, got replaced once the previse particle detection got fixed again).
and back then, the "show middle" view was already contested by others that wanted an option. please keep reusing this option therefore.
Please use this toggle here:
https://github.com/hannibal002/SkyHanni/blob/65ed2250d28c8bb05eb54c452e7ad6bdd572850a/src/main/java/at/hannibal2/skyhanni/config/features/garden/pests/PestWaypointConfig.kt#L37

also maybe we can move the "center" position by setting y to the y of the player, so thats in the same height?

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestParticleWaypoint.kt
rueblimaster: I don't want the current plot, I need the plot the waypoint is inside of. 
a) I can create a new function for getting the plot a location is inside of
b) I leave it like this
what do you prefer?

hannibal002: year fair. i compared it to the old logic - in there we would only start checking if the guess is near the center once you entered the current plot. that shortcut is not possible if we want to cache the state though

rueblimaster: it's also not optimal if we didn't cache it, as it wouldn't recognize it as the middle until you've entered the plot

rueblimaster: I've changed it, I hope this is more like you want it

hannibal002: > it's also not optimal if we didn't cache it, as it wouldn't recognize it as the middle until you've entered the plot

since we want to cache it, we cant use the old approach. therefore your current apprach is fine here still. maybe just change the function to not return the middle but rather return the plot and then access middle in the actual class.

sorry if i was not clear enough before

rueblimaster:
@hannibal002 
I added a toggle for it.
But I don't agree with replacing the showMiddle with what I did as they do the same purpose in completely different ways and the thing I added also isn't invasive in any kind.
Also the showMiddle is default disabled, which I don't think my thing should be.
If we'd do a configfix for it, it would result in negatives

So should I delete the showMiddle entirely or keep both parallel to each other?

hannibal002:
idk anymore. i think the current state is fine.

PR #4044, opened 2025-06-09
Fix: Fast Fairy Souls Pathfinder ignoring if all souls are found

rueblimaster:
## What
The Fast Fairy Soul thing ignored if you have all souls found in the totalFound profile storage, if found (the coordinates of the found souls) doesn't have all.
This case happened for all island except the current one, if you have all fairy souls and you open the quest menu. That's because it only adds the coords for the found souls if you're currently on the island.
This lead to you having to open the quest menu in new island you visited.

[https://discord.com/channels/997079228510117908/1380836500530528357](https://discord.com/channels/997079228510117908/1380836500530528357)

## Changelog Fixes
+ Fixed Fast Fairy Souls Pathfinder not detecting already found fairy souls from the quest menu. - rueblimaster

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/FastFairySoulsPathfinder.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/FastFairySoulsPathfinder.kt
@@ -149,7 +149,9 @@ object FastFairySoulsPathfinder {
 
         fun allFound(state: String) {
             disabled = true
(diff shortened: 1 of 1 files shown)


hannibal002 (changes requested):
The fix properly worked for all islands excep dwarven mines (prob because of the knwon bug)

any way, i just noticed this problem:
image


after marking all souls on the current island as found, the detection logic would not stop (on all islands, not only dwarven mines)

PR #2573, opened 2024-09-22
Feature: Mute High Heat sound in the Magma Fields

martimavocado:
## What
adds a toggle to mute the high heat sound in the magma fields  when at or above 90 heat

[suggestion](https://canary.discord.com/channels/997079228510117908/1287530198094450729)

## Changelog New Features
+ Added option to mute High Heat Sound in Magma Fields. - martimavocado

PR #4049, opened 2025-06-10
Backend: Custom scoreboard heat cleanup

hannibal002:
## Dependencies
- #2573

## What
After we added a new heat detection pattern, the one already in the custom scoreboard is redundant. 
in the process I also cleaned out other duplicate logic around heat in the custom scoreboard.

+ Reduced duplication in Heat Detection and Custom Scoreboard patterns. - hannibal2


hannibal002:
would love to see a review of @martimavocado and @j10a1n15

j10a1n15 (approved):
changes in [c1c80f3](https://github.com/hannibal002/SkyHanni/pull/4049/commits/c1c80f37aa067e5805f1ffed8eb12c9340df2218) look fine

j10a1n15 (changes requested):
heatDisplay will actually never get reset, since it only gets set if the regex matches, not if it fails

hannibal002:
good catch

martimavocado (approved):
looks fine, didn't test ingame

j10a1n15 (approved):

PR #4054, opened 2025-06-11
Backend: Remove unused library

CalMWolfs:
We no longer get discord from team resourceful as of #3963. Now we get it from jitpack. Having this unused repo has caused a few weird issues such as trying to get fabric from it instead of from proper repos.

exclude_from_changelog

### DIFF
--- a/root.gradle.kts
+++ b/root.gradle.kts
@@ -34,7 +34,7 @@ allprojects {
         maven("https://repo.spongepowered.org/repository/maven-public/") // mixin
         maven("https://pkgs.dev.azure.com/djtheredstoner/DevAuth/_packaging/public/maven/v1") // DevAuth
         maven("https://jitpack.io") {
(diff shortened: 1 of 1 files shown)


ThatGravyBoat:
> We no longer get discord from team resourceful as of #3963. Now we get it from jitpack. Having this unused repo has caused a few weird issues such as trying to get fabric from it instead of from proper repos.
> 
> exclude_from_changelog

Btw just for the record I don't mirror fabric at all in that repo idk where you got that from? There are only 3 groups in that repo:
- com.jagrosh.DiscordIPC
- me.owdding
- tech.thatgravyboat

Also if you want to only get specific artifacts you should use ContentDescriptors like includeGroup

CalMWolfs:
> Btw just for the record I don't mirror fabric at all in that repo idk where you got that from?

Idk why the error said that, was likely a weird gradle issue from another thing failing 🤷 
image

PR #4058, opened 2025-06-12
Fix: Custom Scoreboard Party Display not disappearing when a party of 2 players is disbanded due to the leader disconnecting for 5 mins.

NeoNyaa:
## What
Adds the string `§cThe party was disbanded because the party leader disconnected.` to the PartyApi, fixing the Custom Scoreboard not removing the party element when a party of 2 players is disbanded due to the leader disconnecting and not returning within 5 minutes.

## Changelog Fixes
+ Fixed Custom Scoreboard Party Display persisting for 5 minutes after the leader disconnects from a two-player party. - NeoNyaa


ItsEmpa (approved):
out of scope for this pr, but these should be changed for repopatterns

PR #4059, opened 2025-06-14
Fix: Pest spawn timer showing when disabled

Chissl:
## What
Fixes pest spawn timer showing when disabled in config. 

## Changelog Fixes
+ Fixed Pest Spawn Timer displaying incorrectly. - Chissl


hannibal002 (approved):
good catch!

PR #4060, opened 2025-06-15
Backend: Config Val/Var Enforcement

DavidArthurCole:
## What
Adds a detekt rule for enforcing non-mutable config types to `val` instead of `var`. Grabbed some common logic out of the `StorageNeedsExpose` rule and made a util.

+ Added detekt rule enforcing non-mutable config types as val. - Daveed
+ Changed config-referencing helpers to use KProperty0 instead of Mutable. - Daveed


hannibal002 (changes requested):
scrolled throught all classes, looks good.

im just curious, why use the `@Suppress("StorageVarOrVal")` on some dropdown objects and not others? can we isntead we filter to ever show on ConfigEditorDropdown, except when its a `property<x>`?
image

DavidArthurCole:
> scrolled throught all classes, looks good.
> 
> im just curious, why use the `@Suppress("StorageVarOrVal")` on some dropdown objects and not others? can we isntead we filter to ever show on ConfigEditorDropdown, except when its a `property<x>`?

I had a couple to clean up - but the only ones that are suppressed now are the ones where some part of the code is manually setting the config values via some method.

image

image

hannibal002 (approved):

PR #4070, opened 2025-06-15
Backend: Forced Single Registration

DavidArthurCole:
## What
Makes it so funny things like this;
image
Don't actually register functions to listeners multiple times.

+ Prevented duplicate function registration across event parameter combinations. - Daveed


CalMWolfs (approved):

hannibal002 (approved):
im doing this cleanup
image

PR #4090, opened 2025-06-18
Backend: cleanup BucketedItemTrackerData

hannibal002:
## Dependencies
- #4072

## What
using a reflection hack to avoid the parameter of type class.
This could blow up if we were to use another layer of abstraction, but for now it works fine.

+ Removed parameters for isntances of BucketedItemTrackerData. - hannibal2


DavidArthurCole (changes requested):
https://discord.com/channels/997079228510117908/1094190239532208228/1377291383093727331

We intentionally got rid of the reflection, as it's a needless resource usage

hannibal002:
image

PR #4014, opened 2025-05-30
Backend: Alot more of ProfileSpecificStorage working 1.21

NopoTheGamer:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/4012
- https://github.com/hannibal002/SkyHanni/pull/4013

## What
Gets alot of random misc files working for ProfileSpecificStorage, this includes a few trackers
Gets GardenApi and like ~75 related files working idk what this includes fully

<details>
<summary>Images</summary>

image
image
image
image

</details>

## Changelog Technical Details
+ Enabled GardenApi and miscellaneous files for 1.21. - nopo

PR #4094, opened 2025-06-19
Backend: duplicate /ff

hannibal002:
## Dependencies
- #4014

## What
the require() function does not crash in dev env. this should get fixed globally (maybe with mixins idk)

ignore_from_changelog


hannibal002:
replaced by #4099

PR #4099, opened 2025-06-19
Backend: Require Dev Env

ItsEmpa:
## What
Adds a function similar to kotlins `require` function that crashes if value is false and user is in dev environment

+ Added requireDevEnv function. - Empa


CalMWolfs (approved):
code lgtm

hannibal002 (changes requested):
please add a changelog

this pr needs the technical details category way more than many other prs that simply use technical details to not ignore the changelog at all.

j10a1n15 (approved):

hannibal002 (changes requested):
If this PR should fully replace #4094, please also copy this change here 
image

PR #4107, opened 2025-06-20
Fix: Kick Timer not working

Fazfoxy:
## What
I have somehow taken this long to fix it, how.
(somehow I hallucinated it working in dev env, honestly, I don't know either)

## Changelog Fixes
+ Fixed Lobby Kick Timer. - fazfoxy

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/SkyBlockKickDuration.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/SkyBlockKickDuration.kt
@@ -48,26 +48,27 @@ object SkyBlockKickDuration {
         "§cThere was a problem joining SkyBlock, try again in a moment!",
     )
 
+    fun kicked()
+    {
+        kickMessage = false
+        showTime = true
+        lastKickTime = SimpleTimeMark.now()
+    }
+
     @HandleEvent
     fun onChat(event: SkyHanniChatEvent) {
-        if (!isEnabled() || !timeoutKick) return
+        if (!isEnabled() || !(lastKickTime.isFarFuture())) return
 
         if (kickPattern.matches(event.message)) {
             if (SkyBlockUtils.onHypixel && !SkyBlockUtils.inSkyBlock) {
-                kickMessage = false
-                showTime = true
-                timeoutKick = false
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/misc/SkyBlockKickDuration.kt
```diff
@@ -49,12 +50,13 @@ object SkyBlockKickDuration {
 
     @HandleEvent
     fun onChat(event: SkyHanniChatEvent) {
-        if (!isEnabled() || !showTime) return
+        if (!isEnabled() || !timeoutKick) return
 
         if (kickPattern.matches(event.message)) {
             if (SkyBlockUtils.onHypixel && !SkyBlockUtils.inSkyBlock) {
                 kickMessage = false
```
DavidArthurCole: I know this wasn't explicitly your doing, but this logic, and the next if block, are exactly the same - please extract to a function.

src/main/java/at/hannibal2/skyhanni/features/misc/SkyBlockKickDuration.kt
```diff
@@ -49,12 +50,13 @@ object SkyBlockKickDuration {
 
     @HandleEvent
     fun onChat(event: SkyHanniChatEvent) {
-        if (!isEnabled() || !showTime) return
+        if (!isEnabled() || !timeoutKick) return
```
DavidArthurCole: Instead of using `timeoutKick` (the new boolean), compare against `lastKickTime`, and set that to either `null`, or `SimpleTimeMark.farPast()` when the user joins skyblock.

DavidArthurCole (changes requested):

hannibal002 (changes requested):
Davids comment

Fazfoxy:
World Change Removed, based on the logic that this feature is intended to be for marking how long you have to wait until you can re-enter Skyblock, Feature probably should be called something else or be more descriptive based on that, because changing world if you're messing about doing a duel while waiting with someone would prevoiusly do a full timer reset, even though SB rejoin timer is not affected

hannibal002:
image

PR #4117, opened 2025-06-21
Fix: Remove More British

DavidArthurCole:
## What
Oi bruv heard you wanted some updates to your detekt rules

exclude_from_changelog


DavidArthurCole:
(This is gonna fail detekt to justify why it was needed)

lunaynx:
A quick Google search will tell you that both American and British English prefer "dialogue" when referring to a conversation. Sometimes "dialog" is used for conversations in American, but it's not that common. Only "dialogue box" (as opposed to "dialog box") is a specifically British spelling.

This change is also likely to confuse players in the context of a game, as most of them will expect "dialog" to refer to a dialog box.

DavidArthurCole:
> Only "dialogue box" (as opposed to "dialog box") is a specifically British spelling.

Fair enough.

jani270 (changes requested):
Dialogue is even used by Hypixel, i dont think we should change it.

PR #4139, opened 2025-06-23
Fix: Dungeon finder features not working

NopoTheGamer:
## What
the change hannibal did yesterday cooked the feature

## Changelog Fixes
+ Fixed dungeon finder features. - nopo


CalMWolfs (approved):

hannibal002:
what was the issue?

NopoTheGamer:
> what was the issue?

image

hannibal002 (changes requested):
i meant what was the issue in code. when you change it back, sure. but please also add a todo or leave the old code in commented out so we can fix this properly later

NopoTheGamer:
> i meant what was the issue in code. when you change it back, sure. but please also add a todo or leave the old code in commented out so we can fix this properly later

i put a todo

PR #4140, opened 2025-06-23
Backend: Add sodium to /shdebug

NopoTheGamer:
## What
add soidum /shdebug

+ Added sodium to /shdebug. - nopo


src/main/java/at/hannibal2/skyhanni/utils/ComputerEnvDebug.kt
```diff
@@ -223,9 +223,9 @@ object ComputerEnvDebug {
 
     private fun getUptime() = ManagementFactory.getRuntimeMXBean().uptime.milliseconds
 
-    //#if MC == 1.8.9
     private fun performanceMods(event: DebugDataCollectEvent) {
```
j10a1n15, line 224: performanceMods doesnt get called yet since the calling is still preprocessed

NopoTheGamer: smart

PR #4142, opened 2025-06-23
Fix: Text Compat Lag

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1386733302907404338
I think this'll fix the insane lag that happens in beta rn when you try to send a deletable chat message over and over. Draft while testing though.

## Changelog Fixes
+ Fixed lag when using /shnavigate. - Daveed


DavidArthurCole:
Tested myself and noticed no noticeable frame drops when activating fast fairy souls - from the report it seems like it'd be pretty evident if the fix didn't work, so I suppose this is good for review?

CalMWolfs (dismissed):

PR #4149, opened 2025-06-24
Fix: Carnival Zombie Shootout lamps not highlighting in 1.21.5

OSzwed:
## What
Fixed Carnival Zombie Shootout lamp not being highlighted as the logic changed in 1.21. Redstone lamps no longer have lit in name, but they do in properties.

## Changelog Fixes
+ Fixed Carnival Zombie Shootout lamps not highlighting in 1.21. - Skw972

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/event/carnival/CarnivalZombieShootout.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/event/carnival/CarnivalZombieShootout.kt
@@ -29,6 +29,9 @@ import net.minecraft.item.Item
 import net.minecraft.item.ItemStack
 import java.awt.Color
 import kotlin.time.Duration.Companion.seconds
+//#if MC > 1.21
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/event/carnival/CarnivalZombieShootout.kt
```diff
@@ -171,6 +175,20 @@ object CarnivalZombieShootout {
             old == "lit_redstone_lamp" && new == "redstone_lamp" -> null
             else -> lamp
         }
+        //#else
+        //$$ val blockOld = event.old
+        //$$ val blockNew = event.new
+        //$$ println("$block $block2")
```
DavidArthurCole: println statement, please remove

DavidArthurCole (approved):
Thanks for the contribution, just one logging statement that needs removing, otherwise code LGTM.

PR #4150, opened 2025-06-24
Fix: Pet Data (For real real real real real real real this time)

DavidArthurCole:
## What
Trying to do things the right way always bites me in the ass.
Repo pets continue to have either missing fields, or straight up no pet info, so even more default values and parse attempts.

Also redid the logging so that:
- It does not repeat for the same item more than once a session
- The error will not show more than once a minute totally

## Changelog Fixes
+ Fixed additional pet errors in REI search. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
@@ -75,8 +75,8 @@ object SkyBlockItemModifierUtils {
     @KSerializable
     data class PetInfo(
         @Expose val type: String,
-        @Expose val active: Boolean? = null,
-        @Expose val exp: Double,
+        @Expose val active: Boolean = false,
+        @Expose val exp: Double = 0.0,
         @Expose val tier: LorenzRarity,
         @Expose val hideInfo: Boolean = false,
         @Expose val heldItem: NeuInternalName? = null,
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
```diff
@@ -137,18 +138,27 @@ object SkyBlockItemModifierUtils {
     @Suppress("CAST_NEVER_SUCCEEDS")
     inline val ItemStack.cachedData: CachedItemData get() = (this as ItemStackCachedData).skyhanni_cachedData
 
+    val warnedAboutPetParseFailure: MutableSet<String> = mutableSetOf()
+    var lastWarnedParseFailure: SimpleTimeMark = SimpleTimeMark.farPast()
+
     fun ItemStack.getPetInfo(): PetInfo? {
+        val colorlessName = displayName.removeColor()
+        // Repo pets will always return null for PetInfo, don't even attempt to parse it
+        if (colorlessName.contains("→")) return null
```
lunaynx:
```suggestion
        if (colorlessName.contains("→") || colorlessName.contains("{LVL}")) return null
```
https://discord.com/channels/997079228510117908/1387472530989650121/1387472530989650121

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/SkyBlockItemModifierUtils.kt
DavidArthurCole:
```
Extra data:
exception: 'No argument provided for a required parameter: parameter #2 exp of fun `<init>`(kotlin.String, kotlin.Boolean?, kotlin.Double,
```

That was set to default of 0.0 in this PR, so this still would fix that error - the arrow catcher was just a cherry on top - this is probably another good one to include though.

lunaynx (approved):

PR #4131, opened 2025-06-23
Improvement: Visual Words on 1.21

Bloxigus:
## What
image

A bit laggy in the editor, might be a drive issue (it did break recently)

image
image

detekt have mercy on me :pray:

## Changelog Improvements
+ Added Visual Words (/shwords) support for 1.21. - bloxigus

PR #4157, opened 2025-06-25
Feature + Fix: allow hex codes in legacy text & fix chat patches timestamp

ninjaguardian:
## Dependencies
- #4131 (added in 3.12.0)

## What
Fixes [this](https://discord.com/channels/997079228510117908/1387131158017605703) bug reported on Discord.

If any non-vanilla colors (aka not §1 or stuff like it) were sent by a mod or other, SkyHanni would try (and fail) to convert it to a vanilla color code (i.e. §1). This would make it show up as white.

I made it such that if `requiredStyleChangeString` encounters a non-vanilla color, it adds it as such: `§#RRGGBB`.
When converting back, if a `§#` is found, the next six chars are processed and reapplied with `.withColor`.
I also changed the for loop to a while loop so that the six characters can be processed immediately.

## Changelog New Features
+ Made legacy color codes support non-vanilla colors. - ninjaguardian + bloxigus

## Changelog Fixes
+ Fixed Chat Patches timestamps showing as white. - ninjaguardian + bloxigus

### DIFF
--- a/versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
+++ b/versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
@@ -93,24 +93,31 @@ object OrderedTextUtils {
 
         return OrderedText { visitor ->
             var lastStyle = Style.EMPTY
-            var wasLastStyle = false
-
-            for (char in legacyString ?: "") {
-
-                if (char == '§') {
-                    wasLastStyle = true
-                } else if (wasLastStyle) {
-
-                    val formatting = Formatting.byCode(char)
+            val str = legacyString ?: ""
+            var i = 0
+            while (i < str.length) {
+                val c = str[i]
+                if (c == '§' && i + 1 < str.length) {
+                    val code = str[i + 1]
+                    if (code == '#' && i + 7 < str.length) {
+                        val hexPart = str.substring(i + 2, i + 8)
+                        val rgb = hexPart.toIntOrNull(16)
+                        if (rgb != null) {
+                            lastStyle = lastStyle.withColor(TextColor.fromRgb(rgb))
+                            i += 8  // §#RRGGBB
+                            continue
+                        }
+                    }
+                    val formatting = Formatting.byCode(code)
                     if (formatting != null) {
                         lastStyle = lastStyle.withExclusiveFormatting(formatting)
-                    } else if (char == 'z') {
+                    } else if (code == 'z') {
                         lastStyle = lastStyle.withColor(CHROMA_COLOR)
                     }
-
-                    wasLastStyle = false
+                    i += 2
                 } else {
-                    visitor.accept(0, lastStyle, char.code)
+                    visitor.accept(0, lastStyle, c.code)
+                    i += 1
                 }
             }
             true
@@ -138,6 +145,9 @@ object OrderedTextUtils {
             } else {
                 to.color?.toChatFormatting()?.let {
                     sb.append(it.toString())
+                } ?: run {
+                    sb.append("§")
+                    sb.append(to.color?.hexCode)
                 }
             }
         } else if (reset) {
--- a/versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
+++ b/versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
@@ -94,17 +94,17 @@ object OrderedTextUtils {
         return OrderedText { visitor ->
             var lastStyle = Style.EMPTY
             val str = legacyString ?: ""
-            var i = 0
-            while (i < str.length) {
-                val c = str[i]
-                if (c == '§' && i + 1 < str.length) {
-                    val code = str[i + 1]
-                    if (code == '#' && i + 7 < str.length) {
-                        val hexPart = str.substring(i + 2, i + 8)
+            var index = 0
+            while (index < str.length) {
+                val char = str[index]
+                if (char == '§' && index + 1 < str.length) {
+                    val code = str[index + 1]
+                    if (code == '#' && index + 7 < str.length) {
+                        val hexPart = str.substring(index + 2, index + 8)
                         val rgb = hexPart.toIntOrNull(16)
                         if (rgb != null) {
-                            lastStyle = lastStyle.withColor(TextColor.fromRgb(rgb))
-                            i += 8  // §#RRGGBB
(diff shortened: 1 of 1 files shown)


Bloxigus:
:(

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
```diff
@@ -93,24 +93,31 @@ object OrderedTextUtils {
...
+            var i = 0
+            while (i < str.length) {
+                val c = str[i]
+                if (c == '§' && i + 1 < str.length) {
+                    val code = str[i + 1]
+                    if (code == '#' && i + 7 < str.length) {
+                        val hexPart = str.substring(i + 2, i + 8)
+                        val rgb = hexPart.toIntOrNull(16)
+                        if (rgb != null) {
+                            lastStyle = lastStyle.withColor(TextColor.fromRgb(rgb))
```
Bloxigus: probably better to do this
```suggestion
                            lastStyle = lastStyle.withColor(rgb)
```

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
```diff
@@ -93,24 +93,31 @@ object OrderedTextUtils {
...
-                } else if (wasLastStyle) {
-
-                    val formatting = Formatting.byCode(char)
+            val str = legacyString ?: ""
+            var i = 0
+            while (i < str.length) {
+                val c = str[i]
+                if (c == '§' && i + 1 < str.length) {
+                    val code = str[i + 1]
+                    if (code == '#' && i + 7 < str.length) {
```
Bloxigus: This will cause an issue if the color is at the end of the string

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
```diff
@@ -93,24 +93,31 @@ object OrderedTextUtils {
...
+                        val hexPart = str.substring(i + 2, i + 8)
+                        val rgb = hexPart.toIntOrNull(16)
+                        if (rgb != null) {
+                            lastStyle = lastStyle.withColor(TextColor.fromRgb(rgb))
+                            i += 8  // §#RRGGBB
+                            continue
+                        }
+                    }
```
Bloxigus, line 110: The below should be in an else, no formatting code exists with `#`, which is why I was going to use it

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
```diff
@@ -93,24 +93,31 @@ object OrderedTextUtils {
...
-
-            for (char in legacyString ?: "") {
-
-                if (char == '§') {
-                    wasLastStyle = true
-                } else if (wasLastStyle) {
-
-                    val formatting = Formatting.byCode(char)
+            val str = legacyString ?: ""
+            var i = 0
```
Bloxigus: You should probably name this something descriptive, generally avoid single letter variable names
```suggestion
            var index = 0
```

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
```diff
@@ -93,24 +93,31 @@ object OrderedTextUtils {
...
+            while (i < str.length) {
+                val c = str[i]
```
Bloxigus: See above
```suggestion
                val char = str[index]
```

Bloxigus (changes requested):

Bloxigus:
Also this probably doesn't fix that bug, as chat isn't replaced in the same way as everything else in the game
\+ this is just a fix, not an improvement

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
ninjaguardian: If the string was "§#FFFFFF", when index == 0, code == '#'. 0 + 7 < 8, so I don't see how it would error.

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
ninjaguardian: If the chars after the `#` make a valid hex code, `continue` skips that code. If they don't, that code skips over the `§#` and treats the non-hex codes as normal chars. If a `#` was accidentally used instead of a valid color code, this wouldn't leave random text. By adding `else`, it would get stuck in a loop. Even if you added a `index += 1` to fix that, it would just leave a `#RRGGBB` in the middle of the text. It just depends on what is considered more 'user friendly'.

ninjaguardian:
> Also this probably doesn't fix that bug, as chat isn't replaced in the same way as everything else in the game + this is just a fix, not an improvement

I tested the code, and chat patches' timestamps worked fine with non-vanilla colors.

ninjaguardian:
I can make rainbow timestamps with aaron's mod with color #AA5500
image
(at least I think it's arron's mod because it says to use that hex. nonetheless, cool!)

hannibal002 (changes requested):
How can I test this bug?
In beta, I didn't notice this bug:
I already tried `§c§ghey` in clipboards and then sending the cmd `/shtestmessage -clipboard`. And I got what I expected.
image

ninjaguardian:
> How can I test this bug? In beta, I didn't notice this bug: I already tried `§c§ghey` in clipboards and then sending the cmd `/shtestmessage -clipboard`. And I got what I expected. image

Download the chatpatches mod and change the timestamp color to any color that is not a minecraft vanilla color. It will show as white.

hannibal002:
What is the chatpaches mod technically sending in chat?

ninjaguardian:
> What is the chatpaches mod technically sending in chat?

https://github.com/mrbuilder1961/ChatPatches/blob/4c3cfd619c8ad481fe94a5cc2f4da3d009209659/src/main/java/obro1961/chatpatches/config/Config.java#L125
It sends the time and uses `.withColor` to change the color, but SkyHanni removed it if it was not a vanilla color.

hannibal002:
Yield to #4159

PR #4159, opened 2025-06-25
Fix: Aaron's Mod + Meowdding + Any mod that uses custom colors compat

Bloxigus:
Fixes compatibility with Aaron's Mod
Without this change, Aaron's chroma doesn't render correctly (not at all).

i can remove the stuff that overlaps with that other person's changes, i didnt remove it because i already wrote it 🥺 (this is what i get for going to sleep for a test :sob:)

## Changelog Fixes
+ Fixed Visual Words stripping non-Minecraft colors. - bloxigus


hannibal002:
Can we get a bug fix changelog on this?

hannibal002:
Does this PR replaces #4157 ?

ninjaguardian:
> Does this PR replaces #4157 ?

I think so. Both PRs have pretty similar code for the # stuff.

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
```diff
@@ -138,6 +185,13 @@ object OrderedTextUtils {
             } else {
                 to.color?.toChatFormatting()?.let {
                     sb.append(it.toString())
+                } ?: {
+
+                    to.color?.let {
+                        val hexString = it.rgb.toString(16).padStart(6, '0')
```
ninjaguardian: Isn't `it.hexCode` a better way to do this? (Note: hexCode already has a #)

hannibal002:
So what of those two prs should we now keep up?

ninjaguardian:
> So what of those two prs should we now keep up?

Whatever's more optimized I guess. Not sure which one is though...
[This](https://chatgpt.com/s/t_685c6c14823c819184b42be04af98508) is what chatgpt thought (I didn't know what else to ask so AI it is :P)

Maybe the best of both?

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
Bloxigus: Great catch, I didnt see that method when looking for ways to get it as a string

hannibal002 (changes requested):
does not work in my testing with Chat Patches
image

image

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
```diff
@@ -57,29 +56,52 @@ object OrderedTextUtils {
...
-                } else if (char == 'z') {
-                    lastStyle = lastStyle.withColor(CHROMA_COLOR)
+                    val nextChar = legacyString[index + 1]
+
+                    if (legacyString[index + 1] == '#') {
+
+                        if (index + 8 < legacyString.length) {
+                            val hexColor = legacyString.substring(index + 2, index + 8)
+
+                            hexColor.toIntOrNull(16) ?: { color: Int ->
```
ninjaguardian, line 81: This one also needs to be updated
```suggestion
                            hexColor.toIntOrNull(16) ?.let { color: Int ->
```
(I thought `.also` was the right one to use here instead of `.let` because nothing's being returned, but maybe not. dosent really matter anyways.)

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrdereredTextUtils.kt
```diff
@@ -92,26 +114,51 @@ object OrderedTextUtils {
...
+                    if (index + 1 < legacyString.length) {
+
+                        val nextChar = legacyString[index + 1]
+
+                        if (legacyString[index + 1] == '#') {
+
+                            if (index + 8 < legacyString.length) {
+                                val hexColor = legacyString.substring(index + 2, index + 8)
+
+                                hexColor.toIntOrNull(16) ?: { color: Int ->
```
ninjaguardian: Same as above
```suggestion
                                hexColor.toIntOrNull(16) ?.also { color ->
```

hannibal002 (approved):
image

ninjaguardian (commented):
~~Can parts of legacyTextToOrderedText and legacyStringToStringVisitable be combined and put into another function so that it's easier to edit?~~
nvm it's merged

ninjaguardian:
Wait the `?:` still exists :/
legacyStringToStringVisitable still uses `?:` instead of `?.let` or `?.also`

PR #4156, opened 2025-06-25
Feature: Hide status effects from top right

NopoTheGamer:
## What
Hides effects from top right
I also put only legacy on 2 neu config options

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added Hide Effects from top right. - nopo

PR #4163, opened 2025-06-25
Feature: Status effect hider

hannibal002:
## What
saw #4156 too late. anyway, this PR adds both status effect hiders in one.

## Changelog New Features
+ Added Hide Status Effects. - hannibal2
    * Hide the status effect displays at top right of the screen and in inventory.
    * Only in 1.21, as 1.8 doesn't have those status effect displays.


hannibal002:
Yield to https://github.com/nea89o/Firmament/pull/165

PR #4164, opened 2025-06-25
Fix: hide not clickable items dark foreground

hannibal002:
## What
my first try at bringing back the dark foreground. cant get it to work, idk why

image

Probably a better solution would be to wrap this into a version generic utils that then gets special implemented by 1.21

## Changelog Fixes
+ Fixed dark foreground not showing for Hide Not Clickable Entities. - hannibal2


CalMWolfs:
already fixed in #4160

hannibal002:
Replaced by #4160

PR #4171, opened 2025-06-26
Fix: Foraging tracker not hiding when the config is off

NopoTheGamer:
## What
Foraging tracker not hiding when the config is off

## Changelog Fixes
+ Fixed Foraging tracker not hiding when the config is off. - nopo


CalMWolfs (approved):

hannibal002:
there is a bigger problem with foraging tracker and enabled and visibility. will do a pr in a sec

PR #4188, opened 2025-06-27
Backend: Auto Config Bump

Thunderblade73:
## What
Finally got it working with branch protection rules.

@hannibal002 Detailed setup after the images but the TLDA:
before merging we need to setup App under your name, add the 2 secrets from that app to the repo, "install" the app in the repo, then we need to change the current branch portection rule to a rulset (which can do the same but allows for specific bypass list), then add the github app to the bypass list.

Here the info copied from #3346:
We now have a "CONFIG_MOVE_VERSION" which will automatically bump the config version up by one.
On merge with it will be substituted with the version and the config version will be incremented.

In dev enviroment it always performs a config fix for ConfigVersion + 1 to process newly moves.

<details>
<summary>Images</summary>

image

</details>

<details>
<summary>Setup Github App</summary>

Go into your profile settings.
image
image
Give it a name and simply add the hompage URL to the sh repo.
image
Untick webhook.
image
image
Then press create
image
Then add the APP-ID as repository secret "CONFIG_BUMP_ID".
Then create a private Key for the APP and copy and paste the whole content of the file (including whitespace) to repository secret "CONFIG_BUMP_KEY"

</details>

<details>
<summary>Ruleset</summary>

You need to make the branch protection rule a branch ruleset. Work the same way but for now we copy paste the "Require status checks to pass" setting from the branch protection rule to the ruleset (nothing else). And remove them from the branch protection rule. Btw. needs to be set to active for it to apply.

The only thing the workflow needs is that it should be in the bypass list.

image

</details>

+ Added automatic config bumps. - Thunderblade73
    * Use "CONFIG_MOVE_VERSION" to get a version for newly defined config moves.


CalMWolfs:
@Thunderblade73 bump

Thunderblade73:
> @Thunderblade73 bump

I just hate to update for every config move (That's the conflict here + some small change in Module Processor and Contributing.md)

CalMWolfs:
> I just hate to update for every config move (That's the conflict here + some small change in Module Processor and Contributing.md)

Yeah this makes sense, unfortunately i read through the pr and it doesnt seem like something I can add so you'll have to wait for hannibal to be back first

CONTRIBUTING.md
```diff
@@ -262,8 +262,10 @@ Make sure such pull requests have a good explanation in the **What** section.
 - Do not use `fixedRateTimer` when possible and instead use `SecondPassedEvent` to safely execute the repeating event on
   the main thread.
 - When updating a config option variable, use the `ConfigUpdaterMigrator.ConfigFixEvent` with event.move() when moving a value, and
-  event.transform() when updating a
-  value. [For Example](https://github.com/hannibal002/SkyHanni/blob/e88f416c48f9659f89b7047d7629cd9a1d1535bc/src/main/java/at/hannibal2/skyhanni/features/gui/customscoreboard/CustomScoreboard.kt#L276).
+  event.transform() when updating a value.
+    - The config version parameter should be ``CONFIG_MOVE_VERSION`` which will be replaced with the upcoming config version during runtime
```
CalMWolfs:
```suggestion
    - The config version parameter should be `CONFIG_MOVE_VERSION` which will be replaced with the upcoming config version during runtime 
```

Thunderblade73:
I do not want the blame of a security flaw. So I just close this.

PR #4192, opened 2025-06-27
Fix: Pet Data Over-Saving

DavidArthurCole:
## What
Makes pet data store at max every 30 seconds instead of every time there's an update to data.

## Changelog Fixes
+ Fixed lag related to Pet data. - Daveed


CalMWolfs (approved):

hannibal002:
so much extra code just bc we dont have dynamic config separation on load/save :(

PR #4195, opened 2025-06-27
Feature: Add Invisibug highlight

mariethepolarfox:
## What
Adds Invisibug highlighting.

## Changelog New Features
+ Added Invisibug Highlight. - Helium9


versions/1.21.5/src/main/java/at/hannibal2/skyhanni/features/foraging/InvisibugHighlighter.kt
Thunderblade73, line 1: EntityUtils.getAllEntities() is one of the most expensive calls we have (the only more expensive calls is the reloading neu items and doing a TSP). This feature therefore will kill performance, as it calls that function multiple times each render frame. In the particle event you can use EntityUtils.getEntitiesNearby which is better, btw you should add some filter here, as any armorstand will do here. Move the logic of the render event for clearing into a tick event and use getEntities<ArmorStandEntity> at least and use that helper function (Collection Utils is the place where that should go):
```
    fun <T> Sequence<T>.removeNotContainedFromList(list: MutableList<T>) {
        val checkList = Array<Boolean>(list.size) { false }
        for (element in this){
            val i = list.indexOfFirst(element) ?: continue
            checkList[i] = true
        }
        checkList.withIndex().filter { !it.value }.forEach {
            list.removeAt(it.index)
        }
    }
 ```
 As this will remove the multiple filtering of armorstands, as that only happens once with that function.

File: versions/1.21.5/src/main/java/at/hannibal2/skyhanni/features/foraging/InvisibugHighlighter.kt [hunk 1]
```diff
@@ -0,0 +1,68 @@
...
+    @HandleEvent(onlyOnIsland = IslandType.GALATEA)
+    fun onRenderWorld(event: SkyHanniRenderWorldEvent) {
+        if (!isEnabled()) return
+
+        invisibugEntities.removeIf() {
+            !EntityUtils.getAllEntities().filterIsInstance<ArmorStandEntity>().contains(it)
+        }
+
+        for (entity in invisibugEntities) {
+            if (entity.isDead || entity.distanceToPlayer() > 20) continue
```
Thunderblade73: This is a handrolled .canBeSeen() call, so use the function instead.

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/features/foraging/InvisibugHighlighter.kt
```diff
@@ -0,0 +1,68 @@
...
+import net.minecraft.entity.decoration.ArmorStandEntity
+import net.minecraft.particle.ParticleTypes
+import kotlin.time.Duration.Companion.seconds
+
+@SkyHanniModule
+object InvisibugHighlighter {
+    val config get() = SkyHanniMod.feature.foraging.foragingMobHighlight.invisibugHighlight
+
+    private fun isEnabled() = config.enabled && isInIsland()
+    private fun isInIsland() = IslandTypeTags.FORAGING_CUSTOM_TREES.inAny()
```
Thunderblade73: That function does not really make sense to me, as the events below already are filtered to only work on GALATEA.

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/features/foraging/InvisibugHighlighter.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    private val particlesToHighlight = mutableMapOf<LorenzVec, SimpleTimeMark>()
```
Thunderblade73: unused?

Thunderblade73 (changes requested):

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/features/foraging/InvisibugHighlighter.kt
mariethepolarfox: Oh I forgot to remove this from debugging

src/main/java/at/hannibal2/skyhanni/config/features/foraging/InvisibugHighlightConfig.kt
```diff
@@ -0,0 +1,27 @@
...
+class InvisibugHighlightConfig {
+    @Expose
+    @ConfigOption(name = "Highlight Invisibug", desc = "Highlights nearby Invisibug.")
+    @FeatureToggle
+    @ConfigEditorBoolean
+    @OnlyModern
+    var enabled: Boolean = true
+
+    @Expose
+    @ConfigOption(name = "Color", desc = "Color for the Hideonleaf highlight")
```
NopoTheGamer: stole my code :(

mariethepolarfox: Yes im lazy !!

mariethepolarfox: Fixed though

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/features/foraging/InvisibugHighlighter.kt
mariethepolarfox: I think how I did it is how you want it to be? (Also didnt know getAllEntites was so expensive 😭)

src/main/java/at/hannibal2/skyhanni/config/features/foraging/InvisibugHighlightConfig.kt
```diff
@@ -8,19 +8,21 @@ import io.github.notenoughupdates.moulconfig.ChromaColour
 import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorBoolean
 import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorColour
 import io.github.notenoughupdates.moulconfig.annotations.ConfigOption
-import io.github.notenoughupdates.moulconfig.annotations.SearchTag
 import java.awt.Color
 
+/**
+ * Code stolen from @NopoTheGamer (im lazy)
+ */
 class InvisibugHighlightConfig {
```
NopoTheGamer: i was joking dont credit me in ur config file 💀

mariethepolarfox: Consider it as a fun easteregg :D

src/main/java/at/hannibal2/skyhanni/utils/collection/CollectionUtils.kt
```diff
@@ -497,4 +497,17 @@ object CollectionUtils {
...
     }
+
+    fun <T> Sequence<T>.removeNotContainedFromList(list: MutableList<T>) {
+        val checkList = Array<Boolean>(list.size) { false }
+        for (element in this){
+            val i = list.indexOfFirstOrNull { it == element } ?: continue
+            checkList[i] = true
+        }
+        checkList.withIndex().filter { !it.value }
+            .sortedByDescending { it.index }
```
Thunderblade73: ``.reversed()`` is better here, as that should not cost anything. But the sort cost O(log(n)*n).
(It could be called ``.reverse()`` I don't remember which one is for lists.)

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/features/foraging/InvisibugHighlighter.kt
```diff
@@ -3,60 +3,48 @@ package at.hannibal2.skyhanni.features.foraging
...
         val particle = event.type
         if (particle != ParticleTypes.CRIT) return
 
-        val nearestArmorStand = EntityUtils.getAllEntities()
-            .filterIsInstance<ArmorStandEntity>()
+        val nearestArmorStand = EntityUtils.getEntitiesNearby<ArmorStandEntity>(event.location, 5.0)
             .minByOrNull { it.distanceTo(event.location) }
 
-        if (nearestArmorStand == null || nearestArmorStand.distanceTo(event.location) > 5) return
+        if (nearestArmorStand == null) return
```
Thunderblade73: Add an is default check here, as you only look for armorstands that have the default name (looked at the shcopyentitys you send in contributer chat)
```suggestion
        if (nearestArmorStand == null || !nearestArmorStand.isDefaultValue()) return
```

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/features/foraging/InvisibugHighlighter.kt
Thunderblade73: > I think how I did it is how you want it to be?

Yeah, exactly.
> Also didnt know getAllEntites was so expensive 😭

It is a list that has 800+ entries in regular cases, that's why it is bad. We can definitely use it, but we should reduce those calls as much as possible (especially as we already call it to often, imo).

Thunderblade73 (changes requested):

mariethepolarfox:
Looks like frog pet is highlighted? I don't have one so if someone could test would be appreciated

CalMWolfs:
> Looks like frog pet is highlighted? I don't have one so if someone could test would be appreciated

Haven't tested but maybe in addition to checking if the stand is default value which checks the name, you should also check if its armor items are all empty and is holding nothing in its hand

src/main/java/at/hannibal2/skyhanni/config/features/foraging/ForagingConfig.kt
```diff
@@ -44,16 +44,9 @@ class ForagingConfig {
...
-    @Accordion
-    var birriesHighlight = BirriesHighlightConfig()
-
-    @Expose
-    @ConfigOption(name = "Hideonleaf Highlight", desc = "")
-    @OnlyModern
-    @Accordion
-    var hideonleafHighlight = HideonleafHighlightConfig()
+    @Category(name = "Foraging Mob Highlights", desc = "Settings for foraging mob highlights")
+    var foragingMobHighlight = ForagingMobHighlightConfig()
```
hannibal002: remove the rendundant/duplicate "foraging" in the variable name

src/main/java/at/hannibal2/skyhanni/utils/collection/CollectionUtils.kt
```diff
@@ -497,4 +497,17 @@ object CollectionUtils {
             }
         }
     }
+
+    fun <T> Sequence<T>.removeNotContainedFromList(list: MutableList<T>) {
```
hannibal002: kotlin stdlib already has `retainAll`

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/features/foraging/InvisibugHighlighter.kt
```diff
@@ -0,0 +1,64 @@
...
+                extraSize = -0.2
+            )
+        }
+    }
+
+    @HandleEvent(onlyOnIsland = IslandType.GALATEA)
+    fun onTick(event: SkyHanniTickEvent) {
+        if (!config.enabled) return
+
+        EntityUtils.getEntities<ArmorStandEntity>().removeNotContainedFromList(invisibugEntities)
```
hannibal002: this should be a generic function in EntityUtils with the name "remove invalid entities" or something

hannibal002 (changes requested):
code looks overall good, just small things ive noticed

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/features/foraging/InvisibugHighlighter.kt
mariethepolarfox: I added something like this now, lmk if its good or if I should change something

hannibal002:
image

hannibal002 (changes requested):
idk how to test this.
idk what this mob even is.
cant find anything in the galatea isalnd that is highlighted in this color while testing.
i also dont find any sources of what this is on any sb wiki.
am i missing some precondition? do i need to enable someting or do something specific to see those invisibug mobs?

maybe we can include the info what vanilla mc entity this sb mob is?

CalMWolfs (approved):
Works fine for me
image

hannibal002 (approved):
image

aroderg:
bruh

PR #4205, opened 2025-06-28
Fix: Hide vanilla enchants applying outside of SkyBlock

CalMWolfs:
## What
Fixed hide vanilla enchants applying outside of SkyBlock on 1.21

## Changelog Fixes
+ Fixed hide vanilla enchants applying outside SkyBlock on 1.21. - CalMWolfs


ninjaguardian:
I don't understand how this bug is present in the first place.
`EnchantParser.kt`
```kt
    fun onTooltipEvent(event: ItemHoverEvent) {
    
...

    @HandleEvent
    fun onChatHoverEvent(event: ChatHoverEvent) {
    ...
    if (!isEnabled() || !this.enchants.hasEnchantData()) return
```

`isEnabled()` checks if you are in Skyblock

I checked, and this bug did occur.

Am I missing something? (almost definitely yes :/)

Nonetheless, code looks good!

NopoTheGamer:
> I don't understand how this bug is present in the first place. `EnchantParser.kt`
> 
probably because the code isnt related to enchant parser.kt

ninjaguardian:
> > I don't understand how this bug is present in the first place. `EnchantParser.kt`
> 
> > 
> 
> probably because the code isnt related to enchant parser.kt
> 
> 
> 
>

I thought that enchant parser handled hiding vanilla enchants.

CalMWolfs:
> I thought that enchant parser handled hiding vanilla enchants.

It does on 1.8 but we have a different implementation on 1.21 due to them being translatable

PR #4206, opened 2025-06-28
Fix: aaron's mod enchant fixes (and a few other things)

ninjaguardian:
## What
When using enchant formatting from both SkyHanni and Aaron's Mod at the same time, some enchants fail to render. You aren't supposed to use both in the first place, but many players (including me) probably never realize it.

In this pr I do quite a bit of stuff. First, I make it so that SkyHanni can process enchants that have already been processed by Aaron's mod. Second, I add more regex tests because why not. Third, I make some small optimizations with `OrderedTextUtils` (by first checking `sb.isNotEmpty()`, we can skip adding empty formatted strings. It just wastes time and memory). I also rewrote `OtherModsSettings` with some optimizations, fixes (null values should not be assumed to be false), and additions (including handling non-boolean values).

The fix for Aaron's Mod is such: if SkyHanni will format it, then remove Aaron's formatting. If SkyHanni won't, then change the regex to recognize the formatting from Aaron's mod.

This pr may need the API to be updated with the regex for `enchants.aaronchroma` and `enchants.aaronstill`.

(Also, for some reason, the file for `OrderedTextUtils` was called `OrdereredTextUtils.kt`, so I fixed it.)

## Changelog Fixes
+ Fixed processing of enchantments already processed by Aaron's Mod. - ninjaguardian
    * Even though it now works, it's not the best idea to have both mods process enchantments.
    
+ Optimize and improve `OtherModsSettings`. - ninjaguardian
    * Any current uses of `OtherModsSettings` were changed to work with the fix.
+ Added some regex tests for enchant regexes. - ninjaguardian
    * Located in `EnchantParser.kt`.
+ Rename `OrdereredTextUtils.kt` to `OrderedTextUtils.kt`. - ninjaguardian
+ Small optimization to an instance of `StringBuilder`. - ninjaguardian
    * Located in `OrderedTextUtils.kt`.


ninjaguardian:
I'll fix detekt errors tomorrow.

src/main/java/at/hannibal2/skyhanni/utils/OtherModsSettings.kt
```diff
@@ -1,25 +1,108 @@
...
+                runCatching {
+                    clazz.getField(name).apply { isAccessible = true }
+                }.getOrElse {
+                    clazz.getDeclaredField(name).apply { isAccessible = true }
+                }
+            }
+
+        fun clearClassCache() { classCache.clear() }
+        fun clearFieldCache() { fieldCache.clear() }
+        fun clearCaches() {
```
hannibal002, line 27: why/when would we want to clear this cache?

src/main/java/at/hannibal2/skyhanni/utils/OtherModsSettings.kt
```diff
@@ -1,25 +1,108 @@
...
     companion object {
-        fun patcher(): OtherModsSettings = getModPath("club.sk1er.patcher.config.PatcherConfig")
+        private val classCache = ConcurrentHashMap<String, Class<*>>()
+        private val fieldCache = ConcurrentHashMap<Triple<String, Class<*>, String>, Field>()
+
+        fun patcher() = OtherModsSettings("club.sk1er.patcher.config.PatcherConfig")
+        fun aaron() = OtherModsSettings("net.azureaaron.mod.config.AaronModConfigManager")
+
+        internal fun loadClass(path: String): Class<*>? =
+            runCatching { classCache.computeIfAbsent(path) { Class.forName(path) } }
```
hannibal002: we have never used computeIfAbsent anywhere in SkyHanni. please use the kotlin alternate to increase code consistency overall.

src/main/java/at/hannibal2/skyhanni/features/misc/PatcherFixes.kt
```diff
@@ -20,7 +20,7 @@ object PatcherFixes {
         lastCheck = SimpleTimeMark.now()
 
         val patcher = OtherModsSettings.patcher()
-        if (!patcher.getBoolean("parallaxFix")) return
+        if (patcher.getBoolean("parallaxFix") == false) return
```
hannibal002: why do we return nullable here? whats the benefit? can we just assume if the mod reading process fails/is null, it returns false?

src/main/java/at/hannibal2/skyhanni/utils/OtherModsSettings.kt
```diff
@@ -1,25 +1,108 @@
...
+        runCatching { chain.last().set(current, value) }
+    }
+
+    private fun buildFieldChain(path: String): List<Field> {
+        val segments = path.split('.')
+        val rootPair = getOptionPair(segments.first()) ?: return emptyList()
+        val fields = mutableListOf(rootPair.second)
+        var currentClass = rootPair.second.type
+        for (seg in segments.drop(1)) {
+            val f = getField(modConfigPath, currentClass, seg)
```
hannibal002: please rename f to field. we never use one char variable names.

src/main/java/at/hannibal2/skyhanni/utils/OtherModsSettings.kt
```diff
@@ -1,25 +1,108 @@
...
+        }
+        runCatching { chain.last().set(current, value) }
+    }
+
+    private fun buildFieldChain(path: String): List<Field> {
+        val segments = path.split('.')
+        val rootPair = getOptionPair(segments.first()) ?: return emptyList()
+        val fields = mutableListOf(rootPair.second)
+        var currentClass = rootPair.second.type
+        for (seg in segments.drop(1)) {
```
hannibal002: please dont use abbreviations like this that arent very common in the kotlin/hypixel sb scene

hannibal002 (changes requested):
Just, a bit of nitpicking. The overall changes look fine to me.

hannibal002:
> Optimize and improve OtherModsSettings. - ninjaguardian

this is an entirely internal feature. there is no user facing impact of this change. 
please move this to technical details.


> Users, even though it now works, it's not the best idea to have both mods process enchants. Choose one or the other.

please rephrase to not directly talk towards someone. this feels very out of style of skyhanni changelog wordings.


Additionally, when we already detect if the other mods feature is enabled, send a chat message periodically/self replacing, that suggests them to disable one of those two mods. see

src/main/java/at/hannibal2/skyhanni/features/misc/PatcherFixes.kt
ninjaguardian: Well then if getBoolean returns false, you have to guess whether it exists and is false, or does not exist. In some cases, that's fine (i.e. all the existing ones), but if, for example, a mod has a config that won't always have a value initialized, you'll be assuming all those non existent booleans are false.

I just feel like this adds more clarity and could be useful should OtherModSettings be used more.

src/main/java/at/hannibal2/skyhanni/utils/OtherModsSettings.kt
ninjaguardian: Mainly either in testing or to free up space.

Or maybe if somehow the mod is able to modify it’s own config structure (idk how though) you would have to invalidate the cache.

I guess it doesn’t really serve much purpose.

src/main/java/at/hannibal2/skyhanni/utils/OtherModsSettings.kt
ninjaguardian: Replaced with getOrPut

src/main/java/at/hannibal2/skyhanni/utils/OtherModsSettings.kt
ninjaguardian: All 1 letter variables were altered.

src/main/java/at/hannibal2/skyhanni/utils/OtherModsSettings.kt
ninjaguardian: Renamed to segment.

ninjaguardian:
I'm a bit confused by your phrasing. Is the warning already implemented?

Anyways, the other stuff is fixed. I'm not at my computer so I had to code it with a phone, but I'm pretty sure the build will catch any mistakes.

src/main/java/at/hannibal2/skyhanni/features/misc/PatcherFixes.kt
ninjaguardian: As another example, let's say for some reason, a mod has a setting called 'disable feature'. If you want code to run only if that feature is on, you have to run it if the value is false. The issue is, if it returns false, you can't tell if it is actually false or nonexistent (and if it is nonexistent, you'll be running code like it's on).

hannibal002: there is no test command. either remove the cache stuff or create a test command so this is no dead code please. if you like to keep it around, clean it up. there is duplicate clear cache code in this file

hannibal002:
> I'm a bit confused by your phrasing. Is the warning already implemented?

no its not. but since you now have detection if both features are enabled simultaneous its useful to have the chat warning instead of only warn in the config desc or changelog

src/main/java/at/hannibal2/skyhanni/features/misc/PatcherFixes.kt
hannibal002: > but if, for example, a mod has a config that won't always have a value initialized, you'll be assuming all those non existent booleans are false.

If the config is not initialized, then the feature from the other mod is not enabled, and we are right to assume the underlying feature is not working.

src/main/java/at/hannibal2/skyhanni/features/misc/items/enchants/EnchantsJson.kt
```diff
@@ -25,13 +26,18 @@ class EnchantsJson {
         return enchant
     }
 
-    fun containsEnchantment(enchants: Map<String, Int>, line: String): Boolean {
+    @Suppress("MaxLineLength")
```
hannibal002: do not use this suppression. instead, expand the function parameter definion over multiple lines

src/main/java/at/hannibal2/skyhanni/features/misc/items/enchants/EnchantsJson.kt
```diff
@@ -25,13 +26,18 @@ class EnchantsJson {
...
+    fun containsEnchantment(enchants: Map<String, Int>, line: String, enchantmentPattern: Pattern, removeFormattingCodes: Boolean): Boolean {
         val exclusiveMatch = EnchantParser.enchantmentExclusivePattern.matcher(line)
         if (!exclusiveMatch.find()) return false // This is the case that the line is not exclusively enchants
 
-        val matcher = EnchantParser.enchantmentPattern.matcher(line)
+        val matcher = enchantmentPattern.matcher(line)
```
hannibal002, line 39: any reason why we dont use matchMatcher here?

src/main/java/at/hannibal2/skyhanni/features/misc/items/enchants/EnchantsJson.kt
```diff
@@ -25,13 +26,18 @@ class EnchantsJson {
...
+        val formattingCodesRegex = "§[0-9a-fr]".toRegex()
```
hannibal002: i think we already have this list of variables somehwere. if not, at least put it outside the function so its not getting recreated on every function call

src/main/java/at/hannibal2/skyhanni/utils/OtherModsSettings.kt
```diff
@@ -1,25 +1,108 @@
...
 
-    fun getBoolean(optionPath: String): Boolean = getOption(optionPath)?.get(null) as? Boolean ?: false
+    fun isLoaded(): Boolean = loadClass(modConfigPath) != null
+
+    private val chainCache = ConcurrentHashMap<Pair<String, String>, List<Field>>()
 
-    fun setBoolean(optionPath: String, value: Boolean) {
-        getOption(optionPath)?.set(null, value)
+    fun getBoolean(path: String): Boolean? = getNestedValue<Boolean>(path)
+    fun setBoolean(path: String, value: Any) = setNestedValue(path, value)
```
hannibal002: second para should be boolean

hannibal002 (changes requested):
more small things ive noticed (plus reply above)

src/main/java/at/hannibal2/skyhanni/features/misc/PatcherFixes.kt
ninjaguardian: False does not always mean not enabled though.

src/main/java/at/hannibal2/skyhanni/features/misc/items/enchants/EnchantsJson.kt
ninjaguardian: It didn't already use matchMatcher so I didn't add it. Not completely sure how matchMatcher works, but it would probably be fine to add it.

src/main/java/at/hannibal2/skyhanni/features/misc/PatcherFixes.kt
ninjaguardian: > there is no test command. either remove the cache stuff or create a test command so this is no dead code please. if you like to keep it around, clean it up. there is duplicate clear cache code in this file

Duplicates? Do you mean the functions in the Companion?

hannibal002: > False does not always mean not enabled though.

for all our purposes its the same. or when do you want to separate between a mod not being there, not being working, or feature disabled? we only want to know if the other feature is active, why it might not be, is not relevant imo

> Duplicates? Do you mean the functions in the Companion?

yes, but not only that. you currently have a "clear all cache" function, and you dont use it in the companion area, or vice versa. also all the "clear x cache" functions could be inlined into one "clear cache" function. its already rare that we need to clear a cache for test, but I dont see any reason at all to have the filler functions there, we never want to only clear parts of the cache.

src/main/java/at/hannibal2/skyhanni/features/misc/items/enchants/EnchantsJson.kt
hannibal002: Then there was prob a reaosn why it wasnt used before. either i have missed the reason in the beta branch, or your changes removed this reason. either way - using matchMatcher is quite easy. does showing at other usages in the mod (there are hundreds) help you or should others do this for you?

ninjaguardian: I looked at matchMatcher and, from what I can tell, it tries to match the entire line to a pattern, which seems to be different from this code.

EnchantParser.kt also uses a matcher, but I think that has the same issue when converting to matchMatcher.

src/main/java/at/hannibal2/skyhanni/features/misc/items/enchants/EnchantsJson.kt
ninjaguardian: I also did that EnchantParser.kt too

src/main/java/at/hannibal2/skyhanni/features/misc/PatcherFixes.kt
ninjaguardian: I think I fixed the cache stuff.

And I do agree with your statement, but future people using getBoolean may care about the result. Maybe instead, should I make a new function isAvailable or something that changes all null to false like the old getBoolean?

ninjaguardian:
I can do String.replace(regex, "") but I replaced regex with RepoPattern.group(...).pattern() (well, more or less)

I can't use IntelliJ to see how I can use it. I have a guess so I'll try it.

ninjaguardian:
> > I'm a bit confused by your phrasing. Is the warning already implemented?
> 
> 
> 
> no its not. but since you now have detection if both features are enabled simultaneous its useful to have the chat warning instead of only warn in the config desc or changelog

Is there a specific way to send warnings in chat with SkyHanni?

src/main/java/at/hannibal2/skyhanni/features/misc/PatcherFixes.kt
hannibal002: hmm, avaliable sounds also not totally right to me. how about enabled?

src/main/java/at/hannibal2/skyhanni/features/misc/items/enchants/EnchantsJson.kt
hannibal002: good catch! i think we had a "onlny partial matches" logic that is similar to machMatcher, but not relevant enough there. its fine

hannibal002:
> Is there a specific way to send warnings in chat with SkyHanni?

what warnign? a user error, or a mod error?

ninjaguardian:
> > Is there a specific way to send warnings in chat with SkyHanni?
> 
> 
> 
> what warnign? a user error, or a mod error?

A user error. Just like a note that says "Hey! You're using both SkyHanni and Aaron's mod to process enchantments! Please pick one or the other." Or something.

src/main/java/at/hannibal2/skyhanni/features/misc/PatcherFixes.kt
ninjaguardian: I added isEnabled and changed appropriate getBoolean uses to use isEnabled.

hannibal002:
The normal approach here is `ChatUtils.userError`. Or look at https://github.com/hannibal002/SkyHanni/blob/a327eaa79c760ef9d14ba24a8de64712a7d5836b/src/main/java/at/hannibal2/skyhanni/features/misc/PatcherFixes.kt#L28

versions/1.21.5/src/main/java/at/hannibal2/skyhanni/utils/compat/OrderedTextUtils.kt
Bloxigus, line 1: Please don't touch this file, #4210 changes this a lot and achieves better performance in the methods you change. I also dont see how this is in any way related to this PR.

hannibal002 (changes requested):
comments above

AzureAaron:
I don't think this solution is the most sound for this problem, from my understanding this stems from SkyHanni not being able to handle text components properly at which point it probably should with many of the related issues recently; especially given that other mods could also easily do similar things. The detection used for my rainbow enchant mode also will stop working when I introduce translation support to my mod soon.

> > > Is there a specific way to send warnings in chat with SkyHanni?
> > 
> > 
> > what warnign? a user error, or a mod error?
> 
> A user error. Just like a note that says "Hey! You're using both SkyHanni and Aaron's mod to process enchantments! Please pick one or the other." Or something.

I think this is a good message, it is fair and provides the user with two reasonable options to choose from.

ninjaguardian:
> The detection used for my rainbow enchant mode also will stop working

I'll close this pr and make a new one with just the warning message because this has too much clutter.

PR #4209, opened 2025-06-28
Improvement: Make hide fire particles and hide blaze particles work

CalMWolfs:
## What
- Made 2 particle hiding settings do something, didnt get around to hide fireball particles yet
- Removed 2 mixins that were unused on 1.8
- Renamed package that had wrong previous name

## Changelog Improvements
+ Made Hide Fire and Blaze Particles work on 1.21. - CalMWolfs


hannibal002 (changes requested):
the two new mixins will crash the game if called while the config is null. please use events and ideally a hook as well

CalMWolfs:
> the two new mixins will crash the game if called while the config is null. please use events and ideally a hook as well

These 2 mixins are exactly the same as the ones on 1.8 which have not had any problems in the 2+ years they have existed so I don't see a point in adding this extra stuff

hannibal002:
my "overly cautions/concern of easily fixable crashes on startup" have risen since those two years, especially when you remind me that Mojang did couple of code changes in the past 9 years

hannibal002 (approved):
Thanks!

PR #4210, opened 2025-06-28
Improvement: Rewrite the visual words system to use component replacing rather than string replacing

Bloxigus:
Rewrite of the entire Visual Words system to work with mods (and even hypixel maybe) that use custom fonts or custom shadows

Probably needs more testing than I have done yet, it appears to work for me perfectly

## Changelog Improvements
+ Updated Visual Words for supporting custom-font mods. - bloxigus


hannibal002 (changes requested):
> Rewrite of the entire Visual Words system to work

this belongs in the technical detail changelogs

works fine in inventory names, scoreboard, text rendered in world.
does not work with text in chat window

hannibal002 (changes requested):
shwords in chat does work, colors do work, just chroma doesnt

hannibal002:
image

hannibal002 (approved):
Chroma now also works in chat!

i cant get to change my own name in chat, but i think this is related to shmarkplayer

PR #4201, opened 2025-06-28
Backend: EliteBot Resource + Weight Endpoints

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/4055
- https://github.com/hannibal002/SkyHanni/pull/4214

## What
De-coupled the EliteBot Api from FarmingWeight Display.
Added Api endpoints and Serialization classes for a few EliteBot resource endpoints:
- Auctions
- Bazaar
- Items

<details>
<summary>Images</summary>

image
image
image
</details>

## Changelog Technical Details
+ Added resource endpoints to EliteDevApi. - Daveed
    * Items, Auctions, Bazaar.
+ Added farming weight-related endpoints to EliteDevApi. - Daveed
    * Removed these from FarmingWeightDisplay.

PR #4214, opened 2025-06-29
Backend + Fix: Make ApiUtils use IO Couroutine

DavidArthurCole:
## What
Makes all of the outbound calls in ApiUtils use IO coroutines, since we were 50% of the way there already from the using data classes. Pulled this out of #4201. There's a lot that changed, so.. see technical changelog.

## Changelog Fixes
+ Fixed rare lagging issue when checking clock time. - Daveed

+ Added more structure to ApiUtils. - Daveed
    * You can define static Api routes ahead of time as vals.
    * You can specify explicit typing for ApiResponse.
    * Functions will need to be `suspend` marked to use ApiUtils now.
+ Added KDocs to most functions in `ApiUtils`. - Daveed
+ Silenced most errors related to `ApiUtils`. - Daveed
    * If you need to debug, use the "ApiUtils never silent" debug option.
+ Added deprecation to `SkyHanniMod.coroutineScope`. - Daveed
    * Should be using `.launchCoroutine` or `launchIOCoroutine`.
+ Added debug command for fetching moulberry lbin prices. - Daveed
    * `/shfetchmoulblbins`.


CalMWolfs (approved):
code lgtm, havent tested

DavidArthurCole:
I also fixed a possible lag issue with the clock offset in this PR so if we could get it sooned 👉👈

DavidArthurCole:
fire emoji

DavidArthurCole:
Welp, conflicts now that we've temp-reverted the webhook stuff. I'm gonna keep this head alive, and start a second branch that's headed before I merged that in

DavidArthurCole:
Same here.

PR #4215, opened 2025-06-29
Backend: Rename Renderable Params

DavidArthurCole:
## What
Renames the renderable params from `posX`/`posY` -> `mousePosX`/`mousePosY`, and added KDocs to the function so it's clear that the posX/posY is not a position to render at.

+ Renamed renderable params from posX/posY to mousePosX/mousePosY. - Daveed


Thunderblade73:
Technically it is the offset in mouse coordinates that get's passed around not the mouse Position, as that is a global variable.

DavidArthurCole:
> Technically it is the offset in mouse coordinates that get's passed around not the mouse Position, as that is a global variable.

Yeah, this in combination with nopo's comment last night.. Think I'll probably try to come up with something better. `mouseOffsetPosX` feels like too much of a mouthful but...

DavidArthurCole:
> Technically it is the offset in mouse coordinates that get's passed around not the mouse Position, as that is a global variable.

Renamed all instances (again, and apparently I had missed quite a few the first time...) to `mouseOffsetX` and `mouseOffsetY`, as I think that'll be clear enough that it's not a Position per se.

Thunderblade73 (approved):

CalMWolfs (approved):

PR #4223, opened 2025-06-29
Fix: cs error outside galatea

j10a1n15:
## What
idk how doesnt happen to me :shrug:

## Changelog Fixes
+ Fixed Custom Scoreboard error when gaining whispers outside Galatea. - j10a1n15


hannibal002 (changes requested):
I dont see how the code change relates to the fix

DavidArthurCole:
> I dont see how the code change relates to the fix

```
open fun showIsland(): Boolean = true
```

removing the fun makes the lines in question always show

image

j10a1n15:
the patterns dont get loaded if not in that specific island
since you can gain whispers outside of galatea and apparently they show up in the scoreboard [see](https://discord.com/channels/997079228510117908/1388604707534274754), it should get loaded outside aswell

CalMWolfs (approved):

PR #4233, opened 2025-06-30
Backend: `onAnyToggled`

DavidArthurCole:
## What
This PR adds a method to ConditionalUtils that can be used to `onToggle()` every property in a config at once, and it will also iterate down to sub-configs as well. First implemented in Pet Display V2, but I've been finding use cases for it elsewhere.

+ Added `onAnyToggled` to ConditionalUtils. - Daveed
    * Allows you to `onToggle()` an entire config file at once.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/ConditionalUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/ConditionalUtils.kt
@@ -2,6 +2,10 @@ package at.hannibal2.skyhanni.utils
 
 import io.github.notenoughupdates.moulconfig.observer.Observer
 import io.github.notenoughupdates.moulconfig.observer.Property
+import java.lang.reflect.Modifier
+import kotlin.reflect.full.hasAnnotation
+import kotlin.reflect.full.memberProperties
+import kotlin.reflect.jvm.isAccessible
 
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/utils/ConditionalUtils.kt
```diff
@@ -72,4 +76,34 @@ object ConditionalUtils {
         return second1.compareTo(second2)
     }
 
+    /**
+     * Recursively scans each given 'root' for any fields of type Property<*>,
+     * then calls the existing onToggle(...) overload with all discovered properties.
+     */
+    fun onToggleAll(vararg roots: Any, observer: Runnable) = roots.forEach { root ->
```
Thunderblade73: wouldn't it be better called ``onAnyToggled`` ?

DavidArthurCole: Probably, yes

CalMWolfs:
is there an easy way for me to test this?

DavidArthurCole:
> is there an easy way for me to test this?

probably just a Test config with some nested Property<Boolean> options in it, and then do an onToggleAny with a ChatUtils debug

PR #4062, opened 2025-06-15
Feature: Dark Monolith Helpers & Tracker

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/4234

## What
Adds highlighting for Dark Monoliths in Dwarven Mines and a profit tracker.

## Changelog New Features
+ Added tracker, titles, and block highlighting for Dark Monoliths. - Daveed

PR #4234, opened 2025-06-30
Backend: `canSee` Block Faces

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/4233
- https://github.com/hannibal002/SkyHanni/pull/4244

## What
https://discord.com/channels/997079228510117908/1094190239532208228/1389106741815414805
Came up while I was working on #4062 - we had a lot of functions to check if you have line of sight with an entity/vec, but that would almost always fail with a direct use of the LorenzVec of a block, due to the block itself being in the way of the ray.

This PR adds a bunch of backend code for checking against this (`canSeeFace` & `canSeeAnyFace`, as well as a test-suite using `/shtestcanseeface`.

Needed a couple of new World Rendering utils, but nothing that's new should break on 1.21.6 🙏

<details>
<summary>Images</summary>

image

image

image

Example of what a partial LOS would look like:

image

image
</details>

+ Added `LocationUtils` functions for checking LOS to block faces. - Daveed
    * Can specify one specific face (by `EnumFacing`/`Direction`), or any face.
    * Enable in `/sh dev tools`, and `/shtestcanseeface`.
+ Added functions to draw directional rays/highlight block faces in `WorldRenderUtils`. - Daveed


src/main/java/at/hannibal2/skyhanni/test/TestCanSeeFace.kt
```diff
@@ -0,0 +1,291 @@
...
+    fun onCommandRegistration(event: CommandRegistrationEvent) {
+        event.registerBrigadier("shtestcanseeface") {
+            description = "Test if you can see certain faces of a block."
+            category = CommandCategory.DEVELOPER_TEST
+            simpleCallback {
+                if (!enabled) {
+                    ChatUtils.chat(
+                        "The /shtestcanseeface command is disabled. Enable it in the dev tool config.",
+                        replaceSameMessage = true
+                    )
```
CalMWolfs: maybe could link to the option when you click this message

src/main/java/at/hannibal2/skyhanni/config/features/dev/TestCanSeeFaceConfig.kt
```diff
@@ -0,0 +1,119 @@
...
+
+    @Expose
+    @ConfigOption(name = "Step Count", desc = "§c(§4§l!§r§c) Large values may cause issues.")
+    @ConfigEditorSlider(minValue = 0f, maxValue = 60f, minStep = 1f)
+    val stepCount: Property<Int> = Property.of(0)
+
+    @Expose
+    @ConfigOption(name = "Step Density", desc = "§c(§4§l!§r§c) Large values may cause issues.")
+    @ConfigEditorSlider(minValue = 4f, maxValue = 60f, minStep = 1f)
+    val stepDensity: Property<Int> = Property.of(4)
```
CalMWolfs: I know you add a warning here but these probably shouldnt go nearly this high, make the limits lower and then if we ever are testing you can manually enter a larger number anyways

CalMWolfs (changes requested):
Definitely seems like a cool concept but the test isn't quite working as id expect it to yet. There are a few rendering issues + it seems to not always be super accurate as can be seen in the video

When testing I get some z fighting, maybe offset it by a miniscule amount in the direction of the player. It also seems it takes a little bit to update the state of whether a face can be seen.

https://github.com/user-attachments/assets/dd1c3e58-b38b-4358-bf03-3efe3c1e4489

(video from 1.8, with refresh on 1 second, it takes more like 3 seconds to update)

On 1.21 there is less z fighting but still a little. Also the lines on 1.21 seem to not have depth which makes it easier to see if the non visible faces are being marked correctly, this behaviour should be added to 1.8

DavidArthurCole:
> When testing I get some z fighting, maybe offset it by a miniscule amount in the direction of the player. It also seems it takes a little bit to update the state of whether a face can be seen.
> (video from 1.8, with refresh on 1 second, it takes more like 3 seconds to update)

Fixed - old set points weren't being cleared at a good rate.

> On 1.21 there is less z fighting but still a little. Also the lines on 1.21 seem to not have depth which makes it easier to see if the non visible faces are being marked correctly, this behaviour should be added to 1.8

This is a limitation of the 1.8 implementation of `drawFilledBoundingBox`. The 1.21. impl has a `seeThroughBlocks` param;
image

The 1.8 one, does not;
image

DavidArthurCole:
@CalMWolfs bump

DavidArthurCole:
merge conflicts

src/main/java/at/hannibal2/skyhanni/test/TestCanSeeFace.kt
```diff
@@ -0,0 +1,302 @@
...
+import at.hannibal2.skyhanni.utils.renderables.primitives.text
+import net.minecraft.client.player.LocalPlayer
+import net.minecraft.core.Direction
+import net.minecraft.world.phys.AABB
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
+typealias PointSet = TimeLimitedSet<Pair<LorenzVec, Boolean>>
+typealias FacePointEntry = Map.Entry<Direction, PointSet>
+typealias FacePointSet = MutableMap<Direction, PointSet>
```
hannibal002: this should not be defined here. LocationUtils should not depend on a test file.

src/main/java/at/hannibal2/skyhanni/test/TestCanSeeFace.kt
```diff
@@ -0,0 +1,302 @@
...
+    private fun SkyHanniRenderWorldEvent.tryHighlightFace(
+        context: FaceCheckContext,
+        face: Direction,
+    ) {
+        if (!faceHighlightConfig.enabled.get() || faceStates[face] == FaceState.HIDDEN) return
+        val vec1 = context.vec1 ?: return
+        val vec2 = context.vec2 ?: return
+        val points = context.pointSet[face] ?: return
+        val faceSeen = points.any { it.second }
+        val color = if (faceSeen) faceHighlightConfig.seenColor.get() else rayConfig.unSeenColor.get()
```
hannibal002: looks like a bug here. unSeenColor should use faceHighlight, otherwise the config entry is unused.

src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt
```diff
@@ -52,6 +59,165 @@ object LocationUtils {
...
+        val aabb = AABB(min.x, min.y, min.z, max.x, max.y, max.z)
+        val eye = playerEyeLocation()
+        val center = aabb.getBoxCenter()
+        val faceCenter = face.getCenterPos(center, aabb)
+
+        if (eye.distance(faceCenter) > viewDistance.toDouble()) return false
+        val wrappedSuccess = pointFill.wrapCanSee(face, eye, faceCenter, offset)
+
+        return if (wrappedSuccess && pointFill == null) true
+        else if (stepCount == 0) false
```
hannibal002: why cant we see the face when step count is 0?

src/main/java/at/hannibal2/skyhanni/test/TestCanSeeFace.kt
```diff
@@ -0,0 +1,302 @@
...
+            max = vec2,
+            stepCount = config.stepCount.get(),
+            stepDensity = config.stepDensity.get(),
+            pointFill = faceCheckContext.pointSet,
+        )
+        val calculationEndTime = SimpleTimeMark.now()
+        regenDebugRenderable(calculationEndTime - calculationStartTime)
+        faceCheckContext.finished = true
+        DelayedRun.runDelayed(config.refreshInterval.get().seconds) {
+            recalcContext(true)
```
hannibal002, line 262: this does not check enabled again.

src/main/java/at/hannibal2/skyhanni/utils/render/WorldRenderUtils.kt
```diff
@@ -27,11 +27,13 @@ import net.minecraft.client.gui.Font
 import net.minecraft.client.renderer.LightTexture
 import net.minecraft.client.renderer.ShapeRenderer
 import net.minecraft.client.renderer.blockentity.BeaconRenderer
+import net.minecraft.core.Direction
 import net.minecraft.network.chat.Component
 import net.minecraft.world.entity.Entity
 import net.minecraft.world.phys.AABB
 import org.joml.Matrix4f
 import java.awt.Color
+import kotlin.Boolean
```
hannibal002: unused import

src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt
```diff
@@ -1,9 +1,15 @@
 package at.hannibal2.skyhanni.utils
 
+import at.hannibal2.skyhanni.test.FacePointSet
+import at.hannibal2.skyhanni.utils.collection.TimeLimitedSet
 import at.hannibal2.skyhanni.utils.compat.MinecraftCompat
+import net.minecraft.core.Direction
 import net.minecraft.world.entity.Entity
 import net.minecraft.world.phys.AABB
+import kotlin.collections.getOrPut
```
hannibal002: getOrPut and getValue are both unused imports

src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt
```diff
@@ -52,6 +59,165 @@ object LocationUtils {
...
+                val wrappedSuccess = pointFill.wrapCanSee(this, eye, testPoint, offset)
+                if (wrappedSuccess && pointFill == null) return true
+            }
+        }
+        return false
+    }
+
+    private fun FacePointSet?.wrapCanSee(face: Direction, a: LorenzVec, b: LorenzVec, offset: Double?): Boolean {
+        val canSeeResult = canSee(a, b, offset)
+        this?.getOrPut(face) { TimeLimitedSet(5.seconds) }?.add(b to canSeeResult)
```
hannibal002: we should make the 5 seconds dynamic, via parameter. also, do we want to set this to refreshInterval when coming from TestCanSeeFace.canSeeAnyFace? we would need to pass the parameter through the functions tho

src/main/java/at/hannibal2/skyhanni/utils/render/WorldRenderUtils.kt
```diff
@@ -812,6 +814,88 @@ object WorldRenderUtils {
...
+            )
+        }
+
+        val corners = effectiveAABB.getFaceCorners(face)
+        val effectiveAlpha = ((color.alpha / 255f) * alpha * 255).toInt().coerceIn(0, 255)
+        val effectiveColor = Color(color.red, color.green, color.blue, effectiveAlpha)
+        draw(corners[0], corners[1], corners[3], effectiveColor)
+    }
+
+    @Suppress("unused")
```
hannibal002: do we plan on adding see through blocks support? if so, please also add a todo, if not, please remove the parameter + suppress

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt
DavidArthurCole: I'll add a comment in the code, but:

```
        val wrappedSuccess = pointFill.wrapCanSee(face, eye, faceCenter, offset)

        return if (wrappedSuccess && pointFill == null) true
        else if (stepCount == 0) false
```

`wrapCanSee`, when stepCount is 0, will leave pointFill as `null`, if the block can be seen. So, if stepCount is 0, and we make it past that `if`, we can't see the block.

DavidArthurCole: Or, more poignantly, the only other branch after this is to "keep stepping" - when step count is 0, we don't do any stepping.

src/main/java/at/hannibal2/skyhanni/test/TestCanSeeFace.kt
DavidArthurCole: Good catch - yes, was a bug/typo.

src/main/java/at/hannibal2/skyhanni/utils/render/WorldRenderUtils.kt
DavidArthurCole: I think this was a hold-over from the `1.8.9` implementation, where that param existed. This branch is showing its age... removed.

src/main/java/at/hannibal2/skyhanni/config/features/dev/TestCanSeeFaceConfig.kt
```diff
@@ -0,0 +1,119 @@
...
+
+        @Expose
+        @ConfigOption(name = "Seen Color", desc = "Color of rays that are seen by the face.")
+        @ConfigEditorColour
+        val seenColor: Property<ChromaColour> = Property.of(LorenzColor.GREEN.toChromaColor(220))
+
+        @Expose
+        @ConfigOption(name = "Not Seen Color", desc = "Color of rays that are not seen by the face.")
+        @ConfigEditorColour
+        val unSeenColor: Property<ChromaColour> = Property.of(LorenzColor.RED.toChromaColor(220))
```
hannibal002: please rename to unseenColor. "un" is just a prefix, no standalone word

src/main/java/at/hannibal2/skyhanni/test/TestCanSeeFace.kt
```diff
@@ -0,0 +1,299 @@
...
+
+@SkyHanniModule
+object TestCanSeeFace {
+
+    data class FaceCheckContext(
+        var vec1: LorenzVec? = null,
+        var vec2: LorenzVec? = null,
+        var waitingForPunch: Boolean = false,
+        var pointSet: FacePointSet = mutableMapOf(),
+        var generallySeen: Boolean = false,
```
hannibal002: we dont use this. you  plan on using this later or can we remove it? if use later, maybe add a /todo here

src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt
```diff
@@ -52,6 +60,179 @@ object LocationUtils {
...
+        stepDensity: Int,
+        pointFill: FacePointSet?,
+        offset: Double? = null,
+        resultLifespan: Duration = 5.seconds,
+    ): Boolean {
+        val halfX = (aabb.maxX - aabb.minX) / 2
+        val halfY = (aabb.maxY - aabb.minY) / 2
+        val halfZ = (aabb.maxZ - aabb.minZ) / 2
+        val (axis1, axis2, ext1, ext2) = getFaceRayConfig(halfX, halfY, halfZ) ?: return false
+        densityLoop@for (densityIncrement in 0 until stepDensity) {
```
hannibal002: loop label is not used. use or remove please

src/main/java/at/hannibal2/skyhanni/test/TestCanSeeFace.kt
```diff
@@ -0,0 +1,299 @@
...
+                    )
+                    return@simpleCallback
+                }
+                faceCheckContext.reset()
+                faceCheckContext.waitingForPunch = true
+                ChatUtils.chat("The next block you punch will be used for the face check.", replaceSameMessage = true)
+            }
+            literalCallback("stop") {
+                faceCheckContext.reset()
+                lastRenderable = null
```
hannibal002, line 204: no user feedback in chat

hannibal002 (changes requested):

hannibal002 (approved):

PR #4241, opened 2025-07-01
Fix: Pressing edit GUI keybind while searching in REI

mariethepolarfox:
## What
Fixes opening the GUI Editor when pressing the edit keybind while searching the REI searchbar.

## Changelog Fixes
+ Fixed Edit GUI keybind in REI search. - Helium9


NopoTheGamer:
why dont we just prevent the key event from being posted if its focused?

mariethepolarfox:
> why dont we just prevent the key event from being posted if its focused?

Thats a really good Idea I havent thought about that. Will change ^^

NopoTheGamer:
you should also split this into some sort of compat method like how we have the same thing for neu

PR #4243, opened 2025-07-01
Feature: Coral Fish Helper

CalMWolfs:
What
A basic helper for the fish collection at the coral npc on galatea.

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added Coral Fish Helper. - CalMWolfs
    * Shows missing fishes for the NPC, sorted by price.


DavidArthurCole (approved):
image

Is there a reason we're doing yellow `$` instead of gold coins formatting here? Otherwise LGTM.

CalMWolfs:
> Is there a reason we're doing yellow `$` instead of gold coins formatting here? Otherwise LGTM.

I knew something looked weird but I couldn't work it out, should be fixed now 😭

DavidArthurCole:
> I knew something looked weird but I couldn't work it out, should be fixed now 😭

Probably pedantic, and might be not worth changing, but in general I think the `$` does make it look weird (or at least stand out from others), as we generally for profit trackers either use the word "coin[s]" or rely on the gold color to imply it - but I'm good either way.

NopoTheGamer:
> > I knew something looked weird but I couldn't work it out, should be fixed now 😭
> 
> Probably pedantic, and might be not worth changing, but in general I think the `$` does make it look weird (or at least stand out from others), as we generally for profit trackers either use the word "coin[s]" or rely on the gold color to imply it - but I'm good either way.

we shouldnt use the dollar symbol as they arent dollars

PR #4245, opened 2025-07-01
Fix: Hotx

DavidArthurCole:
## What
i hate the jvm. pretty sure there's no way to make an enum class have a delegated repo pattern because enum classes can't extend anything except for interfaces. so fuck it we revert.

## Changelog Fixes
+ Fixed HotM/HotF errors. - Daveed


hannibal002 (changes requested):
you mean hotf and not hotx, right?
image

PR #4246, opened 2025-07-01
Backend: InventoryDetector `inventoryClose`

DavidArthurCole:
## What
Same thing as the `inventoryOpen` param, but for close. Seems like a lot of inventory management classes could use it.

+ Added `inventoryClose` param to InventoryDetector. - Daveed


CalMWolfs (approved):

DavidArthurCole:
Given the change to `openInventory` to make it the event instead of the inv name, I've done the same with the new `closeInventory` here.

PR #4083, opened 2025-06-15
Improvement: Pet Display V2

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/4068
- https://github.com/hannibal002/SkyHanni/pull/4082
- https://github.com/hannibal002/SkyHanni/pull/4114
- https://github.com/hannibal002/SkyHanni/pull/4101
- https://github.com/NopoTheGamer/SkyHanni/pull/13
- https://github.com/hannibal002/SkyHanni/pull/4116
- https://github.com/hannibal002/SkyHanni/pull/4233
- https://github.com/hannibal002/SkyHanni/pull/4249
- https://github.com/hannibal002/SkyHanni/pull/4261
- https://github.com/hannibal002/SkyHanni/pull/4428
- https://github.com/hannibal002/SkyHanni/pull/4454

## What
Completely overhauls the Current Pet Display, and makes it prettier and more customizable, combining the best parts of the other mods' pet displays into one solution.

[Diff between this PR and the PetApi branch](https://github.com/DavidArthurCole/SkyHanni/compare/backend/pet-api-v3...DavidArthurCole:SkyHanni:feat/pet-display-v2)

## Changelog Improvements
+ Completely overhauled current pet display. - Daveed.
    * Pet icon display ("SBE-style"), supports animated skins & spinning.
    * Pet text display ("NEU-style"), can be placed abstractly next to/above/below the icon.
    * Each aspect of the pet display is toggleable and very configurable.

PR #4249, opened 2025-07-02
Fix; AnimatedItemStackRenderable Out of Frame

DavidArthurCole:
## What
Fixed issues with the animated item stack renderable that were causing bounces to go out of the top of the frame. Also pulled the animation stuff out of #4083 while I was here.

exclude_from_changelog


DavidArthurCole:
I think I will be splitting up pet display v2 into 10 +100/-20 PRs until hanni doesn't even realize that's what he's merging anymore 🙏

PR #4253, opened 2025-07-02
Backend: Performance update for bazaar order helper

hannibal002:
## What
instead of checking all items in the bz order menu every frame against multiple mixins, we now only do so once on inventory open.
Also using inventory detector to cache the inventory name.
Also changed the backend of inventory detector to pass the inventory open event over to the parameter.

+ Added caching to BZ Order Helper. - hannibal2


CalMWolfs (approved):
This should also fix a problem I had in something else

hannibal002:
> This should also fix a problem I had in something else

do you remember what problem exactly? can add to the changelog then

CalMWolfs:
> do you remember what problem exactly? can add to the changelog then

its a thing in an open pr so no need

PR #4273, opened 2025-07-04
Fix: Data Watcher and base max health

CalMWolfs:
## What
Data Watcher Api now works and no longer kicks you from Hypixel 🔥 .
This fixes some but not all issues with mobs on 1.21.

## Changelog Fixes
+ Fixed some damage indicator issues on 1.21. - CalMWolfs
+ Fixed mob highlight on 1.21. - CalMWolfs


hannibal002 (changes requested):
even with this pr i dont see a damage indicator here
image

CalMWolfs:
> even with this pr i dont see a damage indicator here

It seems that it hasnt fixed all issues relating to damage indicator on modern versions but it has fixed some e.g. revenant horror tier 2-5. I'd say its still best to merge this as is with the fixes it does have as those are better than nothing but in the coming days/weeks i can fix more of damage indicator. Ive updated the changelog accordingly

PR #4278, opened 2025-07-06
Fix: Blaze Daggers display on 1.21

yhtez:
## What
Added SubtitleS2CPacket to TitleReceivedEvent as titles and subtitles are split into 2 different packets on 1.21 and fixed checking for attunement in title.

## Changelog Fixes
+ Fixed Blaze Daggers display on 1.21. - yhtez

+ Added SubtitleS2CPacket to TitleReceivedEvent on 1.21. - yhtez


CalMWolfs (approved):

hannibal002 (approved):
The fix worked, thanks!


btw the behaviour is still not 100% the same from 1.8. the display should update on right click alone and then only sync up with the title, so that the title doesnt have a delay like the hypixel title has

yhtez:
> btw the behaviour is still not 100% the same from 1.8. the display should update on right click alone and then only sync up with the title, so that the title doesnt have a delay like the hypixel title has

Ah, I see what I missed. I'll look at making another PR to address this soon.

PR #4281, opened 2025-07-06
Feature: Mute Hoppity Egg "Rolling" Sounds

DavidArthurCole:
## What
At some point Hypixel made the egg rolling sounds super fucking goofy, and I don't really like them. This PR adds an option to mute said sounds.

## Changelog New Features
+ Added option to mute sounds while Hoppity Eggs are "rolling". - Daveed
    * Sub-option available to revert to "1.8 sounds" (eating sound).
    * Will only apply to the "new" sounds that happen in 1.20 and above.


CalMWolfs (changes requested):
seeing as this pr has the word "roll"/"rolling" in it I feel that would be an appropriate search tag to have on the option. Also i'd like it to maybe track time since in hoppity inventory and if its less than a second then still cancel the sound as rn if i exit the inv immediately i still get 1-2 notes playing

DavidArthurCole:
> seeing as this pr has the word "roll"/"rolling" in it I feel that would be an appropriate search tag to have on the option. Also i'd like it to maybe track time since in hoppity inventory and if its less than a second then still cancel the sound as rn if i exit the inv immediately i still get 1-2 notes playing

both done as well

CalMWolfs (approved):

PR #4286, opened 2025-07-07
Fix: Unneeded memory use when a OrderedText object was kept in use

Bloxigus:
~~This is such a minor change that I don't think it deserves a changelog entry~~

im gonna cry if you ask me to reformat this again

## Changelog Fixes
+ Fixed memory leak in Visual Words feature. - bloxigus


hannibal002:
consistency win
image

PR #4295, opened 2025-07-08
Fix: Jacob farming contest inventory

CalMWolfs:
## What
Hypixel seems to have made the item we were checking be a feather for a tick or two while data loaded resulting in the mod not knowing we were in the contest inventory. Now we use inventory updated event instead and skip the regex checks if we are already in the inventory.

## Changelog Fixes
+ Fixed Medal Icons not showing in Jacob’s Contest inventory. - CalMWolfs


hannibal002:
code looks good, needs testing

PR #4302, opened 2025-07-08
Backend: Fix intellij warnings

hannibal002:
## What
IntelliJ likes to predict what values some members have. For the MC version string and the legacy boolean, this is confusing. IntelliJ thinks while looking at other classes that those members are constants, and will never change.

+ Disabled IntelliJ’s MC Version/Legacy Prediction to avoid incorrect warnings and suggestions. - hannibal2


NopoTheGamer:
1. add new lines in the ksp thingy 2,. this doesnt compile

CalMWolfs (approved):

PR #4303, opened 2025-07-08
Fix: Custom Scoreboard in the Mines

j10a1n15:
## What
not enough data (and too lazy to get it myself) to create a better regex, this one should do though

## Changelog Fixes
+ Fixed Custom Scoreboard error in the Mines. - j10a1n15


src/main/java/at/hannibal2/skyhanni/features/gui/customscoreboard/ScoreboardPattern.kt
```diff
@@ -397,12 +397,11 @@ object ScoreboardPattern {
     )
 
     /**
-     * REGEX-TEST: Nearby Players: §a0
-     * REGEX-TEST: Nearby Players: §cN/A
+     * REGEX-TEST: Nearby Players: §a5 §cMAX
```
lunaynx, line 402: Why remove the previous regex tests instead of just adding a new one?

j10a1n15: well i assume the previous ones just dont exist anymore and got replaced by some new style

hannibal002: i agree. especially when looking at the new pattern, those old tests are no longer necessary

hannibal002:
code looks fine, testing necessary

j10a1n15:
why dont we just merge it and let the users test it and for the (very unlikely) off chance it doesnt work, we can just create pr? It alr errors ingame so nothing would change if it actually failed

hannibal002:
i try for 10 minutes to compile and test a local branch with this and other 7 prs merged. somehow it doesnt compile  but mc starts instantly (i try to test it, cant. see dc for more infos)

PR #4306, opened 2025-07-08
Backend: NoConfigLink Annotation

DavidArthurCole:
## What
Adds an annotation for marking a `Position` as not having a config link, rather than relying on putting its absolute path in a static list - replaced `ignoredMissingConfigLinks` with this approach.

+ Added `NoConfigLink` annotation. - Daveed


src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -106,22 +106,10 @@ class ConfigManager {
...
-        "features.dev.DebugConfig.trackSoundPosition",
-        "features.dev.DebugConfig.trackParticlePosition",
-        "features.dev.DevConfig.debugPos",
-        "features.dev.DevConfig.debugLocationPos",
-        "features.dev.DevConfig.debugItemPos",
-    )
+    // Use this annotation to mark config elements that don't need a config link.
+    @Retention(AnnotationRetention.RUNTIME)
+    @Target(AnnotationTarget.FIELD)
+    annotation class NoConfigLink
```
CalMWolfs: I feel this could probably be taken outside of the ConfigManager class so that then we annotate with just `@NoConfigLink` instead of `@ConfigManager.NoConfigLink`

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
DavidArthurCole: Done

PR #4266, opened 2025-07-03
Backend: Automatically publish releases to Modrinth

CalMWolfs:
partially reverted with https://github.com/hannibal002/SkyHanni/commit/011c88b869e166f4a894210d585a56e75978613a

correctly fixed via #4310

## Dependencies
- #4229

Hannibal you must create a repo secret that is a modrinth token

Required Permission
https://modrinth.com/settings/pats
image

You then must add that to the repo secrets, with the name `MODRINTH_TOKEN`
https://github.com/hannibal002/SkyHanni/settings/secrets/actions/new

## What
Added automatic uploading to modrinth.
Make sure to follow the instructions!!!!!

<details>
<summary>Images</summary>

image

</details>

## Changelog Technical Details
+ Added automatic uploading to Modrinth. - CalMWolfs

PR #4310, opened 2025-07-09
Fix: Modrinth action breaking intellij

CalMWolfs:
## What
I still have no idea how the old code cooked intellij but it did. So after way too much testing stuff I've come up with a solution that no longer breaks intellij on my laptop and therefore should also not break intellij on other people's devices.
Also did a bit of code cleanup in the action.

Reminder for hannibal,I'm not sure if you ended up adding the modrinth token to the repo so make sure you do that, #4266 has the guide. If this doesnt work I will cry, but obviously dont merge if it breaks your intellij

exclude_from_changelog

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -61,7 +61,4 @@ jobs:
                 run: ./gradlew :sharedVariables:build
 
             -   name: Publish to Modrinth
-                run: ./gradlew publishToModrinth \
-                    -Pchangelog="${{ steps.get_changelog.outputs.changelog }}" \
-                    -PmodVersion="${UPDATE_VERSION}" \
-                    -PmodrinthToken="${MODRINTH_TOKEN}"
+                run: ./gradlew publishToModrinth -Pchangelog="${{ steps.get_changelog.outputs.changelog }}" -PmodVersion="${UPDATE_VERSION}" -PmodrinthToken="${MODRINTH_TOKEN}"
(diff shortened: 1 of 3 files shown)


.github/workflows/post-release.yml
```diff
@@ -61,7 +61,4 @@ jobs:
                 run: ./gradlew :sharedVariables:build
 
             -   name: Publish to Modrinth
-                run: ./gradlew publishToModrinth \
-                    -Pchangelog="${{ steps.get_changelog.outputs.changelog }}" \
-                    -PmodVersion="${UPDATE_VERSION}" \
-                    -PmodrinthToken="${MODRINTH_TOKEN}"
+                run: ./gradlew publishToModrinth -Pchangelog="${{ steps.get_changelog.outputs.changelog }}" -PmodVersion="${UPDATE_VERSION}" -PmodrinthToken="${MODRINTH_TOKEN}"
```
DavidArthurCole, line 64: Not sure what the deps are on this one (on mobile), but might be worth putting this behind the retry-gradle wrapper

.github/workflows/post-release.yml
```diff
@@ -61,7 +61,4 @@ jobs:
                 run: ./gradlew :sharedVariables:build
```
DavidArthurCole, line 61: and this one too

CalMWolfs: Yeah you're probably right but especially because this is a subproject id rather test this with the wrapper a bit later when I have some more time so I wont add this rn. Also this is just run manually so if it fails it can just be quickly re-ran at least for the first few tests but either remind me in a few days or make a pr to add the retry to this in a little bit

PR #4311, opened 2025-07-09
Improvement + Fix: Add config option for self-trigger party commands and fix self-trigger interaction with trust level

yhtez:
## What
Add a config option for whether party commands should be triggered by yourself and fix self-triggered party commands not working if a trust level was set.

## Changelog Improvements
+ Added config option for self-triggered party commands. - yhtez

## Changelog Fixes
+ Fixed self-triggered party commands incorrectly checking friendship status when trust level was set. - yhtez

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/misc/PartyCommandsConfig.java
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/misc/PartyCommandsConfig.java
@@ -18,7 +18,7 @@ public class PartyCommandsConfig {
     @Expose
     @ConfigEditorBoolean
     @ConfigOption(name = "Self-Trigger Commands", desc = "Allow party chat commands to trigger on your own messages.")
-    public boolean selfTriggerCommands = true;
+    public boolean selfTriggerCommands = false;
 
     @Expose
     @ConfigEditorBoolean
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/config/features/misc/PartyCommandsConfig.java
```diff
@@ -14,6 +14,12 @@ public class PartyCommandsConfig {
     @ConfigOption(name = "Trust Level", desc = "Choose who can run party chat commands.")
     public @NotNull TrustedUser defaultRequiredTrustLevel = TrustedUser.FRIENDS;
 
+    // TODO should this be customizable per-command
+    @Expose
+    @ConfigEditorBoolean
+    @ConfigOption(name = "Self-Trigger Commands", desc = "Allow party chat commands to trigger on your own messages.")
+    public boolean selfTriggerCommands = true;
```
DavidArthurCole: This should be defaulted to false IMO, otherwise LGTM.

DavidArthurCole (changes requested):

src/main/java/at/hannibal2/skyhanni/config/features/misc/PartyCommandsConfig.java
yhtez: I only defaulted to true to maintain current behavior but I have no preference either way.

PR #4316, opened 2025-07-10
Improvement: Limit the amount of visual words stored in the visual word caches

Bloxigus:
## What
Uses a size limited cache to only store a maximum of 200k visual words in memory (probably can be lowered)
https://spark.lucko.me/Tn1H9fqSLV <- shows 4.3m cache entries (a lot)

Also prevents visual words from visual wordsing if there are no visual words to visualually word.

## Changelog Improvements
+ Reduced memory usage with a large amount of text rendering. - bloxigus


hannibal002:
> This PR is not mergeworthy yet

moving to 5.0.0

CalMWolfs (approved):

PR #4320, opened 2025-07-10
Fix: Enigma Soul naming conflict

lunaynx:
## What
Refactored Enigma Soul tracker to account for the currently selected area in the menu to avoid conflicts when two Enigma Souls have the same name. Also done some miscellaneous cleanup such as moving things to repo patterns.

## Changelog Fixes
+ Fixed Enigma Soul tracker sometimes sending you to the wrong soul. - Luna


CalMWolfs (approved):
worked for me

PR #4336, opened 2025-07-12
Fix: Make metal detector credits visible to the user

SuperCl4sh:
Changes the metal detector features from an Accordion to a Category so the credits for the features are visible.

exclude_from_changelog


CalMWolfs (changes requested):
As nopo said on discord, I think making a @ConfigEditorInfoText would be better than making this into a full category

CalMWolfs (approved):

PR #4346, opened 2025-07-12
Fix + Improvement: Old SB Menu

DavidArthurCole:
## What
I think the ItemReplaceEvent was being called before the inventory was fully open, so this was a better solution.
Also added more indication when we block a click due to not having an active cookie.

## Changelog Improvements
+ Added warning title when trying to use Old SB Menu buttons without cookie. - Daveed

## Changelog Fixes
+ Fixed Old SB Menu buttons not handling clicks correctly. - Daveed


CalMWolfs (approved):
fixed it for me

PR #4351, opened 2025-07-13
Backend: Added TextRenderable

NopoTheGamer:
## What
Adds a Renderable that uses modern minecrafts Text
Also added a fake Text wrapper that can be used on 1.8
This is how we can render any hex code in text on 1.21

+ Added TextRenderables. - nopo


src/main/java/at/hannibal2/skyhanni/test/renderable/TestText.kt
Thunderblade73, line 1: You don't need the to lazy here. It is only there in the other test you looked at because it was a dynamic element and had real internal state (which Renderables shouldn't have, but that's on me to fix). So just add your code under the overriden function.

Thunderblade73 (changes requested):

CalMWolfs (approved):

PR #4356, opened 2025-07-13
Fix: Minion features and maybe some other stuff as well

CalMWolfs:
## What
We changed how this behaved on 1.8 but never updated 1.21. Might also fix other issues but idk what else uses EntityClickEvent that is broken

<details>
<summary>Images</summary>

image

</details>

## Changelog Fixes
+ Fixed most minion related features on 1.21. - CalMWolfs

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/ItemClickData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/ItemClickData.kt
@@ -77,9 +77,6 @@ object ItemClickData {
                 //#if MC < 1.21
                 val clickedEntity = packet.getEntityFromWorld(MinecraftCompat.localWorld) ?: return
                 //#else
-                //$$ if (packet.type is PlayerInteractEntityC2SPacket.InteractAtHandler) {
-                //$$     return
-                //$$ }
                 //$$ if (packet.type is PlayerInteractEntityC2SPacket.InteractHandler) {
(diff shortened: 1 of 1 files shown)


Thunderblade73 (commented):
I don't understand that pr but nopo added that line so that we do not fire the event 4 times per click. Since in modern when you interact it sends a ``Interact + Right Hand``, ``Interact + Left Hand``, ``Interact At + Right Hand`` and ``Interact At + Left Hand``.

CalMWolfs:
> I don't understand that pr but nopo added that line so that we do not fire the event 4 times per click. Since in modern when you interact it sends a `Interact + Right Hand`, `Interact + Left Hand`, `Interact At + Right Hand` and `Interact At + Left Hand`.

From testing it seems to only be 2 times per click which is what it is on 1.8 as well at least for the entity click events. Except when clicking on minions they only do interact_at and not normal interact

Thunderblade73:
> > I don't understand that pr but nopo added that line so that we do not fire the event 4 times per click. Since in modern when you interact it sends a `Interact + Right Hand`, `Interact + Left Hand`, `Interact At + Right Hand` and `Interact At + Left Hand`.
> 
> From testing it seems to only be 2 times per click which is what it is on 1.8 as well at least for the entity click events. Except when clicking on minions they only do interact_at and not normal interact

The other line below ommits the left hand, so it is 2 times.
But when it always fired twice I guess it is fine then.

PR #4359, opened 2025-07-13
Backend: Handmade string wrapping

Thunderblade73:
## What
Adds our own string wrapping code, which works for 1.8.9, but can be ported to 1.21 if it needs to. This is needed as the native string wrapping code does not like bold text and will produce strings larger than the set limit.

Also updates the Wrapped String Renderable Test to be nicer and cover more cases.

Also added some StringBuilder utility and one Collection Utils to transpose a list of list.

There is some wierd edge case https://discord.com/channels/997079228510117908/1094190239532208228/1394023847178666086 that does not work but not a direct problem of the wrapper as the output looks logicaly sound but somehow the rendering code messes up.

+ Added custom String Wrapper for 1.8.9. - Thunderblade73


CalMWolfs (changes requested):
havent looked through the code at all, just looked at the pr description
>Adds our own string wrapping code, which works for 1.8.9, but can be ported to 1.21 if it needs to

I don't like the idea of making something that only works on 1.8 and not on 1.21

Thunderblade73:
> havent looked through the code at all, just looked at the pr description
> 
> > Adds our own string wrapping code, which works for 1.8.9, but can be ported to 1.21 if it needs to
> 
> I don't like the idea of making something that only works on 1.8 and not on 1.21

I do not even know if the 1.21 native variant is even broke. So I did not make a port for it.

Thunderblade73:
Since this is declared as "not needed". I just close it.

PR #4375, opened 2025-07-15
Fix + Backend: Edge case with shiny orb tracker and EntityClickEvent remove null

CalMWolfs:
## What
Fixed the tracker showing up if you accidentally clicked a pig and there was an orb nearby. This also is prob a tiny performance boost as well.
Made entity in EntityClickEvent no longer nullable because how do you click on an entity but not click on an entity at the same time.

## Changelog Fixes
+ Fixed an edge case where shiny orb tracker would show up when it isn't meant to. - CalMWolfs

+ Made entity in EntityClickEvent no longer nullable . - CalMWolfs

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/events/entity/EntityClickEvent.kt
+++ b/src/main/java/at/hannibal2/skyhanni/events/entity/EntityClickEvent.kt
@@ -6,5 +6,5 @@ import net.minecraft.entity.Entity
 import net.minecraft.item.ItemStack
 import net.minecraft.network.play.client.C02PacketUseEntity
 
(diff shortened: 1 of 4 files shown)


DavidArthurCole:
> Made entity in EntityClickEvent no longer nullable because how do you click on an entity but not click on an entity at the same time.
We're in a world of hurt when mojang implements quantum entangled entities

CalMWolfs:
> > Made entity in EntityClickEvent no longer nullable because how do you click on an entity but not click on an entity at the same time.
> > We're in a world of hurt when mojang implements quantum entangled entities

🥺

DavidArthurCole (approved):
I think the Entity is technically nullable from the raw packet, but we already have a null check before we fire the event, so yeah this LGTM.

PR #4376, opened 2025-07-15
Fix: Vermin Tracker pattern on modern

CalMWolfs:
## What
These lines don't start with white on 1.21.

## Changelog Fixes
+ Fixed vermin tracker not tracking properly on 1.21. - CalMWolfs


CalMWolfs:
didnt see #4371

PR #4379, opened 2025-07-15
Improvement: bait detection accuracy

ninjaguardian:
## What
When fishing, a bait pops out from your bobber. Sometimes, SkyHanni thinks another person's bait is yours.
- This pr makes bait find and save the distance to the player's bobber. 
- It shortens the delay of the check (because it felt long to me).
- It changes `inKuudra()` to `inKuudra` because it was deprecated.
- It imports the correct `TimeLimitedSet`.
- It changes the expiry time on the `TimeLimitedSet` because it was unnecessarily long.
- It removes unused things in the `Bait` data class.
- It inlines `isBait` for the GC.
- It shortens the max distance to minimize false detections (I never had a bait's distance go above 2).
- It makes distance oriented around the bobber and not the player (changed `distanceToPlayer() > 10`).

## Changelog Improvements
+ Improved bait detection accuracy. - ninjaguardian

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/fishing/FishingBaitWarnings.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/fishing/FishingBaitWarnings.kt
@@ -10,8 +10,10 @@ import at.hannibal2.skyhanni.features.nether.kuudra.KuudraApi
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.DelayedRun
+import at.hannibal2.skyhanni.utils.LocationUtils.distanceTo
 import at.hannibal2.skyhanni.utils.SoundUtils
 import at.hannibal2.skyhanni.utils.collection.TimeLimitedSet
+import at.hannibal2.skyhanni.utils.getLorenzVec
 import net.minecraft.entity.item.EntityItem
 import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.Duration.Companion.seconds
@@ -23,7 +25,7 @@ object FishingBaitWarnings {
 
     private data class Bait(
         private val entity: EntityItem,
-        val bobberDistance: Float,
+        val bobberDistance: Double,
         val name: String = entity.entityItem.displayName,
     )
 
@@ -49,11 +51,7 @@ object FishingBaitWarnings {
     @HandleEvent(onlyOnSkyblock = true)
     fun onEntityEnterWorld(event: EntityEnterWorldEvent<EntityItem>) {
         if (KuudraApi.inKuudra || !FishingApi.isFishing()) return
-        //#if MC < 1.16
-        val bobberDistance = event.entity.getDistanceToEntity(FishingApi.bobber ?: return)
-        //#else
-        //$$ val bobberDistance = event.entity.distanceTo(FishingApi.bobber ?: return)
-        //#endif
+        val bobberDistance = event.entity.distanceTo(FishingApi.bobber?.getLorenzVec() ?: return)
         if (bobberDistance > 2) return
         DelayedRun.runNextTick {
             if (event.entity.entityItem.isBait()) {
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/fishing/FishingBaitWarnings.kt
```diff
@@ -46,27 +40,31 @@ object FishingBaitWarnings {
...
     @HandleEvent(onlyOnSkyblock = true)
     fun onEntityEnterWorld(event: EntityEnterWorldEvent<EntityItem>) {
-        if (KuudraApi.inKuudra() || !FishingApi.isFishing()) return
-        if (event.entity.distanceToPlayer() > 10) return
+        if (KuudraApi.inKuudra || !FishingApi.isFishing()) return
+        //#if MC < 1.16
+        val bobberDistance = event.entity.getDistanceToEntity(FishingApi.bobber ?: return)
+        //#else
+        //$$ val bobberDistance = event.entity.distanceTo(FishingApi.bobber ?: return)
+        //#endif
```
CalMWolfs: this should not need preprocessing, we definitely have a method that already works

src/main/java/at/hannibal2/skyhanni/features/fishing/FishingBaitWarnings.kt
```diff
@@ -46,27 +40,31 @@ object FishingBaitWarnings {
...
+                baitEntities += Bait(event.entity, bobberDistance)
             }
         }
     }
 
     private fun checkBait() {
-        val bobber = FishingApi.bobber ?: return
-        val bait = baitEntities.filter { it.distanceTo(bobber) < 8 }.minByOrNull { it.distanceTo(bobber) }?.name
+        FishingApi.bobber ?: return
+        val bait = baitEntities.filter { it.bobberDistance < 2 }.minByOrNull { it.bobberDistance }?.name // If user has no bait, but another player's bait spawns really close, it will be wrong.
```
CalMWolfs: detekt isnt gonna like this line length. early warning

CalMWolfs (changes requested):
looks pretty good, just 2 things

src/main/java/at/hannibal2/skyhanni/features/fishing/FishingBaitWarnings.kt
ninjaguardian: getDistanceToEntity was changed to distanceTo sometime after 1.8.9, and the distanceTo in 1.8.9 calculates it completely differently.

> we definitely have a method that already works

If there is, I can't find it.

src/main/java/at/hannibal2/skyhanni/features/fishing/FishingBaitWarnings.kt
ninjaguardian: 26b14de54a4af3493ae891f5bbd55ae658f86992

CalMWolfs (approved):

ninjaguardian:
Now that both use distanceTo, 1.8.9 uses a different system than the rest. This may cause issues with the if statements. I haven't tested if that type of calculation can sometimes exceed 2 blocks.

(Should be good enough though)

CalMWolfs:
> Now that both use distanceTo, 1.8.9 uses a different system than the rest. This may cause issues with the if statements because I have not tested if that type of calculation can sometimes exceed 2 blocks.

wdym by 1.8.9 uses a different system?

ninjaguardian:
> wdym by 1.8.9 uses a different system?

1.8.9 uses LocatonUtils's distanceTo (SkyHanni, LorenzVec based)

1.16+ uses Entity's distanceTo (Minecraft, math based)

(The results of the calculations may be similar but I haven't tested yet)

CalMWolfs:
> (The results of the calculations may be similar but I haven't tested yet)

I made them both use SkyHanni's distanceTo

ninjaguardian:
> I made them both use SkyHanni's distanceTo

Oh oops didn't see the .getLorenzVec

Thanks!

PR #4384, opened 2025-07-15
Fix + Backend: Agaricus Timer on 1.21 etc.

lunaynx:
## What
Fixed Agaricus Cap timer resetting on 1.21 if you move your camera while still targeting the mushroom, because the ray trace result is no longer a full block coordinate but precise position. Also done some cleanup to nearby code and fixed incorrect logic. Hopefully this doesn't break anything else.

## Changelog Fixes
+ Fixed Agaricus Cap timer incorrectly resetting whenever you move your camera on 1.21. - Luna

+ Improved `BlockUtils.getBlockLookingAt()` and split it up into `getTargetedBlock()` and `getTargetedBlockAtDistance()`. - Luna
    * Coordinates are now explicitly rounded to a full block to match 1.8 behavior.
    * The former gets the targeted block directly from Minecraft using the default reach distance instead of 10, while the latter can still be used for custom distances.
+ Moved `LorenzVec.getBlockBelowPlayer()` to `LocationUtils` and fixed incorrect behavior when the player is standing on a non-full block. - Luna
+ Renamed `LorenzVec.roundLocationToBlock()` to `roundToBlock()` and fixed incorrect rounding. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/BlockUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/BlockUtils.kt
@@ -76,14 +76,14 @@ object BlockUtils {
     fun getTargetedBlock(): LorenzVec? {
         val mouseOverObject = Minecraft.getMinecraft().objectMouseOver ?: return null
         if (mouseOverObject.typeOfHit != MovingObjectPosition.MovingObjectType.BLOCK) return null
-        return mouseOverObject.blockPos.toLorenzVec().roundLocationToBlock()
+        return mouseOverObject.blockPos.toLorenzVec().roundToBlock()
     }
 
     fun getTargetedBlockAtDistance(distance: Double) = rayTrace(
         LocationUtils.playerEyeLocation(),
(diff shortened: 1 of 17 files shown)


src/main/java/at/hannibal2/skyhanni/utils/BlockUtils.kt
```diff
@@ -72,11 +73,21 @@ object BlockUtils {
...
+            mouseOverObject.blockPos.toLorenzVec()
+        } else {
+            // Custom reach distance, so we need to ray trace manually
+            rayTrace(
+                LocationUtils.playerEyeLocation(),
+                MinecraftCompat.localPlayer.lookVec.toLorenzVec(),
+                distance,
+            )
+        }?.roundLocationToBlock()
+    }
```
CalMWolfs: id prob split this up into 2 different functions with the same name, 1 with no arguments and one with just the distance argument

CalMWolfs (changes requested):

CalMWolfs (approved):

PR #4387, opened 2025-07-15
Backend: Renderable Move #1

Thunderblade73:
## Note
DO NOT USE merge to solve merge conflicts!
Please use rebaseing or let it be.

Since this PR is split into two commits: 
[the first](https://github.com/hannibal002/SkyHanni/pull/4387/commits/063b366d49a99a9025211086eedb2a2c5f772d80) for the technical changes to Renderables, 
[the second](https://github.com/hannibal002/SkyHanni/pull/4387/commits/ea31ef9bce3d8c9125177061b99478b6c3e49288) to get everything that uses those working.
## What
This PR is the first of the sequence of moving Renderables to the place where they should be.
This changes mainly focus on Renderables that where moved out of Renderable.kt.

### Overview of Changes
* I moved here all Renderables that where moved out of Renderables.kt to the correct place (where I imaged they should go). 
* Reintroduced the Renderable dot syntax back, as it was never intendet to get rid of that.
* A rename of String (Renderable) as that does not really make sense, since the input is some sort of text (either as ``String`` or as modern ``Text``).
* Cleanuped up the mess with the containers, as they had wierd second class definitions, now they are consitent under one name (and got a varag option more).
* Fixed the TimeDependent mess with ``ItemStackRenderable``.
* Used at some instances ``with(Renderable)`` where a lot of Renderable calls happend and not a lot of others, this gives similar visual notice that there is Rendering stuff there as the direct Renderable calls. 
* Introduced an actual ``empty`` Renderable as Singleton, as there where quite some cases where an ``empty`` Renderable was used / made sense to use. 
* Gave the ``RenderableWrapper`` a rename to ``RenderableDecorator`` and implemented it a bit smoother with an interface instead of an abstract class. 
* Did some cleanup of feature code but be more consitent and use some fancier ways to do the same thing where I noticed.
* Added some todos for stuff that needs changing in the future, which I came accross during the replacement.

+ Cleanup of core Renderables. - Thunderblade73
    * That includes StringRenderable, ItemRenderable, ContainerRenderable and also some misc stuff.


src/main/java/at/hannibal2/skyhanni/utils/collection/RenderableCollectionUtils.kt
```diff
@@ -55,7 +58,7 @@ object RenderableCollectionUtils {
             // Hack to add enchant glint, like Hypixel does it
             itemStack.addEnchantment(EnchantmentsCompat.PROTECTION.enchantment, 1)
         }
-        add(ItemStackRenderable(itemStack, scale = scale))
+        add(Renderable.item(itemStack, scale = scale))
```
DavidArthurCole, line 61: Is this branch up to date? I swear I removed the `Renderable.item()` function in favor of using `ItemStackRenderable` a while back 🤔

DavidArthurCole: (It was just deferring to `ItemStackRenderable` at the time of removal anyways iirc)

Thunderblade73: Look at the Description xD

> Also reintroduced the Renderable dot syntax back, as it was never intendet to get rid of that.

But that instance should actually use addItemStack

DavidArthurCole: Ahh, misunderstood the rebasing comment. Nvm me 🙂

src/main/java/at/hannibal2/skyhanni/utils/renderables/primitives/TextRenderable.kt
```diff
@@ -0,0 +1,88 @@
...
+    // Unsure why we would need that
+    /*
+    constructor(
+        text: String,
+        scale: Double = 1.0,
+        color: Color = Color.WHITE,
+        horizontalAlign: HorizontalAlignment = HorizontalAlignment.LEFT,
+        verticalAlign: VerticalAlignment = VerticalAlignment.CENTER,
+    ) : this(Text.of(text), scale, color, horizontalAlign, verticalAlign)
+     */
```
NopoTheGamer: this can probably go, it was an idea i had before i had better ideas

CalMWolfs (changes requested):
Im a bit confused around the changes like this, I thought we were moving away from the functions like `Renderable.string` or `Renderable.vertical` and instead replacing them with `StringRenderable` or `VerticalContainerRenderable` but now we seem to be going back to the old format that was recently removed, would like some clarification on this

Thunderblade73:
> Im a bit confused around the changes like this, I thought we were moving away from the functions like `Renderable.string` or `Renderable.vertical` and instead replacing them with `StringRenderable` or `VerticalContainerRenderable` but now we seem to be going back to the old format that was recently removed, would like some clarification on this

I never stated in my plan that we move away from it. I explicitly said that we keep that behavior. Daveed just made that decision by himself and the issue is that I did not stop him as I thought it might be better that way, but I came to the conclusion that using the dot Syntax is better to look at when used and also allows for easy Intelli Sense as you just have to type Renderable dot and all possible option will be listed. The class syntax is better for inheritance and that is the only real benefit. Since text length is about the same. But the dot syntax allows for seamless helper function, which gets us a nicer look overall. The conclusion I came was that we keep explicit classes but do not allow instanciation from feature code, there we need to use the Extension Functions. Performance vise there is no real difference, even if there where some we could add an inline to all the extension functions, which makes the code identical with calling the constructor. The only issue is that constructor and the extension function needs to be kept in sync. Which sounds bad but it is just one copy and paste. If need be we could add an detekt rule for that or simply generated those extension functions using ksp.

DavidArthurCole:
> I came to the conclusion that using the dot Syntax is better to look at when used and also allows for easy Intelli Sense as you just have to type Renderable dot and all possible option will be listed.

I disagree on this - there might be a lot of types of Renderables, but nowhere near enough to need to see a list to know what you need to use.

> The class syntax is better for inheritance and that is the only real benefit.

It's a massive benefit - it's what's driving all of the Circular renderables and the item stack ones..

> The only issue is that constructor and the extension function needs to be kept in sync

If we do go down the route of wanting the dot notation, put all of the renderable classes into Renderable.kt and make them defer to the objects - there's no reason to duplicate the code.

DavidArthurCole (changes requested):
see above comment; TLDR fine with dot notation, backend should stay classes, as inheritance is important

Thunderblade73:
> If we do go down the route of wanting the dot notation, put all of the renderable classes into Renderable.kt and make them defer to the objects - there's no reason to duplicate the code.

They do not need to be inside the Renderable.kt file as you can declare extension functions on companion objects, that I already use in that Pr. Since Renderable.kt should be only a small file not the chaos is now. The extension functions are kept as close as possible to the Type they defer. 

> It's a massive benefit - it's what's driving all of the Circular renderables and the item stack ones.

Classes are kept and you can inherit from them, so we do not lose that benifit.

> I disagree on this - there might be a lot of types of Renderables, but nowhere near enough to need to see a list to know what you need to use.

When you use them regular sure you know all the types you need but for new/casual devs it is not that intuitive and Intellij Sense is a big belp, imo.

CalMWolfs (approved):
everything seemed to work

PR #4388, opened 2025-07-16
Improvement: Filter Swoop's Message

NotRoboo:
## What
Chat filter to remove message when hitting Galatea mobs with something that isn't an axe.

<details>
<summary>Images</summary>

SwoopConfig
SwoopChatHistory

</details>

## Changelog Improvements
+ Added filter for Swoop's message. - Roboo


CalMWolfs (approved):
Worked perfectly, thanks. I also removed the search tags as they weren't needed in this case as the config name/description already included both of those words

PR #4397, opened 2025-07-16
BingoNet v2 SH [Feature]

HacktheTime:
This is aimed at integrating the BN Mod into SH soonish. Everything is still broken as of typing and theres are changes everywhere. This will be cleaned up over time, but hanni told me to open a draft PR as soon as possible.

<details>
- Bingo Net & Bingo Brewers (since no modern port) Integrations.
  - These are required for a Bunch of things below. I would disable them by default!
- Splash Announcement Support Ingame
- ChChest Items sharing

There are more Features, but those are invisible to the Normal User and are used for Management Purposes. Some more Features come from Bingo Net, which requires Registration, so not going to be included here.

</details>

## Changelog New Features
+ Bingo Net Integration - Hype_the_Time
    * Keep in mind that Bingo Net requires Registration as parts of its Setup!
    * Splashes, Dynamic Splashes, ChChests, Chest Search, Services, Trades, Bingo Chat, Card / Goal Completion tracking & more.

+ Bingo Brewers Integration (Splashes & ChChest) - Hype_the_Time
    * As of now only Splashes and ChChest are supported

+ Party Commands are now all executed from within the PartyAPI instead of HypixelCommands - Hype_the_Time


mariethepolarfox:
Quick thing I noticed is that you use underscores in variable names, we only use camelCase

HacktheTime:
I will continue working on this, but in which scope it will be delivered to the user is not clear to me yet. We will see.

PR #4402, opened 2025-07-17
Fix: Pet XP Calculations

DavidArthurCole:
## What
Added an index check before we go slicing something that might not exist. Also added a warning that this is caused by NEU repo being fucked, which will go off every 5 minutes when the slice is out of bounds.

## Changelog Fixes
+ Fixed errors when calculating Pet XP. - Daveed


hannibal002 (changes requested):
I think replacing the duplicate bound check and additional slice helper variable with getOrNull calls in the loop might be more elegant here

DavidArthurCole:
> I think replacing the duplicate bound check and additional slice helper variable with getOrNull calls in the loop might be more elegant here

If we don't check the size and just `getOrNull()` and then assume everything's "safe", the user will never know why their XP calculations are so far off - this is point of failure detection

src/main/java/at/hannibal2/skyhanni/utils/PetUtils.kt
```diff
@@ -176,9 +178,19 @@ object PetUtils {
...
-        return getFullLevelingTree(petInternalName)
-            .slice(0 + rarityOffset..<level + rarityOffset - 1)
-            .sumOf { it.toDouble() }
+        val indicesSlice = 0 + rarityOffset..<level + rarityOffset - 1
+        val levelTree = getFullLevelingTree(petInternalName)
+        if (indicesSlice.first < 0 || indicesSlice.last >= levelTree.size) {
+            if (lastRepoWarning.passedSince() > 5.minutes) {
+                ChatUtils.userError(
+                    "§cFailed to load pet levels from NEU repo. " +
+                        "§cYou can try to fix this by running `§e/neuupdaterepo`§c."
```
CalMWolfs: We should tell them to use `ItemUtils.resetCommand` instead of this hard coded string

CalMWolfs (changes requested):
I agree with Daveed here, in the sense that we want the user to know there is an error and have a possible fixing method. And then if it is an issue with their repo not loading correctly, running this command may produce more info for us so that we can fix the root repo issues in the future. 

Just needs this one thing changed

DavidArthurCole:
> Just needs this one thing changed

fixed

src/main/java/at/hannibal2/skyhanni/utils/PetUtils.kt
```diff
@@ -176,9 +178,18 @@ object PetUtils {
...
+            if (lastRepoWarning.passedSince() > 5.minutes) {
+                ChatUtils.userError(
+                    "§cFailed to load pet levels from NEU repo. " +
+                        "§cYou can try to fix this by running `§e/${ItemUtils.resetCommand}`§c."
+                )
+                lastRepoWarning = SimpleTimeMark.now()
+            }
+            return null
+        }
+        return levelTree.subList(rarityOffset, rarityOffset + level).sumOf { it.toDouble() }
```
CalMWolfs, line 192:
```suggestion
        return levelTree.subList(rarityOffset, rarityOffset + level - 1).sumOf { it.toDouble() }
```

I think this is missing

CalMWolfs (changes requested):

CalMWolfs (approved):

PR #4403, opened 2025-07-17
Backend: Full Reset on Repo Manger Update Command

DavidArthurCole:
## What
Won't tell you your hashes are up to date when you use the update command anymore, it'll forcibly nuke the repo and redownload it.

+ Changed update-repo commands to fully remove and re-download the repo. - Daveed
+ Added task to include backup NEU repo in jar. - Daveed


hannibal002 (changes requested):
If we no longer check the hashes for this very context, why do we have them in the first place? I would like to still per default not reset the repo on update command if not entirely necessary. Only actually Update the repo if hashes don't match or any other repo error is indicated. Instead, maybe show a "no update necessary, click here to update anyway" chat message maybe?

NopoTheGamer:
> If we no longer check the hashes for this very context, why do we have them in the first place? I would like to still per default not reset the repo on update command if not entirely necessary. Only actually Update the repo if hashes don't match or any other repo error is indicated. Instead, maybe show a "no update necessary, click here to update anyway" chat message maybe?

the hash is still used for when the game auto updates the repo?

hannibal002:
This auto update logic could aks reuse the additional "is the repo data currently being lagged as invalid?" Check. My point stands, I want to run this command and have it do nothing if rhere is nothing to fix, and then run again to actually force wipe the data. No need to do so immediately if not necessary. The command in question is called update and not reset. Simply put, we don't need to update something that is alr up to date and working @nopothegamer

NopoTheGamer:
>  The command in question is called update and not reset. Simply put, we don't need to update something that is alr up to date and working @NopoTheGamer

clearly we arent asking people to run the command if they arent having issues

hannibal002:
I run this command from time to time simply to see if I'm up to date. I don't want to download the repo again if not necessary

NopoTheGamer:
@DavidArthurCole add a uuid check to make it not redownload the repo

CalMWolfs (approved):
This all seems fine to me but I will wait for @hannibal002's confimation, but i do just want to reiterate, this change here adds the same functionality to SkyHanni as Neu has with this command. 
I think a possible solution is adding an optional boolean argument which is the `forceReset` parameter so then if you do `/neuresetrepo false` it doesn't update unless you are outdated but by default `/neuresetrepo` will reset your repo no matter what

CalMWolfs (approved):

PR #4414, opened 2025-07-17
Backend: Live Plugin for Copying Config Path

DavidArthurCole:
## What
I finally got fed up enough while doing a migration to make a plugin to copy the local config path for a config option with a context action. See images.

<details>
<summary>Images</summary>

image

image
</details>

+ Added live plugin for copying the 'path' to a config option. - Daveed


CalMWolfs (approved):
fire emoji

PR #4417, opened 2025-07-18
Fix: Shorten Coin Amount on < 1 values

CalMWolfs:
## What
If its less than 1 I just show the original number. Also added a search tag and message replace reason

## Changelog Fixes
+ Fixed shorten coin amount behaving incorrectly on values < 1. - CalMWolfs


hannibal002:
Any example where this breaks? Also doesn't this now totally break for cases with long negative numbers?

NopoTheGamer:
> Any example where this breaks? Also doesn't this now totally break for cases with long negative numbers?

anything with 0.5 npc or bazaar price

hannibal002:
Oh good. How about we only ignore if between -500 and 0?

PR #4424, opened 2025-07-18
Fix + Backend: Renderable Move #2 (table)

Thunderblade73:
## What
* Moved the all 4 table renderables out of Renderable.kt.
* Has the same format as #4387 for the Renderable API, which does not really change here as all the table where in the Renderable dot format anyway. So we can merge this independet of the Renderable API discussion, as it won't be more work now or after the PR.
* Added Renderable Table test, to verfiy the behaviour (mostly edge cases therefore it looks somewhat unnatural)
* Rewrote the scrolling logic of the ScrollTable and the SearchableScrollTable, as both where broken in diffrent ways. Both now use one abstracted interface which takes care of the whole window selection, which I hand made every time differently when it was needed. This should be adepted in the 2 scroll List varients, but not the direct point of the PR, so will be done later on.

## Changelog Fixes
+ Fixed small inconsistencies in scrollable Table GUIs. - Thunderblade73
    * Eg: Trackers.

+ Moved table Renderables out of `Renderable.kt`. - Thunderblade73


CalMWolfs (approved):
all seems to work for me and code changes look fine but due to the nature of this will wait for hannibal & daveed to also approve

Thunderblade73:
@hannibal002, @DavidArthurCole pls review🥺

src/main/java/at/hannibal2/skyhanni/features/garden/fortuneguide/pages/UpgradePage.kt
```diff
@@ -37,15 +36,13 @@ class UpgradePage(val crop0: () -> CropType?, sizeX: Int, sizeY: Int, paddingX:
...
-            content = buildList {
-                add(header())
-                val upgradeList = if (crop == null)
-                    FortuneUpgrades.genericUpgrades
-                else
-                    FortuneUpgrades.cropSpecificUpgrades
-                addAll(upgradeList.map { upgrade -> upgrade.print() })
-            },
+            content = upgradeList.map { upgrade -> upgrade.print() },
+            header = header()
```
DavidArthurCole: Not your change, but this is a static list of string renderables - it can probably get val-ed instead of being a function that re-creates renderables every time.
Or just put a `// TODO` on it

src/main/java/at/hannibal2/skyhanni/test/renderable/TestTable.kt
```diff
@@ -0,0 +1,232 @@
...
+import java.awt.Color
+
+@SkyHanniModule(devOnly = true)
+object TestTable : RenderableTestSuite.TestRenderable("table") {
+
+    private val scrollValues = mutableMapOf<Int, ScrollValue>()
+    private val textInput = TextInput()
+
+    @Suppress("LongMethod")
+    override fun renderable(): Renderable = with(Renderable) {
```
DavidArthurCole, line 27: Same principle here - any reason to not make this a val instead of a function? Seems wasteful, even on a testing renderable, to re-do all this (semi?) static stuff each iteration

src/main/java/at/hannibal2/skyhanni/utils/guide/GuideScrollPage.kt
```diff
@@ -11,21 +12,20 @@ abstract class GuideScrollPage(
     paddingY: Int = 0,
     val marginY: Int = 5,
     val velocity: Double = 3.0,
-    val hasHeader: Boolean = true,
 ) : GuideRenderablePage(paddingX, paddingY) {
 
     private val scroll = ScrollValue()
 
-    fun update(content: List<List<Renderable>>) {
+    fun update(content: List<List<Renderable>>, header: List<Renderable>) {
```
DavidArthurCole: Maybe make these have a default of an empty list, so that you don't need to pass something if you don't want to update it.

This would require a couple of inheritance changes down the line, to be able to store, then modify the renderable in place, but I don't think having concrete state management would be a bad thing.

src/main/java/at/hannibal2/skyhanni/utils/renderables/container/table/SearchableScrollTable.kt
```diff
@@ -0,0 +1,166 @@
...
+
+        DrawContextUtils.translate(0f, -renderY.toFloat(), 0f)
+
+    }
+
+    companion object {
+        private val scrollUpTip = Renderable.text("§7§oMore items above (scroll)")
+        private val scrollDownTip = Renderable.text("§7§oMore items below (scroll)")
+
+        fun Renderable.Companion.searchableScrollTable(
```
DavidArthurCole, line 134: Not so much a change request here, but I think this perfectly exemplifies my main issue with the `Renderable.function` approach - we're adding 31 lines of code here to duplicate a constructor.

src/main/java/at/hannibal2/skyhanni/utils/renderables/container/ContainerRenderable.kt
```diff
@@ -1,17 +1,9 @@
...
-abstract class ContainerRenderable(
-    var renderables: Collection<Renderable>,
-    open val spacing: Int = 0,
-    override val horizontalAlign: HorizontalAlignment,
-    override val verticalAlign: VerticalAlignment
-) : Renderable {
-    abstract override val width: Int
-    abstract override val height: Int
-    abstract override fun render(mouseOffsetX: Int, mouseOffsetY: Int)
+abstract class ContainerRenderable : Renderable {
```
DavidArthurCole, line 5: Should either be an interface, or needs some refactoring across the board - `CircularContainerRenderable` should definitely extend this, which as of now is impossible, due to sole sub-classing.

src/main/java/at/hannibal2/skyhanni/utils/renderables/container/SlidingWindow.kt
DavidArthurCole, line 1: Feels a bit weird in the base `container` package, since it itself does not extend container.
Maybe split this into its own `window` package, and each class into its own file?

src/main/java/at/hannibal2/skyhanni/utils/renderables/container/SlidingWindow.kt
```diff
@@ -0,0 +1,114 @@
...
+        }
+}
+
+/**
+ * @param getPostPosition returns the absolute Position of that [P]
+ * @return A Range starting with the first element that is inside the window
+ * and the last element that fits inside the window size, if the previous
+ * ones are stacked on top of the first element
+ */
+fun <P> SlidingWindow.relativProvider(getPostPosition: (post: P) -> Int): (elements: Iterable<P>) -> IntRange = lambda@{ elements ->
```
DavidArthurCole: typo, `relativ` -> `relative`

DavidArthurCole (changes requested):
In general, I can feel the inheritance here getting a bit out of hand.
We have `ScrollTable` and `SearchableScrollTable` both extending _extremely_ similar class sets, we have `CircularContainerRenderable` that does not extend `ContainerRenderable`, etc.

What I might like is to instead of having a mix of interfaces and classes strewn about, we do a hierarchical/combinatorial "ability" approach, where interfaces declare what functions the Renderables are looking to 'serve,' to hopefully prevent the incoming mountain of technical debt I can see over the horizon....

Don't know if that explicitly needs to be done in this PR, but I would like to see this 'solved' at some point, as I don't think the route we're on here is maintainable.

src/main/java/at/hannibal2/skyhanni/features/garden/fortuneguide/pages/UpgradePage.kt
Thunderblade73: This only get's called once. As that is the init function of the page not the render.

src/main/java/at/hannibal2/skyhanni/test/renderable/TestTable.kt
Thunderblade73: This is to allow to hotswaping support as inits won't resolve correctly. So you can teak this function and after hotswapping it will show, that is the reason.

src/main/java/at/hannibal2/skyhanni/utils/guide/GuideScrollPage.kt
Thunderblade73: You know that this is a function to change the renderable, so it is storead already. Also I do not care for a default header empty list as that thing won't be used in the next centerury (since no one makes other ``/ff`` still thingys). Also unrelated to the PR may I add.

src/main/java/at/hannibal2/skyhanni/utils/renderables/container/table/SearchableScrollTable.kt
Thunderblade73: As I sad not the point of the PR to fix that syntax, as we currently have no specification for the Renderable API. So it does not matter to this PR.

src/main/java/at/hannibal2/skyhanni/utils/renderables/container/ContainerRenderable.kt
Thunderblade73: Yah that is the mess you created. I just tried to clean it up a bit. But was not directly the Point of this.

src/main/java/at/hannibal2/skyhanni/utils/renderables/container/SlidingWindow.kt
Thunderblade73: I know it is a bit wierd since it does not even need to be a Renderable since it is that much generic. But we will only us this for Container Renderable as that are the only Renderable where you would need windowing. Also a window package would be insane for one file, imo.

src/main/java/at/hannibal2/skyhanni/utils/renderables/container/SlidingWindow.kt
Thunderblade73: Thx

src/main/java/at/hannibal2/skyhanni/utils/renderables/container/ContainerRenderable.kt
Thunderblade73: Since I tried to make it work with tables so that Table Renderable are also Container Renderable but I could not resolve that cleanly without chaning everything around, I just removed here all the redundant qualifiers.

Thunderblade73:
Sorry that I may sound a bit harsh (I'm quite upset). Most of the issues you have are not the point of the PR and also I explicitly addressed in the description. I just want to get this merged so that we can change the API at once for all instead of me having to fix that PR separately. Also I want to get the bug fix out as well.

> CircularContainerRenderable that does not extend ContainerRenderable, etc.

Btw. ``CircularContainerRenderable`` is not even a container, it is actually a decorator, but I just did not rename it so that I don't break your PR's completely.

> We have ScrollTable and SearchableScrollTable both extending extremely similar class sets, we have

Yeah they are also basically the same thing. I just didn't add the hint behavior out of no where to the ScrollTable (which is only used in one place so far btw.). So why should two basically identically thing not extent similar interfaces?

 > What I might like is to instead of having a mix of interfaces and classes strewn about, we do a hierarchical/combinatorial "ability" approach, where interfaces declare what functions the Renderables are looking to 'serve,' to hopefully prevent the incoming mountain of technical debt I can see over the horizon....
 
 That was the intention I had. That is why SlidingWindow is complelty seperated from Renderable, that is why TimeDepentedRenderable is somewhat detached (I did not find a clean way). 
 For the table Renderable I just made multiple variants since stuff table stuff does not fit in a box cleanly, as some have to act on rows, some only on cells and most have to act on the table call it self. That is why that is so fragmented, I did not find a better way.
 The sliding Window has two ways as one implements it with additional functionality by changing the original functions around/

hannibal002 (approved):
In my testing everything works, and since no more comments from David, ill merge

PR #4425, opened 2025-07-18
Backend: Jump to Config (from ConfigFix)

DavidArthurCole:
## What
Follow-up to #4414 - this adds the other direction. From a string inside an `event.move( )`, there is an `Open Config` option which will try to parse where the path resolves to. See images.

<details>
<summary>Images</summary>

image
image

image
image
</details>

+ Added live plugin for jumping to config from migration path. - Daveed

### DIFF
--- a/.live-plugins/jump-to-definition/plugin.kts
+++ b/.live-plugins/jump-to-definition/plugin.kts
@@ -14,49 +14,73 @@ import org.jetbrains.kotlin.psi.*
 
 registerIntention(NavigateToConfigIntention())
 
-val basePkg = "at.hannibal2.skyhanni.config"
-val baseClass = "at.hannibal2.skyhanni.config.Features"
+val baseConfigPkg = "at.hannibal2.skyhanni.config"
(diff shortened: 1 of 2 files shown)


CalMWolfs (changes requested):
more of a feature request than a problem, i would also like for it to add support for #player and #profile support here to take you to player/profile specific of those paths. Maybe this can be done in a second pr and then add support for these 2 for that other thing you made a day or two ago. lmk if you want to split it up, add to this or dont want to do it at all

DavidArthurCole:
> lmk if you want to split it up, add to this or dont want to do it at all

will pick this up tomorrow. will be a little different confining it to one file but should be doable

DavidArthurCole:
> more of a feature request than a problem, i would also like for it to add support for #player and #profile support

image

image

image

From my very basic testing, seems to be working with `#profile` and `#player` prefixes now too.

CalMWolfs (approved):

PR #4447, opened 2025-07-20
Fix: 1.21 NPE in entity utils

hannibal002:
## What
my first 1.21 code (that i can think of)
hopefully fixed this error here:
https://discord.com/channels/997079228510117908/1396563663908896940

looks like `it.pos` can be null. then `.getBlockEntity()` gets called with null as parameter. Also did some cleanup in the function.

## Changelog Fixes
+ Fixed rare error in Crimson Miniboss Respawn Timer on 1.21. - hannibal2


jani270:
wiht

CalMWolfs (approved):

PR #4454, opened 2025-07-21
Backend: Item Rendering 1.21.7

DavidArthurCole:
## What
Makes spinning items work on 1.21.7 - if I have to do something this matrix intensive again I might actually have a mental breakdown.

+ Ported all item rendering functionality to 1.21.7. - Daveed


DavidArthurCole:
> No it's not working yet, this is getting drafted so I can let this rot while I wait for someone smarter than me to fix it.

Smarter person turned out to be me 6 hours in the future.

DavidArthurCole:
I'm going to mark this as ready for review, as from a literal sense, the items _are_ rendering, spinning, scaling, etc. correctly - however there is a notable change with these items, being that they render from a "true" side-view, rather than that (35 degree?) offset top-down-ish angle that GUI items 'usually' render at.

If this is a dealbreaker, or if we want it to be a 1:1 functionality match, I can see what the static matrix ops would be like for replicating the 'old look.'

src/main/java/at/hannibal2/skyhanni/utils/GuiRenderUtils.kt
```diff
@@ -479,14 +493,64 @@ object GuiRenderUtils {
...
+        //$$ client.gameRenderer.diffuseLighting.setShaderLights(DiffuseLighting.Type.ITEMS_3D)
+        //$$ // I don't know why the magic value is 12f, but it is.
+        //$$ matrices.scale(12f, 12f, 12f)
+        //$$ // 15728880 comes from LightmapTextureManager.MAX_LIGHT_COORDINATE
+        //$$ client.itemRenderer.renderItem(item, ItemDisplayContext.FIXED, 15728880, OverlayTexture.DEFAULT_UV, matrices, consumers, client.world, 0)
+        //$$ consumers.draw()
+        //$$ matrices.pop()
+        //$$ RenderSystem.teardownOverlayColor()
+        //$$ RenderSystem.restoreProjectionMatrix()
+        //#endif
```
CalMWolfs: id like there to be some empty lines in here so it is easier to read this as this is just a massive block of code and comments in one

CalMWolfs (changes requested):
I have a few issues with how normal items now render on 1.21.7
- all skulls are now flat and facing the camera
- seems there is no longer any lighting on the items, particularly visible on the swords
- scale of the items isnt quite right notice how the glass panes fill the whole green square without your changes but with your changes it only is like 90% of the size

with your changes:
image

what it looked like before on 1.21.7 without your changes:
image

another display that now looks wrong due to these changes:
image

src/main/java/at/hannibal2/skyhanni/utils/GuiRenderUtils.kt
```diff
@@ -385,16 +396,28 @@ object GuiRenderUtils {
         )
     }
 
+    // todo, does this actually have to be matching Mojang's projection matrix?
+    //  theirs is 1000 -> 11000 by default, but we only use ~20 layers of that,
+    //  see if we can adjust this to maybe 100f -> 200f.
+    //  if we do change this, the 1.21.6 zT below will need to be adjusted as well.
```
CalMWolfs, line 404: never got around to looking into alternate values of this, you can do this if you want

CalMWolfs (approved):
This all seems to work fine now on every version but will wait for you to look over and verify my changes are fine. I had to remove some of the code you put because it caused problems on other versions such as breaking items with display models on 1.21.5 and as such I added an item with an item model to the item renderable test.

I also have not made it by default use the custom rendering for items with a higher scale as there was a few problems with your code. You had a hacky work around for glass panes but there are many more items like cakes that also need this sort of workaround so a newer alternative needs to be devised for these that isnt as simple as this item didnt render right im adding it to a list of items that need special logic.

Once the issues with random items is fixed I would like to use this custom rendering logic on items that we render scaled up as it will make them look much better

DavidArthurCole:
> You had a hacky work around for glass panes but there are many more items like cakes that also need this sort of workaround so a newer alternative needs to be devised for these that isnt as simple as this item didnt render right im adding it to a list of items that need special logic.

The ideal solution here would be to get the baked model from Mojang, and compare its Z axis to see if it has a truly 3-dimensional model, but in looking at this yesterday, I could not find any way to actually do that. So I don't know if there _is_ a solution to this that doesn't involve some level of manual intervention. If there's a commonality between glass panes and cakes, I can't find it in the code.

src/main/java/at/hannibal2/skyhanni/utils/GuiRenderUtils.kt
```diff
@@ -427,39 +425,22 @@ object GuiRenderUtils {
 
         //#if MC < 1.21
         val baseScale = if (isSkull) (4f / 3f) else 1f
-        //#elseif MC < 1.21.6
-        //$$ val baseScale = if (isSkull) (5f / 4f) else 1f
         //#else
-        //$$ val baseScale = when {
-        //$$     isSkull -> 1f
-        //$$     is3dBlock -> (5f / 4f)
```
DavidArthurCole: why did we undo this? this was necessary to keep skulls the right size, and rescale blocks to the correct size as well...

src/main/java/at/hannibal2/skyhanni/utils/GuiRenderUtils.kt
```diff
@@ -427,39 +425,22 @@ object GuiRenderUtils {
...
-        //#if MC < 1.21.6
-        val translateOffset = if (isSkull) 2.5f * scaleMultiplier.toFloat() else 0f
-        //#else
-        //$$ val translateOffset = when {
-        //$$    // Idk man see field_53210 in HeadFeatureRenderer
-        //$$    isSkull -> -1.1875F * 0.5f * scaleMultiplier.toFloat() // Scale UP instead of down
-        //$$    is3dBlock -> (1f / baseScale) * 2f * scaleMultiplier.toFloat()
-        //$$    else -> 0f
-        //$$ }
-        //#endif
```
DavidArthurCole: Same question here - why did we undo this? The scales of these items when rendered changed drastically between versions - was this removed out of necessity, or?

DavidArthurCole (commented):
At the very least, with these changes, the custom rendering won't match normal item rendering.

src/main/java/at/hannibal2/skyhanni/utils/GuiRenderUtils.kt
```diff
@@ -479,16 +501,119 @@ object GuiRenderUtils {
...
+        //$$ val matrices2D = DrawContextUtils.drawContext.matrices
+        //$$
+        //$$ // And similarly, we need to extract the scaling from the GUI editor as well, since we're building our own stack.
+        //$$ val scaleX = sqrt(matrices2D.m00() * matrices2D.m00() + matrices2D.m01() * matrices2D.m01())
+        //$$ val scaleY = sqrt(matrices2D.m10() * matrices2D.m10() + matrices2D.m11() * matrices2D.m11())
+        //$$ val guiScale = (scaleX + scaleY) * 0.5f
+        //$$ val totalScale = guiScale * finalScale
+        //$$
+        //$$ // TODO once random items don't break this can be re-enabled on items when their scale is high
+        //$$ if (rotationDegrees != null /* || totalScale > 1.5 */) {
```
DavidArthurCole: Also, still wholeheartedly disagree with locking this behind either rotation, or scaling > 1.5 - the performance impact is insanely minimal here. I'm thinking a config option for forcing items to render with better textures.

CalMWolfs: The reason we lock it behind stuff is because I don't want random bugs to do with item rendering, my thoughts is we lock it behind for now so that I can at least merge it, as otherwise this isn't getting merged until I can ensure your changes don't break on random items. I also want there to still be the "simpler" rendering method so that if a future mc update changes around anything that you used for your custom rendering we don't have to solve that before updating to a new mc version.

DavidArthurCole:
Skulls are oversized on 1.21.7
image

And on 1.21.5
image

And in 1.8.9
image

Skulls are being incorrectly scaled to the same size that blocks are.

DavidArthurCole:
After some discussion with Cal, I'm good with this going as it is now. It'll require some editing of things that are using display skulls I think, but conceptually it should be fine.

CalMWolfs (approved):

PR #4461, opened 2025-07-22
Backend: Gradle configuration cache

Thunderblade73:
## Dependencies
- https://github.com/architectury/architectury-loom/pull/291

## What
Enabled configuration cache for gradle, which when hot building skips the whole configuration mess we have (for me it takes 50s to do those, with cache it is 0).
Only needed to fix two incompabtilbe parts. 
First the version coping into our meta files, which was not using the input instead used a direct access to the version, which the configuration cache does not allow.
Secondly a none seriliziable property inside essential-loom, waiting for https://github.com/architectury/architectury-loom/pull/291 to be merged, which then gets merged in to essential-loom afterwards.

<details>
<summary>Images</summary>

Left without the cache, right using the cache. No change is made so it is a bit extreme the values.
image

</details>

+ Enabled Gradle configuration cache. - Thunderblade73
    * Removes the whole configuration stage when hot building, which gives for 1.8.9 a ~50% speedup in build time.


Thunderblade73:
Btw. build failed as that essential-loom version only exist on my local machine 😅

Thunderblade73:
Do not want to work on this.

PR #4462, opened 2025-07-22
backend: move sound of foraging

hannibal002:
## What
wrong sub category
also changed default to false.

ignore_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/foraging/TreesConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/foraging/TreesConfig.kt
@@ -35,12 +35,12 @@ class TreesConfig {
     @ConfigEditorBoolean
     @FeatureToggle
     @OnlyModern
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/config/features/foraging/TreesConfig.kt
```diff
@@ -30,4 +30,17 @@ class TreesConfig {
     @FeatureToggle
     var compactSweepDetails = true
 
+    @Expose
+    @ConfigOption(name = "Mute Tree Breaking", desc = "Mutes the sound of the tree fully breaking.")
+    @ConfigEditorBoolean
+    @FeatureToggle
+    @OnlyModern
+    var muteBreaking = false
```
DavidArthurCole: Don't really know if this _needs_ to be default false - if you wanna change it, not opposed to that either. See discussion in contrib chat.

PR #4463, opened 2025-07-22
Backend: cw waypoint cleanup

hannibal002:
## What
cleanup a bit in the ColeweightWaypointFormat file. just the start? idk, the cw stuff looks to me like it needs a small overhead, and also /shgraph support (like the parkour)

+ Cleaned up some code in ColeweightWaypointFormat . - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/model/waypoints/Waypoints.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/model/waypoints/Waypoints.kt
@@ -12,8 +12,5 @@ class Waypoints<T : Copyable<T>>(
 ) : MutableList<T> by waypoints {
     fun deepCopy() = transform { it.copy() }
 
-    inline fun <R : Copyable<R>> transform(transform: (T) -> R): Waypoints<R> {
-        val waypoints1 = waypoints.map { transform(it) }.toMutableList()
-        return Waypoints(waypoints1)
-    }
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/data/model/waypoints/Waypoints.kt
```diff
@@ -8,7 +8,12 @@ interface Copyable<T> {
...
 class Waypoints<T : Copyable<T>>(
     @Expose
-    val waypoints: MutableList<T> = mutableListOf()
+    val waypoints: MutableList<T> = mutableListOf(),
 ) : MutableList<T> by waypoints {
-    fun deepCopy() = Waypoints(waypoints.map { it.copy() }.toMutableList())
+    fun deepCopy() = transform { it.copy() }
+
+    inline fun <R : Copyable<R>> transform(transform: (T) -> R): Waypoints<R> {
+        val waypoints1 = waypoints.map { transform(it) }.toMutableList()
```
Thunderblade73: Why would we need a ``val waypoint1`` here directly return it would be simpler

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/model/waypoints/Waypoints.kt
hannibal002: good catch thanks!

CalMWolfs (approved):

PR #4468, opened 2025-07-23
Feature + Backend: hide mobs nearby Slayers, entity opacity

hannibal002:
## What
Reused the existing "hoppity egg hide nearby player" code for mobs that are annoying while doing slayers (see screenshot)
instead of code duplication, i have abstracted the "entity opacity render" logic into two new events.

Since the hoppity egg hide nearby" feature doesnt work in 1.21.x, this new feature also only works in 1.8.

<details>
<summary>Images</summary>

image
image

</details>

## Changelog New Features
+ Added Hide Irrelevant Mobs for Slayers. - hannibal2
    * Makes mobs partially transparent so that they don't annoy while having an active slayer quest.
    * Useful for e.g. Magma Cubes in Burning Desert for Tara Slayer.

+ Added abstract logic for changing the opacity of entites with EntityOpacityManager. - hannibal2
    * This also includes the two new events EntityOpacityActiveEvent and EntityOpacityEvent.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/slayer/SlayerConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/slayer/SlayerConfig.kt
@@ -108,7 +108,7 @@ class SlayerConfig {
         desc = "Makes mobs partially transparent so that they dont annoy while having an active slayer quest. " +
             "Useful for e.g. Magma Cubes in Burning Desert for Tara Slayer.",
     )
-    @SearchTag("tarantula, spider")
+    @SearchTag("tarantula, spider opacity")
     @ConfigEditorBoolean
     @FeatureToggle
     var hideIrrelevantMobs: Boolean = false
--- a/src/main/java/at/hannibal2/skyhanni/config/features/slayer/SlayerConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/slayer/SlayerConfig.kt
@@ -108,7 +108,7 @@ class SlayerConfig {
         desc = "Makes mobs partially transparent so that they dont annoy while having an active slayer quest. " +
             "Useful for e.g. Magma Cubes in Burning Desert for Tara Slayer.",
     )
-    @SearchTag("tarantula, spider opacity")
+    @SearchTag("tarantula spider opacity")
     @ConfigEditorBoolean
     @FeatureToggle
     var hideIrrelevantMobs: Boolean = false
@@ -118,7 +118,7 @@ class SlayerConfig {
         name = "Adjust Irrelevant Opacity",
         desc = "Adjust the opacity of irrelevant mobs. (in %)",
     )
-    @SearchTag("magma cube, tarantula, tara, spider, slayer, quest")
+    @SearchTag("magma cube tarantula tara spider slayer quest")
     @ConfigEditorSlider(minValue = 0f, maxValue = 100f, minStep = 1f)
     var hideIrrelevantMobsOpacity: Int = 40
 }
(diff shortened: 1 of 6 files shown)


src/main/java/at/hannibal2/skyhanni/config/features/slayer/SlayerConfig.kt
```diff
@@ -100,4 +101,24 @@ class SlayerConfig {
...
+    @ConfigEditorBoolean
+    @FeatureToggle
+    var hideIrrelevantMobs: Boolean = false
+
+    @Expose
+    @ConfigOption(
+        name = "Adjust Irrelevant Opacity",
+        desc = "Adjust the opacity of irrelevant mobs. (in %)",
+    )
+    @SearchTag("magma cube, tarantula, tara, spider, slayer, quest")
```
NopoTheGamer: remove commas

src/main/java/at/hannibal2/skyhanni/config/features/slayer/SlayerConfig.kt
```diff
@@ -100,4 +101,24 @@ class SlayerConfig {
...
     @FeatureToggle
     var questWarningTitle: Boolean = true
+
+    @Expose
+    @ConfigOption(
+        name = "Hide Irrelevant Mobs",
+        desc = "Makes mobs partially transparent so that they dont annoy while having an active slayer quest. " +
+            "Useful for e.g. Magma Cubes in Burning Desert for Tara Slayer.",
+    )
+    @SearchTag("tarantula, spider opacity")
```
NopoTheGamer: remove commas

NopoTheGamer (changes requested):

src/main/java/at/hannibal2/skyhanni/data/entity/EntityOpacityManager.kt
```diff
@@ -0,0 +1,82 @@
...
+    fun onPostRender(event: SkyHanniRenderEntityEvent.Post<EntityLivingBase>) {
+        if (!active) return
+        if (!canChangeOpacity(event.entity)) return
+
+        GlStateManager.color(1f, 1f, 1f, 1f)
+        GlStateManager.disableBlend()
+    }
+
+    @HandleEvent
+    fun onRenderPlayer(event: EntityRenderLayersEvent.Pre<EntityLivingBase>) {
```
CalMWolfs: is this meant to only be player entities, if so then the stuff in the < > needs to change

src/main/java/at/hannibal2/skyhanni/events/entity/EntityOpacityActiveEvent.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import at.hannibal2.skyhanni.api.event.SkyHanniEvent
+
+/**
+ * Fires once per second, to enable the [EntityOpacityEvent].
+ */
+class EntityOpacityActiveEvent : SkyHanniEvent() {
+    private var status = false
+
+    fun setActive(status: Boolean = true) {
```
NopoTheGamer, line 11: this shouldnt be allowed to set to false

NopoTheGamer (changes requested):

src/main/java/at/hannibal2/skyhanni/events/entity/EntityOpacityActiveEvent.kt
hannibal002: not sure what you mean. the idea is to pass a boolean to that function that might be false, in that case there should nothing happen. instead of that the caller site needs to wrap this function call in a if check

src/main/java/at/hannibal2/skyhanni/data/entity/EntityOpacityManager.kt
hannibal002: no, this is intentionally set to entity living base, as this is the deepest this function allows. we want to make more stuff ttransparent than just players :)

CalMWolfs: If that's the case then I'm just confused by the function name as it mentioned players.

hannibal002: oh. yeah i removed the word "player" from the other two functions, but i forgot for this one. thanks!

PR #4472, opened 2025-07-23
Backend: chroma color!

hannibal002:
## What
used chroma color a lot more

+ Decluttered code by using chroma color in more render functions. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/mob/MobDebug.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/mob/MobDebug.kt
@@ -15,6 +15,7 @@ import at.hannibal2.skyhanni.utils.compat.MinecraftCompat
 import at.hannibal2.skyhanni.utils.render.WorldRenderUtils.drawFilledBoundingBox
 import at.hannibal2.skyhanni.utils.render.WorldRenderUtils.drawString
 import at.hannibal2.skyhanni.utils.render.WorldRenderUtils.expandBlock
+import io.github.notenoughupdates.moulconfig.ChromaColour
 import net.minecraft.client.entity.EntityPlayerSP
 
 @SkyHanniModule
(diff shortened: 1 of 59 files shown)


src/main/java/at/hannibal2/skyhanni/data/mob/MobDebug.kt
```diff
@@ -54,19 +55,25 @@ object MobDebug {
         }
 
         if (config.skyblockMob.isHighlight()) {
-            MobData.skyblockMobs.highlight(event) { if (it.mobType == Mob.Type.BOSS) LorenzColor.DARK_GREEN else LorenzColor.GREEN }
+            MobData.skyblockMobs.highlight(event) { val lorenzColor: ChromaColour =
```
Thunderblade73: the lorenzColor val does not make sense here, directly returning it here is better imo. Also I think you wanted to save the two to Chroma Color calls here, by applying it at the end but didn't do that. (Also does not need a val btw., just a extra bracket.)

src/main/java/at/hannibal2/skyhanni/data/mob/MobDebug.kt
```diff
@@ -54,19 +55,25 @@ object MobDebug {
...
+                lorenzColor
+            }
         }
         if (config.displayNPC.isHighlight()) {
-            MobData.displayNpcs.highlight(event) { LorenzColor.RED }
+            MobData.displayNpcs.highlight(event) { LorenzColor.RED.toChromaColor() }
         }
         if (config.realPlayerHighlight) {
-            MobData.players.highlight(event) { if (it.baseEntity is EntityPlayerSP) LorenzColor.CHROMA else LorenzColor.BLUE }
+            MobData.players.highlight(event) { val color =
```
Thunderblade73: same here, the color val does not make sense.

src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonMobManager.kt
```diff
@@ -133,20 +134,28 @@ object DungeonMobManager {
...
 
-    private fun getStarColor(): Color = starredConfig.color.get().toColor()
+    private fun getStarColor(): ChromaColour = starredConfig.color.get()
 
-    private fun handleStar0(mob: Mob, colour: Color?) {
+    private fun handleStar0(mob: Mob, colour: ChromaColour?) {
         if (mob.name == "Fels") {
             if (mob in felMoving) {
-                mob.highlight(colour)
+                colour?.let {
```
Thunderblade73, line 141: Yeah, why did you remove the null to remove inside Mob?

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonMobManager.kt
hannibal002: if we have two functions `highlight(ChromaColor)` and `highlight(LorenzColor)` or awt color instead of lorenz, and you now call `highlight(null)`, what function gets called?

src/main/java/at/hannibal2/skyhanni/data/mob/MobDebug.kt
hannibal002: thanks for the catch!

PR #4476, opened 2025-07-24
Improvement: Make /shedittracker say the name of the bucket

NopoTheGamer:
## What
It now says no tree selected for foragting tracker instead of bucket
image

## Changelog Improvements
+ Made /shedittracker more clear for bucketed trackers. - nopo


DavidArthurCole (approved):
should default to the enum class name, otherwise lgtm

NopoTheGamer:
> should default to the enum class name, otherwise lgtm

i think people should be forced to set it

CalMWolfs (approved):
i think forcing a name to be set is fine

PR #4427, opened 2025-07-18
Backend: Include Backup NEU Repo

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/4403

## What
Includes a backup version of NEU's repo that the RepoManager can use in the case the repo fails to download for any reason. Will use `assets/notenoughupdates` path if loaded, otherwise SkyHanni's path.

## Changelog Technical Details
+ Added task to include backup NEU repo in jar. - Daveed

PR #4483, opened 2025-07-24
Fix: Even More Repo Issues

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/4427

## What
This PR is notably aimed at fixing issues where the user fully cannot connect to GitHub to fetch the SHA, so it is heavily reliant on the backup repos being correctly in place.

Made more errors transient, and added more logging in the case we catch something going wrong. There's also now somewhat 'real' state management around the Fetch & Unpack process, as it'll give a qualified answer as to what level of success it had.

## Changelog Fixes
+ Fixed repo errors when GitHub is inaccessible. - Daveed


hannibal002:
so far so good (have only broke the link as seen in https://github.com/hannibal002/SkyHanni/pull/4427#issuecomment-3131513214)
image

hannibal002 (approved):
ive deleted both repo files (of sh+neu) and it still worked just fine.
image

PR #4486, opened 2025-07-24
Fix + Improvement: Fix shordered description and add help message when /sho is run with no arguments

SuperCl4sh:
## What
Title.

## Changelog Improvements
+ Added help message when /sho is run with no arguments. - SuperClash

## Changelog Fixes
+ Fixed shordered description in shcommands. - SuperClash


NopoTheGamer (changes requested):
I would much prefer if you could work out some way to just display what /shcommands shows when you run without args

hannibal002:
if the what text is not necessary, you can omit it entirely
image

hannibal002 (changes requested):
please mark it as ready if you think we can review and merge it
image

hannibal002:
stale

PR #4487, opened 2025-07-25
Backend: Abstract Bucket Tracker + Resettable

DavidArthurCole:
## What
The only thing TrackerData was providing was a `reset` function - everything that was using it could be replaced with a data class extending `Resettable`.
Also took this opportunity to make `SkyHanniBucketedTracker` abstract, so its unused overrides can be `final`. This should make it effectively impossible to use the wrong method.

I think this would be a good thing to do to all trackers long-term, as all trackers by virtue of needing to function, have to be Objects, and storing a separate `tracker` val inside those makes no sense - but for the time being, I've just made this change to the bucketed ones.

+ Made `Resettable` an interface instead of an abstract class. - Daveed
+ Replaced `TrackerData` with `Resettable`. - Daveed
    * Trackers should now use data classes that extend Resettable.


hannibal002:
This PR does two things that are independent of each other.

Replace tracker data with resettable: really good, i like the change a lot!
Merging the local internal tracker instance with the sh module/tracker class: interesting change, should work almost everywhere, except Slayer.

hannibal002 (changes requested):
changes do look fine, just one bug ive noticed:


image

src/main/java/at/hannibal2/skyhanni/features/mining/fossilexcavator/ExcavatorProfitTracker.kt
```diff
@@ -218,11 +201,9 @@ object ExcavatorProfitTracker {
     private fun shouldShowDisplay(): Boolean {
         if (!config.enabled) return false
         if (!isEnabled()) return false
-        val inChest = Minecraft.getMinecraft().currentScreen is GuiChest
+        if (Minecraft.getMinecraft().currentScreen !is GuiChest) return false
         // Only show in excavation menu
-        if (inChest && !FossilExcavatorApi.inExcavatorMenu) return false
-
-        return true
+        return FossilExcavatorApi.inExcavatorMenu
```
NopoTheGamer: this is an incorrect change?

NopoTheGamer (changes requested):

hannibal002:
ok the same bug is also in the current beta already

hannibal002:
fixed: https://github.com/hannibal002/SkyHanni/commit/2672544eb11ae004f4d22ccc92d6ba033a0db945

src/main/java/at/hannibal2/skyhanni/features/mining/fossilexcavator/ExcavatorProfitTracker.kt
hannibal002: yes, this change here was wrong.
the wording of the todo might be confusing in the original code. the goal of the function is to show while not in an inventory or while in the scavenger inventory. the new code here can be fixed by changing line 204 to return true

hannibal002 (changes requested):
nopos comment

DavidArthurCole:
> nopos comment

Resolved.

PR #4492, opened 2025-07-25
Backend: Small CurrentPing refactor

lunaynx:
## What
- Added a safety measure to prevent outside callers from mutating `previousPings` on 1.8.
- Switched to a shorter and cleaner way to build the list on 1.21.

exclude_from_changelog


hannibal002:
Is the underline in the name necessary?

lunaynx:
That's the private property, the underline is necessary to not shadow the public one. It could be done in another way, but this is pretty standard convention.

lunaynx:
I could change read-only internal accesses to use the non-underlined version I guess, but arguably consistent usage is cleaner and less confusing.

CalMWolfs (approved):

PR #4497, opened 2025-07-26
Backend: External version file

Thunderblade73:
## What
Adds an external file that holds the current mod version. This allows changing the version without a full invalidation of gradles  configuration cache.

The weird looking get for the version is so that it just a provider which resolve only during execution instead of configuration, needed to make it work with the configuration-cache.

Also added a ``buildTools`` to hide away gradle stuff we need in the future, starting with the version file. As our root directory is already quiet messy.

+ Moved Sh Project Version from Gradle to external file. - Thunderblade73


CalMWolfs (changes requested):
building jars results in their name being like this

image

in game also like this

image

CalMWolfs (approved):
worked for me but will wait for hannibal before merging

PR #4518, opened 2025-07-29
Feature + Fix: Slayer Spawn Timer and Full Slayer Quest timer messages

Fazfoxy:
## What
Adds Slayer Kill Time and Slayer Kill Time

the "Slayer Spawn & Kill Combined Timer" name is ehhh for notes and I'm not fully sure on a better one

## Changelog New Features
+ Added Slayer Time messages. - fazfoxy
    * Sends time to complete (Spawn & Kill) a slayer quest in chat.
+ Added Slayer Kill Timer. - fazfoxy
    * Sends time to kill a slayer in chat.

## Changelog Fixes
+ Fixed wrong color for Tara T5 Name in Damage Indicator. - hannibal2

PR #4443, opened 2025-07-20
Improvement + Fix + Backend: Chat config revamp and filter fixes

NotRoboo:
## What
- Removed the "others" chat filter toggle, which clearly snowballed (see images)...
- Separated the "others" filter into their own toggles
- Added tabs for "Chat Filters", "Compact Chat", "Player Messages", "Rare Drop Messages" and "Sound Responses" (In ABC order)
- Chat Filters tab -> Put filters into categories (Also in ABC order)
- Fixes a few toggles that did not work/had outdated colors e.g. Friend Join/Leave
- Backend changed ChatFilters.kt to sort filters by Category then ABC order
- Change the "Others" feature that filtered out certain slayer RNGs to be its own customizable toggle to pick & choose what you want filtered
- Probably some misc stuff I'm forgetting

Note- I wasn't planning on doing this and to be honest did this on a 3 day bender last week so it may not be the best or even get used, some categories may need to be changed/toggle names changed/descriptions changed (categories were made off "Others" tab features and very little thought was put into it) but I've been using it for ≈ 5 days and figured I'd PR it rather than constantly fixing merges to use it. 

<details>
<summary>Images</summary>

Screenshot 2025-07-18 055056

Screenshot 2025-07-18 055137

Screenshot 2025-07-18 055231

Screenshot 2025-07-18 055255

Screenshot 2025-07-18 055309

Screenshot 2025-07-18 055319

Screenshot 2025-07-18 055346

</details>

## Changelog Improvements
+ Improved "Chat" config tab. - Roboo
    * This removes the "Others" tab so said features will have their own toggles now.

## Changelog Fixes
+ Fixed a handful of chat filters not working. - Roboo

PR #4520, opened 2025-07-29
Improvement: Attempt to Split "Other" Message Hiders

Fazfoxy:
## What
Attempted to split the "other" segment, created a subset for slayers, maybe should be one for Mining as well? I haven't bothered with repo patterns here but it SHOULD be less awful now slightly?

Also, I kind of? Don't want to config fix this into all the new options as it used to be FAR too broad and might not be obvious to users that they can now do more granular control?

## Changelog Improvements
+ Split Other Message Hider into multiple different settings. - fazfoxy

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/chat/ChatConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/chat/ChatConfig.kt
@@ -195,6 +195,7 @@ class ChatConfig {
     @FeatureToggle
     var hideLottery: Boolean = true
 
+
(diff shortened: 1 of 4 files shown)


Fazfoxy:
This, WAS gonna be a smaller addon to #4518 so that the slayer messages could be hidden seperately but then I realised my pure hatred for the Other config toggle.

MTOnline69:
Might want to have a look at #4443

Fazfoxy:
> Might want to have a look at #4443

goddamn I'm a week late

PR #4521, opened 2025-07-29
Fix: Slayer dropped books having both a Hypixel and SH message

Fazfoxy:
## What
Adds a repo pattern for the names of Slayer books to avoid their delayed sending triggering the custom book message

## Changelog Fixes
+ Fixed Slayer books showing both Hypixel and SkyHanni messages. - fazfoxy


hannibal002 (changes requested):
can we just use the check if the name is not "enchatned book" isntead of checking against all those enchantment names?

Fazfoxy:
> can we just use the check if the name is not "enchatned book" isntead of checking against all those enchantment names?

I feel like I'm missing whatever you mean by this, they all have item names already assigned and these aren't different?

PR #4522, opened 2025-07-29
Fix: lack of decimals in Kill timer

Fazfoxy:
## What
Adds back the decimals to TimeToKill (it's kinda pointless being accurate if it only counts in seconds)

## Changelog Fixes
+ Fixed lack of decimal in Slayer Kill Timer. - fazfoxy


Fazfoxy:
how do I do this to myself

PR #4526, opened 2025-07-30
Add "paying" text to price per copper

ptlthg:
## What
This adds the word "paying" in the price per copper thing on hover in a visitor.

The current text of just saying `(3k per)` does not at all communicate *what* this number is to new players. `(paying 3k per)` makes it more clear what this means, it's that by accepting the offer you'd essentially be "paying" 3k coins per copper.

The lack of clarity here has bothered me before, and I just got a new player to the game to tell me what the (3k per) meant in the menu and they had no idea, so I think this helps. 

So instead of (15k per), this will say (paying 15k per)
image

**Notes**
- I just did this edit within GitHub, I don't want to add a util method to get the text please don't make me
- I don't think a "compact" option for this is worth it, it's literally one word
- If you come up with better terminology that's fine too, I just think this should communicate what the number actually means

exclude_from_changelog


Chissl:
I don't know about others, but I prefer the old version. Please add an toggle option to return the compact version.

ptlthg:
> I don't know about others, but I prefer the old version. Please add an toggle option to return the compact version.

`/shwords` `(paying` `(`

You're welcome

Chissl:
> > I don't know about others, but I prefer the old version. Please add an toggle option to return the compact version.
> 
> `/shwords` `(paying` `(`
> 
> You're welcome

:(

ptlthg:
If someone wants to add a toggle to this they can, I just generally think this UI element should communicate what it means better, and I am not going to put more effort into this

PR #4528, opened 2025-07-30
Fix: Broodmother chat message

hannibal002:
## Changelog Fixes
+ Fixed Broodmother spawn message showing even when disabled. - hannibal2


hannibal002:
apparently this breaks the logic entirely idk

PR #4532, opened 2025-07-31
Fix: Actual self boss check in SlayerTimeMessages

Fazfoxy:
## What
Last check fully just always returned null, this is much more definitive
(Excluded from changelog since there's already a changelog about this being fixed from an older broken PR in this beta)

exclude_from_changelog


Fazfoxy:
This creates a fucking race condition because of Hypixel's delay between sending the message that SlayerQuestCompleteEvent triggers based on and Damage Indicator Death Event, Fuck.

Fazfoxy:
Bug is outright fake, I killed a boss too fast for name checks, this is edgecasy af

PR #4373, opened 2025-07-14
Fix: Berberis Highlight on 1.21

lunaynx:
## What
Fixed farmland check for Berberis Highlight being broken on 1.21 due to a height difference, and some miscellaneous cleanup.

I didn't touch `LorenzVec.getBlockBelowPlayer()` due to the risk of potentially breaking other features relying on it. Can be reevaluated later.

## Changelog Fixes
+ Fixed Berberis Highlight not working on 1.21 if "Only on Farmland" is enabled. - Luna

PR #4535, opened 2025-07-31
Fix: Wilted Berberis Helper ghost waypoints

lunaynx:
## What
Reverted a premature optimization in #4373 that changed a list to a set, breaking identity checks due to the items being data classes with mutable properties, often causing a ghost waypoint to stick around after the last Wilted Berberis in a plot is harvested.

## Changelog Fixes
+ Fixed Wilted Berberis Helper occasionally not removing old waypoints. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/rift/area/dreadfarm/RiftWiltedBerberisHelper.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/rift/area/dreadfarm/RiftWiltedBerberisHelper.kt
@@ -33,7 +33,9 @@ object RiftWiltedBerberisHelper {
     private val config get() = RiftApi.config.area.dreadfarm.wiltedBerberis
 
     private val berberisSounds = setOf("mob.horse.donkey.death", "mob.horse.donkey.hit")
-    private val list = mutableSetOf<WiltedBerberis>()
+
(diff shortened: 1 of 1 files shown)


hannibal002 (changes requested):
Set and List have the same identity checks. This has no imipact on data classes or mutable properites.
It's rather the other way around: a Set does not contain the same "identity check" object twice, changing this behavior to a list either does nothing or allow for even more objects to be stored in the collection object then before.

Old waypoints that happens to be idenitcal to new objects might get fixed by switching from list to set, but there is nothing getting removed additonally by swithing from set to list!

This will not fix what you said you hope it does. What is the actual probelm? can we add data to debug event for further testing?

lunaynx:
I was surprised too, but it very much _does_ fix the problem. I think it doesn't happen on 1.21 for whatever reason, but you can easily test before and after on 1.8.

See: https://stackoverflow.com/a/49646322 - basically this is undefined behavior and breaks on Java 8, but happens to work on Java 21 (we should probably not rely on it working even once we drop 1.8 support).

Either way, it was a premature optimization - there is essentially zero memory or performance benefit to using a set here, since added items will never be deduplicated due to their nature.

hannibal002 (approved):
thats a crazy specific logic here, wow

PR #4539, opened 2025-08-02
Improvement: less lags

hannibal002:
## What
made island graph data load and storage data async to avoid small lags during island change and when opening a chest in private island

## Changelog Improvements
+ Slightly improved the mod performance when switching islands or opening private island chests. - hannibal2


hannibal002:
tested. worked for me

PR #4547, opened 2025-08-03
Backend: cleanup one call in ItemUtils

hannibal002:
+ Made one small cleanup call in ItemUtils. - hannibal2


hannibal002:
no longer necessary as those lines got reverted by 740182bf38772861323a89dadb4446011e0b5921

PR #4551, opened 2025-08-03
Backend: Fix RenderableTestSuit Command

Thunderblade73:
## What
Not tested but should work.

+ Fixed Renderable Test Command. - Thunderblade73


hannibal002 (changes requested):
build failed

PR #4554, opened 2025-08-04
Fix: Slayer Tier 5 Mob Detection

Thunderblade73:
## What
Hardcodes the tier 5 names into the regex. 

## Note
The change to the regex does break older versions. It will throw a NullPointerException. That error should be suppressed immediately via repo.

## Changelog Fixes
+ Fixed Tier 5 Slayer compatibility for related features. - Thunderblade73


src/main/java/at/hannibal2/skyhanni/data/mob/MobFactories.kt
```diff
@@ -21,7 +21,7 @@ object MobFactories {
                 armorStand = armorStand,
                 name = this.group("name"),
                 additionalEntities = extraEntityList,
-                levelOrTier = this.group("tier").romanToDecimal()
+                levelOrTier = this.group("tier")?.takeIf { it.isNotEmpty() }?.romanToDecimal() ?: 5
```
hannibal002: `isNotEmpty` should not do anything. we alr catch `[IV]+`. the + means we catch one or more. so the string can never be empty

hannibal002 (changes requested):

PR #4556, opened 2025-08-05
Backend: Something about air idk

CalMWolfs:
## What
I was told this is better idk.

+ Slightly changed how we check vanilla item. - CalMWolfs


NopoTheGamer (changes requested):
This doesnt fix the issue, it just hides the error

PR #4560, opened 2025-08-05
Backend: Moved isAprilFoolsDay out of SkyHanniDebugsAndTests.kt

Thunderblade73:
## What
Move isApirlFoolsDay out of SkyHanniDebugAndTest, Which was the last instance that it was used from another file.
This allows us to ban that file from ever being imported since this file reference a lot of other files that are only referenced here. (Actually the whole test package should be banned to be imported, but that needs more changes)

+ Moved isAprilFoolsDay into TimeUtils. - Thunderblade73


CalMWolfs (changes requested):
I guess i only put this on discord and not on github but I agree that SkyHanniDebugsAndTests is not the correct place for this utils function but I do not think random functions like this should be in SkyHanniMod, maybe TimeUtils would be more appropriate or some other class

hannibal002 (changes requested):
>  I do not think random functions like this should be in SkyHanniMod

i agree, time utils sounds better.

@Thunderblade73

hannibal002 (approved):

PR #4562, opened 2025-08-05
Fix: Pest features not working

NopoTheGamer:
## What
I fabricated 2 regex tests and did test one of the patterns

## Changelog Fixes
+ Fixed Pest Spawn Timer. - nopo


Chissl:
Doesn't work, hypixel removed pluralization for some reason

Chissl:
§6§lYUCK! §25 §2ൠ Pest §7have spawned in §aPlot §7- §b8§7!

lunaynx:
Maybe make the patterns `Pests?` for futureproofing.

(I know we can just change it later via repo, but that doesn't really bloat the patterns much, and it sounds like a thing they may change.)

PR #4577, opened 2025-08-06
Fix: Pet Utils NPE

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1000669238035497022/1402693717130874921
I have no idea how someone managed to hit an NPE inside of what on the surface appears to be a fully safe stack? But alas, they did.. So, this attempts to fix that, without any ability to test if it did...

exclude_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/NeuInternalName.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/NeuInternalName.kt
@@ -102,8 +102,7 @@ class NeuInternalName private constructor(private val internalName: String) {
 
     val isPet: Boolean
         get() = petCache.getOrPut(this) {
-            val isPetCategory = (getItemStackOrNull()?.getItemCategoryOrNull() == ItemCategory.PET)
-            PetUtils.isKnownPetInternalName(this) || isPetCategory
+            PetUtils.isKnownPetInternalName(this) || (getItemStackOrNull()?.getItemCategoryOrNull() == ItemCategory.PET)
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/utils/NeuInternalName.kt
```diff
@@ -102,7 +102,8 @@ class NeuInternalName private constructor(private val internalName: String) {
 
     val isPet: Boolean
         get() = petCache.getOrPut(this) {
-            PetUtils.isKnownPetInternalName(this) || this.getItemStackOrNull()?.getItemCategoryOrNull() == ItemCategory.PET
+            val isPetCategory = (getItemStackOrNull()?.getItemCategoryOrNull() == ItemCategory.PET)
+            PetUtils.isKnownPetInternalName(this) || isPetCategory
```
CalMWolfs: maybe im reading this change wrong but it seems now we are checking category on some item stacks for no reason as before we only did it if it wasnt a known pet internal name but now we do it on everything regardless but only use the result of this boolean if it isnt a known pet internal name

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/NeuInternalName.kt
DavidArthurCole: Yeah you're right - I think that was just me being tired. Fixed.

CalMWolfs (approved):

PR #4582, opened 2025-08-07
Fix: Inaccurate Contest Stats

Chissl:
## What
Fixes an issue where contest stats reported players getting impossible block break counts, caused by stats not getting cleared after every farming contest. Untested as of making this pr but should work. 

## Changelog Fixes
+ Fixed incorrect stats summary for Jacob's Contests. - Chissl

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/contest/JacobContestStatsSummary.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/contest/JacobContestStatsSummary.kt
@@ -42,6 +42,7 @@ object JacobContestStatsSummary {
 
         when (event.phase) {
             FarmingContestPhase.START -> {
+                blocksBroken.clear()
                 ChatUtils.chat("Started tracking your Jacob Contest Blocks Per Second!")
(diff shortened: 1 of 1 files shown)


CalMWolfs (changes requested):
im pretty sure this was made into a map so that it that it could track more than 1 contests data throughout the contest but @DavidArthurCole will have to confirm that, maybe the proper solution is putting this `.clear()` in the Stop part of the when branch

DavidArthurCole:
> im pretty sure this was made into a map so that it that it could track more than 1 contests data throughout the contest but @DavidArthurCole will have to confirm that

It was, yeah - it was a flat Int before ([link](https://github.com/Chissl/SkyHanni/commit/37c7e2509891d04de967d888e881f35eb410f65d#diff-3bceb666e3c4cb3f9356cc2defaa2e5eeffbaa8d3da4d68584c31a28b6a04a3bL19)), ultimately meaning if you switch crop types during the event, SH would "drop" all blocks you had broken previously.

Chissl:
Okay, can confirm this works as intended

PR #4588, opened 2025-08-10
Fix: Tier 5 slayer detection errors

lunaynx:
## What
Fixed issues detecting Atoned Horror and Conjoined Brood, and renamed a method in SlayerType that was unnecessarily "clazz"-ified.

## Changelog Fixes
+ Fixed errors with Atoned Horror (Revenant) and Conjoined Brood (Tarantula) Slayer bosses. - Luna

+ Renamed `SlayerType.getByClazzName()` to `SlayerType.getByClassName()`. - Luna


hannibal002:
Isn't the fix to remove the mob tag from the slayer name?

lunaynx:
No? The `contains` check already accounts for that, it just doesn't check for the special name for Tarantula T5 second phase & Revenant T5. Also, that is just the indicator that it is a slayer boss, not a mob type.

Thunderblade73:
> Isn't the fix to remove the mob tag from the slayer name?

That also needs to be done but doesn't effect the changes here. Since the contains accepts it either way.

CalMWolfs (approved):

PR #4604, opened 2025-08-17
Backend: Changed part of the foundation of the Damage Indicator to utilize the Mob Detection

Thunderblade73:
## What
Made it so that the identification of a entity in the Damage Indicator now runs after the Mob Detection run. So that it can use the data from there instead of doing expensive EntityUtils calls (eg; getNameTagWith). Also reduce the time it is run, as it only runs once on an entity that spawned instead of running every tick on every entity.

I only tested it on the crimson minibosses and it worked fine there. But the rest is hard for me to test since I do not really have a account anymore to do high level stuff (or anything really).

It does not use the Mob Detection to it's full extend but it least it is a start so someone can do stuff there more easily.

+ Changed part of the foundation of the Damage Indicator to utilize the Mob Detection. - Thunderblade73


hannibal002 (changes requested):
does not work for tarantula 5

image

hannibal002 (changes requested):
did a bit of testeing, and so far only the slayer bosses fail to get detected. the second phase bonzo is likely a unrelated bug.
you can easily test yourself with rev t1

image

Thunderblade73:
> Owner

slayer should be fixed. Idk about tier 5 though, I can't test those.

And I look into what is up with bonzo.

hannibal002 (approved):
works for slayers (tested with tara5)

PR #4610, opened 2025-08-20
Fix: Enchanted Book Detection for Corpses and Excavator

mariethepolarfox:
## What
Fixes Enchanted Books not getting detected correctly for Corpses and the Fossil Excavator

## Changelog Fixes
+ Fixed Enchanted Books detection for Corpses and Fossil Excavator. - Siv

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/mining/crystalhollows/CrystalNucleusApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/mining/crystalhollows/CrystalNucleusApi.kt
@@ -11,10 +11,8 @@ import at.hannibal2.skyhanni.events.mining.CrystalNucleusLootEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.ItemUtils
 import at.hannibal2.skyhanni.utils.NeuInternalName
-import at.hannibal2.skyhanni.utils.NeuInternalName.Companion.NONE
 import at.hannibal2.skyhanni.utils.NeuInternalName.Companion.fromItemNameOrNull
 import at.hannibal2.skyhanni.utils.NeuInternalName.Companion.toInternalName
-import at.hannibal2.skyhanni.utils.RegexUtils.matchMatcher
 import at.hannibal2.skyhanni.utils.RegexUtils.matches
 import at.hannibal2.skyhanni.utils.SkyBlockItemModifierUtils.getHypixelEnchantments
 import at.hannibal2.skyhanni.utils.collection.CollectionUtils.addOrPut
--- a/src/main/java/at/hannibal2/skyhanni/features/mining/crystalhollows/CrystalNucleusApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/mining/crystalhollows/CrystalNucleusApi.kt
@@ -11,6 +11,7 @@ import at.hannibal2.skyhanni.events.mining.CrystalNucleusLootEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.ItemUtils
 import at.hannibal2.skyhanni.utils.NeuInternalName
+import at.hannibal2.skyhanni.utils.NeuInternalName.Companion.NONE
 import at.hannibal2.skyhanni.utils.NeuInternalName.Companion.fromItemNameOrNull
 import at.hannibal2.skyhanni.utils.NeuInternalName.Companion.toInternalName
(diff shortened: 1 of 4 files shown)


src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/CorpseApi.kt
```diff
@@ -37,6 +37,14 @@ object CorpseApi {
      */
     private val itemPattern by chatPatternGroup.pattern("item", " {4}§r(?<item>.+)")
 
+    /**
+     * REGEX-TEST:     §r§fEnchanted Book (Ice Cold I§r§f)
+     */
+    private val enchantedBookPattern by chatPatternGroup.pattern(
+        "enchantedBook",
+        " {4}§r§fEnchanted Book \\((?<item>.+)§r§f\\)"
```
hannibal002: we already have a similar regex at https://github.com/hannibal002/SkyHanni/blob/47b43b02cdd337200d72b5d7583d3d8559928757/src/main/java/at/hannibal2/skyhanni/features/mining/crystalhollows/CrystalNucleusApi.kt#L48.

maybe we can merge them together?

src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/CorpseApi.kt
```diff
@@ -74,10 +82,12 @@ object CorpseApi {
...
         } ?: return
-        // Workaround: If it is an enchanted book, we assume it is a paleontologist I book
-        if (pair.first.let { it == "§fEnchanted" || it == "§fEnchanted Book" }) {
-//             pair = "Paleontologist I" to pair.second
-            pair = "§9Ice Cold I" to pair.second
+
+        if (pair.first.startsWith("§fEnchanted Book (")) {
+            val book = enchantedBookPattern.matchMatcher(message) {
+                group("item") ?: "Ice Cold I"
+            } ?: return
```
hannibal002: we only return when the name contains enchatned book (string comparison) but not maches the regex. is this supposed to ever happen? 
if not, then we can move line 90 in the matchMatchehr block and remove the if + startsWith line entirely.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/CorpseApi.kt
mariethepolarfox: I think with the way I did it now this is still needed?

src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/CorpseApi.kt
mariethepolarfox: Merged them into ItemUtils

src/main/java/at/hannibal2/skyhanni/features/mining/crystalhollows/CrystalNucleusApi.kt
```diff
@@ -137,12 +131,10 @@ object CrystalNucleusApi {
...
-            val bookType = bookTypePattern.matchMatcher(itemName) {
-                group("type").lowercase()
-            }
+            val bookType = ItemUtils.readBookType(itemName)
             return when (bookType) {
-                "lapidary" -> LAPIDARY_I_BOOK_ITEM to 1
-                "fortune" -> FORTUNE_IV_BOOK_ITEM to 1
+                "Lapidary I" -> LAPIDARY_I_BOOK_ITEM to 1
+                "Fortune IV" -> FORTUNE_IV_BOOK_ITEM to 1
                 // Fallback to inventory based system
```
mariethepolarfox, line 136: I think this is no longer necessary as hypixel seems to have unified it everywhere now, should I remove the fallback and switch to just parsing the Internal name directly?

src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/CorpseApi.kt
hannibal002: no it is not needed. see https://github.com/hannibal002/SkyHanni/pull/4610/commits/dd7b9a0545f3355e0f8547b5f1b0014506fa127a

src/main/java/at/hannibal2/skyhanni/features/mining/crystalhollows/CrystalNucleusApi.kt
hannibal002: lets do this in another pr

PR #4611, opened 2025-08-21
Feature: Secret Tracker Waypoints

Bloxigus:
## What
Adds a waypoint that shows up when you use the Secret Tracker 3000.
If hypixel fixes it being useless at finding the last secret in some rooms, could be quite useful.

I havent tested on 1.21+ but what could _possibly_ change about this between versions.

## Changelog New Features
+ Added a guess for the Secret Tracker 3000 while in dungeons. - bloxigus


CalMWolfs (approved):
worked well enough for me

PR #4612, opened 2025-08-21
Fix: Cancel Vacuum uses when particles are still being received

Bloxigus:
## What

## Changelog Fixes
+ Prevented use of vacuums when a previous vacuum line is still being traced out. - bloxigus


hannibal002 (changes requested):
image

PR #4613, opened 2025-08-22
Fix: Sticky keybinds being reset while hovering a tracker

NopoTheGamer:
## What
We dont cancel a sticky keybind anymore in text editor or graph editor, i dont think this is a issue at all because its just sneak and sprint

## Changelog Fixes
+ Fixed Toggle Sneak unsneaking when hovering over a SkyHanni Tracker. - nopo


hannibal002 (approved):
works for me in 1.21.5

PR #4622, opened 2025-08-26
Improvement: Dmg Indicator separation

hannibal002:
## What
started moving all sub options of dmg indicator (except hide nametag + dmg over time) out of dmg indicator dependency config wise. this allows leaving the dmg indicator visuals disabled and just enable single config options, like slayer or bacte, that will show at the mob position without showing health + name of dmg indicator.

the next steps are to move the config options into their correct config area (slayer config or rift, etc)
also the long `getCustomHealth` function and all its sub functions should be translated into an event that then gets listened in on their respective classes, making the `DamageIndicatorManager` file way shorter.
Technically, those features will still be using the backend of the dmg indicator to work, but they will finally be independent of the dmg indicator visual toggle, as recently suggested here https://discord.com/channels/997079228510117908/1406721645540016148

<details>
<summary>Images</summary>

image
image

image

</details>

## Changelog Improvements
+ Made many Slayer features, Shuriken/Twilight indicators, and Bacte phase display independent of Damage Indicator. - hannibal2
    * Allows Bacte phase, Eman Slayer blaze timer, Slayer time-to-kill, and Vampire Slayer steak display to show without boss name/health visible.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/combat/damageindicator/DamageIndicatorConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/combat/damageindicator/DamageIndicatorConfig.kt
@@ -10,15 +10,10 @@ import io.github.notenoughupdates.moulconfig.annotations.ConfigOption
 
 class DamageIndicatorConfig {
     @Expose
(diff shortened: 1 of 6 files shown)


hannibal002:
i just tested, when getCustomHealth  and all its sub functions gets extracted into other classes, then DamageIndicatorManager loses half its size
image

PR #4603, opened 2025-08-17
Improvement: Additions to Crown of Avarice Tracker

Tryp0xd:
## Dependencies
- #4626 

## What
Changes how session time works (SimpleTimeMark? -> Stopwatch) to support new feature of pausing the session time (instead of resetting the stuff after 2 minutes). (pls help if the way i do it is absolute garbage and better ways can be done)
Adds Session Time option to see how long your session has been running for (duh). 
You can now toggle whether you want the session to reset when you go to different worlds.

## Changelog Improvements
+ Added Session Time and Reset on World Change toggle to Crown of Avarice Tracker. - Tryp0xd
    * Session Time now pauses after 2 minutes AFK instead of resetting.

PR #4626, opened 2025-08-27
Backend: Stopwatch Util

Chissl:
## What
Adds a simple stopwatch class. Can start, pause, and reset a duration. Has a type adapter for json saving..  
Inspired by #4603 and a couple other features I'm working on that require a similar thing. 

+ Added a stopwatch util class. - Chissl


Thunderblade73 (commented):
Could you add a TypeAdapter to gson so that this can be directly saved in the config/storage? (Save/Load it as it where a duration, but set pause to true on load and have the pause logic run before save)

Chissl:
> Could you add a TypeAdapter to gson so that this can be directly saved in the config/storage? (Save/Load it as it where a duration, but set pause to true on load and have the pause logic run before save)

Done

PR #4631, opened 2025-08-30
Backend + Improvement:  multi-line dynamic text

rueblimaster:
## What
Adds a MultiLineDynamicText and converts drawDynamicText to use that to avoid code duplication.

## Changelog Improvements
+ Improved world text rendering for multi-line dynamic text. - rueblimaster

+ Created `SkyHanniRenderWorldEvent.drawMultiLineDynamicText`. - rueblimaster
    * Simplifies and ensures clear multi-line dynamic text.


rueblimaster:
this doesn't work yet for modern

NopoTheGamer:
so what does this do

rueblimaster:
> so what does this do

so basically instead of having to do multiple dynamic text displays for multiple lines you can now just use 1 and it'll work out of the box without you having to play around with some values and stuff

NopoTheGamer:
I dont think this is particularly useful, and factoring in the mod now supports rendering Components i think this pr adds unnecessary code duplication

PR #4636, opened 2025-09-02
Feature: Shopping List

rueblimaster:
## Dependencies

- #5244

## What
Adds the base frame for the general Shopping List. Only way to interact with it for now is by the `/shshoppinglist` command

### What is coming:

* Categories
* Persistent storage
* Recipes
* Buttons to add stuff to the list

- #4640 
- #4656 

<details>
<summary>Images</summary>

image

</details>

## Changelog New Features
+ Added generic Shopping List. - rueblimaster
    * Use via `/shshoppinglist`.


src/main/java/at/hannibal2/skyhanni/features/inventory/shoppinglist/ShoppingList.kt
```diff
@@ -0,0 +1,116 @@
...
+                }
+            }
+            literalCallback("clear") { clear() }
+        }
+    }
+
+    private fun buildDisplay() {
+        display = buildList {
+            items.forEach { addAll(it.buildDisplay()) }
+        }
```
Thunderblade73:
```suggestion
        display = items.flatMap(ShoppingListItem::buildDisplay)
```

src/main/java/at/hannibal2/skyhanni/features/inventory/shoppinglist/ShoppingList.kt
```diff
@@ -0,0 +1,116 @@
...
+import at.hannibal2.skyhanni.utils.RenderDisplayHelper
+import at.hannibal2.skyhanni.utils.RenderUtils.renderRenderables
+import at.hannibal2.skyhanni.utils.SkyBlockUtils
+import at.hannibal2.skyhanni.utils.renderables.Renderable
+
+@SkyHanniModule
+object ShoppingList {
+    val config = SkyHanniMod.feature.inventory.shoppingList
+
+    val items = mutableListOf<ShoppingListItem>()
```
Thunderblade73: Make this a ``mutableMapOf<NeuInternalName,ShoppingListItem>()`` as you request the internalName often (which with the current implementation runs in O(n) but the map runs in O(1) )

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/shoppinglist/ShoppingList.kt
```diff
@@ -0,0 +1,115 @@
...
+
+    val items = mutableMapOf<NeuInternalName, ShoppingListItem>()
+
+    private var display = listOf<Renderable>()
+
+    fun update() {
+        buildDisplay()
+    }
+
+    private fun getItemOrNull(internalName: NeuInternalName) = items.getOrDefault(internalName, null)
```
Thunderblade73: ``items[internalName]`` is the same thing. So first of all replace the getOrDefault and then inline this function.

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/shoppinglist/ShoppingListItem.kt
```diff
@@ -0,0 +1,22 @@
...
+import at.hannibal2.skyhanni.utils.NeuInternalName
+import at.hannibal2.skyhanni.utils.renderables.Renderable
+import at.hannibal2.skyhanni.utils.renderables.primitives.text
+import com.google.gson.annotations.Expose
+
+class ShoppingListItem(
+    @Expose
+    val internalName: NeuInternalName,
+    @Expose
+    var amount: Double,
```
hannibal002: why is this a double?

hannibal002 (changes requested):
Thanks for splitting the larger feature into smaller PRs, that makes reviewing much easier! A few things came up:

Context in the PR description: Since this is the base frame of a larger feature, it would really help reviewers to mention that in the description and link the follow-up PRs. Right now, it's hard to judge architectural decisions without knowing what's coming next.

BrigadierUtils change: The .map { it.replace(" ", "_") } change looks unrelated to the Shopping List. Could you move that into its own PR?

Config toggle: What does the enabled toggle control here? The feature is entirely command-driven, so it's unclear what disabling it would do. If it controls the overlay rendering, that makes sense, but then the command would still work while the toggle is off, which could be confusing. Worth clarifying what the intended UX is.

Persistence: The items map lives only in memory right now, so the list is lost on restart. Is that intentional for this base PR, with persistence coming in a follow-up? If so, mentioning that in the description would help.

Overall the structure looks clean and follows project patterns well (command registration, RenderDisplayHelper, config accordion). Just needs some clarification on the points above before merging.

src/main/java/at/hannibal2/skyhanni/features/inventory/shoppinglist/ShoppingListItem.kt
rueblimaster: because of item stacks and coins

hannibal002: well, item stacks is no reason to have double :P

but coin is fair, can you add a comment about this in the code, please

rueblimaster: sure

rueblimaster:
@hannibal002 
I addressed all your points

hannibal002 (approved):
code looks good for me.

please expand the changelog wording a bit to explain whats the point of this command, why users might wanna use this.

rueblimaster:
closing in favor of [Skyoceans Crafthelper](https://github.com/meowdding/SkyOcean/tree/main/src/main/kotlin/me/owdding/skyocean/features/recipe/crafthelper), as it already does what this should be doing but maybe not as good as I imagined this to be in the future.  Also I'm not too motivated rn for this and I think I'd rather improve the Skyocean one instead of this one.

PR #4638, opened 2025-09-02
Backend: Improve RenderDisplayHelper

rueblimaster:
## What
Improved the granularity of the RenderDisplayHelper allowing for more precise and cleaner rendering. Should now support all kinds of inventories one could wish to render inside.
Added new parameters to allow this to be turned on and off respectively.
Also added new constant ANY_INVENTORY for if it should be rendered in any inventory.

Added/Improved documentation of the render events.

+ Added and improved documentation of the render events. - rueblimaster
+ Improved RenderDisplayHelper. - rueblimaster
    * Added much more granular control over the situations to render in.


hannibal002 (changes requested):
The refactor introduces a regression for trackers in container inventories and adds infrastructure (inSign/inChat/inIngameMenu) that has no consumer yet.
I'd prefer to keep the current design and add granularity when a concrete feature actually needs it.

If you have a specific use case in mind, let's discuss that first.

rueblimaster:
@hannibal002 how about we keep only the comments?
or the comments and the infrastructure?

src/main/java/at/hannibal2/skyhanni/events/GuiRenderEvent.kt
```diff
@@ -7,14 +7,11 @@ open class GuiRenderEvent(context: GuiGraphics) : RenderingSkyHanniEvent(context
...
      * Renders only while inside an inventory
-     * This event does not render on signs.
-     * Use ScreenDrawnEvent instead.
-     * Also, ensure you do not render with this event while in a sign, as it will override ScreenDrawnEvent.
      */
     class ChestGuiOverlayRenderEvent(context: GuiGraphics) : GuiRenderEvent(context)
 
     /**
-     * Renders always, and while in an inventory it renders a bit darker, gray
+     * Renders always, and while in any inventoy etc. that grays out the background also gets grayed out
```
hannibal002: typo inventoy

src/main/java/at/hannibal2/skyhanni/events/render/gui/ScreenDrawnEvent.kt
```diff
@@ -4,4 +4,9 @@ import at.hannibal2.skyhanni.api.event.RenderingSkyHanniEvent
 import net.minecraft.client.gui.GuiGraphics
 import net.minecraft.client.gui.screens.Screen
 
+/**
+ * Renders only while inside an inventory, a sign, chat or the game menu
+ * While in an inventory it renders a bit darker, gray
+ * Use this for signs and chat
+ */
 class ScreenDrawnEvent(context: GuiGraphics, val gui: Screen?) : RenderingSkyHanniEvent(context)
```
hannibal002, line 11: The docs here are a bit confusing: the visual effect is caused by a different event (DrawBackgroundEvent), and the usage recommendation contradicts line one ("renders in inventories, signs, chat, game menu" vs "use this for signs and chat").
Keeping it to just "when does this event fire" would be better imo.

hannibal002 (changes requested):
We can keep the comments, then please review them all again.

And I don't see a point in shipping unused code here. Please revert those infrastructure changes.

rueblimaster:
closed in favor of #5404
as the RenderDisplaHelper stuff currently isn't really relevant

PR #4590, opened 2025-08-12
Improvement + Fix: Crop Milestones Displays and Backend

Chissl:
## What
This was originally a pretty small pr to add track pest/dicer drops to the crop milestone display. Kept running into issues which eventually compounded into completely redoing most of the crop milestones backend. Did some refactoring as well, feedback about this is appreciated, especially regarding the move of all garden trackers into their own subpackage. 

## Changelog Improvements
+ Added option to show Crop Milestones Display while not holding a farming tool. - Chissl
    * Will show last gained crop by default. 
    * Can set what crop to display by clicking on the tracker while in an inventory. 
+ Added option to set Custom Crop Milestone Goals in config. - Chissl
    * This was previously a command.
+ Mooshroom Cow Display now displays overflow and custom milestones. - Chissl
    * These values are the same as the main Crop Milestone display.
+ Improved overflow milestone calculation performance. - Chissl

## Changelog Fixes
+ Fixed crop milestones not properly tracking pest drops. - Chissl
+ Fixed overflow Crop Milestone level ups occasionally not triggering. - Chissl
+ Fixed overflow Crop Milestone level ups occasionally triggering when they shouldn't. - Chissl
+ Fixed mushrooms gained from Mooshroom Cow calculations assuming the pet was level 100. - Chissl
+ Fixed Best Crop Time display incorrectly telling users with all maxed milestones to farm crops. - Chissl
+ Fixed the wrong milestone incrementing when farming with the wrong tool. - Chissl
+ Fixed milestones not incrementing when farming without a cultivating tool. - Chissl
+ Fixed milestones incrementing too quickly when farming wheat. - Chissl

## Changelog Technical Details
+ Redid backend for Crop Milestones. - Chissl
    * Added caching for frequently calculated values.
    * Fixed some other performance issues and outdated code. 
+ Added crop collection tracking. - Chissl
    * Tracks collection gained from breaking crops, pests, dicer drops, and mooshroom cow.
    * Types that affect Crop Milestones are added to that tracker. 
    * Backend for future crop collection/profit tracker.
+ Added /shaddcropcollection, and /shresetcropmilestones debug commands. - Chissl
+ Added parity between addRenderableNullableButton and addNullableButton. - Chissl
+ Renamed cropCounter to cropMilestoneCounter in ProfileSpecificStorage. - Chissl
    * Planning to add different types of crop counters in the future.
+ Moved and renamed Garden-related files in the data package to sub-packages. - Chissl
[...]

PR #4639, opened 2025-09-02
Fix: Melon

Chissl:
## What
Fixes "no fortune found" tablist warnings, dicer drops (maybe), and CropType lookup caused by hypixel melon name change. Former 2 are regexes and can be updated via repo, while latter requires a beta release. Haven't tested if this actually fixes dicer drops but it works in #4590 so you should definitely merge that right now :)

## Changelog Fixes
+ Fixed fortune display not detecting melon fortune from tablist. - Chissl
+ Fixed dicer drops not being added to tracker/hidden. - Chissl
+ Fixed multiple menus and features throwing an error when encountering melons. - Chissl


CalMWolfs (approved):
Surely this fixes everything

PR #4652, opened 2025-09-06
Backend: Move click cancelling from cancelling outbound packets to cancelling handling of the clicks

Bloxigus:
## What
Currently in SkyHanni, cancelling entity click events and block click events causes the swing to still occur, which may be undesirable in case Hypixel decides to update their anticheat to prevent sending the swing packet when it should not be. One example of improper canceling currently is in the outside of garden visitor display canceling only the entity click and NOT the swing. Some of us gave Skyblocker a lot of heat for sending swing packets at the wrong time, so I think this is something that should be fixed.

This PR has been tested by ThatGravyBoat and cancels the right packets on a 1.21.8 testing server.

Supersedes [my previous PR](https://github.com/hannibal002/SkyHanni/pull/4651)

+ Made in-world attack/use click detection/cancelling not packet-based. - bloxigus


Thunderblade73 (approved):
The code seams to be logical correct. It does change the event behavior by making it more logical and only sending the events once for one action (it was send multiple times, but should not break any feature to my knowledge and imo is way better that way and makes more sense from a feature perspective).
I did not verify if the mixins happen at the "right" place, but I would guess there right.

PR #4655, opened 2025-09-07
Fix: Typo in changelog viewer

CalMWolfs:
## What
Now the text has no typo + is correct.

## Changelog Fixes
+ Fixed a typo in the Changelog Viewer. - CalMWolfs

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangeLogViewerScreen.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangeLogViewerScreen.kt
@@ -136,7 +136,7 @@ class ChangeLogViewerScreen : SkyhanniBaseScreen() {
                         Renderable.text("§aNo changes found", horizontalAlign = RenderUtils.HorizontalAlignment.CENTER)
                     } else if (!ChangelogViewer.shouldShowBeta) {
                         Renderable.text(
-                            "§aOnly Full Releases were added, turn on \"Include Betas\"",
+                            "§aNo Full Releases in specified interval, modify the search or turn on \"Include Betas\"",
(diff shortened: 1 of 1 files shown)


Thunderblade73 (changes requested):
You flipped the meaning. Since the line here is logically: ``There was no Full Release in your specified Interval``
But you say the opposite with that, which is wrong.

CalMWolfs:
> You flipped the meaning. Since the line here is logically: ``There was no Full Release in your specified Interval``
> 
> But you say the opposite with that, which is wrong.

This code only runs if it shouldn't show betas and then the list of changes is empty?

Thunderblade73:
> > You flipped the meaning. Since the line here is logically: `There was no Full Release in your specified Interval`
> > But you say the opposite with that, which is wrong.
> 
> This code only runs if it shouldn't show betas and then the list of changes is empty?

The highlevel branching tree goes:
```kotlin
if(isEmpty()){  // isEmpty on the view
   if(changelogList.isEmpty()){  // isEmpty on the raw data
       return "Interval is empty"
   } else {
      if(ChangelogViewer.shouldShowBeta is not turned on){
         return "Turn on Beta Toggle to See Stuff" // The line you are changing
      } else{
         return "Unreachable Code Reached xD"
      }
   }
} else {
  return this
}
```
You miss understood the second ``isEmpty()`` as the first ``isEmpty()`` (That is in the condition lambda of the transformIf)

I mean the code is not very intuitive to be fair. Having the transform lambda as a real function would help (eg: ``responseOfEmptyView``) and maybe a command to the ``changeloList.isEmpty()`` to hint that is about the underlying data.

Thunderblade73 (approved):

PR #4660, opened 2025-09-08
Improvement: Add Bal Shard to Crystal Nucleus Profit Tracker

lunaynx:
## What
Added Bal Shard to Crystal Nucleus Profit Tracker, which has a chance to drop from the Bal boss on each hit while having a salt active (not tested).

## Changelog Improvements
+ Added Bal Shard to Crystal Nucleus Profit Tracker. - Luna


hannibal002 (changes requested):
cant we just use the shard gain event as implemented here?

https://github.com/hannibal002/SkyHanni/blob/7f11def4315c75ab0803c54bdfc291b4510cacc1/src/main/java/at/hannibal2/skyhanni/features/inventory/attribute/AttributeShardsData.kt#L269

lunaynx:
My bad, didn't realize we had that.

hannibal002 (approved):
not tested, but code looks good

PR #4661, opened 2025-09-09
Fix: Block Strength Guide being unopenable

brooke-gill:
## What
This PR fixes several issues with the Block Strength Guide (/blockstrengthguide). These issues were:

- Fixed the Block Strength Guide not being openable due to the new /stats menu.
- Fixed /blockstrengthguide being runnable outside of SkyBlock.
- Fixed a race condition that prevented the GUI from being openable if the user reloaded it too quickly.
- Fixed some typos and capitalization in the Block Strength Guide.

## Changelog Fixes
+ Fixed Block Strength Guide not opening. - brooke


hannibal002 (approved):
code looks fine & tested in game

PR #4663, opened 2025-09-09
Compilation: Change remap base from 1.8.9 to modern 1.21.7

HacktheTime:
This would mean that modern compilation is a lot quicker.

While the drawback is that 1.8.9 is slower to compile, some features can only be used in modern (Foraging). As such, modern should be prioritised for compilation time.

However, there is another significant drawback. Originally, everyone only had to use the 1.8.9 Style, which can't change anymore. If modern is used as the default, this means all changes have to be learned by all the developers as needed, since it has to be used actively, and only the backport has to be rewritten once.

edit: since AI summary did not get context here, a very easy summary.

before this change the normal src was in 1.8.9. This PR is supposed to change the normal src to 1.21.7 (maybe upgrade further in the future. who knows). This means instead of 1.8.9 being modified to run modern, 1.21.7 would be the base being modified to run 1.8.9. So it's the opposite pre-processing order essentially.


NopoTheGamer:
did you do anything that wasnt just copy the preprocessed code

HacktheTime:
> did you do anything that wasnt just copy the preprocessed code

So far barely. It will need more changes but copying the pre processed to 1.21.7 is one of the first steps

NopoTheGamer:
this is pretty useless unless at bare minimum everything like text compat is removed

hannibal002:
This is the kind of PR that you should have asked us before starting working on it. we want to ditch 1.8 eventually completely, not speed up the 1.21 compile time by making 1.8 way more complex.

PR #4672, opened 2025-09-12
Fix: Skyhanni user luck in sb menu

NopoTheGamer:
## What
hypixel changed the menus

had to make it so item replace event can replace air so had to change some things soz

## Changelog Fixes
+ Fixed SkyHanni User Luck in SB Menu. - nopo

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/OldSkyblockMenu.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/OldSkyblockMenu.kt
@@ -50,8 +50,8 @@ object OldSkyblockMenu {
         if (!isEnabled()) return
 
         val sbButton = slotMap[event.slot]?.takeIf { !it.disabled } ?: return
-        val isAlreadySbButton = event.originalItem?.displayName?.endsWith(sbButton.displayName)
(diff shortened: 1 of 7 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/OldSkyblockMenu.kt
```diff
@@ -50,8 +50,8 @@ object OldSkyblockMenu {
         if (!isEnabled()) return
 
         val sbButton = slotMap[event.slot]?.takeIf { !it.disabled } ?: return
-        val isAlreadySbButton = event.originalItem.displayName.endsWith(sbButton.displayName)
-        if (isAlreadySbButton) return
+        val isAlreadySbButton = event.originalItem?.displayName?.endsWith(sbButton.displayName)
+        if (isAlreadySbButton == true) return
```
lunaynx:
```suggestion
        val isAlreadySbButton = event.originalItem?.displayName?.endsWith(sbButton.displayName) == true
        if (isAlreadySbButton) return
```
Nitpick, but this makes more sense.

lunaynx (changes requested):

lunaynx (approved):

PR #4673, opened 2025-09-12
Feature: Hide all block breaking particles.

ThatOneDevil:
## What
Adds a option in the particle hider section of the config to hide block breaking particles, the intended use being for the garden.

<details>
<summary>Images</summary>

image
</details>

## Changelog New Features
+ Added Block-Breaking Particle Hider. - ThatOneDevil
    * An option in Particle Hider to disable block-breaking particles.
    * Extra option for Garden only.


src/main/java/at/hannibal2/skyhanni/mixins/hooks/BlockRendererDispatcherHook.kt
CalMWolfs, line 1: please revert this files changes

src/main/java/at/hannibal2/skyhanni/features/misc/ParticleHider.kt
```diff
@@ -64,5 +64,6 @@ object ParticleHider {
         event.move(3, "misc.hideCloseRedstoneparticles", "misc.particleHiders.hideCloseRedstoneParticles")
         event.move(3, "misc.hideFireBlockParticles", "misc.particleHiders.hideFireBlockParticles")
         event.move(3, "misc.hideSmokeParticles", "misc.particleHiders.hideSmokeParticles")
+        event.move(3, "misc.hideBlockBreakParticles", "misc.particleHiders.hideBlockBreakParticles")
```
CalMWolfs: This isnt needed as this is a new option so therefore it isnt moving

CalMWolfs (changes requested):
This already exists in patcher & the particles section of video settings on modern.

You mention that this could be used for the garden so i'd see it as fine for being added if it was added and then there was a sub option for it to only work on the garden.

In additon to this I have only briefly looked at the code but I dont think this mixin will work on 1.21

ThatOneDevil:
> This already exists in patcher & the particles section of video settings on modern.
> 
> You mention that this could be used for the garden so i'd see it as fine for being added if it was added and then there was a sub option for it to only work on the garden.
> 
> In additon to this I have only briefly looked at the code but I dont think this mixin will work on 1.21

would it be fine if just add it in the garden config? and make it work for 1.8.9 only? since the mixin works on 1.8.9

CalMWolfs:
> would it be fine if just add it in the garden config? 

I think all the block break particle settings should be in the Particle Hider config but you should make an accordion for block breaking particles and inside that accordion you have the overall toggle then you have a secondary "Only on garden" toggle

> and make it work for 1.8.9 only? since the mixin works on 1.8.9

No, all new features should be added for both versions

ThatOneDevil:
> > would it be fine if just add it in the garden config?
> 
> I think all the block break particle settings should be in the Particle Hider config but you should make an accordion for block breaking particles and inside that accordion you have the overall toggle then you have a secondary "Only on garden" toggle
> 
> > and make it work for 1.8.9 only? since the mixin works on 1.8.9
> 
> No, all new features should be added for both versions

image

like this? i also made it multiversion just need to push and fix some final stuff!

hannibal002:
please move the 1.21 mixin logic into the 1.21 mixin packet as a new file.

hannibal002 (changes requested):
I have not tested this for any version.
please move it in the 1.21 mixin logic into a standalone 1.21 mixin file, so that there is no preprocessed comment stuff there.
Also please create a hook like file that handles the actual logic (so that we have only one code line per function in the java files, and logic stays in kotlin files)

ThatOneDevil:
> I have not tested this for any version. please move it in the 1.21 mixin logic into a standalone 1.21 mixin file, so that there is no preprocessed comment stuff there. Also please create a hook like file that handles the actual logic (so that we have only one code line per function in the java files, and logic stays in kotlin files)

i have tested it on all versions and it works btw. Although if u also want to test you can.

hannibal002:
just tested on 1.21, works for me

hannibal002 (approved):

PR #4678, opened 2025-09-14
Backend: Tablist End Time

Chissl:
## What
Hypixel tablist time always rounds down and excludes seconds until (generally) less than 6 minutes are left, so added a timeutils function to more accurately estimate tablist times. Moved pest timer and plot spray timer over to this since they already did something similar. 

+ Added getTablistEndTime, which better estimates when a tablist timer ends, to TimeUtils. - Chissl


hannibal002 (changes requested):
please review/approve my cleanup changes

Chissl:
> please review/approve my cleanup changes

tested and it works

PR #4687, opened 2025-09-17
Fix: Mob type detection matching parts of mob names

lunaynx:
## What
Fixed the mob type regex to not mistakenly catch parts of mob names as their types for mobs that do not have types (e.g. Rift mobs).

<details>
<summary>Images</summary>

image

</details>

## Changelog Fixes
+ Fixed mob names being incorrectly detected for some mobs without types. - Luna


src/main/java/at/hannibal2/skyhanni/data/mob/MobFilter.kt
```diff
@@ -55,7 +55,7 @@ object MobFilter {
     private val patternGroup = RepoPattern.group("mob.detection")
 
     @Language("RegExp")
-    private val mobType = "(?<mobType>[^\\s✯]+ )?"
+    private val mobType = "(?<mobType>[^a-zA-Z0-9_\\s✯-]+ )?"
```
Thunderblade73:
```suggestion
    private val mobType = "(?<mobType>[^\\w\\s✯-]+ )?"
```

Thunderblade73 (changes requested):

Thunderblade73 (approved):

src/main/java/at/hannibal2/skyhanni/data/mob/MobFilter.kt
```diff
@@ -55,7 +55,7 @@ object MobFilter {
     private val patternGroup = RepoPattern.group("mob.detection")
 
     @Language("RegExp")
-    private val mobType = "(?<mobType>[^\\s✯]+ )?"
+    private val mobType = "(?<mobType>[^\\w\\s✯-]+ )?"
```
Thunderblade73: Actually - should be escaped. Or else you get unintended behavior. But idk why it even is in the list.
```suggestion
    private val mobType = "(?<mobType>[^\\w\\s✯\-]+ )?"
```

Thunderblade73 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/mob/MobFilter.kt
lunaynx: It shouldn't be necessary to escape if it's at the beginning or end of a pattern, that's a special case.

I added it just in case because I could see it being part of a mob name, but it's probably technically not necessary because the word would have to also not contain any alphanumeric characters for it to match.

Thunderblade73: Yeah and if someone who doesn't know about that fact adds a character to the end, unintended behavior will occur.

lunaynx: Fair enough.

Thunderblade73 (approved):

PR #4688, opened 2025-09-17
Fix: Dragon profit concurrent modification error

Stuflo19:
## What
Addresses the dragon profit concurrent modification error

```
SkyHanni 4.18.0 1.8.9: Caught a ConcurrentModificationException in at.hannibal2.skyhanni.features.combat.end.ProfitPerDragon.onTick(at.hannibal2.skyhanni.events.minecraft.SkyHanniTickEvent) at SkyHanniTickEvent: null
 
Caused by java.util.ConcurrentModificationException: null
    at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:711)
    at java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:734)
    at SH.features.combat.end.ProfitPerDragon.scanForLoot(ProfitPerDragon.kt:31)
    at SH.features.combat.end.ProfitPerDragon.onTick(ProfitPerDragon.kt:141)
    at SH.utils.ReflectionUtils$$Lambda$1576/498657274.accept(Unknown Source)
<Skyhanni event post>
```

<details>
<summary>Extreme example of adding all spawned entities to the scanned loot uuid's showing that no error is caused by removing items</summary>

image
</details>

## Changelog Fixes
+ Fixed Dragon Profit Display failing to handle rare drops. - Stuflo19

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/combat/end/ProfitPerDragon.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/combat/end/ProfitPerDragon.kt
@@ -28,15 +28,7 @@ object ProfitPerDragon {
     private fun scanForLoot() {
         val entities = EntityUtils.getEntities<EntityArmorStand>()
 
-        val scannedLootIterator = scannedLootUUIDs.iterator()
-
-        while (scannedLootIterator.hasNext()) {
-            val uuid = scannedLootIterator.next()
-
-            if (entities.none { it.uniqueID == uuid }) {
-                scannedLootIterator.remove()
-            }
-        }
+        scannedLootUUIDs.removeIf { uuid -> entities.none { it.uniqueID == uuid} }
 
         for (entity in entities) {
             val entityName = entity.name
--- a/src/main/java/at/hannibal2/skyhanni/features/combat/end/ProfitPerDragon.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/combat/end/ProfitPerDragon.kt
@@ -28,7 +28,7 @@ object ProfitPerDragon {
     private fun scanForLoot() {
         val entities = EntityUtils.getEntities<EntityArmorStand>()
 
-        scannedLootUUIDs.removeIf { uuid -> entities.none { it.uniqueID == uuid} }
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/combat/end/ProfitPerDragon.kt
```diff
@@ -28,9 +28,13 @@ object ProfitPerDragon {
...
 
-        for (uuid in scannedLootUUIDs) {
+        val scannedLootIterator = scannedLootUUIDs.iterator()
+
+        while (scannedLootIterator.hasNext()) {
+            val uuid = scannedLootIterator.next()
+
             if (entities.none { it.uniqueID == uuid }) {
-                scannedLootUUIDs.remove(uuid)
+                scannedLootIterator.remove()
```
lunaynx:
```suggestion
                scannedLootIterator.remove(uuid)
```
I presume you meant this?

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/features/combat/end/ProfitPerDragon.kt
Stuflo19: > I presume you meant this?

```
Removes from the underlying collection the last element returned by this iterator.
```

The iterator remove function has this definition so the remove by itself will remove the current element that we are on from the list.

Bloxigus:
Please use `removeIf` instead of this.

lunaynx (approved):

hannibal002 (changes requested):
please change the user facing "changelog fix" part so that non technical skyhanni users know whats going on. please avoid using terms like concurrent modification error etc. instead, explain what was the bug the user noticed. e.g. "feature x now works when y is true"

Stuflo19:
> please change the user facing "changelog fix" part so that non technical skyhanni users know whats going on. please avoid using terms like concurrent modification error etc. instead, explain what was the bug the user noticed. e.g. "feature x now works when y is true"

I have updated the user facing changelog text now

CalMWolfs (approved):

hannibal002 (approved):

PR #4695, opened 2025-09-19
Feature + Improvement: /call shortening + Trevor rework

platonvin:
## What
1) Adds `/call` shortening in (almost) the same way as `/warp` shortening works. I hardcoded all the commands, but they should be added to hannibal002/SkyHanni-REPO. I suppose there is an established process of doing so
2) Adds Trevor the trapper feature to `/warp desert` or `/warp trap` on keybind when it is faster (closer to mob than you are), kinda like Diana helpers do. Also, while in "pending" state (started quest, but not accepted yet), keybind now does nothing (it was annoying when you pressed it a little too early and it warped you back instead of accepting)

#### after typing `/ma` and pressing `TAB` with  `/call` shortening enabled
image

## Changelog New Features
+ Added `/call` shortening. - platonvin
    * `/call maddox` -> `/maddox`
+ Added `warp to closest on keybind` feature for Trevor the Trapper. - platonvin
    
+ Removed redundant state variables (e.g. precise timer & in seconds timer => precise timer, truncated when needed), made state transitions more clear, fixed `deprecated` warning

<pre></pre>

Note: i am new here (in both java and minecraft modding), i'd appreciate guidance if you think i need some


NopoTheGamer:
cant believe you would hardcode a list that the mod already gets from the repo

NopoTheGamer:
also dont add two completely different things into the same pr

platonvin:
Should i split into 2 different PRs? I made 2 branches that contain those features separated, but i dont think there is a way to attach them here "natively"
https://github.com/platonvin/SkyHanni/tree/trevor_warp_closest
https://github.com/platonvin/SkyHanni/tree/call_shortening
Also, did i setup my IDE incorrectly? Seems like i am not running Detekt on my side

NopoTheGamer:
> Should i split into 2 different PRs? I made 2 branches that contain those features separated

yes

PR #4698, opened 2025-09-19
Feature: Call shortening

platonvin:
## What
Adds `/call` shortening in the same way as `/warp` shortening works. I hardcoded all the commands, but they should be added to hannibal002/SkyHanni-REPO. I suppose there is an established process of doing so. Do i just open PR there with addition of "AbiphoneContacts" or smth?

<details>
<summary>Image: after typing `/ma` and pressing `TAB` with `/call` shortening enabled</summary>
image
</details>

## Changelog New Features
+ Added `/call` shortening. - platonvin
    * Lets you write `/maddox` instead of `/call maddox`.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/commands/CommandsConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/commands/CommandsConfig.kt
@@ -68,6 +68,16 @@ class CommandsConfig {
     @FeatureToggle
     var shortenCall: Boolean = false
 
+    @ConfigOption(
+        name = "Warp or call on §e/forge",
+        desc = "If both §e/warp§7 and §e/call§7 shortening features enabled, controls §e/forge§7 behaviour\n" +
+            "When disabled, §e/forge§7 does §e/warp forge§7. Otherwise §e/forge§7 does §e/call forge§7. Or you can just use §e/fred§7 ",
+    )
+    @Expose
+    @ConfigEditorBoolean
+    @FeatureToggle
+    var shortenForgeToCallNotWarp: Boolean = false
+
     @Expose
     @ConfigOption(name = "Replace §e/warp is", desc = "Add §e/warp is §7alongside §e/is§7. Idk why. Ask §cKaeso")
     @ConfigEditorBoolean
--- a/src/main/java/at/hannibal2/skyhanni/config/features/commands/CommandsConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/commands/CommandsConfig.kt
@@ -69,14 +69,14 @@ class CommandsConfig {
     var shortenCall: Boolean = false
 
     @ConfigOption(
-        name = "Warp or call on §e/forge",
-        desc = "If both §e/warp§7 and §e/call§7 shortening features enabled, controls §e/forge§7 behaviour\n" +
-            "When disabled, §e/forge§7 does §e/warp forge§7. Otherwise §e/forge§7 does §e/call forge§7. Or you can just use §e/fred§7 ",
+        name = "Warp or call on shortening conflicts (e.g. §e/forge§7)",
+        desc = "If both §e/warp§7 and §e/call§7 shortening features enabled, controls behaviour on conflicts\n" +
+            "Example for §e/forge§7: when disabled, §e/forge§7 does §e/warp forge§7. Otherwise §e/forge§7 does " +
+            "§e/call forge§7. Or you can just use §e/fred§7",
     )
     @Expose
     @ConfigEditorBoolean
-    @FeatureToggle
-    var shortenForgeToCallNotWarp: Boolean = false
+    var preferCallOverWarp: Boolean = false
 
     @Expose
     @ConfigOption(name = "Replace §e/warp is", desc = "Add §e/warp is §7alongside §e/is§7. Idk why. Ask §cKaeso")
--- a/src/main/java/at/hannibal2/skyhanni/config/features/commands/CommandsConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/commands/CommandsConfig.kt
@@ -59,6 +59,15 @@ class CommandsConfig {
     @FeatureToggle
     var shortenWarp: Boolean = false
 
+    @ConfigOption(
+        name = "Shorten §e/call",
+        desc = "Allows calling without the need for the §ecall §7prefix.\n(§e/call maddox §7-> §e/maddox§7)",
+    )
+    @Expose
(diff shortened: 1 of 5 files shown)


lunaynx:
We already have it in the NEU repo. Probably just use `AbiphoneFeatures.abiphoneContacts` (remove the `private` and just make it `private set`).

platonvin:
Now they are not hardcoded, but in NEU repo some contacts are missing some shortcuts (e.g. `Plumber Joe` can be called with just `/call plumber`, but it is not reflected in NEU repo). Also, callNames is missing contact name itself so i add it to the list. Effectively, callNames is only aliases, not names which can be used to call the contact

lunaynx:
The PR adding it to the NEU repo did say it is not intended to contain every single possible name, but `plumber` sounds like it would be reasonably used and could be included.

src/main/java/at/hannibal2/skyhanni/features/misc/AbiphoneFeatures.kt
```diff
@@ -59,13 +59,17 @@ object AbiphoneFeatures {
...
 
     @HandleEvent
     fun onNeuRepoReload(event: NeuRepositoryReloadEvent) {
         val constant = event.getConstant<Map<String, AbiphoneContactInfo>>("abiphone", NeuAbiphoneJson.TYPE)
         abiphoneContacts = constant.flatMap { (key, value) ->
-            value.callNames ?: listOf(key.removeAllNonLettersAndNumbers().replace(" ", ""))
+            // In NEU repo, callNames are optional shortenings and do not include actual contact name
+            val cleanKey = key.removeAllNonLettersAndNumbers().replace(" ", "").lowercase()
+            val aliases = value.callNames.orEmpty().map { it.lowercase() }
+            listOf(cleanKey) + aliases
```
NopoTheGamer: this is not correct, not every normal contact name can be used, i looked at the list i dont think there is any missing names in the callNames?

src/main/java/at/hannibal2/skyhanni/features/commands/ShortenCallCommand.kt
```diff
@@ -0,0 +1,48 @@
...
+        if (!config.shortenCall) return
+
+        val message = event.message
+        if (!message.startsWith("/")) return
+
+        val command = message.lowercase().removePrefix("/").trimEnd()
+
+        val contacts = AbiphoneFeatures.abiphoneContacts
+        if ((contacts != null) && (command in contacts)) {
+            if (!config.shortenForgeToCallNotWarp && command == "forge") return
```
NopoTheGamer: This should probably be a more advanced check seeing as both lists can be updated remotely

src/main/java/at/hannibal2/skyhanni/config/features/commands/CommandsConfig.kt
```diff
@@ -59,6 +59,25 @@ class CommandsConfig {
...
+
+    @ConfigOption(
+        name = "Warp or call on §e/forge",
+        desc = "If both §e/warp§7 and §e/call§7 shortening features enabled, controls §e/forge§7 behaviour\n" +
+            "When disabled, §e/forge§7 does §e/warp forge§7. Otherwise §e/forge§7 does §e/call forge§7. Or you can just use §e/fred§7 ",
+    )
+    @Expose
+    @ConfigEditorBoolean
+    @FeatureToggle
+    var shortenForgeToCallNotWarp: Boolean = false
```
NopoTheGamer: this doesnt get feature toggle as its not a feature

NopoTheGamer (changes requested):

NopoTheGamer:
> Now they are not hardcoded, but in NEU repo some contacts are missing some shortcuts (e.g. `Plumber Joe` can be called with just `/call plumber`, but it is not reflected in NEU repo). 

Feel free to make a pr to add more names to the neu repo

src/main/java/at/hannibal2/skyhanni/features/misc/AbiphoneFeatures.kt
platonvin: Alchemist does not have Alchemist in callNames, that is what i meant. 
Also, i manually checked every contact, and "name without spaces" is only invalid for "Captain Ahone" and "St. Jerry" (e.g. "Maddox the slayer" can be called wiht "/call maddoxtheslayer"). So there is no "analytical" solution here - on Hypixel side they have names and some "call names", which are not properly reflected in NEU repo. So i guess i'll just add all actual call names to callNames in NEU

src/main/java/at/hannibal2/skyhanni/features/commands/ShortenCallCommand.kt
platonvin: wdym? This is exception for forge, i do not see anything else that needs such exception right now. Do you mean creating these settings automatically based on conflicts in warps, calls and other commands instead of hardcoding it for forge?

NopoTheGamer: you should check if the call is also a warp

src/main/java/at/hannibal2/skyhanni/features/misc/AbiphoneFeatures.kt
NopoTheGamer: Alchemist doesnt have callNames so no its not in there

src/main/java/at/hannibal2/skyhanni/config/features/commands/CommandsConfig.kt
platonvin: i removed it and everything still works. I have no idea what it does. Does it mark "a thing" as something that users will be asked about when they update skyhanni and see "the thing" for the first time?

NopoTheGamer: yeah theres a popup that shows all the new features

lunaynx: It basically just means it will be toggled by the enable/disable all buttons in `/shdefaultoptions`.

src/main/java/at/hannibal2/skyhanni/features/commands/ShortenCallCommand.kt
platonvin: yeah i agree. A better way to handle this would be to have a list of names (filled by user) which explicitly sets priority (e.g. list is "forge" and thus `/forge` means `/call forge`, and all other collisions mean `/warp name`)

PR #4674, opened 2025-09-12
Improvement: Individual Tracker Configs

Chissl:
## Dependencies
- #4629 

## What
Gives each tracker its own config settings, instead of solely going off the main one in misc. I wanted this for the "tracker timeout" setting, as some trackers have things added more infrequently than others. Has an option to choose between using that tracker config or the one in misc, renamed to universal tracker config. 

## Changelog Improvements
+ Added config settings for each Skyhanni Tracker. - Chissl
    * These previously all went off the same config, now known as the "Universal Tracker Config".
    * Default settings for each tracker are based on your Universal Tracker Config.
    * Has option to go back to only using Universal Tracker Config.

PR #4700, opened 2025-09-19
Improvement: Garden Tracker Uptime

Chissl:
## Dependencies
- #4674 

## What
As some garden trackers are modified quite infrequently, I moved them over to a system that tracks uptime based on crop breaks, pest kills, and interactions with visitors. Per-tracker config of every garden tracker still shows useless "Afk timeout" setting because I could not figure out how to remove it without redoing the config system, but I added a warning button that brings users to the correct setting. 

## Changelog Improvements
+ Changed the uptime of Garden trackers to be based on farming activities instead of when the tracker last gained items. - Chissl
    * Activities are: Breaking Crops, Visitor interactions, and Killing Pests.
    * Uptime for each activity are saved separately so you can include/exclude them as you wish.


hannibal002:
stale + merge conflicts

PR #4705, opened 2025-09-22
Improvement: Add pest shard drops to pest profit tracker

Chissl:
# What
Adds pest shards to pest profit tracker. Does not bucket them.

## Changelog Improvements
+ Added Pest Shard drops to Pest Profit Tracker. - Chissl

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestProfitTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestProfitTracker.kt
@@ -16,6 +16,7 @@ import at.hannibal2.skyhanni.events.PurseChangeCause
 import at.hannibal2.skyhanni.events.PurseChangeEvent
 import at.hannibal2.skyhanni.events.RepositoryReloadEvent
 import at.hannibal2.skyhanni.events.chat.SkyHanniChatEvent
+import at.hannibal2.skyhanni.events.item.ShardGainEvent
 import at.hannibal2.skyhanni.features.garden.GardenApi
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.test.command.ErrorManager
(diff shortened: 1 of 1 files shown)


The-Deerman:
cant we just use the shard gain event as implemented here?

https://github.com/hannibal002/SkyHanni/blob/7f11def4315c75ab0803c54bdfc291b4510cacc1/src/main/java/at/hannibal2/skyhanni/features/inventory/attribute/AttributeShardsData.kt#L269

Chissl:
> cant we just use the shard gain event as implemented here?
> 
> https://github.com/hannibal002/SkyHanni/blob/7f11def4315c75ab0803c54bdfc291b4510cacc1/src/main/java/at/hannibal2/skyhanni/features/inventory/attribute/AttributeShardsData.kt#L269

In the future maybe, but that has too many false positives to be accurate and I don't want to bother fixing it right now

Chissl:
I'll fix my regex based on its regextests though, thanks

lunaynx:
How exactly does it have false positives?

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestProfitTracker.kt
```diff
@@ -77,7 +77,26 @@ object PestProfitTracker {
...
+        "§.§l\\w{4,5}§7 You charmed a §aPest§7 and captured §9(?<amount>\\d+) Shards? §7from it\\."
+    )
+
+    /**
+     * REGEX-TEST: §aYou caught §7x3 §aPest §aShards§a!
+     */
+    private val pestBlackHoleDropPattern by patternGroup.pattern(
+        "blackholedrop",
+        "§aYou caught §7x(?<amount>\\d+) §aPest §aShards?§a!"
+    )
```
NopoTheGamer: what about naga

Chissl:
> How exactly does it have false positives?

For example, adding shards to hunting box triggers the event

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestProfitTracker.kt
Chissl: image

lunaynx:
Oh, that. I don't know why we would even want that to trigger the event, it should only be needed internally to track the amount in your hunting box.

PR #4706, opened 2025-09-22
Fix: Separate shard events from shard gain event

CalMWolfs:
## What
Separated generic shard stuff for internal tracking from the shard gain event that trackers use.
Fixed the pattern for shard fusions being incorrect
Fixed not subtracting the fusion costs from the stored amount in your hunting box

## Changelog Fixes
+ Fixed some profit trackers double-counting shard gains. - CalMWolfs
+ Fixed shard fusions not being tracked correctly in Attribute Shard features and Item Pickup Log. - CalMWolfs


Chissl:
This doesn't fully fix the issue as syphoning shards will still subtract from trackers

Chissl:
> This doesn't fully fix the issue as syphoning shards will still subtract from trackers

nvm I'm blind

hannibal002 (changes requested):
please move the two events/classes in one file, and add javadoc like comments on both events, explaining the difference.

PR #4708, opened 2025-09-22
Fix: HotX currencies being reset on opening SkyBlock Menu

lunaynx:
## What
Fixes a regression of a previously fixed issue where HotM (and now also HotF) currencies reset upon opening the SkyBlock Menu while on a mining/foraging island—the event was firing in inventories not matched by the InventoryDetector due to unnecessary `@HandleEvent` annotations.

## Changelog Fixes
+ Fixed Total Powder/Forest Whispers resetting upon opening the SkyBlock Menu on mining/foraging islands. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/hotx/HotfData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/hotx/HotfData.kt
@@ -470,12 +470,6 @@ enum class HotfData(
             }
         }
 
-        // This event is fired by an InventoryDetector, and should NOT have @HandleEvent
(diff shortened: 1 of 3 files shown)


Thunderblade73 (changes requested):
I'm pretty sure the function do not need to be overriden in HotF And HotM. So those function can be deleted.

lunaynx:
Yeah, I was thinking it seems redundant, but wasn't sure. I'll double check in a bit.

Thunderblade73 (approved):

hannibal002 (approved):

PR #4726, opened 2025-09-28
Fix: Mooshroom Cow incrementing Mushroom milestone below Garden 9

lunaynx:
## What
Fixed Crop Milestone Display incorrectly incrementing Mushroom milestone if the player is using a Rare or higher Mooshroom Cow Pet but does not have the Mushroom crop unlocked on the Garden (requires Garden Level 9). Below this level, the pet still drops Mushrooms, but they do not count towards the milestone, and the milestone progress starts at 0 once the player reaches Garden Level 9.

<details>
<summary>Images</summary>

</details>

## Changelog Fixes
+ Fixed Crop Milestone Display incrementing milestone from Mooshroom Cow Pet while below Garden Level 9. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCropMilestoneDisplay.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCropMilestoneDisplay.kt
@@ -281,6 +281,14 @@ object GardenCropMilestoneDisplay {
     }
 
     private fun addMushroomCowData() {
+        if (GardenApi.getGardenLevel(overflow = false) < 9) {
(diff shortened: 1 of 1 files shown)


Chissl:
> Fixed Crop Milestone Display incrementing milestone from Mooshroom Cow Pet while below Garden Level 9

I don't see how this pr does that, as mooshroom cow drops are handled in a different class

lunaynx:
Well, you still gain mushrooms, which is relevant for the money per hour display. How else do you propose solving it? I don't see a separate counter for milestone only.

Chissl:
I'll just add it to my milestone pr, our current implementation of milestones is not ideal and not worth working on

PR #4732, opened 2025-09-30
Improvement: Allow duplicate Garden Custom Keybinds on 1.21.9+

lunaynx:
## Dependencies
- #4788

## What
[Minecraft 1.21.9](https://www.minecraft.net/en-us/article/minecraft-java-edition-1-21-9) is now officially released and supported by Hypixel, featuring this change:

> * Multiple key binds that have the same key assigned to them can now work together, while before only one would work
>   * For example, you can now bind `D` to `Jump` and `Strafe Left` to jump and strafe left at the same time
>   * The red warning for assigning the same key to multiple key binds is changed to yellow

Therefore, there is no longer any reason to restrict players from assigning duplicate keybinds in our Garden Custom Keybinds feature.

This has been gated with a Minecraft version check due to concerns expressed by some SkyHanni contributors on Discord, even though I personally believe it should be safe on all versions.

(It is technically 1.21.9 and above, but I wrote 1.21.10 in the changelog, because at this point we do not support 1.21.9 anymore – the 1.21.10 JAR has a hard dependency on the Minecraft version).

## Changelog Improvements
+ Garden Custom Keybinds now allows setting duplicate keybinds on Minecraft 1.21.10 and above. - Luna
    * Minecraft itself allows doing this now, so there is no longer any reason to restrict this.


hannibal002:
I don't know if it's ok to backport this to versions below 1.21.9. It feels like the crop hitbox case, but sk1er or someone talked to Hypixel admins before releasing.

lunaynx:
This behavior is completely indistinguishable to the server. A player holding one key bound to two actions looks the exact same as them holding two keys. It's literally nothing more than an accessibility feature that reduces hand strain. It shouldn't require "permission" from Hypixel admins.

(IMO, it should never have been called a "cheat" in the first place, and it's something you could already do with a controller as well, at least when it comes to movement. But now with it being an official feature, there's no doubt about it being legit.)

hannibal002 (dismissed):
needs further discussions

lunaynx:
Updated to support version limitation due to concerns on Discord (will start out as limited to 1.21.9+ only if the linked repo PR is merged).

src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCustomKeybinds.kt
```diff
@@ -104,21 +109,30 @@ object GardenCustomKeybinds {
...
         KeyBinding.unPressAllKeys()
     }
 
-    private fun calculateDuplicates() {
-        isDuplicate = map.values
-            .filter { it != Keyboard.KEY_NONE }
-            .let { values -> values.size != values.toSet().size }
+    private fun checkDuplicateKeybinds() {
+        hasDisallowedDuplicateKeybinds = !alwaysAllowDuplicateKeybinds &&
+            MCVersion.fromString(PlatformUtils.MC_VERSION) < MCVersion.fromString("1.21.9") &&
```
CalMWolfs: should probably have a lazy boolean for if we are less than 1.21.9

CalMWolfs (changes requested):
havent looked at the rest of the pr, jsut saw this

PR #4733, opened 2025-10-01
Improvement: Add Event EXP to GEXP filter

AfkUserMC:
## What
Change "Guild EXP" chat filter to "Guild/Event EXP"
Previously, it would only hide Event EXP messages if you also gained Guild Exp, now it also hides standalone Event EXP messages.

## Changelog Improvements
+ Added Event EXP to Guild EXP Chat Filter. - AfkUserMC
    * Renamed filter from "Guild EXP" to "Guild/Event EXP".

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
@@ -102,9 +102,10 @@ object ChatFilter {
     /**
      * REGEX-TEST: §aYou earned §r§22 GEXP §r§afrom playing SkyBlock!
      * REGEX-TEST: §aYou earned §r§22 GEXP §r§a+ §r§c210 Event EXP §r§afrom playing SkyBlock!
+     * REGEX-TEST: §aYou earned §r§510 Event EXP §r§afrom playing SkyBlock!
      */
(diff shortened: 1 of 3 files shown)


AfkUserMC:
drafting this to do [5e266fb](https://github.com/hannibal002/SkyHanni/pull/4733/commits/5e266fb58f141fadbefff87c4709f39bcb4be905)

src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
```diff
@@ -705,7 +707,7 @@ object ChatFilter {
         event.move(3, "chat.hypixelHub", "chat.filterType.hypixelHub")
         event.move(3, "chat.empty", "chat.filterType.empty")
         event.move(3, "chat.warping", "chat.filterType.warping")
-        event.move(3, "chat.guildExp", "chat.filterType.guildExp")
+        event.move(3, "chat.guildExp", "chat.filterType.guildEventExp")
```
hannibal002: This change is wrong in principle. If we change the name in config version 3, the name can't be changed in version 107. While this doesn't matter in this specific case, we should never change old config move logic in general.

hannibal002 (changes requested):

PR #4742, opened 2025-10-03
Backend: Improve build process [Failures are expected, read description]

lunaynx:
> [!NOTE]
> Build and Detekt is expected to fail here, due to GitHub still running the old workflows for security reasons. You can test the new workflows locally or by merging to a separate branch first.
> A sample run can be seen here: https://github.com/lunaynx/SkyHanni/actions/runs/18231407642

## What
This PR does two things:
- Makes multi-version build always on. You can still build 1.8.9 only with e.g. `./gradlew 1.8.9:build`. Also updates the documentation related to this.
- Improves the build process in GitHub Actions so that 1.8.9 is no longer built twice, and each version is uploaded in a separate .zip file as soon as the build for it is completed. Also made `gradlew-retry.kts` stream the output of Gradle in real time, and fixed an error with the retry logic.

exclude_from_changelog


Thunderblade73 (changes requested):
You removed the test run in the workflow.

lunaynx:
Actually, I still need to do the matrix stuff, I forgot.

lunaynx:
Matrix didn't work since the jobs can't depend on each other, but I improved it a bit.

lunaynx:
I think the build is pretty much inevitably going to fail in this PR due to it trying to build with the old workflows.

.github/workflows/build.yml
```diff
@@ -14,74 +14,75 @@ on:
     workflow_dispatch:
 permissions: write-all
 jobs:
-    build:
+    build_1_8_9:
+        name: "Build 1.8.9"
         runs-on: ubuntu-latest
-        name: "Build and test"
         steps:
-            -   name: Checkout code
```
NopoTheGamer, line 21: im pretty sure checkout is 1 word

lunaynx: https://www.grammar.com/check_out_vs_checkout

lunaynx:
Closing this, will open an updated PR soon™️.

PR #4743, opened 2025-10-03
Backend: Slayer API cleanup

hannibal002:
## What
fixed stuff around slayer API:
moved state logic out from SlayerQuestWarning into api class.
created ActiveQuestState enum.
created SlayerStateChangeEvent.
renamed some variables in slayer api.
added documentation.

+ Slayer API cleanup. - hannibal2


src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
```diff
@@ -24,7 +27,8 @@ import at.hannibal2.skyhanni.utils.compat.MinecraftCompat
 import at.hannibal2.skyhanni.utils.toLorenzVec
 import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.Duration.Companion.minutes
-import kotlin.time.Duration.Companion.seconds
+
+typealias Type = SlayerType
```
Thunderblade73: you an use ``import as`` instead of an typealias.

Thunderblade73 (changes requested):

PR #4746, opened 2025-10-04
Improvement: Fall back to tab list for slayer quest detection

lunaynx:
## What
SlayerApi will now try to parse slayer quest information from the scoreboard first (as before), but if it is not found (e.g. due to it being overridden by an objective), it will try to use the tab list widget instead.

## Changelog Improvements
+ Improved Slayer quest detection: uses the Tab List widget when Hypixel omits them from the Scoreboard (common in The Park). - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
@@ -23,7 +23,6 @@ import at.hannibal2.skyhanni.utils.RecalculatingValue
 import at.hannibal2.skyhanni.utils.SimpleTimeMark
 import at.hannibal2.skyhanni.utils.SkyBlockUtils
 import at.hannibal2.skyhanni.utils.StringUtils.removeColor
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
```diff
@@ -193,10 +200,19 @@ object SlayerApi {
...
+
+    @HandleEvent(ScoreboardUpdateEvent::class, onlyOnSkyblock = true)
+    fun onScoreboardChange() {
+        updateSlayerState()
+    }
 
+    @HandleEvent(TabListUpdateEvent::class, onlyOnSkyblock = true)
+    fun onTabListUpdate() {
+        updateSlayerState()
     }
```
Thunderblade73: Use the widget event and only the slayer widget. We do not want to process the whole tab list anywhere.
It just takes up way to much performance. Since without the widget abstraction (which parses the tablist once and notifies the features about changes) every feature would need to parse all lines every update (which happens multiple times a second in some cases, eg: Hub).

Thunderblade73 (changes requested):

PR #4761, opened 2025-10-11
Fix: Modified Chat Message Mod Compat

j10a1n15:
## What
also send the old message through the chat event 
5.0.0 milestone pls ty

basically if hypixel sends "WOW! You found a Glacite Mineshaft portal!" skyhanni fetches it before fabric and then modifies it to their liking.
so with pity overlay its "WOW! You found a Glacite Mineshaft portal! (537)" and then that gets send into the fabric pipeline.
So then mods that check explicitly for "WOW! You found a Glacite Mineshaft portal!" fail bc the actual message never gets send in the pipeline.

## Changelog Fixes
+ Fixed SkyHanni conflicting with other mods when modifying chat messages on modern. - juna

### DIFF
--- a/versions/1.21.5/src/main/java/at/hannibal2/skyhanni/mixins/hooks/MessageHandlerHook.kt
+++ b/versions/1.21.5/src/main/java/at/hannibal2/skyhanni/mixins/hooks/MessageHandlerHook.kt
@@ -20,6 +20,9 @@ fun onGameMessage(message: Text, actionBar: Boolean, original: Operation<Void>) 
     }
     val (result, cancel) = ChatManager.onChatReceive(message)
     result?.let {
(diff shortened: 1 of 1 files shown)


hannibal002:
doesnt this cause the fabric event pipeline to receive both the original and the updated message as two events?
below, we only either call original.call or ClientReceiveMessageEvents.invoker, but not both

j10a1n15:
yes it does
i could also split up the chat event into Pre and Post and fully rely on fabric events but that would make the pr like +10k -10k idk if you would want that

this is the best solution that came to mind to make my feature work with default sh settings

hannibal002:
> but that would make the pr like +10k -10k idk if you would want that

not necessary, i agree.

maybe add a todo comment to the line, so that we know for the future that this is not perfect behaviour and whats going on and why. thanks!

PR #4762, opened 2025-10-11
Fix + Backend: Use tab widget instead of hard coded string

CalMWolfs:
## What
Adjusted the profile widget pattern as someone had `§r§r` instead of `§r§a`
Code cleanup to use tab widget information instead of a hard coded string.

## Changelog Fixes
+ Fixed SkyHanni occasionally incorrectly reporting that extra information wasn't found in the tab list. - CalMWolfs

+ No longer use a hard-coded string and instead use Tab Widget information for checks. - CalMWolfs

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/ProfileStorageData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/ProfileStorageData.kt
@@ -18,10 +18,8 @@ import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.DelayedRun
 import at.hannibal2.skyhanni.utils.HypixelCommands
 import at.hannibal2.skyhanni.utils.PlayerUtils
-import at.hannibal2.skyhanni.utils.RegexUtils.matches
 import at.hannibal2.skyhanni.utils.SimpleTimeMark
 import at.hannibal2.skyhanni.utils.SkyBlockUtils
-import at.hannibal2.skyhanni.utils.TabListData
 import kotlin.time.Duration.Companion.seconds
 
 @SkyHanniModule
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/data/ProfileStorageData.kt
```diff
@@ -125,7 +126,7 @@ object ProfileStorageData {
 
         if (noTabListTime.passedSince() < 3.seconds) return
         noTabListTime = SimpleTimeMark.now()
-        val foundSkyBlockTabList = TabListData.getTabList().any { it.contains("§b§lArea:") }
+        val foundSkyBlockTabList = TabListData.getTabList().any { HypixelData.islandNamePattern.matches(it) }
```
Thunderblade73: I see no reason why this isn't using TabWidgets.
It could simply use ``TabWidget.AREA.isActive``.
I checked and it should have the exact same result.

Thunderblade73 (changes requested):

hannibal002 (changes requested):
i dont see the benefit of this pr.
while i agree its a neat technical cleanup, this is no user facing change as far as i see:
it doesnt matter what value `foundSkyBlockTabList` has, in either way we show an red error message in chat. so the changelog wording is wrong.

Thunderblade73:
> i dont see the benefit of this pr. while i agree its a neat technical cleanup, this is no user facing change as far as i see: it doesnt matter what value `foundSkyBlockTabList` has, in either way we show an red error message in chat. so the changelog wording is wrong.

It used and old "regex" (string compare) that did not work for eg; dungeon. So this is a valid change. Look at the first commit of this PR for what happend really. I just suggested to clean it up as well.

CalMWolfs:
I've now found the underlying issue so now this should be good

PR #4767, opened 2025-10-13
Feature: keyboard control

platonvin:
## What
Adds a structured way to define keybinds for skyblock menus, some keybinds themself, keybind rendering plus generic "selector" ("slot cursor") to make no-mouse menu interaction always an option.

<details>
<summary>Images</summary>
/bz -> pressing "4" -> pressing "3"
blue square is "selector" ("slot cursor"), can be moved with arrow keys

image
image
image

*idk why first one is not dark* 
</details>

## Changelog New Features
+ Added keyboard control feature. - platonvin
    * Bazaar, Auction House, Sacks, and Recipes (Supercraft) menus.
    * Ability to display available keybinds.
    * "Selector" ("slot cursor") controlled with keyboard that works in every menu.

## Changelog Improvements
+ Thickness selection for immediate mode border rendering. - platonvin

+ Centralized keybinds definition and made it declarative. - platonvin

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/inventory/InventoryConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/inventory/InventoryConfig.kt
@@ -31,6 +31,10 @@ class InventoryConfig {
     val bazaar: BazaarConfig = BazaarConfig()
 
     @Expose
+    @Category(name = "Keyboard Control", desc = "Lets you control different menus with keyboard")
+    val keyboardControl: KeyboardControlConfig = KeyboardControlConfig()
+
+    @Expose
     @Category(name = "Experimentation Table", desc = "QOL features for the Experimentation Table.")
     val experimentationTable: ExperimentationTableConfig = ExperimentationTableConfig()
 
--- a/src/main/java/at/hannibal2/skyhanni/config/features/inventory/KeyboardControlConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/inventory/KeyboardControlConfig.kt
@@ -1,0 +1,313 @@
+package at.hannibal2.skyhanni.config.features.inventory
+
+import at.hannibal2.skyhanni.config.FeatureToggle
+import com.google.gson.annotations.Expose
+import io.github.notenoughupdates.moulconfig.annotations.Accordion
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorBoolean
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorKeybind
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorSlider
+import io.github.notenoughupdates.moulconfig.annotations.ConfigOption
+import io.github.notenoughupdates.moulconfig.annotations.SearchTag
+import org.lwjgl.input.Keyboard
+
+class KeyboardControlConfig {
+
+    @Expose
+    @ConfigOption(name = "Keybinds", desc = "Enable keyboard controls for Bazaar, Auction House, Sacks, and other inventories.")
+    @ConfigEditorBoolean
+    @SearchTag("keyboard")
+    @FeatureToggle
+    var keybindsEnabled: Boolean = false
+
+    @Expose
+    @ConfigOption(name = "Keybind cooldown", desc = "Set minimum time (ms) between two distinct clicks in the same inventory.")
+    @ConfigEditorSlider(minValue = 0f, maxValue = 300f, minStep = 1f)
+    var clickCooldown: Float = 150f
+
+    @Expose
+    @ConfigOption(name = "Render", desc = "Render available keybinds on corresponding slots in inventories.")
+    @ConfigEditorBoolean
+    var renderEnabled: Boolean = false
+
+    @Expose
+    @ConfigOption(name = "Selector", desc = "Enable keyboard controlled selector that works in all inventories.")
+    @ConfigEditorBoolean
+    var selectorEnabled: Boolean = false
+
+    @Expose
+    @ConfigOption(name = "Shared Keybinds", desc = "")
+    @Accordion
+    var shared: SharedConfig = SharedConfig()
+
+    class SharedConfig {
+        @Expose
+        @ConfigOption(name = "Back", desc = "Go back button.")
+        @ConfigEditorKeybind(defaultKey = Keyboard.KEY_C)
+        var back: Int = Keyboard.KEY_C
+
+        @Expose
+        @ConfigOption(name = "Confirm", desc = "Confirm actions like purchases, bids, orders, etc.")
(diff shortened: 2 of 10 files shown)


platonvin:
I have some issues:

* ~minecraft events and rendering system~

*
  * Keybinds for the most need to be "clicks". There exists "is clicked" method for keys, but you cant use it in "key press" events. However, the only cancellable event is GuiKeyPressEvent. So we need "is clicked", but we also need GuiKeyPressEvent, and those are not compatible. I just added delay to it
Question is: why is it so? Why cant we have key events **with** key state change (so you know if its pressed/unpressed/repeated) that are also cancellable?

  * GuiKeyPressEvent does not have pressed key code, which would be very nice to not have to iterate over each possible key. For 1.8.9 forge there is eventKeyCode tho

  * On OS level, you have some sort of callback or polling and you get "**which** key down / key up / repeated" events. You sit between OS and minecraft input, and if "event" was cancelled, rest of the code does not get notified. What am i missing that makes it impossible to combine all this? 

* This (in pr) keyboard manager treats every inventory in the same way no matter what previously opened inventory was. Hypixel decided they are gonna give some confirmation menus proper names, but give other ones generic "Confirm" name, which leads to problems*. We can add some sort of "history" to it but i believe it would be better to have some common utility / manager that would identify which inventory we are in based on ... *well, everything*, and would provide common utilities - interface for interacting with buttons, matching inventory to actions for it, extracting information, etc. That is kinda what de-centralized features already do, but there is a lot of redundant work in parsing titles and items lore and names (and no clear structure)
*for given example it is possible to use lore and match buttons strictly, but confirmation menus keybinds everywhere does not harm anyone so i did not bother
```kotlin
interface SkyblockMenu {
    // mostly always presented for this inventory
    var regularButtons: Map<ButtonId, Button>
    // ones representing "data" for this inventory
    var irregularButtons: Map<ButtonId, Button>
    // newly added buttons that this version of mod cannot parse    
    var unknownButtons: Map<UniqueUnknownButtonId, Button>
    var name: String // exact name of opened inventory
}
typealias UniqueUnknownButtonId = Int // just a way to support ids for unknown buttons
data class Button(val slot: Int, val name: String, val lore: ArrayList<String>)

enum class ButtonId { GO_BACK, CANCEL, BZ_MANAGE_ORDERS, ...}

data class MainBazaarScreen(
    override var regularButtons: Map<ButtonType, Button>,
    // holds search results / subcategories buttons
    override var irregularButtons: Map<ButtonType, Button>,
    override var unknownButtons: Map<ButtonType, Button>,
    override val name: String, // override here if constant
    val manageOrders: Button,
    val search: Button,
    val categories: ArrayList<Button>
    ...
) : SkyblockMenu
```
Some menus, would have a range of titles, mapped to some properties. Like categories for bazaar. Others, like skyblock menu, would be strictly named

so no matter how many bazaar related features we have, we would only parse menu once and avoid redundant work
and then, instead of features checking in which inventory we are, we can reverse it and make them subscribe to only be called in certain inventories

* I could not find key combination in config. Same for key list. How should it be done?

* Somehow, 2 ways to access inventory slots are out of sync (Inventory Fully Opened event and InventoryUtils.getItemsInOpenChest). Also, currentScreen.inventorySlots.inventorySlots contain empty slots, while InventoryUpdatedEvent.inventoryItems does not. But those are not out of sync??? I am pretty sure something in my code will break when executed in wrong time. Multiple sources of truth for the same state makes things really complicated

* You cant point to same regex pattern from different places, which forces duplication event if you need the same match. And when Hypixel updates bazaar, you have to modify not one, but every single such pattern

* Compile times are atrocious. It takes like 8 minutes for incremental rebuild for 1.21.7 and 2 minutes for 1.8.9. My machine can build latest Linux kernel from scratch (not incrementally) in 5 minutes. **In WSL**. SkyHanni is also larger than kernel. I do not think SkyHanni is comparible in complexity to kernel to be this large and this slow to build. Something is wrong
I believe problem is partially in preprocessor, which seems to trigger less incremental builds (it touches every file or smth)

Bloxigus (changes requested):
no

Chissl:
Did your mouse break

platonvin:
> Did your mouse break

Using keyboard is very nice. You should try typing with keyboard, it is a lot faster then using mouse!
On serious note, there is surprisingly a lot of usecases for that in skyblock. Have you ever done a lot of bazaar manipulations? Using purely mouse for moving tons of items from buy order to sacks is very slow and even painful. With selector + position caching you just have to repeat same sequance of key presses (like Bazaar keybind -> enter -> right shift -> esc -> sacks keybind -> enter, that is how it looks for my keybinds) till its done
It is also very nice for quickly buying some consumables, like gummy bears, experience bottles or god potions - no roundtrip from keyboard to mouse
Exploring new inventories is not that great because tooltip is not shown on selector, i forgot to mention it. What is best way to do so?

Chissl:
> > Did your mouse break
> 
> Using keyboard is very nice. You should try typing with keyboard, it is a lot faster then using mouse!
> On serious note, there is surprisingly a lot of usecases for that in skyblock. Have you ever done a lot of bazaar manipulations? Using purely mouse for moving tons of items from buy order to sacks is very slow and even painful. With selector + position caching you just have to repeat same sequance of key presses (like Bazaar keybind -> enter -> right shift -> esc -> sacks keybind -> enter, that is how it looks for my keybinds) till its done
> It is also very nice for quickly buying some consumables, like gummy bears, experience bottles or god potions - no roundtrip from keyboard to mouse
> Exploring new inventories is not that great because tooltip is not shown on selector, i forgot to mention it. What is best way to do so?
> 

This may help you https://www.wikihow.com/Use-a-Computer-Mouse

Bloxigus (changes requested):
Several things, without looking at the code:
1) How do you handle when keybinds overlap
2) When they overlap how is the user supposed to know which item is associated with each keybind
3) Are the keybinds hiding any important details
4) How do you handle ships
5) How do you handle conflicting keybinds
6) Do you see anyone using this feature outside of yourself
7) Could this be a standalone mod
8) Thats not an improvement
9) Thats not something you touched

Bloxigus (changes requested):
You haven't changed anything???? Why would you request a review?????

platonvin:
1) I dont. Thats a problem, best fix is probably a warning like some other skyhanni features do
2) -> 1
3) ???
4) ???
5) -> 1
6) Yes. Some people prefer controlling ui with keyboard, thats the point
7) It could be a standalone mod, like most other features. But skyblock would pretty much be the only real use case
8) ???
9) ???

@hannibal002 quick ping - could you explain whats going on? Is there some background Im not aware of that makes me so unwelcome here? Why are they reacting so negatively?

platonvin:
Actually, i think im done. I dont want to contribute to a project where sarcastic mockery instead of reviews is tolerated

PR #4771, opened 2025-10-13
Fix + Remove: Walk Speed Rounding + Sneak Handling

lunaynx:
## What
Fixed Garden optimal speed calculation sometimes estimating the player's speed incorrectly due to two reasons:
* Floating point errors (e.g. 308 would register as 307.99999833106995, which is then rounded down to 307).
* Walking forwards while sprinting was incorrectly applying a 1.3x speed multiplier on 1.21.

Note: I purposely didn't change this to just round to the nearest integer. We intentionally round down, as due to how farming mechanics work, being slightly over the optimal speed is pretty much always preferable to being slightly under. Therefore, now we round to 2 decimal places first to avoid floating point errors. (Players can have fractional speeds such as 307.5 if they are not using Rancher's Boots, which we want to continue rounding down to 307.)

Also cleaned up the code in general, including removal of the sneak handling I added myself a while back, because I believe it was misguided and is more confusing than useful. For example, if the player wants to farm a 93 speed farm with Fermento Boots, which requires sneaking to achieve, it's better for them to just set their optimal speed to what they will see, which is 310, rather than SkyHanni trying to interpolate it and interpreting their speed as 93 due to sneaking. We don't do that for other things like Depth Strider or soul sand either.

## Changelog Fixes
+ Fixed Garden Optimal Speed sometimes reporting the wrong speed. - Luna

+ Removed special handling for sneaking in Garden Optimal Speed. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/GardenOptimalSpeed.kt
@@ -19,11 +19,13 @@ import at.hannibal2.skyhanni.utils.ItemUtils.getInternalNameOrNull
 import at.hannibal2.skyhanni.utils.LorenzColor
 import at.hannibal2.skyhanni.utils.NeuInternalName.Companion.toInternalName
 import at.hannibal2.skyhanni.utils.PlayerUtils
+import at.hannibal2.skyhanni.utils.RecalculatingValue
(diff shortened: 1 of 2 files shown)


lunaynx:
Actually, this needs further fixes. Sprinting is incorrectly affecting the reported speed. (Not my fault, but I still want to fix it in this PR.)

PR #4788, opened 2025-10-18
Backend: Add MCVersion class

lunaynx:
## What
Adds an `MCVersion` class for easier runtime comparison of Minecraft version numbers.

+ Added `MCVersion` class. - Luna


NopoTheGamer:
you should add a `PlatformUtils.isMcAbove("1.21.5")`  or something so people dont have to manually use it

PR #4799, opened 2025-10-19
Fix + Improvement: Foraging Tracker Forest Essence + faster update

lunaynx:
## What
Fixed Foraging Tracker not tracking Forest Essence due to Hypixel putting the formatting codes in a different place.

While we're at it, I also removed the ranged item logic, because these "ranged" rewards are just bugged to say "0-n", they always drop the highest amount listed (which is obviously intended behavior, because why would they not tell you the exact amount when it's what you *just* got). It was fixed by Hypixel once to not show the "0-" but then they unfixed it. The way I did it, it *should* work fine on old mod versions as well, since the ranged logic will just never trigger.

I also cleaned up the code a bit, including removing redundant regex tests. I left some without `§r` because I don't have a copy of an up-to-date message containing those rewards on hand, plus it's a good test to ensure we are being lenient enough with the formatting codes anyway.

## Changelog Improvements
+ Foraging Tracker now adds all Tree Gift rewards to the tracker immediately instead of waiting for a sack update. - Luna

## Changelog Fixes
+ Fixed Forest Essence from Tree Gifts not being tracked on Foraging Tracker. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/foraging/ForagingTrackerLegacy.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/foraging/ForagingTrackerLegacy.kt
@@ -113,33 +113,19 @@ object ForagingTrackerLegacy {
     )
 
     /**
-     * REGEX-TEST: §2Forest Essence §8x4
-     * REGEX-TEST: §2Forest Essence §8x6
-     * REGEX-TEST: §2Forest Essence §8x12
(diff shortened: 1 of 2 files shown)


Nealkitt (approved):

CalMWolfs (changes requested):
image

CalMWolfs (approved):

PR #4803, opened 2025-10-20
Fix: Craft Room Holographic Entities

Chissl:
## What
Fixes performance issues caused by the craft room holographic mob feature (it was scanning and adding every entity in the room to a list every frame for no reason). Also disabled head yaw on the mobs as that looked weird since we don't rotate the bodies

## Changelog Fixes
+ Fixed major performance issues in the Rift caused by Mirrorverse Holographic Entities. - Chissl


hannibal002 (approved):
works well in 1.8, seems like it never worked in 1.21

PR #4806, opened 2025-10-21
fix: use the proper config directory from forge

Rigner:
## What
FML passes the config directory to use in the preinit event, since it is configurable. SkyHanni didn't check that value before, it is now properly passed around.

+ Fix mod config not using Forge's configured config directory - Rigner

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
+++ b/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
@@ -50,6 +50,7 @@ import net.minecraft.client.gui.GuiScreen
 import org.apache.logging.log4j.Level
 import org.apache.logging.log4j.LogManager
 import org.apache.logging.log4j.Logger
+import java.io.File
 import kotlin.coroutines.cancellation.CancellationException
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
@@ -57,12 +58,16 @@ import kotlin.time.Duration.Companion.seconds
 @SkyHanniModule
 object SkyHanniMod {
 
-    fun preInit() {
+    fun preInit(configDirectory: File?) {
         PlatformUtils.checkIfNeuIsLoaded()
 
         LoadedModules.modules.forEach { SkyHanniModLoader.loadModule(it) }
 
         SkyHanniEvents.init(modules)
+
+        if (configDirectory != null) {
+            ConfigManager.configDirectory = File(configDirectory, "skyhanni")
+        }
 
         PreInitFinishedEvent.post()
     }
--- a/src/main/java/at/hannibal2/skyhanni/SkyHanniModLoader.kt
+++ b/src/main/java/at/hannibal2/skyhanni/SkyHanniModLoader.kt
@@ -20,7 +20,7 @@ class SkyHanniModLoader {
     @Mod.EventHandler
     fun preInit(event: FMLPreInitializationEvent?) {
         HotswapSupport.load()
-        SkyHanniMod.preInit()
+        SkyHanniMod.preInit(event?.modConfigurationDirectory)
         loadedClasses.clear()
     }
 
--- a/src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
@@ -57,7 +57,7 @@ class ConfigManager {
 //             .registerIfBeta(FeatureTogglesByDefaultAdapter)
             .create()
 
-        val configDirectory = File("config/skyhanni")
+        var configDirectory = File("config/skyhanni")
     }
 
     private val logger = LorenzLogger("config_manager")
--- a/versions/1.21.5/src/main/java/at/hannibal2/skyhanni/SkyHanniModLoader.kt
+++ b/versions/1.21.5/src/main/java/at/hannibal2/skyhanni/SkyHanniModLoader.kt
@@ -2,11 +2,12 @@ package at.hannibal2.skyhanni
 
 import at.hannibal2.skyhanni.SkyHanniMod.modules
 import net.fabricmc.api.ModInitializer
+import net.fabricmc.loader.api.FabricLoader;
 
 class SkyHanniModLoader : ModInitializer {
 
     override fun onInitialize() {
-        SkyHanniMod.preInit()
+        SkyHanniMod.preInit(FabricLoader.getInstance().getConfigDir().toFile())
         SkyHanniMod.init()
         loadedClasses.clear()
     }


NopoTheGamer:
1. this doesnt build 2. doesnt this wipe the config of anyone effected by the bug? 3. why?

Rigner:
Hey,

~~It builds, I tested it locally and verified the fix works as intended with another person as well🤔 Even the github action seems to like it...~~

Technically yes, it does, but I'm pretty sure noone is affected by this issue **yet**.

Besides this being needed by the upcoming (next few weeks?) Forge support in Lunar Client, which will bring a ton of new users to these skyblock mods, it is still a good practice to use the correct config directory. Just that last part should be enough.

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
```diff
@@ -50,20 +50,25 @@ import net.minecraft.client.gui.GuiScreen
...
+import java.io.File
 import kotlin.coroutines.cancellation.CancellationException
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
 
 @SkyHanniModule
 object SkyHanniMod {
 
-    fun preInit() {
+    fun preInit(configDirectory: File?) {
```
CalMWolfs, line 61:
```suggestion
    fun preInit(configDirectory: File? = null) {
```

CalMWolfs (changes requested):
required for multiversion

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
Rigner: Let me add proper support for Fabric config directory as well

CalMWolfs: > Let me add proper support for Fabric config directory as well

Would this here break the config for existing users? As it sounds like the 1.8 change wont break anything as very few people would be using whatever you guys are testing right now but there are numerous people who use skyhanni on fabric with lunar already

NopoTheGamer:
> Besides this being needed by the upcoming (next few weeks?) Forge support in Lunar Client, which will bring a ton of new users to these skyblock mods

lets hope it works better than fabric on lunar client...

Rigner:
Should be good now with proper Fabric support!


> > Besides this being needed by the upcoming (next few weeks?) Forge support in Lunar Client, which will bring a ton of new users to these skyblock mods
> 
> lets hope it works better than fabric on lunar client...

Feel free to report any issues 😄 but yeah Forge support, even though it was more tricky (mainly because of Java 8 vs 17), should be more stable since 1.8.9 / 1.12.2 are old updates with no new content!

hannibal002:
Hey, I'm pleased to see technical bug fixes by increasing our use of good practices in code.
Though, I'm unsure if this is necessary here.

Ignoring the Lunar Client update you mentioned (I go into this further below), this change has no benefit for SkyHanni users but has a very slight chance to reset their data entirely. If we want to go forward with this PR, we should at least enforce some sanity check. E.g., confirming if the folder does not yet exist but the old, hard-coded path does. Then we copy/migrate the old folder over to avoid data loss for the user.
That's the only big point I see wrong here.


Additional thoughts and questions I have independently of this PR, or even SkyHanni:

Is the default (and properly used) config location for either Fabric or Forge something else than `.minecraft/config/<mod-id>`?

How/why does an upcoming Lunar Client update require this change? Are you planning on changing the default path? If so, why and to what path?
This looks like creating a problem if users just want to drag and drop their folders around between different profiles or launchers.

Why do you want to deny this option for mods generally? Why not let mods create folders wherever they want? Otherwise, you break every mod that doesn't follow this more strict enforcement- again, for no benefit of the user, afaik.

Thanks for clarifying!

Rigner:
Hey, these are indeed legit questions!

I can definitely try to add securities in the code to not use the "correct" directory if a config exists in the other path (literally a one line change).

You're right, the default path is always `.minecraft/config/` in both Forge and Fabric, but can be configured.

As per why it's needed by Lunar Client, It's because the process is running in another directory (specific to LC), but configured to use `.minecraft` files as the game directory (using the `--gamedir` args for example), this is a common practice by launchers supporting many versions or modpacks, to keep some runtime files separate.
I'm not sure about how it used to work but It's possible this feature was originally made for ready-to-use modpacks with preconfigured settings, so they do not conflict with the user's. To be confirmed.

Mods can use their own directory, sure, but I believe this is the recommended way to go. We tested quite a bunch of mods (both for skyblock and "custom content") and SkyHanni was part of the few ones with that issue.

hannibal002:
Thanks for the reply!

I've looked a bit closer into this topic. Now I'm confused by Lunar's design choice here.

Instead of trying to implement the "fix" in the bunch of mods that use hard-coded paths, how about normalizing the different paths on Lunar's end? That's what Modrinth and Prism/MultiMC do (and vanilla MC, but who counts them), so proper instancing or security can't be an issue.

Are there any other reasons you decided to split up `process dir`, `gameDir` and `instance folder`?

Adding support for this on our end now feels odd if there is nothing broken right now and no need to change anything on your end either.

j10a1n15:
> Feel free to report any issues 😄 

Hi, so I’ve reported that kotlin mods which use an object as their main class don’t load twice (about 3 months and 2 months ago respectively) and have been told twice that it’s been fixed (it still hasn’t). When changing the mod to [use a class instead]( https://github.com/meowdding/SkyCubed), it loads fine. The Ticket id of the report with a bit more information is #207118. Could you please look into it? 🥺

Rigner:
> Thanks for the reply!
> 
> I've looked a bit closer into this topic. Now I'm confused by Lunar's design choice here.
> 
> Instead of trying to implement the "fix" in the bunch of mods that use hard-coded paths, how about normalizing the different paths on Lunar's end? That's what Modrinth and Prism/MultiMC do (and vanilla MC, but who counts them), so proper instancing or security can't be an issue.
> 
> Are there any other reasons you decided to split up `process dir`, `gameDir` and `instance folder`?
> 
> Adding support for this on our end now feels odd if there is nothing broken right now and no need to change anything on your end either.

We're discussing that internally at the moment, I'll report back with any info! If we can just move the `cwd` to `.minecraft`, without causing any issues (e.g. running multiple instances of the game or other mod conflicts), we'll try that approach instead.

> > Feel free to report any issues 😄
> 
> Hi, so I’ve reported that kotlin mods which use an object as their main class don’t load twice (about 3 months and 2 months ago respectively) and have been told twice that it’s been fixed (it still hasn’t). When changing the mod to [use a class instead](https://github.com/meowdding/SkyCubed), it loads fine. The Ticket id of the report with a bit more information is #207118. Could you please look into it? 🥺

Thanks, I'll take a look asap!

NopoTheGamer:
are you still interested in this pr?

PR #4807, opened 2025-10-22
Fix: Weird tuba ability cooldown on 1.21

NopoTheGamer:
## What
idk the sound is different 

## Changelog Fixes
+ Fixed Weird Tuba ability cooldown in 1.21. - nopo

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/itemabilities/abilitycooldown/ItemAbilityCooldown.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/itemabilities/abilitycooldown/ItemAbilityCooldown.kt
@@ -160,7 +160,7 @@ object ItemAbilityCooldown {
                 }
             }
             // Weird Tuba & Weirder Tuba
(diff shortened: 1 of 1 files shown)


CalMWolfs (approved):

lunaynx:
Yeah, I noticed this was broken too, but how is it not already handled by this?

<https://github.com/hannibal002/SkyHanni/blob/e2bdf017a5b0a723f27fcf96b25dd1d2d16c5a8c/src/main/java/at/hannibal2/skyhanni/utils/compat/SoundCompat.kt#L39>

NopoTheGamer:
> Yeah, I noticed this was broken too, but how is it not already handled by this?
> 
> https://github.com/hannibal002/SkyHanni/blob/e2bdf017a5b0a723f27fcf96b25dd1d2d16c5a8c/src/main/java/at/hannibal2/skyhanni/utils/compat/SoundCompat.kt#L39

it does firstOrNull and theres 2 mappings so the 2nd one is useless

PR #4811, opened 2025-10-23
Fix: Ping API always showing network charts on 1.21.9+

lunaynx:
## What
Fixes Ping API being enabled causing network and ping charts to always show while F3 is open on 1.21.9/1.21.10.

## Changelog Fixes
+ Fixed not being able to hide network and ping charts on the F3 Debug Screen in 1.21.9/1.21.10. - Luna


CalMWolfs (approved):
worked for me

PR #4820, opened 2025-10-28
Fix: Pest Vinyl Stuff

lunaynx:
## What
Fixed various issues related to active pest vinyl detection and refactored the code a bit.

The nullability changes aren't arbitrary, the idea is:
* The stored active vinyl should never be `null`, it is better to just default to `VinylType.NONE` if we don't know, `null` suggests an error state.
* The `getByNameOrNull` and `getByInternalNameOrNull` methods can return `null`, because there is a distinction:
  * A return value of `VinylType.NONE` means "the inputted name represents the state where no vinyl is active"
  * A return value of `null` means "the inputted name does not represent any known vinyl"

## Changelog Fixes
+ Fixed Stereo Harmony Display not updating the active vinyl when changed outside the Garden or when SkyHanni is not installed. - Luna
+ Fixed Replace Menu Icons in Stereo Harmony sometimes highlighting the wrong vinyl as active. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
@@ -594,7 +594,7 @@ class ProfileSpecificStorage(
         var pestProfitTracker: PestProfitTracker.BucketData = PestProfitTracker.BucketData()
 
         @Expose
-        var activeVinyl: VinylType = VinylType.NONE
+        var activeVinyl: VinylType? = null
     }
 
     // - gui
--- a/src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
@@ -594,7 +594,7 @@ class ProfileSpecificStorage(
         var pestProfitTracker: PestProfitTracker.BucketData = PestProfitTracker.BucketData()
 
         @Expose
-        var activeVinyl: VinylType? = null
+        var activeVinyl: VinylType = VinylType.NONE
     }
 
     // - gui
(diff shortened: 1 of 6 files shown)


src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
```diff
@@ -594,7 +594,7 @@ class ProfileSpecificStorage(
         var pestProfitTracker: PestProfitTracker.BucketData = PestProfitTracker.BucketData()
 
         @Expose
-        var activeVinyl: VinylType? = null
+        var activeVinyl: VinylType = VinylType.NONE
```
CalMWolfs: This should still be marked as being possible to be null, otherwise I think it will wipe configs

CalMWolfs (changes requested):
The rest seemed to work fine

src/main/java/at/hannibal2/skyhanni/features/garden/pests/stereo/VinylType.kt
```diff
@@ -13,19 +13,18 @@ enum class VinylType(val displayName: String) {
...
+            VinylType.entries.firstOrNull { it.displayName == name }
 
-        fun getByName(name: String) = VinylType.entries.firstOrNull { it.displayName == name } ?: NONE
-        fun getByInternalNameOrNull(internalName: NeuInternalName) = when (internalName) {
-            BEETLE_VINYL -> NOT_JUST_A_PEST
-            else -> VinylType.entries.firstOrNull {
-                internalName.asString() == "VINYL_${it.name}"
-            }
-        }
+        fun getByInternalNameOrNull(internalName: NeuInternalName): VinylType? =
```
hannibal002, line 27: what happened to BEETLE_VINYL -> NOT_JUST_A_PEST ?

hannibal002 (changes requested):
code looks alright

src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
hannibal002: ive removed none. doesnt seem to have any benefit differentiating between unknown, buggy, and known as being not avtive

NopoTheGamer: i think you put the comment on the wrong class

src/main/java/at/hannibal2/skyhanni/features/garden/pests/stereo/VinylType.kt
lunaynx: That's handled by `internalNameOverride`, see lines 6 and 16.

PR #4842, opened 2025-11-03
Fix: Revert #4652 and its "fixes"

Chissl:
## What
This PR broke block click event, and its fixes didn't actually fix the issue; the first fix caused double counting on 1.21, and the second one caused the event to sometimes not be posted. No relevant features use this so I don't see the point of keeping this in until it can be extensively tested and fixed for all game versions.

## Changelog Fixes
+ Fixed crop milestone bps display and other features that relied on block break detection. - Chissl

+ Restored BlockBreakEvent and ItemClickEvent to their functionality before #4652. - Chissl


Thunderblade73 (changes requested):
Nice. I sure hope we will have fun with the second ban wave (you reverted the change that prevented the banning for our clicking)

Chissl:
> Nice. I sure hope we will have fun with the second ban wave (you reverted the change that prevented the banning for our clicking)

That pr took over a month to merge so it clearly was not that important

lunaynx:
It wasn't that important because there is no other feature currently in SkyHanni that cancels left click interactions with an item, but it's a ticking time bomb. (And also because Hannibal didn't understand it was an important fix so it was in the 6.0.0 milestone for a while.)

At the very least if you're going to revert it you should prevent cancellation of the event on left clicks for now.

Chissl:
> It wasn't that important because there is no other feature currently in SkyHanni that cancels left click interactions with an item, but it's a ticking time bomb. (And also because Hannibal didn't understand it was an important fix so it was in the 6.0.0 milestone for a while.)
> 
> At the very least if you're going to revert it you should prevent cancellation of the event on left clicks for now.

Fair enough, I cba to do that so I'll just close this and hope one of the next 5 fixes to this actually works

NopoTheGamer:
> (And also because Hannibal didn't understand it was an important fix so it was in the 6.0.0 milestone for a while.)

shame the important fix doesnt work whatsoever

Chissl:
> > (And also because Hannibal didn't understand it was an important fix so it was in the 6.0.0 milestone for a while.)
> 
> shame the important fix doesnt work whatsoever

If the event doesn't post you can't cancel it

PR #4845, opened 2025-11-04
Fix: Update CroesusChestTracker patterns

Thunderblade73:
## What
idk if it did get changed for all 3 patterns but the first one definitively uses a small c.

## Changelog Fixes
+ Fixed Croesus Chest Tracker after Hypixel update. - Thunderblade73


Pugzy:
I believe the regexes you have updated are the fallbacks, they would need to be updated in the [regex section of the repo](https://github.com/hannibal002/SkyHanni-REPO/blob/main/constants/regexes.json#L317-L319) too.

Bloxigus:
> I believe the regexes you have updated are the fallbacks, they would need to be updated in the [regex section of the repo](https://github.com/hannibal002/SkyHanni-REPO/blob/main/constants/regexes.json#L317-L319) too.

not how this works (it gets automagically updated on merge)

CalMWolfs (approved):

PR #4847, opened 2025-11-06
Fix: Singular Croesus Pattern

Fazfoxy:
## What
This would cause F0 to get saved into config since the pattern  didn't match

exclude_from_changelog


CalMWolfs (approved):
Im gonna assume this doesn't break anything

PR #4850, opened 2025-11-07
Fix: NullPointerException with DianaV2

Bloxigus:
## Changelog Fixes
+ Fixed an error when reading Diana repo data. - bloxigus


hannibal002:
is this the `Cannot invoke "kotlin.Lazy.getValue()" because "<local1>" is null` error?

PR #4859, opened 2025-11-09
Backend: Removed hoppity egg field

hannibal002:
## What
`eggLocations` is useless since we have the data duped in `apiEggLocations`.
Currently, we need to add new egg locations at two spots because of this.
With this PR, the field `eggLocations` becomes redundant, and we do not need to update the repo anymore once this PR gets into a full version.

+ Cleanup in HoppityEggLocations. - hannibal2


CalMWolfs (approved):
worked for me

PR #4867, opened 2025-11-14
Backend: Use fabric chat events

NopoTheGamer:
## What
the pr worked in my small testing, there might be possible issues with other mods
the old system caused issues with other mods (so sorry sbo for causing the double mob count 🙂 )

+ Use fabric chat events instead of a mixin. - nopo


CalMWolfs (approved):
seemed to work

PR #4871, opened 2025-11-14
Fix: Tooltip Color Stripping

lunaynx:
## What
Fixes modern tooltip color stripping code to strip `§5§o` instead of just `§5`. The latter happens to work fine in vanilla because Hypixel starts every line with a color code, which also happens to reset the `§o` formatting code, but can break if certain mods, e.g. Catharsis, modify the tooltip to strip the `§5§o` prefix.

This is a quick band-aid fix and does not address several other underlying issues with the implementation (e.g. breaking modern components like fonts, the scuffed check for modified lines, and potentially incorrectly stripping formatting once again if Hypixel *did* ever genuinely start a lore line with `§5§o`).

## Changelog Fixes
+ Fixed some features incorrectly stripping purple color from item tooltips when used with certain mods. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/ToolTipData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/ToolTipData.kt
@@ -24,8 +24,8 @@ object ToolTipData {
     //$$     stack: ItemStack,
     //$$     originalToolTip: MutableList<Text>,
     //$$ ): MutableList<Text> {
-    //$$     val tooltip = originalToolTip.map { it.formattedTextCompatLessResets().removePrefix("§5") }.toMutableList()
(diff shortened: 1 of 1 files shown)


CalMWolfs (changes requested):
With your pr
Image

Without your pr
Image

Upgrade Price does not work anymore, this is because none of the tooltips have `§5§o` and they all have `§5` like the code currently handles

`original: §f§o§aComposter Speed VII, §5§7Increases composter production, §5§7speed by §a140%§7., §5, §5§7Next Tier: §a160%, §5, §5§7Upgrade Cost:, §5  §9Enchanted Golden Carrot §8x16, §5  §9Cropie §8x3, §5  §c500 Copper, §5, §5§8+§b2 SkyBlock XP, §5, §5§cMissing required items!`

PR #4873, opened 2025-11-15
Fix: Disable Farming Fortune warnings if display is disabled

lunaynx:
## What
Don't send warnings about missing Farming Fortune/Crop Fortune if Farming Fortune Display is disabled.

## Changelog Fixes
+ Fixed warnings about missing Farming Fortune in tab list being sent even if Farming Fortune Display is disabled. - Luna


NopoTheGamer (changes requested):
The data seems to be used for more than just this display (GardenCropMilestoneDisplay) i think maybe it doesnt check is enabled for a reason

lunaynx:
Fair, someone just wanted a way to not get warned because they weren't using the feature anyway. I'll take a look later if this is salvageable by adding more checks or something.

NopoTheGamer:
> Fair, someone just wanted a way to not get warned because they weren't using the feature anyway. I'll take a look later if this is salvageable by adding more checks or something.

i think ideally the data collection wouldnt be inside of a feature file

PR #4877, opened 2025-11-17
Improvement: Entity Distance Performance

ItsEmpa:
## What
This PR makes it so that the functions that get entities within a distance internally first filter using the bounding box function I added on another PR. From my testing, this seemed to improve the actual performance of the function by roughly 10x.

## Changelog Improvements
+ Improved performance of Mob Detection and other features that detect entities by 10x. - Empa


NopoTheGamer (approved):
seems to work, didnt test performance

PR #4885, opened 2025-11-22
Fix: Basic Kuudra Key cost not being added

Fazfoxy:
## What
This should fix the kuudra key part of https://discord.com/channels/997079228510117908/1441587782295490600 I'll look into wheel of fate when I'm not playing a different thing

## Changelog Fixes
+ Fixed Instance Chest Profit ignoring Basic Kuudra Key Cost. - Fazfoxy


NopoTheGamer:
i love exclude from changelog for fixes??

CalMWolfs (changes requested):
can we get a regex test for the string that this now matches

CalMWolfs (approved):

PR #4887, opened 2025-11-22
Fix: Wheel of Fate being erroneously added as profit

Fazfoxy:
## What
I don't wanna keep running solo Kuudra to test these, I'm gonna just hope this one works
Also this changelog is possibly too vague? but the last idea I had was too long

## Changelog Fixes
+ Fixed Wheel of Fate being erroneously added as chest profit. - Fazfoxy

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/combat/InstanceChestProfit.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/combat/InstanceChestProfit.kt
@@ -126,7 +126,6 @@ object InstanceChestProfit {
         val itemsWithCost: MutableMap<String, Double> = mutableMapOf()
         items.forEach {
             if (fakeWheelofFate.matches(it.value.displayName)) return
-                // This is intended to check against Wheel of Fate in Kuudra chests since it both passes the item name check & is in the chest Type
                 if (it.value.getInternalNameOrNull() != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/combat/InstanceChestProfit.kt
```diff
@@ -116,23 +116,26 @@ object InstanceChestProfit {
...
-                itemsWithCost.addOrPut(it.value.displayName, price)
-            }
-            essencePattern.matchMatcher(it.value.displayName) {
-                val name = group("name")
-                val rawCount = group("count").toInt()
-                val count = if (name == "Crimson") rawCount * (1 + getKuudraEssenceBonus())
-                else rawCount.toDouble()
-                val price = count * (NeuInternalName.fromItemName(name).getPriceOrNull(config.priceSource) ?: 0.0)
-                itemsWithCost.addOrPut(it.value.displayName, price)
+            if (items[51]?.equals(it) == false) {
```
lunaynx: Pretty sure this is supposed to be `it.value`. But you can also simplify the check and change it to an early return:
```kt
if (it.key == 51) return@forEach
```
(This is equivalent to a `continue` in a normal `for` loop.)

Could also make it check *only* specific slots instead, but I'm not sure if you ever get reward items in multiple rows.

lunaynx (dismissed):

PR #4888, opened 2025-11-23
Improvement: Stop Metal Detector Solver After All Tools Found

AfkUserMC:
## What
Adds a toggle to stop looking for treasure after you have already found all scavenged tools in the Mines of Divan
(code is heavily based on (pretty much identical to) features.mining.crystalhollows.MetalDetectorAllToolsAlert because i am. not good at coding)

## Why
makes my adhd brain happy :3

## Very important to consider
this change makes the cyclomatic complexity of the metal detector function 26, 1 over the limit.
currently, the warning is suppressed. i really need to fix it but right now i'm in a limbo of just looking at it and confusing myself

-# ~~would it just be better if i made this a draft for the time being~~

## Changelog Improvements
+ Added a toggle to stop Metal Detector Solver when having all tools. - AfkUserMC

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/ConfigUpdaterMigrator.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/ConfigUpdaterMigrator.kt
@@ -11,7 +11,7 @@ import com.google.gson.JsonPrimitive
 object ConfigUpdaterMigrator {
 
     val logger = LorenzLogger("ConfigMigration")
-    const val CONFIG_VERSION = 110
+    const val CONFIG_VERSION = 111
     fun JsonElement.at(chain: List<String>, init: Boolean): JsonElement? {
         if (chain.isEmpty()) return this
         if (this !is JsonObject) return null
--- a/src/main/java/at/hannibal2/skyhanni/config/ConfigUpdaterMigrator.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/ConfigUpdaterMigrator.kt
@@ -11,7 +11,7 @@ import com.google.gson.JsonPrimitive
 object ConfigUpdaterMigrator {
 
     val logger = LorenzLogger("ConfigMigration")
-    const val CONFIG_VERSION = 111
+    const val CONFIG_VERSION = 110
     fun JsonElement.at(chain: List<String>, init: Boolean): JsonElement? {
         if (chain.isEmpty()) return this
         if (this !is JsonObject) return null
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/features/mining/crystalhollows/MetalDetectorSolver.kt
```diff
@@ -66,17 +69,35 @@ object MetalDetectorSolver {
...
+            var hasLapis = false
+            var hasDiamond = false
+            var hasEmerald = false
+            var hasGold = false
+            InventoryUtils.getItemsInOwnInventory().forEach {
+                val internalName = it.getInternalName()
+                if (internalName == DWARVEN_LAPIS_SWORD) hasLapis = true
+                if (internalName == DWARVEN_DIAMOND_AXE) hasDiamond = true
+                if (internalName == DWARVEN_EMERALD_HAMMER) hasEmerald = true
+                if (internalName == DWARVEN_GOLD_HAMMER) hasGold = true
```
ItsEmpa: use a when statement instead of so many if statements
```kt
when (internalName) {
	DWARVEN_LAPIS_SWORD -> hasLapis = true
	DWARVEN_DIAMOND_AXE -> hasDiamond = true
	DWARVEN_EMERALD_HAMMER -> hasEmerald = true
	DWARVEN_GOLD_HAMMER -> hasGold = true
}
```

something like this

ItsEmpa (changes requested):

src/main/java/at/hannibal2/skyhanni/config/ConfigUpdaterMigrator.kt
```diff
@@ -11,7 +11,7 @@ import com.google.gson.JsonPrimitive
 object ConfigUpdaterMigrator {
 
     val logger = LorenzLogger("ConfigMigration")
-    const val CONFIG_VERSION = 110
+    const val CONFIG_VERSION = 111
```
NopoTheGamer: why are you changing the config version?

NopoTheGamer (changes requested):
the feature worked for me
i had to do a nuc run to test this and got [SkyHanni] +Tracker Drop: ✧ Flawless Topaz Gemstone (2.3M) :( alloy when

src/main/java/at/hannibal2/skyhanni/config/ConfigUpdaterMigrator.kt
AfkUserMC: sorry, i was under the impression additions to the configuration update the version.
Was it already updated?

NopoTheGamer: its only used if you use the config migration

PR #4897, opened 2025-11-27
Improvement: Add Aura Mayor

lunaynx:
## What
Added Aura as a valid mayor, with perks from API. Did not add any logic for mob HP yet, as we don't have a crystal ball to know if/how Hypixel will round the numbers.

## Changelog Improvements
+ Added support for Aura mayor. - Luna


NopoTheGamer:
do you know if this will error if the perks change?

lunaynx:
Not sure.

PR #4899, opened 2025-11-27
Improvement: Add NPC sell toggle for Not Clickable Items

lunaynx:
## What
Adjusted the Not Clickable Items feature so that by default, in NPC trade menus, only items that cannot be sold or are recombobulated are grayed out, but added a default off toggle to also gray out items that "should not" be sold (our whitelist for that is quite outdated and hasn't been touched in ages, but it can be improved in the future).

## Changelog Improvements
+ Improved the "Not Clickable Items" feature in NPC trade menus. - Luna
    * By default, it only grays out items that cannot be sold to NPCs, or are recombobulated.
    * You can re-enable graying out items that are not commonly sold to NPCs in the config.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/HideNotClickableItems.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/HideNotClickableItems.kt
@@ -20,6 +20,7 @@ import at.hannibal2.skyhanni.test.command.ErrorManager
 import at.hannibal2.skyhanni.utils.InventoryUtils
 import at.hannibal2.skyhanni.utils.InventoryUtils.getLowerItems
 import at.hannibal2.skyhanni.utils.ItemCategory
+import at.hannibal2.skyhanni.utils.ItemPriceUtils.getNpcPriceOrNull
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/HideNotClickableItems.kt
```diff
@@ -482,25 +485,31 @@ object HideNotClickableItems {
...
+            hideReason = "This item cannot be sold at the NPC! (Donated to Museum)"
+            return true
+        }
+
+        if (ItemUtils.isRecombobulated(stack)) {
+            hideReason = "This item should not be sold at the NPC! (Recombobulated)"
+            return true
+        }
 
+        if (config.protectRarelySoldItems) {
```
hannibal002: could use early return

hannibal002 (changes requested):
i like the changes!

for the future, we could compare npc price with bz/ah price and only show if the diff per/percentage wise is "worth enough" to sell via npc instead

PR #4900, opened 2025-11-28
Improvement + Fix: More Entities Performance

ItsEmpa:
## What
Refactors a lot of features that use entities for more performance.

Mainly makes features either use the new functions for getting entities in bounding boxes, or makes them use entity events instead. Also adds TODOs for some other features that would be too complex to change here.

Haven't tested it yet, but i will use this pr's build for the next few days to test it.

## Changelog Improvements
+ Improved performance of entity features. - Empa

## Changelog Fixes
+ Fixed Fire Pillar Display disappearing sometimes wrongly. - Empa


src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorFeatures.kt
Thunderblade73, line 1: I think both usages here can use ``MobData.displayNpcs`` and query by name of the mob.

ItsEmpa: mob detection can be a bit brittle sometimes, i think its fine to use it like this

also, I would need to handle the jacob visitor thing then, and id rather just keep the logic here closer to the original but optimized in this case

hannibal002:
I've added this missing line to the changelog:

+ Fixed Fire Pillar Display disappearing sometimes wrongly. - Empa

image

hannibal002 (dismissed):
Please review my changes  @ItsEmpa

NopoTheGamer (approved):
lets hope empa is good at testing

PR #4908, opened 2025-11-29
Backend: Add @Hidden, make @OnlyLegacy/@OnlyModern mutually exclusive

lunaynx:
## What
Banned the hack of using `@OnlyLegacy` and `@OnlyModern` simultaneously to hide config options, and instead added a dedicated `@Hidden` annotation for it.

+ Added `@Hidden` annotation to hide config options from the GUI. - Luna
    * They can still be edited via `/shconfig`.
+ Using `@OnlyLegacy` and `@OnlyModern` together on the same config option now throws a compile-time (and runtime) error. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
@@ -317,16 +317,11 @@ class BlockingMoulConfigProcessor : MoulConfigProcessor<Features>(SkyHanniMod.fe
 
         val onlyLegacy = field.isAnnotationPresent(OnlyLegacy::class.java)
         val onlyModern = field.isAnnotationPresent(OnlyModern::class.java)
-        if (PlatformUtils.IS_LEGACY) {
-            if (onlyModern) {
-                if (onlyLegacy) throw IllegalStateException("@OnlyLegacy and @OnlyModern cannot be used together on the same option")
-                return GuiOptionEditorHidden(default)
-            }
-        } else {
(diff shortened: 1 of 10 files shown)


src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -311,12 +311,20 @@ class BlockingMoulConfigProcessor : MoulConfigProcessor<Features>(SkyHanniMod.fe
...
 
+        if (field.isAnnotationPresent(Hidden::class.java)) {
+            return GuiOptionEditorHidden(default)
+        }
+
+        val onlyLegacy = field.isAnnotationPresent(OnlyLegacy::class.java)
+        val onlyModern = field.isAnnotationPresent(OnlyModern::class.java)
         if (PlatformUtils.IS_LEGACY) {
-            if (field.isAnnotationPresent(OnlyModern::class.java)) {
+            if (onlyModern) {
```
ItsEmpa: cant you just move the check to be outside the PlatformUtils.IS_LEGACY if statement? and you just do

```kt
if (onlyLegacy && onlyModern) // throw exception

if (PlatformUtils.IS_LEGACY) {
// ...
```

ItsEmpa (changes requested):

NopoTheGamer:
not a fan of my ideas i see... Anyway i was going to replace this code soon with 1 annotation that uses the MCVersion class

ItsEmpa (approved):

PR #4911, opened 2025-12-02
Feature: add an option to force diana

SteinGaming:
## What
Adds an option to forcibly enable Diana. This just adds the config option in the ``Events -> Diana`` category and modifies the ``isDoingDiana`` function.

This does allow any player to use Diana features when none is detected, but also might cause some negligible performance usages (e.g. the player enables the option but doesn't disable it afterwards).

Only tested on 1.21.8, but should work on any. Defaults to being disabled.

<details>
<summary>Images</summary>
image
</details>

## Changelog New Features
+ Enable toggle to forcibly enable Diana. - SteinGaming

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaApi.kt
@@ -32,7 +32,7 @@ object DianaApi {
 
     fun hasGriffinPet() = CurrentPetApi.isCurrentPet("Griffin")
 
-    fun isDoingDiana() = IslandType.HUB.isCurrent() && isRitualActive() && hasSpadeInInventory()
+    fun isDoingDiana() = SkyHanniMod.feature.event.diana.forceEnabled || (IslandType.HUB.isCurrent() && isRitualActive() && hasSpadeInInventory())
 
     val ItemStack.isDianaSpade get() = getInternalName() in spades
 
--- a/src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaApi.kt
@@ -27,8 +27,9 @@ object DianaApi {
 
     fun hasSpadeInHand() = InventoryUtils.itemInHandId in spades
 
-    private fun isRitualActive() = SkyHanniMod.feature.event.diana.forceEnabled || ((Perk.MYTHOLOGICAL_RITUAL.isActive || Perk.PERKPOCALYPSE.isActive) ||
-        SkyHanniMod.feature.dev.debug.assumeMayor.get() == ElectionCandidate.DIANA)
+    private fun isRitualActive() = (Perk.MYTHOLOGICAL_RITUAL.isActive || Perk.PERKPOCALYPSE.isActive) ||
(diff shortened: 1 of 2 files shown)


SteinGaming:
Fixed the changelog text, whoops

lunaynx:
There is no real reason to add this as a toggle, we already have an Assume Mayor option. (It *might* break some mob detection logic due to it no longer seeing Aura as a mayor, but it doesn't appear to break Diana mob detection at least.)

src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaApi.kt
```diff
@@ -32,7 +32,7 @@ object DianaApi {
 
     fun hasGriffinPet() = CurrentPetApi.isCurrentPet("Griffin")
 
-    fun isDoingDiana() = IslandType.HUB.isCurrent() && isRitualActive() && hasSpadeInInventory()
+    fun isDoingDiana() = SkyHanniMod.feature.event.diana.forceEnabled || (IslandType.HUB.isCurrent() && isRitualActive() && hasSpadeInInventory())
```
Bloxigus: If you are going to do this correctly, put it in `isRitualActive()`. I however doubt this pr will be of any use given the limited nature of this event, among other things.

SteinGaming:
> There is no real reason to add this as a toggle, we already have an Assume Mayor option. (It _might_ break some mob detection logic due to it no longer seeing Aura as a mayor, but it doesn't appear to break Diana mob detection at least.)

Hmmm, but it's hidden inside of dev options (that's why I didn't even try to search there for an existing feature). IMO, no user would find it by themselves, except if they got the instructions from someone/somewhere.

src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaApi.kt
SteinGaming: > If you are going to do this correctly, put it in isRitualActive().

Fixed, thanks!

> I however doubt this pr will be of any use given the limited nature of this event, among other things.

I think this is also for future-proofing. The devs have proven, that they can (and will) arbitrarily have mayor features without their mayor and even event active. I agree that this very, very likely become useless after the next election, so it's up to you on what you want to do.

lunaynx:
> Hmmm, but it's hidden inside of dev options (that's why I didn't even try to search there for an existing feature). IMO, no user would find it by themselves, except if they got the instructions from someone/somewhere.

That's fair, it may actually make sense to add this, but then ideally we expose similar settings for all relevant mayors in the future.

For what it's worth, I do plan to add a feature to remotely force enable certain mayor perks for all players via the SkyHanni repo (will require a mod update the first time around).

SteinGaming:
> For what it's worth, I do plan to add a feature to remotely force enable certain mayor perks for all players via the SkyHanni repo (will require a mod update the first time around).

Yeah, that will completely eliminate the need of this option lol. But this is just supposed to be an "emergency" option for others who are currently wanting to do the mythological event. 

Anyway, I have another idea to just make this PR completely useless: Make an announcement on the discord, so everyone knows how to temporarily enable this option (and how to undo it)!
There's probably already players telling others how to enable it, but that's not going to reach everybody.

NopoTheGamer:
this option used to exist and got replaced with assume mayor

PR #4919, opened 2025-12-08
Fix: Slayer Cost not being counted

Fazfoxy:
## What
Admins Blew up all of Aura's perks in API breaking the smarter version of this (also Aura might just, lose this perk? In the future, so I'm doing this instead) and at worst this gives the Slayer spawn feature a slightly larger margin of error for the future

## Changelog Fixes
+ Fixed Slayer Spawn Cost not Counting during Aura. - fazfoxy


NopoTheGamer:
you love exclude from changelog

Fazfoxy:
> you love exclude from changelog

does this make you more happy

PR #4921, opened 2025-12-09
Fix: Fix tooltip format feature

Obsidianninja11:
## What
Fixes the Tooltip Format feature so that default actually turns it off.
This feature was [hiding gemstone fortune](https://discord.com/channels/997079228510117908/1094190239532208228/1447835166847533086), which is fixed, but it's still using a method that could very easily hide stats in future updates when this feature is turned on.

Also moves the tool fortune in the FF breakdown feature to be with the other crop specific stats.

I haven't tested this in game, but it's pretty minor changes, so it should work.

<details>
<summary>Images</summary>

image

</details>

## Changelog Fixes
+ Fixed farming tool tooltip format hiding gemstones and not having an off option. - Obsidian


NopoTheGamer:
i have just commented out this function in beta, idk what to do with this pr

NopoTheGamer:
im just going to close this pr, someone will have to fully rewrite the feature before it gets enabled again anyway

PR #4932, opened 2025-12-13
Fix: Skill Level Parsing

lunaynx:
## What
Fixed a `.romanToDecimal()` call to use `.romanToDecimalIfNecessary()` instead. This happened to silently work before #4927.

## Changelog Fixes
+ Fixed error when opening Your Skills menu with Roman numerals disabled in Hypixel settings. - Luna


NopoTheGamer:
knew this would happen

PR #4950, opened 2025-12-15
Fix: Skill error duing crop fever

NopoTheGamer:
## What
lets put random colour codes in the action bar "good idea guys, ill get right on it"

## Changelog Fixes
+ Fixed Skill error during crop fever. - nopo


Nealkitt:
lgtm

yes i actually tested it this time
image

CalMWolfs (approved):

PR #4955, opened 2025-12-16
Fix: jacob contest spam around new years

NopoTheGamer:
## What
big science went into this one, not really tested but probably works

## Changelog Fixes
+ Fixed Jacob's Contest spam around SB new year. - nopo


ptlthg:
I am Kaeso, and I approve of this message. (It was tested fairly well, should work fine)

PR #4843, opened 2025-11-04
Fix: Make BlockClickEvent use the right block

Bloxigus:
## What
For some reason `MovingObjectPosition.blockPos` gets mapped into `HitResult.pos` and not `BlockHitResult.blockPos`

## Changelog Fixes
+ Fixed an issue with clicks. - bloxigus

PR #4962, opened 2025-12-16
Fix: Timite Evolution Timer resetting when moving mouse

MTOnline69:
## What
BlockClickEvent for right click was firing with the exact collision point as the position rather than the block's position. This is because https://github.com/hannibal002/SkyHanni/pull/4843 mapped `MovingObjectPosition.blockPos` to `BlockHitResult.blockPos` for left click but not right click.

Fixing this fixes the Timite Evolution Timer on 1.21 resetting every time you move your mouse, even if you don't move your mouse away from the glass pane. This might also fix other issues, such as [Carnival "Double Click to Start"](https://discord.com/channels/997079228510117908/1447943730354061462).

I also took the opportunity to refactor `MinecraftInputHook.kt` as it seemed like this issue was caused by the code duplication the file had.

## Changelog Fixes
+ Fixed the Timite Evolution Timer feature resetting every time you moved your mouse. - MTOnline


lunaynx (approved):
Works fine for me, tested on 1.21.5 and 1.21.10.

MTOnline69:
> Works fine for me, tested on 1.21.5 and 1.21.10.

where's my 1.21.8 representation 💔

PR #4971, opened 2025-12-18
Feature: Aura Surveillance Goon Check-In Reminder

lunaynx:
## What
Added a check-in reminder for Candypat's new Public Surveillance minister perk. Made it toggleable via DisabledFeatures as well, so that in case it appears multiple times, we can more cleanly disable and re-enable it without forcing the reminder on players.

Also removed the old disabled Aura Propaganda feature entirely while I'm at it.

## Changelog New Features
+ Added Aura Surveillance Goon Check-In Reminder. - Luna


lunaynx:
Well, the perk is gone and is probably not happening again...

PR #4989, opened 2025-12-20
Fix + Improvement: Powder Chest Timer false positives + open detection

lunaynx:
## What
Reworked Powder Chest Timer a bit to fix false positives and made the highlight removal faster.

The openness check also serves as a way to remove structure chests once opened in case they do somehow still get misdetected, since they normally just stay there in an open state and never disappear, unlike powder chests.

## Changelog Improvements
+ Chests from powder mining are now unhighlighted faster (as soon as you open them). - Luna

## Changelog Fixes
+ Fixed structure chests sometimes being incorrectly highlighted as chests dug up from powder mining. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/mining/powdertracker/PowderChestTimer.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/mining/powdertracker/PowderChestTimer.kt
@@ -26,12 +26,14 @@ import at.hannibal2.skyhanni.utils.StringUtils
 import at.hannibal2.skyhanni.utils.TimeUnit
 import at.hannibal2.skyhanni.utils.TimeUtils.format
 import at.hannibal2.skyhanni.utils.collection.TimeLimitedCache
+import at.hannibal2.skyhanni.utils.collection.TimeLimitedSet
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/mining/powdertracker/PowderChestTimer.kt
```diff
@@ -74,22 +77,28 @@ object PowderChestTimer {
...
-        val isNewChest = event.newState.isChest()
-        val isOldChest = event.oldState.isChest()
 
-        if (isNewChest && !isOldChest) {
-            if (arePlayersNearby && lastSound.passedSince() > 200.milliseconds) return
-            chests[location] = maxDuration.fromNow()
-        } else if (isOldChest && !isNewChest) {
-            chests.remove(location)
+        when {
+            event.oldState.`is`(Blocks.STONE) && event.newState.`is`(Blocks.AIR) -> {
```
NopoTheGamer, line 82: .is is so annoying... thx mojang

NopoTheGamer (commented):
lets hope it works

PR #5005, opened 2025-12-23
Fix: Item Resolution for New Shards

lunaynx:
## What
Fixed ItemResolutionQuery not being able to get the output shard's internal name on the Confirm Fusion screen when there is a "NEW SHARD" text after its name. Also cleaned up some code in the file regarding the pet pattern and made it use RepoPatterns.

## Changelog Fixes
+ Fixed fused shards not being detected when it is a "NEW SHARD". - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/enoughupdates/ItemResolutionQuery.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/enoughupdates/ItemResolutionQuery.kt
@@ -13,6 +13,7 @@ import at.hannibal2.skyhanni.utils.ItemUtils.getLore
 import at.hannibal2.skyhanni.utils.NumberUtil.romanToDecimal
 import at.hannibal2.skyhanni.utils.RegexUtils.firstMatcher
 import at.hannibal2.skyhanni.utils.RegexUtils.matchMatcher
+import at.hannibal2.skyhanni.utils.RegexUtils.matches
 import at.hannibal2.skyhanni.utils.StringUtils.cleanString
 import at.hannibal2.skyhanni.utils.StringUtils.removeColor
 import at.hannibal2.skyhanni.utils.UtilsPatterns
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/api/enoughupdates/ItemResolutionQuery.kt
```diff
@@ -89,10 +108,9 @@ class ItemResolutionQuery {
             val isPet = itemName.contains("[Lvl ")
```
DavidArthurCole: This still being here somewhat defeats the purpose of moving the pattern to repo.
Higher-level, I might suggest moving some, or all of the Pet components of this file into `PetUtils`. This file is getting borderline monolithic, and PetUtils already has some lookup-type functions defined in it.

DavidArthurCole (changes requested):

src/main/java/at/hannibal2/skyhanni/api/enoughupdates/ItemResolutionQuery.kt
lunaynx: Yeah, I've fixed some leftover stuff like that now. Not sure I want to deal with a bigger refactor in this PR, although it is a good idea.

src/main/java/at/hannibal2/skyhanni/api/enoughupdates/ItemResolutionQuery.kt
```diff
@@ -102,14 +118,11 @@ class ItemResolutionQuery {
                 val unCleanItemDisplayName: String = EnoughUpdatesManager.getDisplayName(internalName)
                 var cleanItemDisplayName = unCleanItemDisplayName.removeColor()
                 if (cleanItemDisplayName.isEmpty()) continue
-                if (isPet) {
+                if (petPattern.matches(itemName)) {
                     if (!cleanItemDisplayName.contains("[Lvl {LVL}] ")) continue
                     cleanItemDisplayName = cleanItemDisplayName.replace("[Lvl {LVL}] ", "")
```
DavidArthurCole, line 123: Image

Better for sure - but there is still room to "combine" these two branches. The outputs of the matcher are not used anywhere besides that conditional, so they can be "conjoined", to avoid pattern-matching more than we need to.

Essentially, use the `matchMatcher` _as_ the if statement.
I can explain more in Discord, if that was nonsensical...

DavidArthurCole (changes requested):

PR #5014, opened 2025-12-28
Backend: Remove legacy shader code

VixidDev:
## What
Remove all shader code that was used for 1.8.9 and that is no longer needed for 1.21

+ Removed 1.8.9 shader code. - Vixid


src/main/java/at/hannibal2/skyhanni/shader/RoundedShader.kt
```diff
@@ -1,11 +1,9 @@
 package at.hannibal2.skyhanni.shader
 
 import at.hannibal2.skyhanni.utils.compat.GuiScreenUtils
-import at.hannibal2.skyhanni.utils.shader.Shader
-import at.hannibal2.skyhanni.utils.shader.Uniform
 import org.joml.Matrix4f
 
-abstract class RoundedShader<Self : RoundedShader<Self>>(vertex: String, fragment: String) : Shader(vertex, fragment) {
+abstract class RoundedShader<Self : RoundedShader<Self>>() {
```
NopoTheGamer: "An empty default constructor can be removed."

PR #5015, opened 2025-12-28
Backend: Clean up stereo harmony disc replacer

NopoTheGamer:
## What
Fixes https://github.com/SkyblockerMod/Skyblocker/pull/1980
I do wish it got actually reported to us seeing as this is something we can fix now
of course i havent tested it

+ Changed Stereo Disc replacer to use Components. - nopo

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestApi.kt
@@ -162,7 +162,7 @@ object PestApi {
      */
     val stereoPlayingItemPattern by patternGroup.pattern(
         "stereo.playing.item",
-        "§a§lPLAYING",
+        "PLAYING",
     )
 
     private var gardenJoinTime = SimpleTimeMark.farPast()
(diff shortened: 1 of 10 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestApi.kt
```diff
@@ -158,11 +158,11 @@ object PestApi {
     )
 
     /**
-     * REGEX-TEST: §a§lPLAYING
+     * REGEX-TEST: PLAYING
      */
     val stereoPlayingItemPattern by patternGroup.pattern(
         "stereo.playing.item",
-        "§a§lPLAYING",
+        "PLAYING",
```
lunaynx, line 165: The pattern should be renamed to not break old versions.

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestApi.kt
NopoTheGamer: eh, it doesnt effect 6.0 so it only breaks for people on the beta

PR #5020, opened 2025-12-29
Backend: Another ItemHoverEvent injection point

j10a1n15:
## What
ItemStacks can also render their tooltips through GuiGraphics.renderTooltip, which SbPv uses soon:tm: to allow for sh chroma in pv itself.

+ Added secondary ItemHoverEvent injection point. - juna, mona


NopoTheGamer:
this is great and all but we are trying to delete itemhoverevent...

does ItemTipTextEvent not post on your stuff? if it does then just wait for enchant parsing to be moved over to that

j10a1n15:
ah okay, i think it should work with the fabric event too

PR #5022, opened 2025-12-31
Feature: Warning when "overpaying" significantly with Super Crafting

HacktheTime:
## What
When using super craft it shows the user a Warning if they overpay significantly.

Default for significantly is 10m and 5m if cookie Buff is active. This can be adjusted in the config.

This was inspired by the fact that a lot of enchanted resources are WAY more expensive than in the enchanted block version. With this I hope to adapt the changes so these are also going down in price.

The algorithm will always ignore when any of the items aren’t in the bazaar.

<details>
<summary>Images</summary>

Bildschirmfoto_20251231_011601

</details>

## Changelog New Features
+ Added Anti Super Crafting Overpay. - Hype_the_Time
    * Blocks Super Crafts (bypassable) if you can save more than X coins by instantly selling the mats and instant buying the item directly.


ItsEmpa:
i think it could be nice to make this feature not work if the player is playing ironman, you can use `SkyBlockUtils.isIronmanProfile` for this

NopoTheGamer:
> i think it could be nice to make this feature not work if the player is playing ironman, you can use `SkyBlockUtils.isIronmanProfile` for this

`if (HypixelData.noTrade) return` ?

hannibal002:
not looked at the code yet, just the image: i think the second smaller message about how to avoid wasting this amount of money should belong in the chat instead (use new feature chatutils.optional or disable for this once its merged wiht one or my current open prs)

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,158 @@
...
+        if (HypixelData.noTrade) return
+        val craftingAmount = getSuperCraftingCount() ?: return
+        val maxCraftingAmount = getSuperCraftingMaxCount() ?: return
+        val profit = getProfit(craftingAmount, maxCraftingAmount) ?: return
+        if (event.clickedButton != 0) return
+        if (blockWasteClick(profit, craftingAmount, maxCraftingAmount)) {
+            SoundUtils.playErrorSound()
+            TitleManager.sendTitle(
+                "§cCraft-click Prevented (Big Loss Detected)",
+                subtitleText = "§7Hold §eControl §7to bypass. You can safe §c${String.format("%,.1f", -profit)}§6 Coins§7 by " +
```
NopoTheGamer: "safe" is not the correct word

src/main/java/at/hannibal2/skyhanni/utils/ItemPriceUtils.kt
```diff
@@ -9,6 +9,7 @@ import at.hannibal2.skyhanni.config.commands.brigadier.BrigadierArguments
 import at.hannibal2.skyhanni.events.SecondPassedEvent
 import at.hannibal2.skyhanni.features.inventory.bazaar.BazaarApi.getBazaarData
 import at.hannibal2.skyhanni.features.inventory.bazaar.HypixelItemApi
+import at.hannibal2.skyhanni.features.mining.crystalhollows.ChChestMessageAnalyser.items
```
NopoTheGamer: this cant be right

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingCoinWasteConfig.kt
```diff
@@ -0,0 +1,44 @@
...
+        "for wanted) to get a warning.")
+    @ConfigEditorSlider(minValue = 0.1f, maxValue = 50.0f, minStep = 0.1f)
+    val warnCoinWaste: Double = 20.0
+
+    @Expose
+    @ConfigOption(name = "Minimum Amount (With Booster Cookie active)", desc = "The minimum amount of coins (in millions) you must save " +
+        "(instant sell and insta buy for wanted) to get a warning while having Booster Cookie Buff.")
+    @ConfigEditorSlider(minValue = 0.1f, maxValue = 50.0f, minStep = 0.1f)
+    val warnCoinWasteWithCookie: Double = 10.0
+
```
NopoTheGamer: why would there be a separate option for when you have a cookie?

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,158 @@
...
+import net.minecraft.world.item.AirItem
+import kotlin.time.Duration.Companion.seconds
+
+@SkyHanniModule
+object SuperCraftingInventory {
+    val config = SkyHanniMod.feature.inventory.superCraftingCoinWaste
+    val invDetector = InventoryDetector(
+        onOpenInventory = { },
+        checkInventoryName = { name ->
+            name.matches(".* Recipe".toRegex())
```
NopoTheGamer: repo pattern

src/main/java/at/hannibal2/skyhanni/utils/ItemPriceUtils.kt
```diff
@@ -215,4 +216,26 @@ object ItemPriceUtils {
             }
         }
     }
+
+
+    fun NeuInternalName.getSumPriceForCount(
+        count: Int,
+        priceSource: ItemPriceSource,
+    ): Double? {
+        if (priceSource== ItemPriceSource.NPC_SELL) {
```
NopoTheGamer: formatting

src/main/java/at/hannibal2/skyhanni/utils/ItemPriceUtils.kt
```diff
@@ -215,4 +216,26 @@ object ItemPriceUtils {
             }
         }
     }
+
+
+    fun NeuInternalName.getSumPriceForCount(
```
NopoTheGamer: surely you can think of a better name for this

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,158 @@
...
+        "crafting.count",
+        ".*Crafting (?<count>[0-9,]+) item.*",
+    )
+    val craftingResourcePattern by craftingPatternGroup.pattern(
+        "crafting.resource",
+        " *([✔✖]) [0-9,]+/[0-9,]+ \\((?<amount>[0-9,]+)x\\) (?<resource>.+)",
+    )
+
+    private fun blockWasteClick(profit: Double, craftingAmount: Int, maxCraftingAmount: Int): Boolean {
+        if (!config.warnCoinWasteEnabled) return false
```
NopoTheGamer: why does this check the config again

src/main/java/at/hannibal2/skyhanni/config/features/inventory/InventoryConfig.kt
```diff
@@ -90,6 +93,12 @@ class InventoryConfig {
     @Accordion
     val rngMeter: RngMeterConfig = RngMeterConfig()
 
+    @Expose
+    @ConfigOption(name = "Super Crafting Coin Waste", desc = "")
+    @Accordion
+    val superCraftingCoinWaste: SuperCraftingCoinWasteConfig = SuperCraftingCoinWasteConfig()
+
+
```
NopoTheGamer: why 2 new lines

NopoTheGamer (changes requested):

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingCoinWasteConfig.kt
HacktheTime: Having cookie means remote access to bazaar with /bz. While most people have it not everyone does. If you have to leave the lobby this can matter.

NopoTheGamer: then make an option to disable it when you dont have a cookie?

HacktheTime:
> not looked at the code yet, just the image: i think the second smaller message about how to avoid wasting this amount of money should belong in the chat instead (use new feature chatutils.optional or disable for this once its merged wiht one or my current open prs)

If you have to close the menu to see the message fully or interact with it, it means if you have specified a custom amount you need to re-enter it. This sounds sub optimal to me. Are you sure this is the best way?

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingCoinWasteConfig.kt
HacktheTime: If you would loose 50m + its getting worth it to leave the lobby in most cases no matter what. But I should increase the max of the amounts.

NopoTheGamer: if you have enough materials that the difference from crafting is 50mil you have a cookie

HacktheTime: Not necessarily.

I have a ~ level 400 profile and I dont always have cookie since I dont afk and dont make enough money of it for it to be worth it. The times I have cookie is best described with every half a year 1 cookie for derpy.


I dont see why the extra customizability would be a downside for those who can use it. Even if you always have cookie you should not assume  everyone else does too.

NopoTheGamer: image

1. the option is unreadable
2. i think its just more confusing than having a toggle for enabled while cookie
3. why would you want it to be a different amount anyway? you can bypass it if you dont want to do it

HacktheTime: > image
> 
> 1. the option is unreadable
> 2. i think its just more confusing than having a toggle for enabled while cookie
> 3. why would you want it to be a different amount anyway? you can bypass it if you dont want to do it

For 1 idrk. Maybe with 2?

2 maybe outsource them further down the chain with an accordion for the with cookie ones? Description of the cookie part could be at accordion level.

3) because you can still get annoyed by it if you have to bypass it often since its a small amount and you dont have cookie.

HacktheTime:
@NopoTheGamer 

I hope to have addressed the changes you requested however I am uncertain whether all of them are as you imagine. My name creativity is pretty low. If you have better suggestions for names I'm open for it.

i hope that the issues with the cookie / no cookie are solved in a way we both agree to be fine.

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingCoinWasteConfig.kt
hannibal002: make the name shorter. the name is not the desc

HacktheTime: Have you checked the current names?

Both remaining issues that I didn't mark as solved were addressed but I'm not sure whether they go far enough.

If they don't I suggest exact name suggestions because I have no ideas.

hannibal002: > Have you checked the current names?

no. i just skimmed the pr reviews, saw the image with not readable text, saw your question about it, and agreed that the text should be readable. to make unreadable text readable that is unreadable because it is too long you need to make it shorter so it becomes readable again, thats all i suggested, make it shorter. no more context.

hannibal002: now i looked at the code.
multiple variables that needs to be updated here.
mainly, the whole config alr has the name "super crafting coin waste", so any option inside this categroy doesnt need any of those terms used again. the first boolean here has FeatureToggle, this one should  then  be renamed to "enabled". see other features for similar naming strucures.

especially "warnCoinWasteMaxResourcesWithCookie" is unnecesary long.
just remove words that are alr clear by the category name until you are left with what this double does exactly, inside this category.

also the sub class WithoutCookieSuperCraftingCoinWasteConfig belongs into a new file.

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,163 @@
...
+import at.hannibal2.skyhanni.utils.RegexUtils.matches
+import at.hannibal2.skyhanni.utils.SoundUtils
+import at.hannibal2.skyhanni.utils.StringUtils.removeColor
+import at.hannibal2.skyhanni.utils.collection.CollectionUtils.sumAllValues
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPatternGroup
+import net.minecraft.world.item.AirItem
+import kotlin.time.Duration.Companion.seconds
+
+@SkyHanniModule
+object SuperCraftingInventory {
```
hannibal002, line 32: please look at the problems intellij suggests for you.

image

src/main/java/at/hannibal2/skyhanni/utils/ItemPriceUtils.kt
```diff
@@ -215,4 +215,25 @@ object ItemPriceUtils {
             }
         }
     }
+
+    fun NeuInternalName.getTotalPriceForCount(
```
hannibal002: this function name should not be accessible via internal name and should not be in item price utils. this looks like something that belongs in the bz api, and should be renamed to something along the lines "calcualte price off avalibalbe bz offers" or someting?

src/main/java/at/hannibal2/skyhanni/config/features/inventory/InventoryConfig.kt
```diff
@@ -90,6 +90,11 @@ class InventoryConfig {
     @Accordion
     val rngMeter: RngMeterConfig = RngMeterConfig()
 
+    @Expose
+    @ConfigOption(name = "Super Crafting Coin Waste", desc = "")
+    @Accordion
+    val superCraftingCoinWaste: SuperCraftingCoinWasteConfig = SuperCraftingCoinWasteConfig()
```
hannibal002: maybe put this inside a new category super crafting".
we have other options that relate to super craft, e.g. in garden visitor shopping list and get from sack. not sure tho

hannibal002 (changes requested):
some stuff

hannibal002:
> If you have to close the menu to see the message fully or interact with it, it means if you have specified a custom amount you need to re-enter it. This sounds sub optimal to me. Are you sure this is the best way?

i think we talk about different stuff. i dont want the user to need to close the menu to do the thing you want the user to do by the feature. tho the info "why" the feature is doing what it is doing (plus the info how to disble it) should belong in the chat. 

this follos a quite new, but not really properly implemented strucutre of warning titles, espeically in inv, that i thought up:

the user should always know what mod is doing what. so if there is e.g. a title, it should be accompanied by a chat message. that message can then not only explain the feature but also allow to click the msg to disable said feature.

we alr do this for all the features that are powered by "chat utils.or disable".

src/main/java/at/hannibal2/skyhanni/config/features/inventory/InventoryConfig.kt
HacktheTime: i did not move stuff like the garden over.

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingWasteConfig.kt
```diff
@@ -0,0 +1,57 @@
...
+    )
+    @ConfigEditorSlider(minValue = 0.1f, maxValue = 50.0f, minStep = 0.1f)
+    val maxResource: Double = 5.0
+
+    @Expose
+    @Accordion
+    @ConfigOption(name = "Values without Cookie", desc = "Like the others but when you don't have Cookie Buff active → no /bz access")
+    val withoutCookieValues = WithoutCookie()
+
+    class WithoutCookie {
```
hannibal002: this belongs in a new file

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,171 @@
...
+
+        val recipeMultiplier = resultItem.second
+
+        val itemsPrice = materials.mapValues {
+            it.value * (craftingAmount / recipeMultiplier)
+        }.mapValues {
+            //The materials are always summed up already by the getRecipeMaterials function
+            val key = it.key!!
+            val price = BazaarApi.calculatePriceOffAvailableOrders(key, it.value, BazaarApi.SimpleTransactionType.BUY_ORDER)
+            return@mapValues price?: return null
```
hannibal002: prss control + alt + L to auto reformat, then small formatting issues like here are gone

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,171 @@
...
+        if (materials.containsKey(null)) return null
+        val resultItem = getResultItem() ?: return null
+
+        val recipeMultiplier = resultItem.second
+
+        val itemsPrice = materials.mapValues {
+            it.value * (craftingAmount / recipeMultiplier)
+        }.mapValues {
+            //The materials are always summed up already by the getRecipeMaterials function
+            val key = it.key!!
```
hannibal002: do not use !!

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,171 @@
...
+                wasteConfig::enabled )
+            event.cancel()
+        }
+    }
+
+    private fun getSuperCraftingMaxCount(): Int {
+        val slots = InventoryUtils.getItemsInOpenChestWithNull()
+        val pickaxeSlot = slots[32]
+        val minimum = pickaxeSlot.item.getLore().mapNotNull {
+            val it = it.removeColor()
```
hannibal002: please use a better name

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,171 @@
...
+    @HandleEvent
+    fun onClick(event: GuiContainerEvent.SlotClickEvent) {
+        if (!invDetector.isInside()) return
+        if (!wasteConfig.enabled) return
+        if (HypixelData.noTrade) return
+        val craftingAmount = getSuperCraftingCount() ?: return
+        val maxCraftingAmount = getSuperCraftingMaxCount()
+        val profit = getProfit(craftingAmount) ?: return
+        if (event.clickedButton != 0) return
+        if (blockWasteClick(profit, craftingAmount, maxCraftingAmount)) {
```
hannibal002: early return

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,171 @@
...
+    val craftingResourcePattern by craftingPatternGroup.pattern(
+        "crafting.resource",
+        " *([✔✖]) [0-9,]+/[0-9,]+ \\((?<amount>[0-9,]+)x\\) (?<resource>.+)",
+    )
+
+    val inventoryPattern by craftingPatternGroup.pattern(
+        "inventory.name",
+        "(?<itemname>.*) Recipe",
+    )
+    val invDetector = InventoryDetector(
```
hannibal002: this (any maybe other variables above) can be private here

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,171 @@
...
+import at.hannibal2.skyhanni.utils.collection.CollectionUtils.sumAllValues
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPatternGroup
+import net.minecraft.world.item.AirItem
+import kotlin.time.Duration.Companion.seconds
+
+@SkyHanniModule
+object SuperCraftingInventory {
+    val wasteConfig = SkyHanniMod.feature.inventory.superCrafting.waste
+
+    val craftingPatternGroup = RepoPatternGroup("supercraftinginventory")
```
hannibal002: can be renamed to group

hannibal002: also please add `-` betweeen words in the group name string so that there is no more typo warning

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,171 @@
...
+            BazaarApi.SimpleTransactionType.SELL_OFFER,
+        ) ?: return null
+
+        return totalResultPrice - itemsPrice
+    }
+
+    fun getRecipeMaterials(): Map<NeuInternalName?, Int> {
+        val slots = InventoryUtils.getItemsInOpenChestWithNull()
+        return listOf(
+            slots[10], slots[11], slots[12],
```
hannibal002: create a list slotIds outside this function then map it here, instad of hard coding the variables inside the function like this.
ideally, even put the slot ids in repo, tho not necessary rn

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,171 @@
...
+            slots[19], slots[20], slots[21],
+            slots[28], slots[29], slots[30],
+        ).map {
+            val name = it.item.getInternalNameOrNull()
+            if (name != null) return@map name to it.item.count
+            if (it.item.item is AirItem) return@map NeuInternalName.NONE to 0
```
hannibal002: why do we need the air check here? what are we doing here exactly? what do we try to avoid?

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,171 @@
...
+            else return@map null to it.item.count
+        }.groupBy { it.first }.mapValues { it ->
+            it.value.sumOf {
+                it.second
```
hannibal002: the new lines in front and behind this line can  be collapsed

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,171 @@
...
+            }
+        }.filter { it.key != NeuInternalName.NONE }
```
hannibal002: instead of none, we could work with null directly. then we can use filterNotNull

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    fun getSuperCraftingCount(): Int? {
+        val slots = InventoryUtils.getItemsInOpenChestWithNull()
+        val pickaxeSlot = slots[32]
+        val lore = pickaxeSlot.item.getSingleLineLore().removeColor()
+        val craftingCount = craftingCount.matchMatcher(lore) {
+            groupOrNull("count")?.replace(",", "")?.toIntOrNull()
+        }
+        return craftingCount
```
hannibal002: you can inline the function craftingCount directly

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    fun getResultItem(): Pair<NeuInternalName, Int>? {
+        val slots = InventoryUtils.getItemsInOpenChestWithNull()
+        val resultSlot = slots[25]
+        return resultSlot.item.getInternalNameOrNull().let {
```
hannibal002: create a new variable for resultSlots.item, then inline restultsSlot

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,171 @@
...
+            if (resultSlot.item.item is AirItem) return null
+            (it ?: NeuInternalName.NONE) to resultSlot.item.count
+        }
+    }
+
+    private fun blockWasteClick(profit: Double, craftingAmount: Int, maxCraftingAmount: Int): Boolean {
```
hannibal002: please add explanations to the lines, explaining what we are blocking exactly or why

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,171 @@
...
+            }
+        }.filter { it.key != NeuInternalName.NONE }
+    }
+
+    fun getSuperCraftingCount(): Int? {
+        val slots = InventoryUtils.getItemsInOpenChestWithNull()
+        val pickaxeSlot = slots[32]
+        val lore = pickaxeSlot.item.getSingleLineLore().removeColor()
+        val craftingCount = craftingCount.matchMatcher(lore) {
+            groupOrNull("count")?.replace(",", "")?.toIntOrNull()
```
hannibal002: String.formatIntOrNull can be used instead of replace and toIntoOrNull

hannibal002 (changes requested):
looked a bit more in the code,  not tested yet. i like the changes you did, they look good!

couple more stuff i noticed, mostly small stuff.

tho, whats going on with AirItem and neu internal name none? why do we need to deal with or null entries in the item list but not proper work with null here? this feels weird to me

HacktheTime:
> looked a bit more in the code,  not tested yet. i like the changes you did, they look good!
> 
> couple more stuff i noticed, mostly small stuff.
> 
> tho, whats going on with AirItem and neu internal name none? why do we need to deal with or null entries in the item list but not proper work with null here? this feels weird to me

Im not 100% sure on how the neu internal name with null works here tbh.

I have to ensure that the indexes I use to get the items work. So even empty slots etc matter.

But I also have to ensure that the item is not just missing in neu repo and is actually empty.

Also I noticed yesterday that there is probably a swap of instant but and instant sell in the code logic. Didnt get to fixing it yet.

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
HacktheTime: Normally I would turn on auto reformat but that fucks up the imports with the pre processor afair.

HacktheTime:
for the block waste click explenation there is a chatutils with config if you mean to the user or should I add kdoc for developers?

src/main/java/at/hannibal2/skyhanni/config/features/inventory/InventoryConfig.kt
```diff
@@ -90,6 +90,11 @@ class InventoryConfig {
     @Accordion
     val rngMeter: RngMeterConfig = RngMeterConfig()
 
+    @Expose
+    @ConfigOption(name = "Super Crafting", desc = "")
+    @Accordion
+    val superCrafting: SuperCrafting = SuperCrafting()
```
hannibal002: the suffix "Config" in the class name is missing

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingWasteConfig.kt
```diff
@@ -0,0 +1,37 @@
...
+        name = "Minimum Amount if Max Resource Usage",
+        desc = "Minimum amount of coins when compacting items due to space reasons.",
+    )
+    @ConfigEditorSlider(minValue = 0.1f, maxValue = 50.0f, minStep = 0.1f)
+    val maxResource: Double = 5.0
+
+    @Expose
+    @Accordion
+    @ConfigOption(name = "Values without Cookie", desc = "Like the others but when you don't have Cookie Buff active → no /bz access")
+    val withoutCookieValues = SuperCraftingWasteConfigWithoutCookie()
```
hannibal002: the suffix "Config" in the class name is missing

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingWasteConfigWithoutCookie.kt
```diff
@@ -0,0 +1,25 @@
...
+class SuperCraftingWasteConfigWithoutCookie {
+
+    @Expose
+    @ConfigOption(
+        name = "Minimum Amount",
+        desc = "The minimum amount of coins (in millions) you must save (instant sell and insta buy " +
+            "for wanted) to get a warning.",
+    )
+    @ConfigEditorSlider(minValue = 0.1f, maxValue = 100.0f, minStep = 0.1f)
+    val normal: Double = 20.0
```
hannibal002: is this now normal or miniam amount? please be consistent in names

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingWithoutCookieConfig.kt
```diff
@@ -0,0 +1,25 @@
...
+        name = "Minimum Amount",
+        desc = "The minimum amount of coins (in millions) you must save (instant sell and insta buy " +
+            "for wanted) to get a warning.",
+    )
+    @ConfigEditorSlider(minValue = 0.1f, maxValue = 100.0f, minStep = 0.1f)
+    val normal: Double = 20.0
+
+    @Expose
+    @ConfigOption(
+        name = "Minimum Amount if Max Resource Usage",
```
hannibal002: the name should be too long i think

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingWasteConfig.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.google.gson.annotations.Expose
+import io.github.notenoughupdates.moulconfig.annotations.Accordion
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorBoolean
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorSlider
+import io.github.notenoughupdates.moulconfig.annotations.ConfigOption
+
+class SuperCraftingWasteConfig {
+    @Expose
+    @ConfigOption(
+        name = "Warn about Super Crafting Coin Waste",
```
hannibal002: too long a name. how about just naming it "enabled" if you cant make it shorter

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingWasteConfig.kt
```diff
@@ -0,0 +1,37 @@
...
+    val enabled: Boolean = true
+
+    @Expose
+    @ConfigOption(name = "Minimum Amount", desc = "The minimum amount of coins (in millions) you must save.")
+    @ConfigEditorSlider(minValue = 0.1f, maxValue = 50.0f, minStep = 0.1f)
+    val normal: Double = 10.0
+
+    @Expose
+    @ConfigOption(
+        name = "Minimum Amount if Max Resource Usage",
```
hannibal002: name too long

hannibal002 (changes requested):
again, your changes look good, just the null/air thingy remaining and small cleanups

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,170 @@
...
+        val lore = slots[PICKAXE_SLOT].item.getSingleLineLore().removeColor()
+        return craftingCount.matchMatcher(lore) {
+            groupOrNull("count")?.formatIntOrNull()
+        }
+    }
+
+    fun getResultItem(): Pair<NeuInternalName, Int>? {
+        val slots = InventoryUtils.getItemsInOpenChestWithNull()
+        val item = slots[RESULT_SLOT].item
+        return item.getInternalNameOrNull().let {
```
hannibal002: `let` can be removed here

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,170 @@
...
+            )
+            event.cancel()
+        }
+    }
+
+    private fun getSuperCraftingMaxCount(): Int {
+        val slots = InventoryUtils.getItemsInOpenChestWithNull()
+        val pickaxeSlot = slots[32]
+        val minimum = pickaxeSlot.item.getLore().mapNotNull { loreLine ->
+            val loreLine = loreLine.removeColor()
```
hannibal002: please use a different variable name

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/BazaarApi.kt
```diff
@@ -115,7 +115,7 @@ object BazaarApi {
 
     private var taxRate: Double
         get() = storage?.taxRate ?: 1.25
-        private set(value) {
+        set(value) {
```
hannibal002, line 118: why is this public now?

HacktheTime:
> again, your changes look good, just the null/air thingy remaining and small cleanups

I changed the code regarding the null and air. If this isnt what you expect I need an explenation of what you mean.

Ill work on the requested changes in the meanwhile.

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/BazaarApi.kt
```diff
@@ -286,4 +286,27 @@ object BazaarApi {
...
+            val takeAmount = offer.amount.coerceAtMost(remaining.toLong()).toInt()
+            totalPrice += takeAmount * offer.pricePerUnit
+            remaining -= takeAmount
+            if (remaining <= 0) break
+        }
+        if (remaining > 0) return null
+        return totalPrice
+    }
+
+    enum class SimpleTransactionType {
```
hannibal002, line 314: ItemPriceSource does already exist

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/BazaarApi.kt
```diff
@@ -286,4 +286,27 @@ object BazaarApi {
...
+    fun calculatePriceOffAvailableOrders(
+        item: NeuInternalName, count: Int,
+        priceSource: SimpleTransactionType,
+    ): Double? {
+        val bazaarData = item.getBazaarData()?.product ?: return null
+        val offers = if (priceSource == SimpleTransactionType.BUY_ORDER) bazaarData.buySummary else bazaarData.sellSummary
+        var remaining = count
+        var totalPrice = 0.0
+        for (offer in offers) {
+            val takeAmount = offer.amount.coerceAtMost(remaining.toLong()).toInt()
```
hannibal002: to long and then to int looks wrong.

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/BazaarApi.kt
```diff
@@ -286,4 +286,27 @@ object BazaarApi {
     }
 
     fun isBazaarOrderInventory(inventoryName: String): Boolean = inventoryBazaarOrdersPattern.matches(inventoryName)
+
+    fun calculatePriceOffAvailableOrders(
```
hannibal002: im still not happy with the function name. at least add a desc for now please.

hannibal002 (changes requested):
more small stuff

hannibal002:
> I have to ensure that the indexes I use to get the items work. So even empty slots etc matter.

why? a null item is as meaningful as a key that doesnt exist in the map. or what are you doing here exactly?

> But I also have to ensure that the item is not just missing in neu repo and is actually empty.

thats a valid concern. tho you could just throw the user facing error right here and there, and then not work with air/null items below anymore


> I changed the code regarding the null and air. If this isnt what you expect I need an explenation of what you mean.

i think the issue is more fundamental. think of it: we dont use the air item check anywhere else. thats one issue. the other is the way you use none/null. you should check it at the first entry point, then throw an error if something is invalid. then proceed the remaining functions without invalid data.
the current impl looks like you keep invalid data inside the maps you pass along and need to check for them at every step along the way.

or do i miss some crucial part of the logic as to why you truly need the null/air/none data there?

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/BazaarApi.kt
HacktheTime: uh isnt the property private definition making it redundant?

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/BazaarApi.kt
HacktheTime: but has NPC Price entry which I want to exclude via possible type instead of runtime error.

hannibal002: you could simply return the normal npc price logic if the enum is set to npc price, no error needed. since you control the two spots where you call the function, i dont see the issue

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingWasteConfigWithoutCookie.kt
HacktheTime: normal in form of the normal way. not the max resource. its still minimum amount. but in the normal scope. and as I was told Im not supposed to include existing info in names again.

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingWithoutCookieConfig.kt
HacktheTime: then I want a name and description from you. Because in my opinion its to unprecise otherwise.

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingWasteConfig.kt
HacktheTime: same as other with the same name



then I want a name and description from you. Because in my opinion its to unprecise otherwise.

hannibal002: the name is not the desc!

the name is to distinguish the option from other toggles in the same category.
it does not have to be precise, it has to not be wrong.
check other spots in the code where we also have toggles that do complex stuff and still have readable names.
im sure you can find a shorter name. confusions as of what the toggle does are fixed by reading the description anyway.

HacktheTime:
> > I have to ensure that the indexes I use to get the items work. So even empty slots etc matter.
> 
> why? a null item is as meaningful as a key that doesnt exist in the map. or what are you doing here exactly?
> 
> > But I also have to ensure that the item is not just missing in neu repo and is actually empty.
> 
> thats a valid concern. tho you could just throw the user facing error right here and there, and then not work with air/null items below anymore
> 
> > I changed the code regarding the null and air. If this isnt what you expect I need an explenation of what you mean.
> 
> i think the issue is more fundamental. think of it: we dont use the air item check anywhere else. thats one issue. the other is the way you use none/null. you should check it at the first entry point, then throw an error if something is invalid. then proceed the remaining functions without invalid data. the current impl looks like you keep invalid data inside the maps you pass along and need to check for them at every step along the way.
> 
> or do i miss some crucial part of the logic as to why you truly need the null/air/none data there?



> > I have to ensure that the indexes I use to get the items work. So even empty slots etc matter.
> 
> why? a null item is as meaningful as a key that doesnt exist in the map. or what are you doing here exactly?
> 
> > But I also have to ensure that the item is not just missing in neu repo and is actually empty.
> 
> thats a valid concern. tho you could just throw the user facing error right here and there, and then not work with air/null items below anymore
> 
> > I changed the code regarding the null and air. If this isnt what you expect I need an explenation of what you mean.
> 
> i think the issue is more fundamental. think of it: we dont use the air item check anywhere else. thats one issue. the other is the way you use none/null. you should check it at the first entry point, then throw an error if something is invalid. then proceed the remaining functions without invalid data. the current impl looks like you keep invalid data inside the maps you pass along and need to check for them at every step along the way.
> 
> or do i miss some crucial part of the logic as to why you truly need the null/air/none data there?

I need to differentiate between a slot being empty and a slot being with an unknown item.

If its a unknown item I have to exit while if it is just an empty slot i just have to ignore that slot.

If there is a new item idk yet I cant say oh it costs you 1 coin since the only thing I found in the recipe is 32 redstone.

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/BazaarApi.kt
HacktheTime: coerceAtMost needs a long and returns a long. therefore i must convert into long and back,

hannibal002:
> I need to differentiate between a slot being empty and a slot being with an unknown item.

if its an unknown item, you just error here and there, yo dont proceed or hande this case further down anymore.

HacktheTime:
> > I need to differentiate between a slot being empty and a slot being with an unknown item.
> 
> if its an unknown item, you just error here and there, yo dont proceed or hande this case further down anymore.

right now I exit since I dont want to spam the user. But Idk how I have to do it so I do get a difference between empty slot and unknown item different otherwise.

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/BazaarApi.kt
hannibal002: ur right, thats fine. keep the change.

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingWasteConfigWithoutCookie.kt
hannibal002: That's not what I meant. Don't repeat the parent category, but the name should still be meaningful. normal tells me nothing. threshold or minSavings would work.

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/BazaarApi.kt
hannibal002: ill fix it

hannibal002:
Ive fixed some issues, the missing repo pattern tests should be the only ones remaining (and the still open suggestions above)

HacktheTime:
> Ive fixed some issues, the missing repo pattern tests should be the only ones remaining (and the still open suggestions above)

Ill add the test parts later. But wdym with the open suggestions above?

HacktheTime:
In your recent change you renamed wasteConfig to config.

Isnt that bad since were in SuoerCraftingInventory and its the specialised config for waste only?

hannibal002:
> Ill add the test parts later. But wdym with the open suggestions above?

the "requested changes"/issues here on the pr review page.

most important things i think still need changing:

- too long send title subtitle
- unclear remaining null policy: you return null at many different functions. what is the intended logic here? some of those seem like they should just throw errors instead.
-  ItemPriceSource instead of SimpleTransactionType
- config names too long and confusing/not consistent
- repo pattern tests

hannibal002:
> Isnt that bad since were in SuoerCraftingInventory and its the specialised config for waste only?

this is the project standard. we always name the config instance in the feature class just "config". out of context, its alr clear that the config here refers to the relevant config, and nothing else

image

HacktheTime:
Throwing errors is expensive and it feels unecessary if you know the issue that caused it. Items not being defined yet is sth I see as an indirect developer mistake. Not the users.

For the Itemprice aource the bz api order sum cant use the itempricesource npc

hannibal002:
Throwing errors is expensive and it feels unecessary if you know the issue that caused it. Items not being defined yet is sth I see as an indirect developer mistake. Not the users.

i see some flaws in your approach:

> throwing errors is expensive

the error is more expensive than not doing anything at "this very moment for this single user", yes.
but we want to be able to maintain the logic. imagine there would be a design change on hypxiel side, and the item slot ids would change, or the names, lores, whatever.
in this scenario, with your solution of "hiding errors" by returning null/ignoring, we would never know that this is the issue, unless some user reports that the feature doesnt work. then we need to start investigating whats wrong. if we can replicate, etc etc.
on the other hand, with proper error handling, we immediately see at what line and why is there an issue. this is a user facing chat message, that the user can report way easier to us, and we have a easier time fixing it.

also, we already throw errors everywhere in skyhanni, thats the consistent approach here.

> items not being defined yet

if this is the error, the user sees a "repo error" in chat instead. thats also intended.

> indirect developer mistake. Not the users.
all errors we print to chat are dev errors. but the only way for us to see them is by showing them to the user and asking them to send it to us. thats how our error logic works.

only exception is ChatUtils.userError. but thats not relevant here.

HacktheTime:
>     * unclear remaining null policy: you return null at many different functions. what is the intended logic here? some of those seem like they should just throw errors instead. 
didnt you change this already the way you want?

>     * ItemPriceSource instead of SimpleTransactionType
I will not change this because doing so seems straight up bad design. ItemPriceSource can be things that arent bazaar order types (such as npc price) and TransactionType has flipping and order types. I think showing errors at compile time is better than at runtime.

>     * too long send title subtitle
>     * config names too long and confusing/not consistent
If you want other names either say the names or straight up commit them in changed form.
I know when I have to tell others to do sth when Ik that I suck at it.

>     * repo pattern tests
Added / fixed.

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,166 @@
...
+        return if (BitsApi.hasCookieBuff()) config.maxResource
+        else config.withoutCookieValues.maxResource
+    }
+
+    @HandleEvent
+    fun onClick(event: GuiContainerEvent.SlotClickEvent) {
+        if (!invDetector.isInside()) return
+        if (!config.enabled) return
+        if (HypixelData.noTrade) return
+        if (event.clickedButton != 0) return
```
hannibal002, line 86: this only blocks left clicks. can we super craft by shift click? if so, we should block this as well

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/BazaarApi.kt
```diff
@@ -286,4 +286,33 @@ object BazaarApi {
     }
 
     fun isBazaarOrderInventory(inventoryName: String): Boolean = inventoryBazaarOrdersPattern.matches(inventoryName)
+
+    /**
+     * Gets the Sum of Coins you can get when using the specified sell type to sell the item.
+     * For example if you want to instant sell 71680 enchanted coal from sack you cant use top offer only.
+     * The best offers may be inflated, and you are selling all the items so you must use the prices in each order as such.
+     */
+    fun calculatePriceOffAvailableOrders(
```
CalMWolfs: should be `calculatePriceOfAvailableOrders`

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingWasteConfig.kt
```diff
@@ -0,0 +1,37 @@
...
+
+import at.hannibal2.skyhanni.config.FeatureToggle
+import com.google.gson.annotations.Expose
+import io.github.notenoughupdates.moulconfig.annotations.Accordion
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorBoolean
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorSlider
+import io.github.notenoughupdates.moulconfig.annotations.ConfigOption
+
+class SuperCraftingWasteConfig {
+    @Expose
```
CalMWolfs, line 12:
```suggestion
class SuperCraftingWasteConfig {

    @Expose
```

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
NopoTheGamer: shift click maxes the amount

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftingConfig.kt
```diff
@@ -0,0 +1,12 @@
+package at.hannibal2.skyhanni.config.features.inventory
+
+import com.google.gson.annotations.Expose
+import io.github.notenoughupdates.moulconfig.annotations.Accordion
+import io.github.notenoughupdates.moulconfig.annotations.ConfigOption
+
+class SuperCraftingConfig {
+    @Expose
```
CalMWolfs, line 9:
```suggestion
class SuperCraftingConfig {

    @Expose
```

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -0,0 +1,177 @@
...
+        config.withoutCookie.bulkThreshold
+    } * 1_000_000L
+
+    @HandleEvent
+    fun onSlotClick(event: GuiContainerEvent.SlotClickEvent) {
+        if (!invDetector.isInside()) return
+        if (!config.enabled) return
+        if (HypixelData.noTrade) return
+        if (event.clickedButton != 0) return
+        val slots = InventoryUtils.getItemsInOpenChestWithNull()
```
hannibal002, line 88: do we check if we click the craft button or any other slot in the inv? we dont want to trigger the whole logic (math + error msg) when we click on any other slot than the craft button, right?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
HacktheTime: Right now it reacts on clicking any slot.

Could matter indeed. When going down the chain of materials.

I vaguely remeber some stuff when going down the chain which could break the regexes too. Forgot about that.

PR #5040, opened 2026-01-05
Fix: Diana Mob Name not sharing in Rare Mob Waypoints

Nealkitt:
## What
fixes diana mobs name not being shared from sharing the waypoint, ty nopo for the real fix, i just got rid of an extra space to fix the other part of it

## Changelog Fixes
+ Fixed Diana Rare Mob Waypoints not sharing the rare mob's name. - Stella


NopoTheGamer:
so why is this exchange from changelog? its a bug fix

PR #5043, opened 2026-01-05
Fix: dont wipe matrix stack when rendering items in hud 1.21.8

Soopyboo32:
## What
Removes the call to `RenderSystem.setupDefaultState()` when rendering items,
this call wipes the modelViewStack and the textureMatrix and the item rendering logic does not restore them after
This pr instead pushes to the modelViewStack and saves a copy of the textureMatrix, then restoring the state after rendering

That code is what's causing the crash with lunar when entering garden (and a few other situations) as lunar pushes before hud rendering and later fails to pop cus the matrix was wiped
This may also cause issues with other mods doing similar things around wrapping hud rendering

<details>
<summary>Images</summary>

</details>

## Changelog Fixes
+ Fixed crash with Lunar when entering the Garden. - Soopyboo32


hannibal002:
Once ill get IntelliJ to run again ill test this. code looks good tho, good job!

PR #5057, opened 2026-01-08
Fix: Ubik Timer Offset

lunaynx:
## What
The Ubik's Cube cooldown actually starts counting as soon as you open the GUI. Even if you just close it, you get like 10 seconds to get back into the game, and if you don't, it forces Split, ends the game, and puts you in cooldown. I fixed the timer to account for this instead of starting after the final round is over.

## Changelog Fixes
+ Fixed Ubik's Cube Timer showing as ready slightly too late (up to about 30 seconds). - Luna


src/main/java/at/hannibal2/skyhanni/features/rift/everywhere/UbikReminder.kt
```diff
@@ -53,25 +56,24 @@ object UbikReminder {
         "SPLIT! You need to wait (?<duration>.+) before you can play again\\.",
     )
 
+    init {
+        InventoryDetector(
+            onOpenInventory = { storage?.ubikRemindTime = 2.hours.fromNow() },
```
hannibal002: missing config check

hannibal002 (changes requested):

PR #5058, opened 2026-01-08
Backend: Show Total Tracker Uptime by default

Chissl:
Tracker uptime on total was disabled by default since it'd be wrong for people who had trackers prior to its inclusion. Enough time has passed that most people with prior tracker data have updated so there's no point keeping it off. 

exclude_from_changelog


NopoTheGamer:
can you make the afk timeout default slightly higher and add a searchtag for uptime

PR #5064, opened 2026-01-10
Improvement: Add bits to pest profit tracker

Chissl:
## What
Added option to include bits in pest profit tracker, off by default

## Changelog Improvements
+ Added option to include bits from pest kills to pest profit tracker. - Chissl

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestProfitTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestProfitTracker.kt
@@ -35,7 +35,6 @@ import at.hannibal2.skyhanni.utils.NumberUtil.shortFormat
 import at.hannibal2.skyhanni.utils.RegexUtils.matchGroup
 import at.hannibal2.skyhanni.utils.RegexUtils.matchMatcher
 import at.hannibal2.skyhanni.utils.SimpleTimeMark
-import at.hannibal2.skyhanni.utils.SkyBlockUtils
 import at.hannibal2.skyhanni.utils.chat.TextHelper.asComponent
 import at.hannibal2.skyhanni.utils.collection.CollectionUtils.addOrPut
 import at.hannibal2.skyhanni.utils.collection.RenderableCollectionUtils.addSearchString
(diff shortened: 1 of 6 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestProfitTracker.kt
```diff
@@ -104,6 +111,21 @@ object PestProfitTracker : SkyHanniBucketedItemTracker<PestType, PestProfitTrack
...
+        override fun getCustomPricePer(internalName: NeuInternalName): Double {
+            return if (internalName == BITS) {
+                getBitsPrice()
+            } else {
+                super.getCustomPricePer(internalName)
+            }
+        }
+
+        private fun getBitsPrice(): Double {
+            return if (SkyBlockUtils.noTradeMode) 0.0 else config.coinsPerBit.get().toDouble()
```
NopoTheGamer: this should check if its set to npc prices or not i think (instead of ironman)

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestProfitTracker.kt
```diff
@@ -140,6 +162,19 @@ object PestProfitTracker : SkyHanniBucketedItemTracker<PestType, PestProfitTrack
         event.checkSprayChats()
     }
 
+    @HandleEvent
+    fun onPestKill(event: PestKillEvent) {
+        if (BitsApi.bitsAvailable > 0) {
+            val bitsAmount = KILL_BITS * BitsApi.bitsMultiplier()
+            addItem(event.pestType, BITS, bitsAmount.toInt(), false)
+        }
+    }
```
NopoTheGamer, line 174: is there a reason you dont use the currency event? it would track doubles from the bit talisman

NopoTheGamer (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestProfitTracker.kt
Chissl: Too inconsistent; it would sometimes record visitor bits or other bit gains as pests

Chissl: Also means that we couldn't add bits to specific buckets. I guess I can account for bits talismans if we have an accessory bag api?

NopoTheGamer: idt we do. its fine i was just wondering

PR #5070, opened 2026-01-10
Improvement: Setting to hide/show PestProfitTracker while farming

f4bii:
## What
Added ability to show/hide pest profit tracker while farming

## Changelog Improvements
+ Added Option to Show/Hide Pest Profit Tracker while farming. - FabiHBBBT

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/garden/pests/PestProfitTrackerConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/garden/pests/PestProfitTrackerConfig.kt
@@ -22,6 +22,11 @@ class PestProfitTrackerConfig {
     var hideChat: Boolean = true
 
     @Expose
+    @ConfigOption(name = "Hide while farming", desc = "Hide profit tracker while farming.")
+    @ConfigEditorBoolean
+    var hideWhileFarming: Boolean = true
+
+    @Expose
     @ConfigOption(name = "Time Displayed", desc = "Time displayed after killing a pest.")
     @ConfigEditorSlider(minValue = 5f, maxValue = 60f, minStep = 1f)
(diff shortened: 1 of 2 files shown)


Chissl:
I don't think this is particularly useful at this stage as pest profit tracker will only show when using a vacuum/killing a pest, would recommend modifying display options to be more similar to pest timer

f4bii:
So changing the Settings to include:
 - Enabled
 - Only When Holding?
 - Hide messages
 - Kill Timeout (When not killing a pest for a duration)
 
 Dont know if we need a Text setting thing or do you want to completely change the design?

Chissl:
> Dont know if we need a Text setting thing or do you want to completely change the design?

Yeah, draggable list would be preferred as the options would take up a lot of space without it, basically the same options as what's in the pest timer config with the addition of kill timeout

I'd also recommend keeping the current "don't show when farming" functionality as a separate toggle

f4bii:
Oh i get it prob now. You want to remove the list of items? And the rest as draggable. 

So we have after the remove:
 - Pest Killed
 - Sprays used:
 - Session Profit:
 - Profit Per Hour:
 - Session Uptime:

image

Chissl:
No, I'd recommend looking at https://github.com/hannibal002/SkyHanni/blob/d01aa9fdd705b376e4a623af22127a0153b7cc3e/src/main/java/at/hannibal2/skyhanni/config/features/garden/pests/PestTimerConfig.kt#L21 and implement something similar for when to show the pest profit tracker display, with the additional option of "show after pest kill"

Currently, your pr only overrides the "hide when farming" option which isn't very impactful as the display currently only shows after killing a pest or holding a vacuum

f4bii:
Probably should change some wordings. But im bad with wording

PR #5091, opened 2026-01-14
Fix: Increased Wardrobe pages to 3

f4bii:
## What
Increased Wardrobe pages to 3

<details>
<summary>Images</summary>

image
image

</details>

## Changelog Fixes
+ Fixed Custom Wardrobe not having 3 pages. - FabiHBBBT

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/WardrobeApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/WardrobeApi.kt
@@ -53,7 +53,7 @@ object WardrobeApi {
     private const val FIRST_LEGGINGS_SLOT = 18
     private const val FIRST_BOOTS_SLOT = 27
     const val MAX_SLOT_PER_PAGE = 9
-    const val MAX_PAGES = 2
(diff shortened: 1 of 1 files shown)


hannibal002:
i think there is something wrong with this impl.
we need to proper render this.
the existing wardrobe code expects only two lines existing. Have you tested this? i assume this must look ugly without additional fixes

NopoTheGamer:
> Have you tested this? 

have you looked at the attached image?

i do agree this should probably have functionality to use commands to instantly go between pages though

f4bii:
Commands would be nice so you dont need to click 3 times for armors in 3rd page. But Hypixel currently only supports opening the first and second page with commands

NopoTheGamer:
> Commands would be nice so you dont need to click 3 times for armors in 3rd page. But Hypixel currently only supports opening the first and second page with commands

you can make it do `/wd 1` when you are on page 3, and even make the logic for `/wd 3` but its locked behind a repo flag

NopoTheGamer:
well i did some science and hypixel ignores /wd commands while the wardrobe is open, probably as an anti dupe mechanic

PR #5095, opened 2026-01-15
Fix + Backend: Replace user-presented logs with internal logging fixing Discord Rich Presence

NetheriteMiner:
## What
Uses LorenzLogger to log errors in ChatProgressUpdates, therefore not prompting the user if an error happens. Previously, the error caused something to go wrong in DiscordRPCManager, which prevented the feature from working if certain fields were null. Now, logging with this class should never result in the feature crashing.

The filePath in the logger and the specific message being logged can be changed as desired.

## Changelog Fixes
+ Fixed Discord Rich Presence erroring on startup. - NetheriteMiner

+ Changed ChatProgressUpdates to log silently. - NetheriteMiner


hannibal002:
I'm not sure this does what we want.

The error functions here don't throw errors, they only log them to begin with. The error message happens when the logging system is used wrongly, not when the data that is logged is wrong. This is a logical error in the discord rcp logic that should get fixed instead of hiding it by replacing it with the outdated Lorenz logger.

By merging the pr as is we lose proper debug informations about discord rcp if something went wrong.

NetheriteMiner:
Replaced with better fix in PR #5142

PR #5096, opened 2026-01-15
Feature: Nautilus SC Health Overlay & Kill Time & LS Sphere

Fazfoxy:
## Dependencies
- #5098
- https://github.com/hannibal002/SkyHanni-REPO/pull/604

## What
Ports some of the [Nautilus](https://github.com/ItsEmpa/Nautilus/tree/main) Features & Required Backend Features (needs better integration into SH Files probably?) will be a dependency for future PRs taking NT features as a lot have reliance on random helper functions Empa made

<details>
<summary>Images</summary>

image
image

(the kill timer has a Rare check I temporarily turned it off for the sake of getting this screenshot)
</details>

## Changelog New Features
+ Added Sea Creature Kill Time. - Empa + Fazfoxy
+ Added Sea Creature Health Overlay. - Empa + Fazfoxy
+ Added Sea Creature Lootshare Sphere. - Empa + Fazfoxy


src/main/java/at/hannibal2/skyhanni/config/features/fishing/SeaCreatureHealthDisplayConfig.kt
```diff
@@ -0,0 +1,36 @@
...
+import at.hannibal2.skyhanni.utils.ConfigUtils.asProperty
+import com.google.gson.annotations.Expose
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorBoolean
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorSlider
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorText
+import io.github.notenoughupdates.moulconfig.annotations.ConfigLink
+import io.github.notenoughupdates.moulconfig.annotations.ConfigOption
+
+class SeaCreatureHealthDisplayConfig {
+    @Expose
```
NopoTheGamer, line 14: new line

src/main/java/at/hannibal2/skyhanni/utils/EntityUtils.kt
```diff
@@ -233,4 +234,9 @@ object EntityUtils {
     // TODO use derpy() on every use case
     val LivingEntity.baseMaxHealth: Int
         get() = this.getAttributeBaseValue(Attributes.MAX_HEALTH).toInt()
+
+    inline val LivingEntity.hasDied get() = isDeadOrDying
```
NopoTheGamer: what is the point of this

src/main/java/at/hannibal2/skyhanni/utils/PlayerPosData.kt
```diff
@@ -0,0 +1,52 @@
...
+    val positions: List<LorenzVec> get() = playerPositions
+
+    @HandleEvent(priority = HandleEvent.HIGHEST)
+    fun onTick() {
+        playerPositions.addFirst(LocationUtils.playerLocation())
+        if (playerPositions.size > SIZE) playerPositions.removeLast()
+    }
+
+    @HandleEvent
+    fun onWorldChange() = playerPositions.clear()
```
NopoTheGamer: i am personally not a fan of strange functions on handle events

NopoTheGamer (changes requested):
alot of todos..

src/main/java/at/hannibal2/skyhanni/config/features/fishing/FishingConfig.kt
```diff
@@ -123,4 +130,27 @@ class FishingConfig {
...
     var muteBanshee: Boolean = false
+
+    @Expose
+    @ConfigOption(
+        name = "Sea Creature Kill Time",
+        desc = "Sends a chat message that shows how much time it took to kill a specific sea creature."
+    )
+    @ConfigEditorBoolean
+    @FeatureToggle
+    var seaCreatureKillTimer: Boolean = false
```
NopoTheGamer, line 146: there should be an option to only show for ur own mobs, and then maybe make this on by default

NopoTheGamer (changes requested):

src/main/java/at/hannibal2/skyhanni/config/features/fishing/SeaCreatureHealthDisplayConfig.kt
```diff
@@ -0,0 +1,30 @@
...
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorInfoText
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorSlider
+import io.github.notenoughupdates.moulconfig.annotations.ConfigLink
+import io.github.notenoughupdates.moulconfig.annotations.ConfigOption
+
+class SeaCreatureHealthDisplayConfig {
+
+    @Expose
+    @ConfigOption(name = "Enabled", desc = "Shows a GUI with the health of the sea creatures shown.")
+    @ConfigEditorBoolean
```
ItsEmpa, line 16: feature toggle

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,286 @@
...
+        if (lastPos.distance(playerPos) > MAX_WAIT_DEATH_DISTANCE) return false
+        val timeAroundPos = PlayerPosData.timeAtPos(lastPos, MAX_WAIT_DEATH_DISTANCE) ?: return false
+        if (timeAroundPos < 5.seconds) return false
+        return true
+    }
+
+    @HandleEvent
+    fun onWorldChange() = reset()
+
+    val BABY_MAGMA_SLUG = SeaCreature(
```
ItsEmpa: this was just a workaround i added in nautilus, you should probably just add it to repo instead

ItsEmpa (changes requested):
theres probably some of my old todos that you can remove also

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
```diff
@@ -0,0 +1,123 @@
...
+import at.hannibal2.skyhanni.utils.LorenzRarity
+import at.hannibal2.skyhanni.utils.LorenzVec
+import at.hannibal2.skyhanni.utils.MobUtils.getLorenzVec
+import at.hannibal2.skyhanni.utils.SimpleTimeMark
+import at.hannibal2.skyhanni.utils.render.WorldRenderUtils.exactBoundingBoxExtraEntities
+import at.hannibal2.skyhanni.utils.render.WorldRenderUtils.exactLocation
+import net.minecraft.world.entity.LivingEntity
+import net.minecraft.world.phys.AABB
+
+data class LivingSeaCreatureData(
```
hannibal002: a data class with var properties looks sus. make all val or remove the data class. what should the comparison actually compare? i assume just the entity id, right?

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
```diff
@@ -0,0 +1,123 @@
...
+        SeaCreatureEvent.Remove(this).post()
+    }
+
+    private var canBeSeenCache = false
+
+    fun canBeSeen(): Boolean = canBeSeenCache
+
+    private fun updateCanBeSeen(): Boolean {
+        val mob = mob ?: return false
+        mob.baseEntity.canBeSeen()
```
hannibal002: dead code?

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    inline val displayName: String get() = seaCreature.displayName
+
+    inline val isRare: Boolean get() = seaCreature.rare
+
+    inline val rarity: LorenzRarity get() = seaCreature.rarity
+
+    inline val health: Int? get() = mob?.health?.toInt()
+
+    inline val despawnTime: SimpleTimeMark get() = spawnTime + SeaCreatureDetectionApi.DESPAWN_TIME
```
hannibal002: unused

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
```diff
@@ -0,0 +1,123 @@
...
+    fun update(renderWorld: SkyHanniRenderWorldEvent) {
+        lastUpdate = SimpleTimeMark.now()
+        val mob = mob ?: return
+        actualLastPos = mob.getLorenzVec()
+        if (!updateCanBeSeen()) return
+        aabb = renderWorld.exactBoundingBoxExtraEntities(mob)
+        pos = renderWorld.exactLocation(mob)
+    }
+
+    override fun toString(): String {
```
hannibal002: unused, since this is a data class. you might want to remove the data class entirely, so that this custom tostring works

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
```diff
@@ -0,0 +1,123 @@
...
+    init {
+        if (canBeSeen()) {
+            aabb = mob?.boundingBox
+            pos = mob?.getLorenzVec()
+        } else {
+            aabb = null
+            pos = null
+        }
+        updateCanBeSeen()
+        actualLastPos = mob?.getLorenzVec() ?: LorenzVec(0.0, 0.0, 0.0)
```
hannibal002: using location 000 is an error prone design. can we just set it to null, and make actualLastPos nullable?

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    /** This tracks the last bounding box of the sea creature that the user was able to see. */
+    var aabb: AABB?
+        private set
+
+    var lastUpdate: SimpleTimeMark = SimpleTimeMark.now()
+        private set
+
+    /** This tracks the real last position of the sea creature. Don't display this to the user */
+    internal var actualLastPos: LorenzVec
```
hannibal002: why do we need this? cant we just always calculate `mob?.getLorenzVec()` ?

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,276 @@
...
+        seaCreatures[mob] = data
+        entityIdToData[mob.entityId] = data
+        SeaCreatureEvent.Spawn(data).post()
+    }
+
+    private fun handleOwnMob() {
+        if (lastSeaCreatureFished.passedSince() > 1.seconds) return
+        val name = lastNameFished ?: return
+        val lastBobber = lastBobberLocation ?: return
+        // TODO: create a sortedByFiltered function that removes elements when the comparator returns null
```
hannibal002: impl the todo

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,276 @@
...
+        val mobs = recentMobs.asSequence().filter { (mob, data) -> mob.name == name && data.passedSince() < 1.5.seconds }.map {
+            it to it.key.baseEntity.distanceTo(lastBobber)
+        }.filter { it.second <= 3 }
+            .sortedBy { it.second }
+            .take(mobsToFind).toList()
+
+        if (mobs.isEmpty()) return
```
hannibal002: duplicate code with inside handleBabySlugs()

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,276 @@
...
+        isOwn: Boolean = false,
+    ) {
+        val seaCreature = SeaCreatureManager.allFishingMobs[mob.name] ?: return
+        val data = LivingSeaCreatureData(isOwn, seaCreature, mob.entityId, time, mob)
+        seaCreatures[mob] = data
+        entityIdToData[mob.entityId] = data
+        SeaCreatureEvent.Spawn(data).post()
+    }
+
+    private fun handleOwnMob() {
```
hannibal002, line 136: what about the edge case when in a 1.5s time frame two ppl throw the bobber? in that case we should not detect anything at all, rather than having a 50% chance of false positive imo

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,276 @@
...
+        if (lastSeaCreatureFished.passedSince() > 1.seconds) return
+        val name = lastNameFished ?: return
+        val lastBobber = lastBobberLocation ?: return
```
hannibal002: this value only gets updated per tick. is this always null when the onMobSpawn gets fired?

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,276 @@
...
+        val mob = event.mob
+        val data = entityIdToData[mob.entityId]
+        if (data != null) {
+            seaCreatures[mob] = data
+            data.mob = mob
+            SeaCreatureEvent.ReDetect(data).post()
+            return
+        }
+
+        if (mob.name == "Baby Magma Slug") {
```
hannibal002: repo

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,276 @@
...
+            recentBabyMagmaSlugs[mob] = SimpleTimeMark.now()
+            // TODO: test if baby magma slugs work without delayed run
```
hannibal002: impl todo

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,276 @@
...
+    @HandleEvent(onlyOnSkyblock = true)
+    fun onSecondPassed() {
+        val playerPos = LocationUtils.playerLocation()
+        for ((_, data) in entityIdToData) {
+            if (!assumeDeathIfAreaLeft(data, playerPos)) continue
+            data.sendDeath(false)
+        }
+    }
+
+    fun assumeDeathIfAreaLeft(data: LivingSeaCreatureData, playerPos: LorenzVec): Boolean {
```
hannibal002: can be private

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,276 @@
...
+                if (mob.name == "Magma Slug") {
+                    lastMagmaSlugLocation = mob.getLorenzVec()
+                    babyMagmaSlugsToFind += 3
+                    lastMagmaSlugTime = SimpleTimeMark.now()
+                    handleBabySlugs()
+                }
+            }
+            data.sendDeath()
+            return
+        } else if (oldId != newId) { // we update the entity id in case the baseEntity has changed at some point
```
hannibal002, line 108: why would this be the case? the mob detection logic uses entity id as key, so base entity is always same entity id. or am i missing an edge case? if so, what?

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    // TODO: remove workaround
+    private fun isActive(): Boolean = !DungeonApi.inDungeon()
+
+    @HandleEvent
+    fun onDebug(event: DebugDataCollectEvent) {
+        event.title("Sea Creatures")
+        event.addIrrelevant {
+            "entityIdToData" to entityIdToData.entries
```
hannibal002: only thing you do here is create a pair, not adding it to the list of debug lines.
this does nothing rn. you need to call add(Pair).

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,276 @@
...
+        babyMagmaSlugsToFind = 0
+        lastMagmaSlugTime = SimpleTimeMark.farPast()
+        lastSeaCreatureFished = SimpleTimeMark.farPast()
+        lastNameFished = null
+        mobsToFind = 0
+    }
+
+    @HandleEvent
+    fun onCommand(event: CommandRegistrationEvent) {
+        event.registerBrigadier("skyhanniresetdata") {
```
hannibal002: generic name

src/main/java/at/hannibal2/skyhanni/utils/PlayerPosData.kt
```diff
@@ -0,0 +1,54 @@
...
+    private const val TIME_SECONDS = 30
+    private const val SIZE = TIME_SECONDS * 20
+
+    val time = TIME_SECONDS.seconds
+
+    private val playerPositions = ArrayDeque<LorenzVec>()
+    val positions: List<LorenzVec> get() = playerPositions
+
+    @HandleEvent(priority = HandleEvent.HIGHEST)
+    fun onTick() {
```
hannibal002: should use EntityMoveEvent instead

src/main/java/at/hannibal2/skyhanni/utils/PlayerPosData.kt
```diff
@@ -0,0 +1,54 @@
...
+import at.hannibal2.skyhanni.utils.TimeUtils.ticks
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
+/**
+ * Tracks the player's position over the last 30 seconds. THe index of the position in the list
+ * corresponds to the time in ticks since the position was recorded, with the most recent position being at index 0.
+ * This makes it so that the time can inaccurate by up to 1 tick.
+ */
+@Suppress("unused")
```
hannibal002: why is this unused suppressed here? only timeSinceLastAt and positions are actually unused

src/main/java/at/hannibal2/skyhanni/utils/EntityUtils.kt
```diff
@@ -233,4 +234,7 @@ object EntityUtils {
     // TODO use derpy() on every use case
     val LivingEntity.baseMaxHealth: Int
         get() = this.getAttributeBaseValue(Attributes.MAX_HEALTH).toInt()
+
+    inline val Entity.spawnTime: SimpleTimeMark get() = SimpleTimeMark.now() - tickCount.ticks
```
hannibal002: simple time mark uses real time, tick count is just an approximation. this should not be accurate with small lags. we have ServerTimeMark for this

src/main/java/at/hannibal2/skyhanni/utils/MobUtils.kt
```diff
@@ -86,6 +86,10 @@ object MobUtils {
         return possibleEntities.distinct().sortedBy { it.baseEntity.distanceTo(pos) }
     }
 
+    fun Mob.getLorenzVec() = baseEntity.getLorenzVec()
```
hannibal002: belongs in mob file directly. no reason for extended function

src/main/java/at/hannibal2/skyhanni/utils/MobUtils.kt
```diff
@@ -86,6 +86,10 @@ object MobUtils {
...
+
+    inline val Mob.entityId get() = baseEntity.id
```
hannibal002: unnecessary. id already exists in mob class

src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
```diff
@@ -56,6 +57,16 @@ object ConfigUtils {
         return true
     }
 
+    fun <T : Any> T.asProperty(): Property<T> = Property.of(this)
```
hannibal002: all three are unused

src/main/java/at/hannibal2/skyhanni/events/fishing/SeaCreatureEvent.kt
```diff
@@ -0,0 +1,26 @@
+package at.hannibal2.skyhanni.events.fishing
+
+import at.hannibal2.skyhanni.api.event.SkyHanniEvent
+import at.hannibal2.skyhanni.features.fishing.LivingSeaCreatureData
+
+sealed class SeaCreatureEvent(val seaCreature: LivingSeaCreatureData) : SkyHanniEvent() {
```
hannibal002, line 6: either duplciate the structure of MobEvent (especially the names of the events) or better, directly merge this into MobEvent.

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,276 @@
...
+            recentBabyMagmaSlugs.remove(mob)
+        }
+        if (babyMagmaSlugsToFind == 0) {
+            lastMagmaSlugLocation = null
+        }
+    }
+
+    @HandleEvent(onlyOnSkyblock = true, priority = HandleEvent.HIGHEST)
+    fun onRenderWorld(event: SkyHanniRenderWorldEvent) {
+        for (data in seaCreatures.values) data.update(event)
```
hannibal002: this update should not happen per frame, only per tick. then, the cached data should get fired in the render event. but no can be seen checks in the render logic please

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,276 @@
...
+            val time = mob.baseEntity.spawnTime
+            addMob(mob, time, isOwn = true)
+            recentBabyMagmaSlugs.remove(mob)
+        }
+        if (babyMagmaSlugsToFind == 0) {
+            lastMagmaSlugLocation = null
+        }
+    }
+
+    @HandleEvent(onlyOnSkyblock = true, priority = HandleEvent.HIGHEST)
```
hannibal002, line 184: why highest?

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
```diff
@@ -0,0 +1,123 @@
...
+        val mob = mob ?: return
+        actualLastPos = mob.getLorenzVec()
+        if (!updateCanBeSeen()) return
+        aabb = renderWorld.exactBoundingBoxExtraEntities(mob)
+        pos = renderWorld.exactLocation(mob)
+    }
+
+    override fun toString(): String {
+        return buildString {
+            fun appendInfo(string: String) = append("$string, ")
```
hannibal002: if the file keeps its custom function, how about using a format like this to avoid the inner function?

```kt
override fun toString(): String = listOf(
"isOwn=$isOwn",
"seaCreature=$seaCreature",
"entityId=$entityId",
"spawnTime=$spawnTime",
"mob=$mob",
"pos=$pos",
"aabb=$aabb",
).joinToString(", ", "SeaCreatureData(", ")")
```

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -277,6 +278,7 @@ enum class ConfigFileType(val fileName: String, val clazz: Class<*>, val propert
     STORAGE("storage", StorageData::class.java, SkyHanniMod::storageData),
     ROUTES("routes", OrderedWaypointsRoutes::class.java, SkyHanniMod::orderedWaypointsRoutesData),
     CUSTOM_TODOS("custom_todos", CustomTodosStorage::class.java, SkyHanniMod::customTodos),
+    SEA_CREATURES("SeaCreatureSettings", SpecificSeaCreatureStorage::class.java, SkyHanniMod::seaCreatureStorage),
```
hannibal002: wrong file name format

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    inline val isRare: Boolean get() = seaCreature.rare
+
+    inline val rarity: LorenzRarity get() = seaCreature.rarity
+
+    inline val health: Int? get() = mob?.health?.toInt()
+
+    inline val despawnTime: SimpleTimeMark get() = spawnTime + SeaCreatureDetectionApi.DESPAWN_TIME
+
+    fun isLoaded(): Boolean = entity != null
```
hannibal002: how about renaming this to exists() ?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
Fazfoxy: Hanni I need to understand your thought process here because, for all intents and purposes from all my understanding of Empa's writing here, own bobber is always accurate as Minecraft provides us an ownership info for the bobber, so from my perspective the only way this could be inaccurate is if 2 people hooked up the same mob, while overlapping bobbers within an extremely low period of time since this is only called within on fish events, which trigger from a Hypixel sent chat message which should be impossible for another player to trigger, or from entity entering world events, which should primarily be ignored unless it's the same mob as you fish EXTREMELY close time, but they're sorted by distance to the bobber which is theoretically difficult for someone to make closer than your own bobber should be?

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
Fazfoxy: This is meant to persist post death for detecting stuff like Death time if the player gets close and the Entity is not present close enough to the position it should be when the area around it re-appears which I don't believe mob?.getLorenzVec() does?

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
ItsEmpa: the priority is highest so that if other features want to, for example, render something with the position of a sea creature, they can use the updated position of the current event

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
ItsEmpa: i dont think t his needs to be repo personally, the name of baby magma slugs isnt gonna change.

in fact, [FishingTimer](https://github.com/hannibal002/SkyHanni/blob/cc3883c2c915359423fdd408ff835905975b9431/src/main/java/at/hannibal2/skyhanni/features/fishing/FishingTimer.kt#L93) already uses a hardcoded Baby Magma Slug

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
ItsEmpa: the base entity of mobs can change, its not immutable

src/main/java/at/hannibal2/skyhanni/events/fishing/SeaCreatureEvent.kt
ItsEmpa: wdym duplicate the structure of mob event?

also this cant really be merged into MobEvent

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
ItsEmpa: exists is too generic of a name imo

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
ItsEmpa: what

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
hannibal002: As long as you don't remove the deleted/invalid mob detection data, I don't think we manually remove the location data in it on despawn.

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
hannibal002: Oh, I overlooked the check with the own bobber.

Was just thinking about the many small bugs we had with bait used detection bc of many users throwing their rods next to each other. But yeah, the bait is way easier to differentiate bc of the clear ownership tag

Fazfoxy: I have also kinda messed with this to make it not require ontick updating for itself and instead updates a holding value with the previous bobber's location which, should? be more performant

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
ItsEmpa: mob.getLorenzVec() uses the base entity as backing data, so if you move far away enough that its out of render distance, you cant use that

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
hannibal002: The value last bobber location gets set in the update function somewhere else. This call here happens in the mob spawn event. Without testing, this reads as if we try to work with a variable that is always nulll at this point in time.

Or am I missing something? Is the update function getting called before the mob spawn event fired this code block?

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
hannibal002: Yeaj fair. Maybe we move both cases of this name to a had coded string somewhere tho? E.g. fishing API?

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
hannibal002: This sounds really bad. Is this some edge case hypxiel broke consistency for fishing? 

Don't remember having this issue somewhere else in the mod so far when working with entity ids

src/main/java/at/hannibal2/skyhanni/events/fishing/SeaCreatureEvent.kt
hannibal002: If it's too different in the impl of mob event then leave it separate, i agree.

But about the different names:
Look at the event names in this file. Then look at the event names in mob event. They use different terms for the same state.

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
Fazfoxy: > i dont think t his needs to be repo personally, the name of baby magma slugs isnt gonna change.

idk if I, fully, agree, Hypixel have shown to change mob names like Loch Emperor, it's not entirely impossible they rename a mob, but that would require a more full difference adding like a "Hypixel Name" to Repo for mobs that HAVE changed names or something idk

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
hannibal002: Fair

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
hannibal002: Yeah I agree repo is prob overkill at this point. Tho we now have at two different files the same string for the same purpose. We should really put this in a shared variable so that we don't later fix it just in one file and forget about the other file. This is not a refactor as large as adding s new repo system for just one string.

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
hannibal002: Interesting. How about we fix this in mob detection directly instead of here? I assume the same problem will arise at other features..

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
ItsEmpa: no, its just something that can happen with mob detection, and it has always been like this

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
ItsEmpa: MobDetectino creates a new Mob instance if you get out of render distance and then enter it again. however, this actually keeps the same sea creature data, so Mob shouldnt actually have this

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
Fazfoxy: Moved to FishingApi for both Fishing Timer & This.

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
hannibal002: When exactly does mob detection create a new instance? Only when you come close enough again, right?
So for the time while we were stuck with an invalid mob detection object, we could store the last known location inside mob detection, and once the new instance is there and gets used by the fishing data object here it swaps internal object and inntheory the location should not change at all. Or is this that impractical to store the invalid mob object data? Like, we already store the invalid (assumed) mob location.

This change would just avoid this one variable here and instead allow for mob detection instances to also return the last known location if the Minecraft entity is not in render distance.

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
hannibal002: Hmm. Please add a to-do here so we can some time later fix the actual bug in mob detection

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
ItsEmpa: mob detection is currently set up so that you should ALWAYS discard the Mob instance when it despawns. changing this is wayyy more complicated than the scope of this pr

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
ItsEmpa: i think this is intended and not really a bug?  you would need to ask thunder for more info about this i think

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
hannibal002: fair

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
Fazfoxy: > Or am I missing something? Is the update function getting called before the mob spawn event fired this code block?

no you are entirely right, the hook death trigger was happening later than I expected, this was "fixed" by making it use current bobber but I don't entirely trust that in terms of actually being accurate so at Empa's suggestion I changed it to storing the last known bobber update position & when it was last updated and checking against a timemark stored at the same time within a pair to avoid race conditions between events being fired

src/main/java/at/hannibal2/skyhanni/events/fishing/SeaCreatureEvent.kt
Fazfoxy: >  Look at the event names in this file. Then look at the event names in mob event. They use different terms for the same state.

as mentioned in Discord neither me/Empa can tell what you mean by them using the same terms for different statuses the only real difference is that this is intended to persist information between when a mob despawns to when it seems to die which mob detection does not do. Have marked this as resolved since I can't fully tell what you mean with the comments.

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
Fazfoxy: Changed to null, I think? This should be impossible to hit anyway.

Fazfoxy:
whenever this merges it would be preferable to have a known bugs post about how cocooned mobs aren't marked as your own because of Hypixel and that is ISN'T a mod bug

I'd say something like "Hypixel Currently Removes ownership data from mobs that leave cocoons, and only gives full drops/XP to the user who performs the final hit with anyone else who has hit lootshare conditions lootsharing as expected, due to this logic stuff like Sea Creature kill timer & Lootshare Sphere will act as if it is not your sea creature once it leaves the cocoon."

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,280 @@
...
+
+    private val recentMobs = mutableMapOf<Mob, ServerTimeMark>()
+
+    private var babyMagmaSlugsToFind = 0
+    private var lastMagmaSlugLocation: LorenzVec? = null
+    private var lastMagmaSlugTime = SimpleTimeMark.farPast()
+
+    private val recentBabyMagmaSlugs = mutableMapOf<Mob, ServerTimeMark>()
+
+    var lastBobberLocation: LorenzVec? = null
```
hannibal002, line 59: does no longer get reset on world change. intended?

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureDetectionApi.kt
```diff
@@ -0,0 +1,280 @@
...
+            recentMobs.remove(mob)
+        }
+
+        if (mobsToFind == 0) {
+            lastNameFished = null
+            lastBobberLocation = null
+        }
+    }
+
+    private fun findMobs(mobs: List<Pair<Map.Entry<Mob, ServerTimeMark>, Double>>, toBeFound: Int): Pair<Int, List<Mob>>? {
```
hannibal002: this line is hard to read. create a small class or different data structure to increase maintainable.

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    @Suppress("HandleEventInspection")
+    fun updateWorld(renderWorld: SkyHanniRenderWorldEvent) {
+        val mob = mob ?: return
+        if (!canBeSeenCache) return
+        aabb = renderWorld.exactBoundingBoxExtraEntities(mob)
+        pos = renderWorld.exactLocation(mob)
+    }
+
+    override fun toString(): String = listOf(
```
hannibal002: have you tested what this toString function's result looks like?  looks ugly to me.

also, why do we even want this? we dont have similar tostring functions for other custom data types. if we wanna debug the data, we can use DebugDataCollectEvent

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/fishing/LivingSeaCreatureData.kt
Fazfoxy: This is used under the debug collection event in DetectionApi for an output of (with 1 mob)


== Sea Creatures ==
 EntityIdToData: [347608=SeaCreatureData(SeaCreatureData(, isOwn=true, seaCreature=SeaCreature(name=Rider of the Deep, fishingExperience=270, chatColor=§b, rare=false, rarity=UNCOMMON), entityId=347608, spawnTime=ServerTimeMark(ticks=1779, now=1848), mob=Rider of the Deep - 347608, pos=LorenzVec(x=45.410400390625, y=75.00341796875, z=-12.67041015625), aabb=AABB[45.17863638401032, 75.00341796875, -12.877983140945435] -> [45.61040039360523, 76.32841799855233, -12.470410153269768], ))]
 seaCreatures: [Rider of the Deep - 347608=SeaCreatureData(SeaCreatureData(, isOwn=true, seaCreature=SeaCreature(name=Rider of the Deep, fishingExperience=270, chatColor=§b, rare=false, rarity=UNCOMMON), entityId=347608, spawnTime=ServerTimeMark(ticks=1779, now=1848), mob=Rider of the Deep - 347608, pos=LorenzVec(x=45.410400390625, y=75.00341796875, z=-12.67041015625), aabb=AABB[45.17863638401032, 75.00341796875, -12.877983140945435] -> [45.61040039360523, 76.32841799855233, -12.470410153269768], ))]
 lastNameFished null
 mobsToFind: 0
 lastSeaCreatureFished 2026-02-25T00:02:25.941Z
 recentMobs []
 babyMagmaSlugsToFind 0
 lastMagmaSlugLocation null
 lastMagmaSlugTime The Far Past
 recentBabyMagmaSlugs []

hannibal002: > EntityIdToData: [347608=SeaCreatureData(SeaCreatureData(, isOwn=true, 

`SeaCreatureData(SeaCreatureData(,`
yeah idk i think the format is a bit broken here

hannibal002: i think its better to rather manually print the info we want in the debug collection event, then say exactly what we actualy want. and send it in a better readable format, sort locations by 1 digit, etc.

hannibal002 (approved):

PR #5111, opened 2026-01-18
Improvement: Ubik's cube cooldown accuracy

larswmh:
## What
This PR improves the Ubik's Cube reminder in the Rift Dimension.

The current implementation triggers the two hour cooldown based on the final round chat message. However, it has been observed that the cooldown does not reset at that point. Interacting with the NPC immediately afterwards shows a different remaining time, which then gets corrected by the resync logic. This results in the reminder being slightly delayed.

This PR changes the cooldown tracking to start when the “Split or Steal” inventory is opened, which appears to be the point at which the server updates the Ubik's Cube cooldown.

As a result, the reminder timing is now accurate and no longer offset.

## Changelog Improvements
+ Improved Ubik's Cube reminder accuracy - larswmh


NopoTheGamer:
#5057

PR #5114, opened 2026-01-18
Improvement: Add SkyBlock ID to crop ItemStacks

lunaynx:
## What
Added SkyBlock IDs to the crop ItemStacks we render so that resource packs like [PHQ Garden Flowers](https://modrinth.com/resourcepack/phq-garden-flowers) can retexture them.

## Changelog Improvements
+ Added support for resource packs to retexture crop icons rendered by SkyHanni. - Luna
    * This is mainly relevant for Moonflowers, as they need to be distinguished from Blue Orchids.


NopoTheGamer:
does this cause issues if the repo is loaded later?

lunaynx:
Hmm, yeah, good question, caching it as lazy might be a problem here.

CalMWolfs:
> Hmm, yeah, good question, caching it as lazy might be a problem here.

`NeuItemStackProvider` might work idk

PR #5124, opened 2026-01-20
Fix: Thunder Sparks not being Highlighted

Fazfoxy:
## What
Fixes Thunder Sparks not being Highlighted due to incorrect method selection (Seriously how many times to Hypixel ACTUALLY make it Held instead of worn)

## Changelog Fixes
+ Fixed Thunder Spark Highlight not working. - Fazfoxy


NopoTheGamer (changes requested):
please actually make the event work properly

PR #5141, opened 2026-01-22
Backend: Force addChatMessageToChat to run on client thread

lunaynx:
## What
Added a `DelayedRun.runOrNextTick` in `TextCompat.addChatMessageToChat`, which should hopefully fix calls such as `ChatUtils.chat` and `ChatUtils.debug` randomly causing "Rendersystem called from wrong thread" errors. Also changed `TextCompat.addDeletableMessageToChat` to use our utility method for consistency.

This probably obsoletes some calls being wrapped in `run(Or)NextTick` but I can't be bothered to go through all of them and figure out which ones are safe to remove right now.

+ Made `TextCompat.addChatMessageToChat` always run on the client thread to avoid rendersystem errors on 1.21.10 and above. - Luna


lunaynx:
image

NopoTheGamer:
does this cause lag

lunaynx:
Not sure, we should probably test it. I'll run my instance with it for a bit. But I'd be kinda surprised if it did, since `addDeletableMessageToChat` already did this (just called differently).

NopoTheGamer (approved):
seems to work without chat patches

PR #5151, opened 2026-01-24
Improvement: Add warning for Feather/Lunar Client

lunaynx:
## What
Added a warning that triggers on profile join if Feather Client or Lunar Client is detected.

## Changelog Improvements
+ Added a warning if you are using Feather Client or Lunar Client. - Luna
    * These clients are closed source and can cause various issues we are unable to fix.


hannibal002:
does both exist on 1.21? feather and lunuar?

NopoTheGamer:
i personally dont like the concept, these types of warnings only really serve to annoy users and dont really solve problems

hannibal002 (changes requested):
i agree with nopo. id say we close the pr

PR #5155, opened 2026-01-25
Fix: Not Clickable Items graying out composter items with overlay disabled

lunaynx:
## What
Fixed Not Clickable Items graying out items in Composter even if Composter Overlay is disabled.

## Changelog Fixes
+ Fixed "Only insert the selected items!" showing even when Composter Overlay is disabled. - Luna


src/main/java/at/hannibal2/skyhanni/features/inventory/HideNotClickableItems.kt
```diff
@@ -286,9 +286,8 @@ object HideNotClickableItems {
         return true
     }
 
-    @Suppress("UnusedParameter")
```
NopoTheGamer, line 289: who supressed this instead of just removing it :sob:

PR #5173, opened 2026-01-29
Improvement: Customizable plot border key

lunaynx:
## What
This should have been changed long ago for modern, but here we go.

## Changelog Improvements
+ The Garden Plot Borders feature is now activated with a custom keybind instead of F3+G. - Luna
    * If you previously had this option enabled, you will have to set a keybind in the config to continue using it.


NopoTheGamer:
this should probably just be on a customizable keybind with no f3 prefix anymore

lunaynx:
Yeah I could just do that I guess.

lunaynx:
There, seems to work.

src/main/java/at/hannibal2/skyhanni/features/garden/GardenPlotBorders.kt
```diff
@@ -5,45 +5,38 @@ import at.hannibal2.skyhanni.events.minecraft.KeyPressEvent
...
 @SkyHanniModule
 object GardenPlotBorders {
+    private val config get() = GardenApi.config
 
-    private val config get() = GardenApi.config.plotBorders
-    private var timeLastSaved = SimpleTimeMark.farPast()
     private var showBorders = false
 
     @HandleEvent
     fun onKeyPress(event: KeyPressEvent) {
```
NopoTheGamer: this should be key down event

NopoTheGamer (changes requested):

PR #5174, opened 2026-01-29
Fix: higher quantity crafts (e.g. bait)

HacktheTime:
## What
Fixed an exception for resource line not found if an item is crafted in single amounts with higher quantities. Essentially 16x Bait crafts like worm or hot bait.

## Changelog Fixes
+ Fixed Super Crafting Anti Waste not working for recipes that craft multiple items. - Hype_the_Time
    * Also covers coin waste for multi-craft recipes like 16x bait.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
@@ -89,8 +89,9 @@ object SuperCraftingInventory {
         if (event.slotId != PICKAXE_SLOT) return
         val slots = InventoryUtils.getItemsInOpenChestWithNull()
         val craftingAmount = getSuperCraftingCount(slots) ?: return
-        val craftMultiplier = getResultItem(slots).amount
-        val profit = getProfit(slots, craftingAmount, craftMultiplier) ?: return
+        val result = getResultItem(slots)
+        val craftMultiplier = result.amount
+        val profit = getProfit(slots, craftingAmount, craftMultiplier, result) ?: return
         val maxCraftingAmount = getSuperCraftingMaxCount(slots, craftingAmount, craftMultiplier)
         if (!blockWasteClick(profit, craftingAmount, maxCraftingAmount)) return
         SoundUtils.playErrorSound()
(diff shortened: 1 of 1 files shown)


hannibal002:
unused imports. please fix.

hannibal002:
@HacktheTime

hannibal002 (changes requested):
hi

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -108,27 +109,28 @@ object SuperCraftingInventory {
...
+            owned / matsPerCraft
+        }
 
-    private fun getProfit(slots: List<Slot>, craftingAmount: Long): Double? {
+    private fun getProfit(slots: List<Slot>, craftingAmount: Long, craftMultiplier: Int): Double? {
         val materials = getRecipeMaterials(slots)
-        val resultItem = getResultItem(slots)
 
-        val recipeMultiplier = resultItem.amount
+        val recipeMultiplier = craftMultiplier
```
hannibal002: rendundant variable

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -108,27 +109,28 @@ object SuperCraftingInventory {
...
+        val resultItem = getResultItem(slots)
```
hannibal002: getResultItem gets already called in onSlotClick. how about reuse the variable there and pass it in getProfit as parameter, so that we dont need to call getResultItem twice per SlotClickEvent?

hannibal002 (changes requested):
please follow the pr template for the proper format.

the phrase "Fixed Non 1x Crafts in Anti Waste Super Crafting" sounds confusing. how about
"Fixed Super Crafting Anti Waste not working for recipes that craft multiple items.".

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftingInventory.kt
```diff
@@ -89,8 +89,10 @@ object SuperCraftingInventory {
         if (event.slotId != PICKAXE_SLOT) return
         val slots = InventoryUtils.getItemsInOpenChestWithNull()
         val craftingAmount = getSuperCraftingCount(slots) ?: return
-        val profit = getProfit(slots, craftingAmount) ?: return
-        val maxCraftingAmount = getSuperCraftingMaxCount(slots, craftingAmount)
+        val result = getResultItem(slots)
+        val craftMultiplier = result.amount
+        val profit = getProfit(slots, craftingAmount, craftMultiplier, result) ?: return
```
hannibal002, line 98: The error check for `craftMultiplier == 0` is currently buried inside `getProfit`, but the value is already available here in `onSlotClick`. Moving the check right after `val craftMultiplier = result.amount` would make the validation easier to follow and simplify `getProfit`.

hannibal002 (changes requested):

hannibal002 (approved):

PR #5185, opened 2026-02-01
Fix: MobHighlight

MTOnline69:
## What
Revival of #2114

Refactors the MobHighlight class to (mostly) use mob detection. For highlighting zealots the old EntityMaxHealthUpdateEvent is used instead, as mob detection currently does not work with mobs that are far away.

Corrupted highlight still doesn't work the best since it should probably use nametags instead, but somebody else can do that.

## Changelog Fixes
+ Fixed Corleone, Arachne, and Zealot-related highlight features not working. - Empa & MTOnline
+ Fixed Line to Arachne not working. - Empa & MTOnline

+ Refactored MobHighlight to use Mob Detection. - Empa & MTOnline
    * With the exception of Zealot-related highlights.


src/main/java/at/hannibal2/skyhanni/features/combat/mobs/MobHighlight.kt
```diff
@@ -11,171 +12,98 @@ import at.hannibal2.skyhanni.mixins.hooks.RenderLivingEntityHelper
...
+            else -> return
         }
 
-        if (entity is Spider) {
-            checkArachne(entity)
-        }
+        RenderLivingEntityHelper.setEntityColorWithNoHurtTime(
+            entity,
+            color.toColor().addAlpha(127),
+        ) { true }
```
hannibal002: true here as condition is wrong. if the player toggles the option, the color would stay there.

src/main/java/at/hannibal2/skyhanni/features/combat/mobs/MobHighlight.kt
```diff
@@ -11,171 +12,98 @@ import at.hannibal2.skyhanni.mixins.hooks.RenderLivingEntityHelper
...
-        ) { config.arachneBossHighlighter }
-    }
-
-    private fun markArachne(entity: LivingEntity) {
-        RenderLivingEntityHelper.setEntityColorWithNoHurtTime(
-            entity,
-            LorenzColor.RED.toColor().addAlpha(50),
-        ) { config.arachneBossHighlighter }
-    }
+    private fun LivingEntity.isZealotOrBruiser() = baseMaxHealth == 13_000 || baseMaxHealth == 65_000
```
hannibal002: missing runic checks. old code did check if x or x * 4

src/main/java/at/hannibal2/skyhanni/features/combat/mobs/MobHighlight.kt
```diff
@@ -11,171 +12,98 @@ import at.hannibal2.skyhanni.mixins.hooks.RenderLivingEntityHelper
...
+            entity,
+            LorenzColor.DARK_PURPLE.toColor().addAlpha(127),
+        ) { config.corruptedMobHighlight }
     }
 
-    @HandleEvent(onlyOnSkyblock = true)
-    fun onEntityHealthUpdate(event: EntityMaxHealthUpdateEvent) {
+    // Mob detection isn't used here to allow for highlighting Zealots from further away.
+    @HandleEvent(onlyOnIsland = IslandType.THE_END)
+    fun onEntityJoinWorld(event: EntityMaxHealthUpdateEvent) {
```
hannibal002: event and function name mismatch

src/main/java/at/hannibal2/skyhanni/features/combat/mobs/MobHighlight.kt
```diff
@@ -11,171 +12,98 @@ import at.hannibal2.skyhanni.mixins.hooks.RenderLivingEntityHelper
...
+                arachne = mob
+                LorenzColor.RED
+            }
+
+            mob.isRunic && config.runicMobHighlight -> LorenzColor.LIGHT_PURPLE
+
+            else -> return
+        }.toChromaColor()
+
+        mob.highlight(color)
```
hannibal002: using the new .highlight() instead of the old setEntityColorWithNoHurtTime causes the color to be way darker. in the long run, the color should be done in the config anyway, but for now the color should just not be different because of this pr.

hannibal002 (changes requested):
the cleanup and migration to mob detection looks good!

small suggetions, and i assume you have tested that the code works well with death/despawn/world switch?

src/main/java/at/hannibal2/skyhanni/features/combat/mobs/MobHighlight.kt
MTOnline69: Doesn't baseMaxHealth stay the same no matter whether the mob is runic, corrupted or neither?

lunaynx: No, it's misleadingly named. It's simply the server max health attribute, which often ends up being higher than usual if the mob is corrupted, runic, etc.

(Corrupted should be an impossible state for mobs that are neither sea creatures nor Crimson Isle mobs now, and corrupting sea creatures no longer increases their health or max health anyway, so only runic needs to be checked here.)

src/main/java/at/hannibal2/skyhanni/features/combat/mobs/MobHighlight.kt
```diff
@@ -11,171 +12,113 @@ import at.hannibal2.skyhanni.mixins.hooks.RenderLivingEntityHelper
...
-            ) { config.corleoneHighlighter }
-        }
+        val heldBlock = entity.getBlockInHand()?.block
 
-        if (entity is EnderMan) {
-            val isZealot = maxHealth == 13_000 || maxHealth == 13_000 * 4 // runic
-            val isBruiser = maxHealth == 65_000 || maxHealth == 65_000 * 4 // runic
+        val (color, isEnabled) = when {
+            heldBlock == Blocks.END_PORTAL_FRAME ->
+                LorenzColor.DARK_RED to config.specialZealotHighlighter
```
hannibal002: before this patch, speical zealot was with alpha 50, and everything else with alpha 127. is this change intended?

src/main/java/at/hannibal2/skyhanni/features/combat/mobs/MobHighlight.kt
```diff
@@ -11,171 +12,113 @@ import at.hannibal2.skyhanni.mixins.hooks.RenderLivingEntityHelper
...
+            else -> return
         }
 
-        if (entity is Spider) {
-            checkArachne(entity)
-        }
+        RenderLivingEntityHelper.setEntityColorWithNoHurtTime(
+            entity,
+            color.toColor().addAlpha(127),
+        ) { isEnabled }
```
hannibal002: This will set the value to whatever it was the moment this event gets updated, not a periodic check. You need to return the function call in when block so that the render living entity helper can calculate the live value of the config option itself. One easy way to do this is this: 
image

src/main/java/at/hannibal2/skyhanni/features/combat/mobs/MobHighlight.kt
```diff
@@ -11,171 +12,113 @@ import at.hannibal2.skyhanni.mixins.hooks.RenderLivingEntityHelper
...
+                LorenzColor.LIGHT_PURPLE to config.runicMobHighlight
+
+            else -> return
         }
-        toHighlightRunicMobs.clear()
+
+        RenderLivingEntityHelper.setEntityColorWithNoHurtTime(
+            mob.baseEntity,
+            color.toColor().addAlpha(127),
+        ) { isEnabled }
```
hannibal002: same bug here.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/combat/mobs/MobHighlight.kt
NopoTheGamer: ill be so real i dont think the alpha changes anything for glowing

hannibal002 (approved):

PR #5188, opened 2026-02-01
Improvement: remove more 1.8/version specific stuff from docs

Nealkitt:
## What
i feel like, we don't need to say what version we are every single time anymore because it will only be the supported versions on hypixel, we're a hypixel skyblock mod...

exclude_from_changelog

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -306,6 +306,12 @@ written in [Kotlin DSL](https://docs.gradle.org/current/userguide/kotlin_dsl.htm
 
 This start script will automatically download all required libraries.
 
+### NotEnoughUpdates
+
+Although NEU is not a modern version mod, we use NEU API's to get auction house and bazaar price data for items and to read 
+the [NEU Item Repo](https://github.com/NotEnoughUpdates/NotEnoughUpdates-REPO) for item internal names, display names
+and recipes.
+
 ### Config
 
 SkyHanni stores the config (settings and user data) as a json object in a single text file.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -306,13 +306,6 @@ written in [Kotlin DSL](https://docs.gradle.org/current/userguide/kotlin_dsl.htm
 
 This start script will automatically download all required libraries.
 
-### NotEnoughUpdates
-
-SkyHanni requires **[NotEnoughUpdates](https://github.com/NotEnoughUpdates/NotEnoughUpdates/)**.
-We use NEU to get auction house and bazaar price data for items and to read
-the [NEU Item Repo](https://github.com/NotEnoughUpdates/NotEnoughUpdates-REPO) for item internal names, display names
-and recipes.
```
DavidArthurCole, line 314: These lines are still accurate.

Nealkitt: is this a better wording to fit how we use it?, we don't actually use it, just the services around it

CONTRIBUTING.md
```diff
@@ -308,8 +308,7 @@ This start script will automatically download all required libraries.
 
 ### NotEnoughUpdates
 
-SkyHanni requires **[NotEnoughUpdates](https://github.com/NotEnoughUpdates/NotEnoughUpdates/)**.
-We use NEU to get auction house and bazaar price data for items and to read
+Although NEU is not a modern version mod, we use NEU API's to get auction house and bazaar price data for items and read
```
NopoTheGamer: just auction house prices

NopoTheGamer (changes requested):

PR #5190, opened 2026-02-02
Backend: Use interface injection + related cleanup

lunaynx:
## What
Switched from accessWidener to classTweaker, added interface injection, replaced some old accessors with access widening, and some general cleanup.

+ Bumped minimum Fabric Loader version to 0.18.0. - Luna
+ Switched from accessWidener to classTweaker. - Luna
+ Added interface injection to eliminate casts to interfaces. - Luna


lunaynx:
~~Causes crash on launch outside dev env, need to investigate.~~ Fixed.

stonecutter.gradle.kts
```diff
@@ -134,7 +136,6 @@ stonecutter parameters {
     replacements {
         string(current.parsed >= "1.21.11") {
             replace("com.google.gson.internal.`\$Gson\$Types`", "com.google.gson.internal.GsonTypes")
-            replace("isF3Visible", "isOverlayVisible")
```
NopoTheGamer, line 137: how come?

NopoTheGamer (changes requested):

stonecutter.gradle.kts
lunaynx: I don't think it makes sense to declare a replacement for something that is used in a grand total of one place in a compat class, I replaced it with a comment. But I can change it back if you really want me to.

NopoTheGamer: no its ok

NopoTheGamer (approved):
cant wait for the support posts

PR #5192, opened 2026-02-02
Fix: Packet Cancellation Issues

lunaynx:
## What
Fixes issues with cancelling received packets due to bundles by moving relevant logic into mixins.

## Changelog Fixes
+ Fixed errors with features that mute sounds or hide particles/titles. - Luna

+ `PacketReceivedEvent` is no longer cancellable. - Luna
    * Due to issues posed by bundle packets and us being wary of modifying them, all relevant cancellation logic must now be done via mixins or other appropriate methods.
+ `SoundEvent` now also fires for client-side sounds. - Luna


src/main/java/at/hannibal2/skyhanni/mixins/transformers/gui/MixinGui.java
```diff
@@ -84,4 +88,12 @@ private boolean modifyRenderText(boolean bool) {
         return bool;
     }
 
+    @WrapMethod(method = {"setTitle", "setSubtitle"})
+    private void handleTitle(Component component, Operation<Void> original) {
+        String formattedText = TextCompatKt.formattedTextCompat(component);
+        if (!new TitleReceivedEvent(formattedText).post()) {
+            original.call(component);
+        }
+    }
```
NopoTheGamer: i know we dont currently use it but maybe set a boolean for title/subtitle?

NopoTheGamer (commented):
looks good

src/main/java/at/hannibal2/skyhanni/mixins/transformers/gui/MixinGui.java
lunaynx: Yeah we should, I just kept it as-is for now but I can probably add this tomorrow.

PR #5198, opened 2026-02-03
Fix: 3d skin layers compat issue

Bloxigus:
## What
Makes the injection point for removing armor skulls more specific to reduce the possibility of mod conflicts

Before:
image

After:
image

## Changelog Fixes
+ Fixed Hide Armor hiding the outer layer of some skulls occasionally when using 3d skin layers. - bloxigus


NopoTheGamer (changes requested):
this makes it so block helmets arent hidden

Bloxigus:
> this makes it so block helmets arent hidden

Oops, forgot skyblock did that

PR #5202, opened 2026-02-03
Fix: CME in GardenCropBreakTracker

lunaynx:
## What
Fixes a ConcurrentModificationException in GardenCropBreakTracker when you rapidly switch from farming one crop to another.

## Changelog Fixes
+ Fixed error when quickly switching between crops while farming. - Luna


NopoTheGamer (approved):
this mod is a cme magnet

PR #5211, opened 2026-02-05
Backend: redo the player name formatter with components

NopoTheGamer:
## What
cus i felt likt it
still has issues not yet fixed

## Changelog Fixes
+ Fixed something in Player Name Formatter. - nopo

+ redid the code for player name formatter. - nopo


hannibal002:
Moving to draft.

There's commented-out code that needs to be either properly ported or removed, throw Error() should use ErrorManager, and the PR description needs to explain properly why we change this and mention the known remaining issues properly.

hannibal002:
stale

PR #5217, opened 2026-02-06
Fix: Skill Progress Display not working

NopoTheGamer:
## What
i didnt mean to change this oops

## Changelog Fixes
+ Fixed Skill Progress Display not working. - nopo
+ Fixed Performance issue with the scoreboard. - nopo

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/SkillApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/SkillApi.kt
@@ -122,7 +122,7 @@ object SkillApi {
 
     @HandleEvent
     fun onActionBarUpdate(event: ActionBarUpdateEvent) {
-        val actionBar = event.chatComponent.string
+        val actionBar = event.chatComponent.string.removeColor()
         val components = SPACE_SPLITTER.splitToList(actionBar)
         for (component in components) {
             val matcher = listOf(skillPercentPattern, skillMultiplierPattern).firstOrNull {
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/api/SkillApi.kt
```diff
@@ -122,7 +122,7 @@ object SkillApi {
 
     @HandleEvent
     fun onActionBarUpdate(event: ActionBarUpdateEvent) {
-        val actionBar = event.chatComponent.string
+        val actionBar = event.chatComponent.string.removeColor()
```
hannibal002, line 125: In #5193 you changed `event.actionBar.removeColor()` to `event.chatComponent.string`. Now you want to change `event.chatComponent.string` to `event.chatComponent.string.removeColor()`.

So the intended change was `event.actionBar.removeColor()` to `event.chatComponent.string.removeColor()`?

Can't we directly change `actionBar` to `chatComponent.string` inside the event class itself?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/api/SkillApi.kt
NopoTheGamer: i didnt really mean to change it then.
the string with no colour codes will always be the same between them but the formatting could be different

PR #5223, opened 2026-02-09
Fix: Remove unnecessary cooldown on all burrows

SidOfThe7Cs:
## What
removes the warp delay for non spade guesses, the delay for spade is there to prevent you from warping while the spade particles are still being detected and was not removed, the delay for all guesses was added to prevent you from warping before the arrow particles themselves were detected but this simply doesnt happen as it waits for the chat message before removing the burrow (and triggering a target update allowing warp) which should arrive at the same time as particles and then you have to wait for the warp to actually be registered by the server before you get tped away

## Changelog Fixes
+ Fixed warping in diana having an unnecessary delay on non-spade guesses. - SidOfThe7Cs

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/event/diana/BurrowApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/event/diana/BurrowApi.kt
@@ -2,7 +2,6 @@ package at.hannibal2.skyhanni.features.event.diana
 
 import at.hannibal2.skyhanni.utils.LorenzVec
 import at.hannibal2.skyhanni.utils.SimpleTimeMark
-import kotlin.time.Duration.Companion.milliseconds
 
 object BurrowApi {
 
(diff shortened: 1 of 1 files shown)


hannibal002 (changes requested):
What problem does this solve, and are you sure this doesn't cause bugs?

The 400ms warp block was added intentionally to prevent warping immediately after clicking on a burrow. Removing it without explanation makes it hard to evaluate whether this is safe, and what's the reasoning that this won't cause issues?

Also, this is a user-facing behaviour change, so `exclude_from_changelog` doesn't apply here. From the [contributing guidelines](https://github.com/hannibal002/SkyHanni/blob/beta/CONTRIBUTING.md#no-category)

> Some changes don't fit any categories. E.g. when reverting pull requests or doing quick fixes to PRs merged immediately beforehand but not yet released in a beta.

This PR needs a changelog entry (likely under **Fixes** or **Improvements**) and a filled out **What** section explaining why removing the cooldown is safe.

PR #5225, opened 2026-02-09
Fix: Rei search bar focus check

SidOfThe7Cs:
## What
Cast to the base mc class the required method is in so there arnt mapping issues
screen null check is because it seems to still return true if you esc out of the screen while the search bar is focused

## Changelog Fixes
+ Fixed GUI Editor opening while typing in REI search box. - SidOfThe7Cs

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/compat/ReiCompat.kt
+++ b/src/main/java/at/hannibal2/skyhanni/compat/ReiCompat.kt
@@ -18,8 +18,18 @@ object ReiCompat {
     @JvmStatic
     fun searchHasFocus(): Boolean {
         if (!isReiLoaded) return false
+        if (Minecraft.getInstance().screen == null) return false
         return try {
-            REIRuntime.getInstance().searchTextField?.isFocused == true
+            val searchBar = REIRuntime.getInstance().searchTextField
+            try {
+                searchBar?.isFocused == true
+            } catch (e: NoSuchMethodError) {
+                try {
+                    searchBar?.javaClass?.getMethod("method_25370")?.invoke(searchBar) as Boolean? == true
+                } catch (e: Throwable) {
+                    false
+                }
+            }
         } catch (e: Throwable) {
             false
         }
--- a/src/main/java/at/hannibal2/skyhanni/compat/ReiCompat.kt
+++ b/src/main/java/at/hannibal2/skyhanni/compat/ReiCompat.kt
@@ -20,16 +20,7 @@ object ReiCompat {
         if (!isReiLoaded) return false
         if (Minecraft.getInstance().screen == null) return false
         return try {
-            val searchBar = REIRuntime.getInstance().searchTextField
-            try {
-                searchBar?.isFocused == true
-            } catch (e: NoSuchMethodError) {
-                try {
-                    searchBar?.javaClass?.getMethod("method_25370")?.invoke(searchBar) as Boolean? == true
-                } catch (e: Throwable) {
-                    false
-                }
-            }
+            (REIRuntime.getInstance().searchTextField as? GuiEventListener)?.isFocused == true
         } catch (e: Throwable) {
             false
         }


src/main/java/at/hannibal2/skyhanni/compat/ReiCompat.kt
```diff
@@ -18,8 +18,18 @@ object ReiCompat {
...
         if (!isReiLoaded) return false
+        if (Minecraft.getInstance().screen == null) return false
         return try {
-            REIRuntime.getInstance().searchTextField?.isFocused == true
+            val searchBar = REIRuntime.getInstance().searchTextField
+            try {
+                searchBar?.isFocused == true
+            } catch (e: NoSuchMethodError) {
+                try {
+                    searchBar?.javaClass?.getMethod("method_25370")?.invoke(searchBar) as Boolean? == true
```
Bloxigus: could this be an accesswidener or accessor?

SidOfThe7Cs: how would i know

NopoTheGamer: i dont believe you can aw other mods

SidOfThe7Cs: the actual method call is from 	net.minecraft.client.gui.components.events.GuiEventListener but idk if that matters

Bloxigus: `method_25370` looks an awful lot like an intermediary mapping, which can be awed

hannibal002 (changes requested):
This is a workaround for a problem that needs to be properly understood first.

Why does `isFocused` throw `NoSuchMethodError` in the first place?
That is the actual root cause, which has not been investigated. This underlying issue should be investigated before adding a workaround.

A cleaner approach might be to cast to `GuiEventListener` instead of using reflection:
```kotlin
(searchBar as? GuiEventListener)?.isFocused == true
```

Have you tested something like this already?

SidOfThe7Cs:
that does seem to just work not sure how i forgot to try that first lol
but for my own understanding, this happenes because rei is compiled with yarn mappings and so *at compile* it inherits the obfuscated method names? but when we cast to the base mc class it uses the local version in official mojang mappings?

hannibal002:
TextField.isFocused() is from REI, not Minecraft code. So as long as they don't obfuscate the code like cheats do, this is not related to many mappings, neither moaning nor yarn. The function is just in their raw plain name during runtime.

hannibal002:
TextField.isFocused() is from REI, not Minecraft code. So as long as they don't obfuscate the code like cheats do, this is not related to many mappings, neither moaning nor yarn. The function is just in their raw plain name during runtime.

hannibal002:
TextField.isFocused() is from REI, not Minecraft code. So as long as they don't obfuscate the code like cheats do, this is not related to many mappings, neither moaning nor yarn. The function is just in their raw plain name during runtime.

hannibal002:
TextField.isFocused() is from REI, not Minecraft code. So as long as they don't obfuscate the code like cheats do, this is not related to many mappings, neither moaning nor yarn. The function is just in their raw plain name during runtime.

hannibal002:
> that does seem to just work not sure how i forgot to try that first lol
> but for my own understanding, this happenes because rei is compiled with yarn mappings and so *at compile* it inherits the obfuscated method names? but when we cast to the base mc class it uses the local version in official mojang mappings?

TextField.isFocused() is from REI, not Minecraft code. So as long as they don't obfuscate the code like cheats do, this is not related to many mappings, neither moaning nor yarn. The function is just in their raw plain name during runtime.

hannibal002:
> that does seem to just work not sure how i forgot to try that first lol
> but for my own understanding, this happenes because rei is compiled with yarn mappings and so *at compile* it inherits the obfuscated method names? but when we cast to the base mc class it uses the local version in official mojang mappings?

TextField.isFocused() is from REI, not Minecraft code. So as long as they don't obfuscate the code like cheats do, this is not related to many mappings, neither moaning nor yarn. The function is just in their raw plain name during runtime.

hannibal002:
Who said using GitHub for Android is a good idea? I'm not gonna delete the other messages now our of spite. Sorry.

ThatGravyBoat:
> > that does seem to just work not sure how i forgot to try that first lol
> > but for my own understanding, this happenes because rei is compiled with yarn mappings and so _at compile_ it inherits the obfuscated method names? but when we cast to the base mc class it uses the local version in official mojang mappings?
> 
> TextField.isFocused() is from REI, not Minecraft code. So as long as they don't obfuscate the code like cheats do, this is not related to many mappings, neither moaning nor yarn. The function is just in their raw plain name during runtime.

This is not the case due to how loom handles reobfuscation of mc code, and due to REI implementing `TextField` on a widget that also implements `GuiEventListener` it automatically in built reobufscated jars changes it from `isFocused` in `TextField` to the `GuiEventListener` method name of `method_25370`. Attached a screenshot of `TextField` decompiled from the latest fabric jar for REI 
image

SidOfThe7Cs:
yea what he said 
        println(REIRuntime.getInstance().searchTextField?.javaClass?.simpleName)
        prints -> OverlaySearchField
        which both implements rei's TextField and extends minecrafts ContainerEventHandler

hannibal002 (approved):
Thanks for the explanation @ThatGravyBoat, that makes sense.

The `GuiEventListener` cast resolves this cleanly. Code looks good to me now.

PR #5227, opened 2026-02-10
Fix: Praying mantis leaderboard

ptlthg:
## What

Praying mantis leaderboard was using the id of "praying mantis" directly in the URL, which was erroring out because there's a space in the URL. The proper url is just `mantis`. 

Also adds support for if I change the backend api at some point to return non-sequential upcoming players which may be necessary if SkyHanni is hammering the API too much. Kinda impossible to test right now, but it seems to work.

A non sequential response might include upcoming players at ranks like: 
`49998, 49997, 49996, 49995, 49994, 49500, 49000, 48500`
Such that there will likely be *some* player in the request immediately to use rather than loop with requests, we'll see if it's needed though.

<details>
<summary>Images</summary>

If the upcoming player isn't the next one a grayed out rank shows here:
image

</details>

## Changelog Fixes
+ Fixed praying mantis leaderboard erroring out. - Ke5o

+ Added support for non-sequential upcoming player lists in leaderboards. - Ke5o

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/leaderboarddisplays/EliteLeaderboardDisplayBase.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/leaderboarddisplays/EliteLeaderboardDisplayBase.kt
@@ -40,10 +40,11 @@ import kotlin.time.Duration.Companion.seconds
 abstract class EliteLeaderboardDisplayBase<E : Enum<E>, T : EliteLeaderboardType.WithEnum<E>>(
     private val typeClass: KClass<out T>,
     private val createType: (E, EliteLeaderboardMode) -> EliteLeaderboardType,
-    private val name: String
+    private val name: String,
 ) {
     protected val configBase get() = GardenApi.config.eliteFarmersLeaderboards
     private val config get() = baseClass?.let { getConfigFromClass(it) }
+
     @Suppress("Unchecked_cast")
     private val baseClass: KClass<out EliteLeaderboardType>?
         get() = typeClass as? KClass<out EliteLeaderboardType>
@@ -164,9 +165,9 @@ abstract class EliteLeaderboardDisplayBase<E : Enum<E>, T : EliteLeaderboardType
         val currentRank = leaderboardPos
         val shouldShowRank = playerRank != null && currentRank != null && (
             useRankGoal || // Always show when using rank goal
-            (!getLastPlayer && playerRank != currentRank - 1) || // Show for next player if non-sequential
-            (getLastPlayer && playerRank != currentRank + 1) // Show for last player if non-sequential
-        )
+                (!getLastPlayer && playerRank != currentRank - 1) || // Show for next player if non-sequential
+                (getLastPlayer && playerRank != currentRank + 1) // Show for last player if non-sequential
+            )
 
         if (useRankGoal && rankGoal != null) {
             nextName += " §7[§b#${rankGoal.addSeparators()}§7]"
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/leaderboarddisplays/EliteLeaderboardDisplayBase.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/leaderboarddisplays/EliteLeaderboardDisplayBase.kt
@@ -82,13 +82,11 @@ abstract class EliteLeaderboardDisplayBase<E : Enum<E>, T : EliteLeaderboardType
     open val currentLeaderboardType: EliteLeaderboardType?
         get() {
             val enum = currentEnum ?: run {
-                // Apply debounce when auto-switching (currentEnum is null)
+                // Debounce leaderboard when auto switching
                 val now = SimpleTimeMark.now()
                 if (lastEnumSwitchTime.passedSince() < 5.seconds) {
-                    // Use cached value during debounce period
                     lastAutoSelectedEnum
                 } else {
-                    // Update cache after debounce period
                     val newDefault = getDefaultEnum()
                     if (newDefault != lastAutoSelectedEnum) {
                         lastEnumSwitchTime = now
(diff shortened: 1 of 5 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/leaderboarddisplays/EliteLeaderboardDisplayBase.kt
```diff
@@ -133,24 +153,35 @@ abstract class EliteLeaderboardDisplayBase<E : Enum<E>, T : EliteLeaderboardType
...
+        var nextName = playerName
+
+        val currentRank = leaderboardPos
+        val shouldShowRank = playerRank != null && currentRank != null && (
+            useRankGoal || // Always show when using rank goal
+                (!getLastPlayer && playerRank != currentRank - 1) || // Show for next player if non-sequential
+                (getLastPlayer && playerRank != currentRank + 1) // Show for last player if non-sequential
+            )
+
+        if (useRankGoal && rankGoal != null) {
```
hannibal002: `rankGoal != null` is always true

src/main/java/at/hannibal2/skyhanni/features/garden/leaderboarddisplays/EliteLeaderboardDisplayBase.kt
```diff
@@ -133,24 +153,35 @@ abstract class EliteLeaderboardDisplayBase<E : Enum<E>, T : EliteLeaderboardType
...
+            nextName += " §7[§b#${rankGoal.addSeparators()}§7]"
+        } else if (shouldShowRank) {
+            nextName += " §7[#${playerRank?.addSeparators()}]"
```
hannibal002: `?` can go

src/main/java/at/hannibal2/skyhanni/features/garden/leaderboarddisplays/EliteLeaderboardDisplayBase.kt
```diff
@@ -133,24 +153,35 @@ abstract class EliteLeaderboardDisplayBase<E : Enum<E>, T : EliteLeaderboardType
...
+        val playerInfo: LeaderboardPlayerInfo? = if (getLastPlayer) getLastPlayer(leaderboardType) else getNextPlayer(leaderboardType)
 
         val rankGoal = getRankGoal(leaderboardType)
         val useRankGoal = useEtaGoalRank() && rankGoal != null
         if (useRankGoal && getLastPlayer) return Renderable.empty()
 
-        var (nextName, amountUntil) = next ?: return nullNextPlayerRenderable(leaderboardType)
-        if (useRankGoal) {
-            nextName += " §7[§b#${rankGoal?.addSeparators()}§7]"
+        val (playerName, amountUntil, playerRank) = playerInfo ?: return nullNextPlayerRenderable(leaderboardType)
```
hannibal002, line 161: different name: amountUntil here vs amountDifference in LeaderboardPlayerInfo. can we make this consistent please?

src/main/java/at/hannibal2/skyhanni/features/garden/leaderboarddisplays/EliteLeaderboardDisplayBase.kt
```diff
@@ -77,7 +81,23 @@ abstract class EliteLeaderboardDisplayBase<E : Enum<E>, T : EliteLeaderboardType
     }
 
     open val currentLeaderboardType: EliteLeaderboardType?
```
hannibal002: this value access feels like it should be just a function instead.

hannibal002 (changes requested):
im a bit lost, is the fix just on your side? if so, please mention this in the pr desc somewhere

ptlthg:
> im a bit lost, is the fix just on your side? if so, please mention this in the pr desc somewhere

The praying mantis leaderboard fix is in this PR, it's just one line that sets the leaderboard name.

The other code is just so that I could in the future return ranks from my API that aren't sequential if I need to for performance reasons. 

(Sorry for mixing them, I do not like writing code in Kotlin)

hannibal002:
> > im a bit lost, is the fix just on your side? if so, please mention this in the pr desc somewhere
> 
> The praying mantis leaderboard fix is in this PR, it's just one line that sets the leaderboard name.
> 
> The other code is just so that I could in the future return ranks from my API that aren't sequential if I need to for performance reasons. 
> 
> (Sorry for mixing them, I do not like writing code in Kotlin)

All fine, thanks for explaining it further!

src/main/java/at/hannibal2/skyhanni/features/garden/leaderboarddisplays/EliteLeaderboardDisplayBase.kt
ptlthg: I blame chissl

Chissl: I agree

NopoTheGamer:
where is the mantis fix??

ptlthg:
> where is the mantis fix??

I am actually so confused, that was the first thing I did on this branch, but apparently not? Pushed that now.

NopoTheGamer:
it went away in the 2nd commit it looks like, its why hannibal was so confused

ptlthg:
> it went away in the 2nd commit it looks like, its why hannibal was so confused

I thought hannibal was just blind, my bad

ptlthg:
image
Can confirm the fix works (I just don't have enough mantis kills to be ranked)

hannibal002:
> it went away in the 2nd commit it looks like, its why hannibal was so confused 

Thanks of providing a valid explanation for one of the many moments where I get confused by code. The agreement here helps waay more than it should.

PR #5229, opened 2026-02-10
Improvement: Pet Display V2

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/5262
- https://github.com/hannibal002/SkyHanni/pull/5264
-  https://github.com/hannibal002/SkyHanni/pull/5265
- https://github.com/hannibal002/SkyHanni/pull/5274
- https://github.com/NotEnoughUpdates/MoulConfig/pull/70

## What
See #4083.
Completely overhauls the Current Pet Display, and makes it prettier and more customizable, combining the best parts of the other mods' pet displays into one solution.

## Changelog Improvements
+ Completely overhauled current pet display. - Daveed.
    * Pet icon display ("SBE-style"), supports animated skins & spinning.
    * Pet text display ("NEU-style"), can be placed abstractly next to/above/below the icon.
    * Each aspect of the pet display is toggleable and very configurable.


lunaynx:
@DavidArthurCole I'll comment here since I'm restricted on Discord and can only communicate via reactions.

https://discord.com/channels/997079228510117908/1094190239532208228/1473085583365640223

You're abusing extension functions in an unintended way that is *not* extending the event class, simply to not have to use `event.` on each call. I think it is unnecessary to change it at all, but if you really wanted to, the proper way is to use `event.apply { ... }` (however, that does introduce an additional level of indentation).

You are bringing up examples of "the same pattern" across the code, except those are legitimate extension functions and not abusing a language feature, so this argument is disingenous. If there *are* any actual abuses of extension functions in our code similar to what you're doing, I would say they should be eliminated.

DavidArthurCole:
image

Soon...

DavidArthurCole:
> You are bringing up examples of "the same pattern" across the code, except those are legitimate extension functions and not abusing a language feature, so this argument is disingenous. If there *are* any actual abuses of extension functions in our code similar to what you're doing, I would say they should be eliminated.


Please look at what my suggested Kotlin sample would compile to in the JVM and let me know how it's abusing a language feature.

DavidArthurCole:
> Please look at what my suggested Kotlin sample would compile to in the JVM and let me know how it's abusing a language feature.

(spoiler alert because I know that was a rather generic statement), every single one of these:

```
// HandleEvent class specifier
@HandleEvent(EventClass::class)
fun onEvent()

// Using primary function
@HandleEvent
fun primaryFunctionNme()

// Specifying parameter type
@HandleEvent
fun onEvent(event: EventClass)

// Extension function
@HandleEvent
private fun EventClass.onEvent()
```

Compiles to the same thing in the JVM - a virtual method which takes the event type as a parameter, which is what event handlers are.

If you're telling me that I'm misusing a feature of the language, I think I would counter that you're underutilizing it? If it compiles to the same code, what's the difference? If there's a shorter form way to write something that changes logically nothing about how the code runs, why not use it?

And one more time, I would like to ask anyone to explain the complexion in what it is I'm doing, that's causing so much apprehension here..? The Event is still being specified by name, directly as a part of the method. The functions would be private, they wouldn't show up in auto-complete. And generally, I cannot imagine anyone looking into the codebase without also being able to make the connection there, when there's a `@HandleEvent` annotation?

NopoTheGamer:
really

DavidArthurCole:
> really

thanks, nopo, you're as helpful as ever! I really appreciate the positive attitude you bring when I'm trying to have a conversation.

NopoTheGamer:
ok well for your little example you are wrong, 2 of those compile to one thing, the 3rd compiles to something else and the 4th doesnt compile. anyway that doesnt matter as no one was making a point that it "compiled to the jvm differently". The point is doing this doesnt achieve anything you couldnt do otherwise. Yes it does show in autocomplete, an argument i didnt even think of before so thanks for that. and yes last time you put one of these in the code i did get confused by it.

overall: remove it

DavidArthurCole:
so are you just going to ignore the entire part of the conversation that happened in contrib-chat about me planning to address private event handlers, or do you just prefer to vilify me?

> Yes it does show in autocomplete

No. It wouldn't.

> i did get confused by it.

Skill. Issue. Are you actually trying to tell me that these two:
image
image

Are so _wildly_ different, that you were perplexed enough to not know what was happening? Because, nopo, I do not believe that for a fucking second.

DavidArthurCole:
image

In other news.
+/- 2 lines for 1.21.11.

DavidArthurCole:
> so are you just going to ignore the entire part of the conversation that happened in contrib-chat about me planning to address private event handlers, or do you just prefer to vilify me?

@NopoTheGamer I'm waiting... This is a genuine question. Since that seems to be one of your main points of contention, I'm interested to know what your holdbacks are, still.

DavidArthurCole:
Schrödinger's pet display. This is either ready for review, and everything will work, or there will be one billion bugs.

DavidArthurCole:
This is pending a moul update to fix config option order - pulling back to draft until that's done.

PR #5236, opened 2026-02-12
Fix: Remove canBeSeen check from Fishing Hook Display

lunaynx:
## What
Just removing it for now because it's breaking legitimate use cases where the original armor stand is actually clearly visible to the player due to how visible modern vanilla clients make armor stand name tags through walls, and it likely didn't address the original macro check issue it was added in response to anyway. Kept the `isCustomNameVisible` check because that should be harmless, at worst it does nothing to help.

https://discord.com/channels/997079228510117908/1117409804382642296/1471338356972720260

## Changelog Fixes
+ Fixed a recent change causing Fishing Hook Display to sometimes not show when it should. - Luna


hannibal002 (commented):
the changelog is too long/complex
how about `Fixed Fishing Hook Display sometimes not appearing. - Luna` ?

also , the second line belogns in the what block, not in the user facing changelog imo

hannibal002 (changes requested):

lunaynx:
Yeah true, I reworded the whole thing now.

PR #5237, opened 2026-02-12
Backend: fix player death event

NopoTheGamer:
## What
50% of the uses of the event werent working

+ Fixed player death event not firing on yourself. - nopo


hannibal002 (approved):
lgtm

PR #5240, opened 2026-02-12
Improvement: Move Diana warps to repo

NopoTheGamer:
## What
i heard they might make a new hub so i added this just in case

## Changelog Improvements
+ Added support for changing diana warps via the repo. - nopo


hannibal002 (changes requested):
your_name_here

:P

PR #5232, opened 2026-02-11
Improvement: Different color for non-maxed Ultimate Enchants

Germanminer:
## What

Adds an option to use a different color for non-maxed Ultimate enchants.

Also adds an advanced override color for maxed Ultimate enchants.

Makes it easier to tell at a glance whether an Ultimate enchant is maxed or not.

## Screenshots
image

## Changelog Improvements
+ Cleaned up Ultimate enchant style logic to support separate maxed/non-maxed colors. - germanminer

PR #5241, opened 2026-02-12
Fix + Improvement: Jacob Contest Warning Features

lunaynx:
## What
Updated Jacob Contest warning features:
- Warning sending no longer depends on the display, and now has a separate "Warn Outside Garden" option (previously tied to "Show Outside Garden")
- Switched from JOptionPane (does not work on modern Minecraft due to it running with java.awt.headless=true) to TinyFileDialogs (LWJGL, bundled with Minecraft)

## Changelog Improvements
+ Jacob Contest warnings can now be sent even if the display is disabled. - Luna
+ Added a separate "Warn Outside Garden" option for Jacob Contests. - Luna
    * This means it is now possible to show the upcoming contest everywhere but only warn while on the Garden.

## Changelog Fixes
+ Fixed Jacob Contest popup warning not working. - Luna


src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
```diff
@@ -215,8 +217,10 @@ object GardenNextJacobContest {
 
     @HandleEvent(SecondPassedEvent::class)
     fun onSecondPassed() {
-        if (!isEnabled() || calendarDetector.isInside()) return
-        update()
+        if (isEnabled() && !calendarDetector.isInside()) {
+            update()
+        }
+        nextContest?.warnAbout()
```
hannibal002: this now also runs outside sb. the config option desc for `warnOutsideGarden` does only mention "outside garden". either add missing outside sb check, or make in the config option desc clear that this doesnt meant "inside sb", but always while the game is open.

showOutsideGarden for example mentions "everywhere in skyblcok" and has a "in skyblcok" check.

please add consistency to the desc format as well

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
```diff
@@ -539,5 +541,21 @@ object GardenNextJacobContest {
                 add(JsonPrimitive(CropType.WILD_ROSE.name))
             }
         }
+
+        event.add(124, "$base.warnOutsideGarden") {
+            JsonPrimitive(config.showOutsideGarden)
```
hannibal002: this does not work. i have tested the pr.

i see your intion as to copy the existing other toggle's value here. sadly, this cant work as you wrote it since during the config fix event the config is not yet filled with data.

easiest solution is to just leave warnOutsideGarden default disabled just like showOutsideGarden. then also mention in the changelog that this new option is default enabled/separate from the other one.

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
```diff
@@ -430,27 +433,26 @@ object GardenNextJacobContest {
         SoundUtils.playBeepSound()
 
         val cropTextNoColor = crops.joinToString(", ") {
-            if (it == boostedCrop) "<b>${it.cropName}</b>" else it.cropName
+            if (it == boostedCrop) "*${it.cropName}*" else it.cropName
```
hannibal002: `'` does not get formatted for me. see:

Image

```kt
DialogUtils.openPopupWindow(
    title = "SkyHanni Test",
    message = "Normal text and *starred text* here",
)
```

if this is on windows different, please add a check in DialogUtils that removes stars in the string when on linux, or remove it entirely if its also not working on windows.

src/main/java/at/hannibal2/skyhanni/utils/DialogUtils.kt
```diff
@@ -1,46 +1,26 @@
...
-import javax.swing.JOptionPane
-import javax.swing.UIManager
+import org.lwjgl.util.tinyfd.TinyFileDialogs
 
 object DialogUtils {
 
-    private val closeListener = object : MouseAdapter() {
-        override fun mouseClicked(event: MouseEvent) {
-            (event.source as? JFrame)?.isVisible = false
+    fun openPopupWindow(title: String, message: String) = SkyHanniMod.launchCoroutine("openPopupWindow $title") {
```
hannibal002: please give the launchCoroutine function the parameter function with infinite. so that the coroutine knows that this thread can run indefinetly. whilte this does not cause an error, its more clear for future contributors whats the idea here.

hannibal002 (changes requested):
code looks good, just small suggestions.

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
hannibal002: good news! in

- #5232 

we add a new duplciate funtion, that is exaclty what you want to use here.
see [this commit](https://github.com/hannibal002/SkyHanni/pull/5232/changes/09607d25f64a3862c16a1fc6378b33074973a849)

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
lunaynx: Yeah, it was not meant to actually format it, I don't think TinyFileDialogs supports that. It was meant to be visual emphasis. But on second thought, I'll probably just add a separate line like "Boosted crop: Wheat" to make it more clear.

src/main/java/at/hannibal2/skyhanni/config/features/garden/NextJacobContestConfig.kt
```diff
@@ -87,6 +87,14 @@ class NextJacobContestConfig {
     @ConfigEditorSlider(minValue = 10f, maxValue = 300f, minStep = 1f)
     var warnTime: Int = 120
 
+    @Expose
+    @ConfigOption(
+        name = "Warn Outside Garden",
+        desc = "Warn about upcoming contests not only in the Garden but everywhere in SkyBlock.",
+    )
+    @ConfigEditorBoolean
+    var warnOutsideGarden: Boolean = false
```
hannibal002, line 96: this is unused now.

hannibal002 (changes requested):
`warnOutsideGarden` is never actually checked now.

hannibal002:
@lunaynx this pr has the fix tag, i assume this is more important than other prs. please look at the pr again if possible!

hannibal002 (changes requested):
bump

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
```diff
@@ -429,36 +431,41 @@ object GardenNextJacobContest {
...
         val display = display ?: return
         config.inventoryPosition.renderRenderable(display, posLabel = "Load SkyBlock Calendar")
     }
 
     private fun sbEnabled() = SkyBlockUtils.inSkyBlock && (GardenApi.inGarden() || config.showOutsideGarden)
     private fun outsideSbEnabled() = OutsideSBFeature.NEXT_JACOB_CONTEST.isSelected() && !SkyBlockUtils.inSkyBlock
-    private fun isEnabled() = config.display && (sbEnabled() || outsideSbEnabled())
+    private fun inCorrectArea() = sbEnabled() || outsideSbEnabled()
+    private fun isDisplayEnabled() = inCorrectArea() && config.display
+    private fun isWarningEnabled() = inCorrectArea() && config.warn && (GardenApi.inGarden() || config.warnOutsideGarden)
```
hannibal002: warnOutsideGarden will be ignored if showOutsideGarden is false

src/main/java/at/hannibal2/skyhanni/config/features/dev/DebugConfig.kt
```diff
@@ -226,6 +226,11 @@ class DebugConfig {
     @ConfigEditorBoolean
     val forceGreatSpook: Property<Boolean> = Property.of(false)
 
+    @Expose
+    @ConfigOption(name = "Always Unfocused", desc = "Always act as if the Minecraft window is unfocused.")
```
hannibal002, line 220: how about renaming to "forced unfocused"?

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
```diff
@@ -215,8 +217,10 @@ object GardenNextJacobContest {
 
     @HandleEvent(SecondPassedEvent::class)
     fun onSecondPassed() {
-        if (!isEnabled() || calendarDetector.isInside()) return
-        update()
+        if (isDisplayEnabled() && !calendarDetector.isInside()) {
+            update()
```
hannibal002: fetchContestsIfAble is only getting called when isDisplayEnabled is true. this means, the "Warning Time" feature (that is now independent of the main toggle) will not work at all.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
lunaynx: Fixed.

src/main/java/at/hannibal2/skyhanni/config/features/dev/DebugConfig.kt
lunaynx: Not sure, I like the current name better. Forcing sounds like it's actually making the window unfocused, and "always X" is a common pattern in config alreayd.

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
lunaynx: I think this is now fixed by the fix to the above.

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
```diff
@@ -541,5 +551,24 @@ object GardenNextJacobContest {
                 add(JsonPrimitive(CropType.WILD_ROSE.name))
             }
         }
+
+        event.duplicate(
+            126,
```
hannibal002: wrong version

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
hannibal002: no, fetchContestsIfAble is still not getting called if display is disabled. here in the on second passed event it should also get called if the condition here is false. that would fix it

src/main/java/at/hannibal2/skyhanni/utils/DialogUtils.kt
```diff
@@ -1,46 +1,28 @@
...
+import org.lwjgl.util.tinyfd.TinyFileDialogs
+import kotlin.time.Duration
 
 object DialogUtils {
 
-    private val closeListener = object : MouseAdapter() {
-        override fun mouseClicked(event: MouseEvent) {
-            (event.source as? JFrame)?.isVisible = false
+    fun openPopupWindow(title: String, message: String) =
+        SkyHanniMod.launchCoroutine("openPopupWindow $title", timeout = Duration.INFINITE) {
```
hannibal002: launchCoroutine is deprecated

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
```diff
@@ -431,38 +432,47 @@ object GardenNextJacobContest {
...
+    fun onChestGuiRender() {
         if (!config.display || !calendarDetector.isInside()) return
         val display = display ?: return
         config.inventoryPosition.renderRenderable(display, posLabel = "Load SkyBlock Calendar")
     }
 
-    private fun sbEnabled() = SkyBlockUtils.inSkyBlock && (GardenApi.inGarden() || config.showOutsideGarden)
-    private fun outsideSbEnabled() = OutsideSBFeature.NEXT_JACOB_CONTEST.isSelected() && !SkyBlockUtils.inSkyBlock
-    private fun isEnabled() = config.display && (sbEnabled() || outsideSbEnabled())
+    private fun sbEnabled() = SkyBlockUtils.inSkyBlock &&
```
hannibal002: please add docs on what this function now does. warn outside garden here feels wrong. this would mean the display shows outside garden even if the user just wants the warnings to show outside

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
lunaynx: I've refactored a bit, hopefully it's more self-explanatory and less confusing names now.

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
```diff
@@ -431,50 +432,72 @@ object GardenNextJacobContest {
...
-    private fun outsideSbEnabled() = OutsideSBFeature.NEXT_JACOB_CONTEST.isSelected() && !SkyBlockUtils.inSkyBlock
-    private fun isEnabled() = config.display && (sbEnabled() || outsideSbEnabled())
-    private fun isFetchEnabled() = isEnabled() && config.fetchAutomatically
-    private fun isSendEnabled() = isFetchEnabled() && config.shareAutomatically != ShareContestsEntry.DISABLED
+    // <editor-fold="Conditionals">
+    private fun inRelevantArea(): Boolean {
+        if (!SkyBlockUtils.inSkyBlock &&
+            !OutsideSBFeature.NEXT_JACOB_CONTEST.isSelected()
+        ) return false
+        return GardenApi.inGarden() || (config.showOutsideGarden || config.warnOutsideGarden)
```
hannibal002, line 469: this will break the logic, still. we cant combine show and warn checks that way, otherwise selecting one of them toggles for both. needs clear separation as it was prior to the pr

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
```diff
@@ -431,50 +432,72 @@ object GardenNextJacobContest {
...
         val display = display ?: return
         config.inventoryPosition.renderRenderable(display, posLabel = "Load SkyBlock Calendar")
     }
 
-    private fun sbEnabled() = SkyBlockUtils.inSkyBlock && (GardenApi.inGarden() || config.showOutsideGarden)
-    private fun outsideSbEnabled() = OutsideSBFeature.NEXT_JACOB_CONTEST.isSelected() && !SkyBlockUtils.inSkyBlock
-    private fun isEnabled() = config.display && (sbEnabled() || outsideSbEnabled())
-    private fun isFetchEnabled() = isEnabled() && config.fetchAutomatically
-    private fun isSendEnabled() = isFetchEnabled() && config.shareAutomatically != ShareContestsEntry.DISABLED
+    // <editor-fold="Conditionals">
```
hannibal002, line 464: the proper formatting is `<editor-fold desc=`, then, while folded, the text gets properly shown as "Conditionals", otherwise, this one line gets shown raw.

hannibal002 (changes requested):

lunaynx:
Will reopen this as smaller separate PRs to avoid holding up the main bugfix over the other stuff.

PR #5243, opened 2026-02-13
Backend: Sealed Events

ItsEmpa:
## What
Changes many events to use sealed classes instead of open. Sealed classes are generally preferred over open classes because they provide controlled inheritance of your class hierarchies, and all of their subclasses are known at compile time

+ Changed some events to be sealed instead of open classes. - Empa


src/main/java/at/hannibal2/skyhanni/events/SkyHanniRenderEntityEvent.kt
```diff
@@ -10,7 +10,7 @@ import net.minecraft.world.entity.LivingEntity
  * For normal cases of "hide this entity" rather use [CheckRenderEntityEvent].
  */
 @Deprecated("use CheckRenderEntityEvent instead")
-open class SkyHanniRenderEntityEvent<T : LivingEntity>(
```
hannibal002, line 13: unnecessary change here imo, since the whole file is due to get deleted anyway

hannibal002 (changes requested):
the "What" section in the pr desc could include why sealed is preferable here. Helps other devs looking at this pr to understand the motivation at a glance

src/main/java/at/hannibal2/skyhanni/events/SkyHanniRenderEntityEvent.kt
ItsEmpa: i dont wanna have to open intellij again to change this back 😭

hannibal002 (approved):

PR #5245, opened 2026-02-13
Fix: Mute Banshee sound name

lunaynx:
## What
Fixed the wrong sound name being used for Mute Banshee, causing it to not work. (Hypixel may have changed it at some point, since it used to be the wrong sound on modern, though I could have sworn it was changed *before* the feature was added.

## Changelog Fixes
+ Fixed Mute Banshee not working. - Luna


hannibal002:
A slightly bit of outside the scope of this pr, but we should handle sounds the same way we handle skins. In the repo, with one unique identifier, that gets backed via the repo by storing a sound name, a sound pitch and a sound amplitude (or whatever the third parameter we used to identify sounds was named).

PR #5246, opened 2026-02-14
Fix: /viewrecipe autocomplete missing items

lunaynx:
## What
It was only checking for the `recipe` key, but several items use `recipes` now, even single-recipe ones, since Firmament export uses that format.

## Changelog Fixes
+ Fixed `/viewrecipe` autocomplete not showing some items. - Luna


src/main/java/at/hannibal2/skyhanni/features/commands/ViewRecipeCommand.kt
```diff
@@ -56,15 +58,21 @@ object ViewRecipeCommand {
         HypixelCommands.viewRecipe(item.toInternalName(), page)
     }
 
-    val list by lazy {
-        val list = mutableListOf<String>()
-        for ((key, value) in NeuItems.allNeuRepoItems()) {
-            if (value.has("recipe")) {
-                list.add(key.lowercase())
+    val list: List<String> by lazy {
```
NopoTheGamer: this list should probably be recreated if the repo is reloaded?

lunaynx: My attempt at this broke the completion entirely, will try more tomorrow.

lunaynx:
What happened here...

src/main/java/at/hannibal2/skyhanni/features/commands/ViewRecipeCommand.kt
lunaynx: Fixed it. Turns out my attempt was fine, I just had the option disabled for some reason.

PR #5258, opened 2026-02-18
Fix: Death messages with Mark Own Name

lunaynx:
## What
Fixed own death messages being changed from "You" to the player name if Mark Own Name is enabled or the player marked themselves to /shmarkplayer, sometimes also leading to grammatically incorrect messages such as "mia_uw were killed by Empyrean Harpy."

Includes some cleanup of relevant files as well, and adding separate sub-events of PlayerDeathEvent for Allow and Modify, with the base event being there for consumers that do not care about the chat event.

## Changelog Fixes
+ Fixed own death messages being incorrectly modified with Mark Own Name enabled. - Luna

+ Split `PlayerDeathEvent` into `PlayerDeathEvent.Allow` and `PlayerDeathEvent.Modify`. - Luna
+ Added `isSelf` convenience property to `PlayerDeathEvent`. - Luna
+ Made a small optimization to `PlayerUtils.getName` to avoid a round-trip from
  `String` to `Component` and back to `String`. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/MiningApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/MiningApi.kt
@@ -281,7 +281,7 @@ object MiningApi {
 
     @HandleEvent
     fun onPlayerDeath(event: PlayerDeathEvent) {
-        if (event.name == PlayerUtils.getName()) {
+        if (event.isSelf) {
             updateCold(0)
             updateHeat(0)
             lastColdReset = SimpleTimeMark.now()
--- a/src/main/java/at/hannibal2/skyhanni/data/MiningApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/MiningApi.kt
@@ -279,7 +279,7 @@ object MiningApi {
     }
 
     @HandleEvent
-    fun onPlayerDeath(event: PlayerDeathEvent) {
+    fun onPlayerDeath(event: PlayerDeathEvent.Allow) {
         if (event.isSelf) {
             updateCold(0)
             updateHeat(0)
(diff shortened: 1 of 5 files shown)


src/main/java/at/hannibal2/skyhanni/data/MiningApi.kt
```diff
@@ -281,7 +280,7 @@ object MiningApi {
 
     @HandleEvent
     fun onPlayerDeath(event: PlayerDeathEvent) {
```
hannibal002: this listener gets called three times. specify with allowed or modify.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/MiningApi.kt
lunaynx: Does it? The intent was to provide a base class for cases where the consumer doesn't care about the chat event, but if it doesn't work as I expected I can probably either create a `PlayerDeathEvent.Base` or just replace those use cases with `PlayerDeathEvent.Allow`.

hannibal002: im not entirely sure, tho i assumed so.

lunaynx: I'll test it in a bit.

NopoTheGamer: allow should be used in every case where you arent modifying the message

lunaynx: Yeah ok, I just did that.

hannibal002 (approved):

PR #5262, opened 2026-02-21
Backend: Modern Custom Item Rendering

DavidArthurCole:
## What
With lots of help from Vixid, this PR gets the custom (better quality, spinning, moving) item rendering working for both 1.21.10, and 1.21.11. We're now using PIP and state submission, combined with a texture atlas, to improve quality, while still remaining performant.

Right now, the criteria for using the custom item rendering is any item with either:
- Scale > 1 & uses block lighting
- Any rotation

This branch was cherry-picked from #5229, so if anything looks out of place, let me know.

+ Added higher-quality item rendering. - Daveed + Vixid
+ Fixed `AnimatedItemStackRenderable` not being animated. - Daveed + Vixid
+ Refactored TestRenderableSuite to optionally take an event type. - Daveed
    * Needed as our new item renderer needs to run on a specific render event.
    * If type is not specified, GuiRenderEvent.GuiOnTopRenderEvent is inferred.
+ Added SkyHanni Picture-in-Picture render handler. - Daveed + Vixid
+ Added `renderState` and `bufferSource` to the Mixin for `GuiRenderer`. - Daveed + Vixid
    * Allows for direct interfacing with render pipelines.


src/main/java/at/hannibal2/skyhanni/utils/render/item/atlas/SkyHanniItemAtlasKey.kt
```diff
@@ -0,0 +1,51 @@
...
+    override fun equals(other: Any?): Boolean =
+        if (other !is SkyHanniAtlasKey) false
+        else if (this === other) true
+        else modelIdentity == other.modelIdentity &&
+            quantizedRotationVector == other.quantizedRotationVector &&
+            scale == other.scale &&
+            guiScale == other.guiScale &&
+            stableId == other.stableId
+
+    override fun hashCode(): Int = Objects.hash(modelIdentity, quantizedRotationVector, scale, guiScale)
```
hannibal002, line 39: this does not check for stableId, while equals does. intended?

src/main/java/at/hannibal2/skyhanni/utils/render/item/atlas/SkyHanniItemAtlasKey.kt
```diff
@@ -0,0 +1,51 @@
...
+    val frameNumber: Int,
+) : SkyHanniAtlasKey(modelIdentity, scale, guiScale, stableId, rotationVector) {
+    override val rotationSnapDegrees: Float = 0.125f
+
+    override fun equals(other: Any?): Boolean =
+        if (this === other) true
+        else if (other !is SkyHanniAnimatedAtlasKey) false
+        else super.equals(other) && frameNumber == other.frameNumber
+
+    override fun hashCode(): Int = Objects.hash(super.hashCode(), stableId)
```
hannibal002: should also check for frameNumber

src/main/java/at/hannibal2/skyhanni/utils/render/item/atlas/SkyHanniItemAtlasKey.kt
```diff
@@ -0,0 +1,51 @@
...
+    override val modelIdentity: Any,
+    override val scale: Float,
+    override val guiScale: Int,
+    override val stableId: Int,
+    override val rotationVector: Vec3,
+    val frameNumber: Int,
+) : SkyHanniAtlasKey(modelIdentity, scale, guiScale, stableId, rotationVector) {
+    override val rotationSnapDegrees: Float = 0.125f
+
+    override fun equals(other: Any?): Boolean =
```
hannibal002, line 56: should also check for stableId

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniItemRenderCoordinator.kt
```diff
@@ -0,0 +1,83 @@
...
+    private val projectionBuffer = CachedOrthoProjectionMatrixBuffer("SkyHanni items", -1000.0f, 1000.0f, true)
+    private val settleTracker = HashMap<SkyHanniAtlasKey, SettleEntry>()
+    private val atlas = SkyHanniItemAtlas()
+
+    fun invalidateAtlas() {
+        atlas.invalidate()
+        settleTracker.clear()
+    }
+
+    fun closeAtlas() {
```
hannibal002, line 28: unused. should this not get called when e.g. world change or something?

src/main/java/at/hannibal2/skyhanni/utils/compat/DrawContextUtils.kt
```diff
@@ -78,6 +78,16 @@ object DrawContextUtils {
         popMatrix()
     }
 
+    /**
+     * Push and pop the matrix stack, running the action in between, and returning the result of the action.
+     */
+    @Suppress("DEPRECATION")
+    inline fun <T> pushPopResult(action: () -> T): T {
```
hannibal002, line 88: add error handling. if action throws an erorr, pop matrix will never be called.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinGuiRenderer.java
```diff
@@ -51,4 +65,53 @@ public RenderPipeline replacePipeline(GuiElementRenderState state, Operation<Ren
...
+    // Inject at the tail of preparePictureInPictureState, after Fabric's
+    // postPrepareSpecialElements has run. At this point all PIP states have
+    // been submitted, our collector has gathered them, and renderState is
+    // ready to accept new BlitRenderState submissions before sortElements runs.
+    @Inject(
+        method = "preparePictureInPictureState",
+        at = @At("TAIL")
+    )
+    private void skyhanni$prepareSkyHanniItems(CallbackInfo ci) {
+        PictureInPictureRenderer<?> renderer = pictureInPictureRenderers.get(SkyHanniGuiItemRenderState.class);
```
hannibal002: pleas move this code into a hook/kotlin context.

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniItemRenderCoordinator.kt
```diff
@@ -0,0 +1,83 @@
...
+internal object SkyHanniItemRenderCoordinator {
+
+    // items actively spinning re-render every frame, same as mojang's isAnimated path.
+    // items that have been stable for this many frames are committed to the atlas.
+    private const val SETTLE_FRAMES = 4
+    private val projectionBuffer = CachedOrthoProjectionMatrixBuffer("SkyHanni items", -1000.0f, 1000.0f, true)
+    private val settleTracker = HashMap<SkyHanniAtlasKey, SettleEntry>()
+    private val atlas = SkyHanniItemAtlas()
+
+    fun invalidateAtlas() {
```
hannibal002, line 23: this only gets called on resource pack reload. shouldn't we try to keep the maps of data inside small over the whole session, like deleting entries that aren't used for a long time, or something similar? or is this here truly irrelevant as we only have a small list of entries at any time?

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniItemRenderCoordinator.kt
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        // Only set up atlas if we have animated items
+        if (animatedStates.isNotEmpty()) SkyHanniItemRenderContext(
+            animatedStates, staticFallbackStates, guiRenderState,
+            bufferSource, featureRenderDispatcher,
+            frameNumber, guiScale,
+        ).setupRendering()
+
+        val fallbackRenderer = SkyHanniItemRenderer(bufferSource)
```
hannibal002: we create two SkyHanniItemRenderer instances per prepare call. is this intended? if so, please add a small comment somewhere here to explain why.

src/main/java/at/hannibal2/skyhanni/utils/render/item/atlas/SkyHanniItemAtlas.kt
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    private fun checkFull(): Boolean = isFull().also { if (it) needsGrowing = true }
+    private fun checkRowFull(): Boolean = isRowFull().also { if (it) newRow() }
+
+    private fun isFull() = cursorY + slotSize > sizePixels ||
+        (cursorX + slotSize > sizePixels && cursorY + slotSize * 2 > sizePixels)
+    private fun isRowFull() = cursorX + slotSize > sizePixels
+
+    private fun recordPosition(key: SkyHanniAtlasKey, slotX: Int, slotY: Int): SkyHanniItemAtlasEntry? {
```
hannibal002: this function does not need to return nullable. we always have the object in the map.

src/main/java/at/hannibal2/skyhanni/utils/renderables/Renderable.kt
```diff
@@ -394,21 +394,30 @@ interface Renderable {
             bottomLayer: Renderable,
             topLayer: Renderable,
             blockBottomHover: Boolean = true,
+            forceBottomRenderFirst: Boolean = false,
```
hannibal002, line 409: please explain in this function docs when this forceBottomRenderFirst value can be true/explain why this is necessary.

src/main/java/at/hannibal2/skyhanni/utils/renderables/animated/OrbitSystemRenderable.kt
```diff
@@ -10,9 +10,13 @@ import kotlin.math.cos
 import kotlin.math.sin
 import kotlin.time.Duration
 
-enum class OrbitDirection {
-    CLOCKWISE,
-    COUNTER_CLOCKWISE
+enum class OrbitDirection(private val displayName: String, val dirFactor: Int) {
+    NONE("None", 0),
```
hannibal002: what does none represent? its at least not used right now. Do we need this?

hannibal002 (changes requested):
I've not tested the PR in game, just looked at the code.
I like the changes, I have mostly just small formatting suggestions and clarification questions!

src/main/java/at/hannibal2/skyhanni/utils/render/item/atlas/SkyHanniItemAtlasKey.kt
DavidArthurCole: This is intended, but I'll put a comment in the code as well.
Because `stableId` is unique per instantiated renderable, it would make the AtlasKey _too_ specific if we used that as part of the hash.

I.e., if two renderables generate the same state, we want both of them to "share" the atlas space.

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniItemRenderCoordinator.kt
DavidArthurCole:
```kt
    fun SkyHanniItemRenderContext.setupAtlasRendering(
        frameNumber: Int,
        projectionBuffer: CachedOrthoProjectionMatrixBuffer,
    ) {
        pruneFrames(frameNumber)
```

Pruning occurs during setup, which clears stale frames, so yeah, the latter - we never need to clear it wholesale apart from texture reloads.

src/main/java/at/hannibal2/skyhanni/utils/renderables/animated/OrbitSystemRenderable.kt
DavidArthurCole: Almost certainly not, no.
This was cherry picked from a very old commit of Pet Display, will remove.

src/main/java/at/hannibal2/skyhanni/utils/render/item/atlas/SkyHanniItemAtlasKey.kt
hannibal002: Thanks for the explanation! The thing is, HashMap uses equals for key lookup, so as long as stableId is in equals, two renderables with different stableId will still get separate atlas entries and never actually share space. Could you clarify which behavior you want: if they should truly share, stableId should come out of equals too, and if they should stay separate, it should go into hashCode to avoid unnecessary hash collisions.

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniItemRenderCoordinator.kt
DavidArthurCole: Added a shutdown hook to the runtime that'll call this - we don't want the atlas to close on world change, it wants to be when the render engine shuts down.

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniItemRenderCoordinator.kt
DavidArthurCole: Not intended! Not sure how that got jumbled, but, fixed in next commit.

src/main/java/at/hannibal2/skyhanni/utils/render/item/atlas/SkyHanniItemAtlasKey.kt
DavidArthurCole: Should be accounted for now.

src/main/java/at/hannibal2/skyhanni/utils/compat/DrawContextUtils.kt
```diff
@@ -74,8 +74,38 @@ object DrawContextUtils {
...
-        action()
-        popMatrix()
+        try {
+            action()
+        } catch (e: Exception) {
+            ErrorManager.logErrorStateWithData(
+                "Error in pushPop action",
+                "Exception: ${e.message}"
+            )
+            throw e
```
hannibal002: i think this should work, no need to cathc, or skyhanni's error mananger.
if a function causes an error, the error shows up further up in the stack.
we just want to make sure that the pop matrix call happens, everything else is not relevant here imo


```kt
pushMatrix()
try {
    action()
} finally {
    popMatrix()
}
```

src/main/java/at/hannibal2/skyhanni/utils/render/item/atlas/SkyHanniItemAtlas.kt
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    private fun checkFull(): Boolean = isFull().also { if (it) needsGrowing = true }
+    private fun checkRowFull(): Boolean = isRowFull().also { if (it) newRow() }
+
+    private fun isFull() = cursorY + slotSize > sizePixels ||
+        (cursorX + slotSize > sizePixels && cursorY + slotSize * 2 > sizePixels)
+    private fun isRowFull() = cursorX + slotSize > sizePixels
+
+    private fun recordPosition(key: SkyHanniAtlasKey, slotX: Int, slotY: Int): SkyHanniItemAtlasEntry {
```
hannibal002, line 83: please use `.also {` instead of the impossible error, two times.

src/main/java/at/hannibal2/skyhanni/utils/compat/DrawContextUtils.kt
```diff
@@ -74,8 +74,38 @@ object DrawContextUtils {
...
+        } catch (e: Exception) {
+            ErrorManager.logErrorStateWithData(
+                "Error in pushPopResult action",
+                "Exception: ${e.message}"
+            )
+            throw e
+        } finally {
+            popMatrix()
+        }
+        return result ?: throw IllegalStateException("Result of pushPopResult action cannot be null")
```
hannibal002:
```kt
pushMatrix()
try {
    return action()
} finally {
    popMatrix()
}
```

hannibal002 (changes requested):

hannibal002 (approved):
code looks good to me

PR #5273, opened 2026-02-23
Fix: Tiki mask getting added to the tracker every time u take it off your head

NopoTheGamer:
## What
the code isnt ideal because the itemstack isnt passed through the events
also untested cus i dont got one 

## Changelog Fixes
+ Fixed Tiki Mask being added to the fishing profit tracker every time it was taken off your head. - nopo

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/FishingProfitTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/FishingProfitTracker.kt
@@ -30,7 +30,7 @@ import at.hannibal2.skyhanni.utils.NumberUtil.shortFormat
 import at.hannibal2.skyhanni.utils.RegexUtils.matchMatcher
 import at.hannibal2.skyhanni.utils.RenderDisplayHelper
 import at.hannibal2.skyhanni.utils.SimpleTimeMark
-import at.hannibal2.skyhanni.utils.SkyBlockItemModifierUtils.getReforgeModifier
+import at.hannibal2.skyhanni.utils.SkyBlockItemModifierUtils.getTimestamp
 import at.hannibal2.skyhanni.utils.SkyBlockUtils
 import at.hannibal2.skyhanni.utils.StringUtils
 import at.hannibal2.skyhanni.utils.collection.RenderableCollectionUtils.addSearchString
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/FishingProfitTracker.kt
```diff
@@ -243,12 +246,26 @@ object FishingProfitTracker {
...
+
     private fun tryAddItem(internalName: NeuInternalName, amount: Int, command: Boolean) {
         if (!FishingApi.isFishing(checkRodInHand = false)) return
         if (!isAllowedItem(internalName)) {
             ChatUtils.debug("Ignored non-fishing item pickup: $internalName'")
             return
         }
+        if (internalName == TIKI_MASK) {
+            var foundCleanTikiMask = false
+            for (stack in InventoryUtils.getItemsInOwnInventory()) {
```
hannibal002: This could be 
```kt
val foundCleanTikiMask = InventoryUtils.getItemsInOwnInventory().any {
    it.getInternalName() == TIKI_MASK && it.getReforgeModifier().isNullOrBlank()
}
```

hannibal002 (changes requested):
If someone never reforged their Tiki Mask and takes it off their head, a clean mask ends up in the inventory and the check would still let it through. Is this a case that can be safely ignored, or does it need handling?

Fazfoxy:
This might be a silly question to ask of you nopo but, couldn't you also check the timestamp of an item and cancel the addition to trackers in general if it's like, more than 2 minutes old (if hypixel somehow even creates an item that late) for items that have timestamps

lunaynx:
Yeah, I was about to suggest the timestamp thing. It *should* have a timestamp since it's a UUID item. 2 minutes is probably good enough to account for most reasonable clock drifts. Could also add a TimeLimitedCache for already seen UUIDs to further reduce false positives close to the original obtainment.

NopoTheGamer:
we dont have the item stack

Fazfoxy:
> we dont have the item stack

Couldn't we split trackers into a seperate event of some kind that ignores items that are "too old" though?

hannibal002 (changes requested):
The reforge check doesn't check for unreforged masks. A timestamp-based solution  (skip items older than ~2 minutes) would handle all cases.

NopoTheGamer:
that would be great but idk a good way to make that happen for all trackers

hannibal002:
wdym all trackers? this only is relevant for the fishing profit tracker here, isnt it?

hannibal002:
Closed due to #5294

PR #5278, opened 2026-02-24
Fix: Chroma preview text is always white

HyperKids:
## What
The preview text previously showed as white at all times. It now shows as white when disabled and as chroma when enabled, which is consistent with the last time this issue was fixed in #2302.

I am aware that chroma will appear to "freeze" when in a singleplayer world or when accessing configuration from the main menu - this is because the [animation relies on ClientEvents.totalTicks](https://github.com/hannibal002/SkyHanni/blob/458bc04ecb6ed24466b2cde18bdc8e040b10aa7f/src/main/java/at/hannibal2/skyhanni/mixins/hooks/GuiRendererHook.kt#L26) which seeming only increments when the game is not frozen (i.e. on a multiplayer server). This was previously resolved in #2302 by disabling the preview when outside of SkyBlock.

In my opinion, it's okay to leave this as-is given that (1) there is already a disclaimer in the menu ("Only on SkyBlock") and (2) this fix is much less complex than needing to conditionally enable/disable chroma. If you really want this fixed one way or another, there are two options:
1. Follow #2302 and display the preview text as white when not in SkyBlock
2. Modify how ticks are calculated, tying them to system time or some other external method

Tested on 1.21.10 and works fine. Let me know your thoughts about whether you want the "freezing in menu" fixed or if you're okay as it is.

## Changelog Fixes
+ Fixed Chroma preview text always displaying as white. - HyperKids

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/chroma/ChromaConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/chroma/ChromaConfig.kt
@@ -1,6 +1,7 @@
 package at.hannibal2.skyhanni.config.features.chroma
 
 import at.hannibal2.skyhanni.config.FeatureToggle
+import at.hannibal2.skyhanni.features.chroma.CHROMA_PREVIEW_COLOR_CODE
 import at.hannibal2.skyhanni.features.chroma.ChromaManager
 import com.google.gson.annotations.Expose
 import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorBoolean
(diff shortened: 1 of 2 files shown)


NopoTheGamer:
1. yes many things dont work correctly in a single player world dont worry about it 
2. the reason this has been an issue for so long is because we want it to only show for when OUR chroma is enabled and not like sba, but i will agree this is better than leaving it broken

HyperKids:
I went ahead and added unicode private use area character \uE002 () as a color code only used for the chroma preview. This fixes the concern of other mods interfering with the preview.

I did try to replicate the method used in #2302 (directly matching the preview string) but was unable to as "text" in onColorCodeCheck in MixinTextVisitFactory doesn't actually return the full string - it only returns the two characters that make up the color code (in this case, "§f").

Regardless, I think the method I used here is much simpler and doesn't add any complexity.

VixidDev:
At one point #2302 did fix the issue, but later it broke again since MoulConfig didn't return the full preview text from the config field, iirc it stripped formatting or something before arriving at the place where I checked if the string currently being rendered is the same as the value in the preview field, and it sounds like its still doing something like that.

In any case this seems good enough, and works since you are not using a Minecraft formatting code and persists when it gets to setChromaColorStyle, so clearly MoulConfig isn't stripping it.

Like Nopo said don't worry about the freezing in singleplayer worlds, in multiplayer no GUI ever pauses the game so its never an issue.

hannibal002 (changes requested):
please define the magic number somewhere in ChromaFontManager, with a good description text that explains the reason behind.

hannibal002 (approved):

PR #5280, opened 2026-02-25
Fix: Game crashes when clicking jump-to buttons in configuration

HyperKids:
## What
When the user has Fabric Language Kotlin installed, SkyHanni crashes when clicking one of two buttons:
- Inventory > Enchant Parsing > Chroma Warning > Go
- Garden > Pests > Pest Trap > Display > Go

Tested on version 1.21.10, with Fabric Language Kotlin v1.13.7 and v1.13.9.

This PR fixes the issue by using Java reflection instead of Kotlin reflection.

Of note, I can't replicate this crash when using runClient from Gradle - it only crashes when creating a fresh install with SkyHanni + Fabric Language Kotlin + Fabric API, or when downloading an existing modpack. I don't know if this issue affects other parts of the mod or not - this issue is the one that I happened to run into.

`Dungeon > Instance Chest Profit > Go` does use the same class but works - I think this has something to do with the two instances of the crash using Property\<T>, while the one that works uses Boolean.

Logs:
- Minimal install: https://mclo.gs/Ad3AeZm
- Fresh install of the Skyblock Enhanced modpack: https://mclo.gs/73QVlSV

## Changelog Fixes
+ Fixed game crash when clicking some buttons in the configuration UI. - HyperKids

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
@@ -32,6 +32,8 @@ object ConfigUtils {
     }
 
     private fun KProperty0<*>.tryFindEditor(editor: MoulConfigEditor<*>): ProcessedOption? {
+        // Java reflection is used because MoulConfig is relocated at build time, causing Kotlin reflection
+        // (this.javaField) to fail to resolve property descriptors in the production build.
         val receiver = (this as? CallableReference)?.boundReceiver
             ?.takeIf { it !== CallableReference.NO_RECEIVER }
             ?: return null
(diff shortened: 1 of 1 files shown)


lunaynx:
Yeah, the issue is that we're relocating MoulConfig (which is necessary to prevent mod incompatibilities as it has no API compatibility guarantees) but Kotlin reflection is still looking for the original package name (`io.github.notenoughupdates.moulconfig` instead of `at.hannibal2.skyhanni.deps.moulconfig`). It doesn't crash in the development environment because shadowing is not done there.

The only way I figured to fix it was to publish our own fork of MoulConfig which gets relocated at build time, and then depend on that, but this will probably do the trick.

src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
```diff
@@ -32,7 +32,14 @@ object ConfigUtils {
     }
 
     private fun KProperty0<*>.tryFindEditor(editor: MoulConfigEditor<*>): ProcessedOption? {
-        return editor.getOptionFromField(this.javaField ?: return null)
+        val receiver = (this as? CallableReference)?.boundReceiver
```
hannibal002, line 37: A brief code comment explaining why Java Reflection is used here instead of javaField would be helpful for future readers.

hannibal002 (changes requested):
I agree with the fix

hannibal002 (approved):

PR #5284, opened 2026-02-27
Backend: Upload Jar Artifacts

DavidArthurCole:
Changes `upload-artifact` to `v7`, and changes all artifact uploads to be `archive: false`, so it uploads `.jar` and `.json` files now.

exclude_from_changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,16 +33,17 @@ jobs:
                 uses: hannibal002/SkyHanni/.github/actions/gradle-retry@beta
                 with:
                     gradle-command: assemble -x test --stacktrace -PskipDetekt=true
-            -   uses: actions/upload-artifact@v4
+            -   uses: actions/upload-artifact@v7
                 name: Upload development build
                 with:
                     name: "Development Build"
                     path: build/libs/*.jar
+                    archive: false
             -   name: Test with Gradle w/retry
                 uses: hannibal002/SkyHanni/.github/actions/gradle-retry@beta
                 with:
                     gradle-command: test
-            -   uses: actions/upload-artifact@v4
+            -   uses: actions/upload-artifact@v7
                 name: "Upload test report"
                 if: ${{ !cancelled() }}
                 with:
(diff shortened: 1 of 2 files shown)


DavidArthurCole:
https://github.com/DavidArthurCole/SkyHanni/actions/runs/22469069994
for regexes upload

.github/workflows/build.yml
```diff
@@ -83,9 +86,52 @@ jobs:
...
+            fail-fast: false
+            matrix: ${{ fromJSON(needs.preprocess.outputs.jar-matrix) }}
+        steps:
+            -   name: Download staged JARs
+                uses: actions/download-artifact@v4
+                with:
+                    name: jars-transfer
+
+            -   name: Upload ${{ matrix.jar }} unarchived
+                continue-on-error: true
```
hannibal002, line 118: if we remove this line, we should see if the upload failed

.github/workflows/build.yml
```diff
@@ -33,16 +33,17 @@ jobs:
                 uses: hannibal002/SkyHanni/.github/actions/gradle-retry@beta
                 with:
                     gradle-command: assemble -x test --stacktrace -PskipDetekt=true
-            -   uses: actions/upload-artifact@v4
+            -   uses: actions/upload-artifact@v7
                 name: Upload development build
                 with:
                     name: "Development Build"
                     path: build/libs/*.jar
+                    archive: false
```
hannibal002, line 41: this only works for one single file, the path above can catch multiple files

.github/workflows/generate-constants.yaml
```diff
@@ -34,11 +34,12 @@ jobs:
                 uses: ./.github/actions/gradle-retry
                 with:
                     gradle-command: generateRepoPatterns --stacktrace
-            -   uses: actions/upload-artifact@v4
+            -   uses: actions/upload-artifact@v7
                 name: Upload generated repo regexes
                 with:
                     name: Repo Regexes
```
hannibal002, line 40: The two `name: Repo Regexes` lines are wrong now. The upload name is ignored when `archive: false` (the filename is used instead), so the download step can no longer find the artifact

hannibal002 (changes requested):
see

https://github.com/actions/upload-artifact/blob/main/action.yml#L49

DavidArthurCole:
Will revisit when Luna does multiversion stuff.

PR #5288, opened 2026-02-27
Fix: Revert Modern Custom Item Rendering for now

lunaynx:
## What
Reverts #5262 for now, because it causes VRAM memory leaks, leading to stuttering and crashes, sometimes even system-wide graphics rendering breakdowns when items are rendered.

## Changelog Fixes
+ Fixed VRAM memory leak in SkyHanni 7.1.0 causing freezes and crashes. - Luna


lunaynx:
Closing in favor of #5291.

PR #5290, opened 2026-02-27
Fix: Having multiple mouse buttons as custom garden keybinds only uses the last pressed button.

Taubsie:
## What
When setting multiple mouse buttons in the Garden feature "custom keybinds", only the last input is taken into consideration, overwriting all other held buttons.

For example: I did bind `LMB` to `Attack`, `MB4` (mouse button 4) to `Strafe Left` and `MB5` to `Strafe Right`. This allows me to only use my mouse while farming, without needing my keyboard.
Without this bug fix, I can currently only do one of the actions (`Attack` or `Strafe Left`/`Strafe Right`) at once. (whichever was clicked last).

<details>
<summary>Implementation detail</summary>

This needed a change to the current `MouseCompat` handler to be able to save multiple mouse button inputs - for this, I used an array of size 6 that that saves the state of each mouse button.
The mixin needed to be adapted as well, only setting the state to true when the mouse is held down (action = 1), and set to false once it's released (action != 1).

</details>

## Changelog Fixes
+ Fixed Custom Keybinds for Garden not working when setting multiple mouse buttons. - Taubsie


src/main/java/at/hannibal2/skyhanni/utils/compat/MouseCompat.kt
```diff
@@ -1,21 +1,32 @@
...
 import net.minecraft.client.Minecraft
+import net.minecraft.client.input.MouseButtonInfo
 
 object MouseCompat {
     var deltaMouseY = 0.0
     var deltaMouseX = 0.0
     var scroll = 0.0
     var timeDelta = 0.0
-    var lastEventButton = -1
+    val buttonStates = BooleanArray(6) { false }
```
hannibal002: make private

hannibal002: extract magic number 6 to a new constant. give it a name explaining what this number represents

hannibal002: the lambda can be removed, since the boolean array per default gets initialized as false already.

hannibal002 (changes requested):
The approach is good. Just a few small fixes, then this is ready to get merged.

src/main/java/at/hannibal2/skyhanni/utils/compat/MouseCompat.kt
Taubsie: Thanks for the quick reply, I changed those :)
Honestly didn't realize that the BooleanArray is initialized to false, I thought it has more of a behavior of `java.lang.Boolean`

hannibal002 (approved):

PR #5291, opened 2026-02-27
Fix: Custom item rendering causing high RAM and VRAM usage

VixidDev:
## What
Makes 'static' custom rendered items use the 'animated' code path for rendering which utilises an atlas instead of continously creating a fallback renderer every frame with no clean up. (The whole idea of instantiating the fallback renderer is flawed, so correctly cleaning it up wouldn't solve the core issue anyway).

This is just a temporary fix until a better and more complete solution is made for the item renderer.

## Changelog Fixes
+ Fixed some item rendering causing extreme performance issues. - Vixid


lunaynx (changes requested):
You now have an if/else that does the same thing in both branches, the condition can be eliminated. Maybe leave the old code commented out if it's something that should be revisited in the future.

lunaynx (approved):

PR #5294, opened 2026-02-28
Fix + Backend: Include armor slots in item add detection

hannibal002:
## What
`OwnInventoryData.getCurrentItems()` only counted items in the main inventory slots. Moving an armor piece from an armor slot to the inventory (or vice versa) was detected as a new item pickup. This caused issues like the Tiki Mask being added to the fishing profit tracker every time it was taken off the head.

The fix includes armor slots in the item count so that moving items between armor and inventory does not change the total. Additionally, closing the wardrobe now temporarily ignores armor items to prevent false detections from armor set swaps.

Supersedes #5273.

## Changelog Fixes
+ Fixed Tiki Mask being added to the Fishing Profit Tracker every time it was taken off the head. - hannibal2

+ Added `InventoryUtils.getArmorInternalNames()` utility method. - hannibal2
+ Included armor slots in `OwnInventoryData.getCurrentItems()` item tracking. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/OwnInventoryData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/OwnInventoryData.kt
@@ -35,6 +35,7 @@ object OwnInventoryData {
 
     private var itemAmounts = mapOf<NeuInternalName, Int>()
     private var dirty = false
+    private var lastWardrobeClose = SimpleTimeMark.farPast()
 
     /**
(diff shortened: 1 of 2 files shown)


NopoTheGamer:
are you sure this causes no issues?

hannibal002:
This does less than we did before. this is just a fancier way of "avoid item add events when the user moves the armor items around".

the only added behaviour possible is if another hypixel feature would change the armor slots. then, in this case we would get the event, and if this happens with the tiki mask, we would again add the item wrongly to the tracker.

once this happens, we can fix it again, tho.

PR #5295, opened 2026-02-28
Fix: Some trackers not showing due to invalid stopwatch config

NopoTheGamer:
## What
if you downgraded ur config it could leave null stopwatch data it seems

## Changelog Fixes
+ Fixed some Garden displays not rendering if you downgraded the mod. - nopo


hannibal002:
The original fix guards one access site.
I loved the cleanup into `migrateData()` so all null entries are removed once before any further access. The safe calls in `filterAndRemove` are no longer needed now.

hannibal002 (changes requested):
what do you think of my approach?

hannibal002:
your approach with nullable fields in the config was alr the correct one. i partially reverted and reworked it a bit so that we dont need to deal with nullable values everywhere

hannibal002 (approved):

PR #5296, opened 2026-03-02
Feature: Quantity preset overlay for supercrafting menu

HyperKids:
## What
Adds an quantity preset overlay to the supercrafting GUI. Default preset quantities can be changed in configuration, which accepts a comma-separated list of integers. Similar code to the optimal angles overlay / rancher boots overlay.
<details>
<summary>Images</summary>
image
</details>

## Changelog New Features
+ Added Supercrafting menu quantity preset overlay. - HyperKids

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftPresetsConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftPresetsConfig.kt
@@ -19,7 +19,7 @@ class SuperCraftPresetsConfig {
     var enabled: Boolean = false
 
     @Expose
-    var presets: MutableList<Int> = mutableListOf(4, 8, 16, 32, 64, 128, 256, 512)
+    val presets: MutableList<Int> = mutableListOf(4, 8, 16, 32, 64, 128, 256, 512)
 
     @Expose
     @ConfigLink(owner = SuperCraftPresetsConfig::class, field = "enabled")
--- a/src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftPresetsConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftPresetsConfig.kt
@@ -1,0 +1,35 @@
+package at.hannibal2.skyhanni.config.features.inventory
+
+import at.hannibal2.skyhanni.config.FeatureToggle
+import at.hannibal2.skyhanni.config.core.config.Position
+import com.google.gson.annotations.Expose
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorBoolean
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorText
+import io.github.notenoughupdates.moulconfig.annotations.ConfigLink
+import io.github.notenoughupdates.moulconfig.annotations.ConfigOption
+
+class SuperCraftPresetsConfig {
+
+    @Expose
+    @ConfigOption(
+        name = "Enabled",
+        desc = "Show preset craft amounts in the Supercrafting sign menu."
+    )
+    @ConfigEditorBoolean
+    @FeatureToggle
+    var enabled: Boolean = false
+
+    @Expose
+    @ConfigOption(
+        name = "Preset Values",
+        desc = "Comma-separated list of supercraft amounts to display."
+    )
+    @ConfigEditorText
+    var presets: String = "4,8,16,32,64,128,256,512"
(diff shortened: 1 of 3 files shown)


HyperKids:
I set this feature to default to off, but would love to have it default on - I'm just not sure if it's convention to have newly added features enabled or disabled, let me know.

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftPresetsConfig.kt
```diff
@@ -0,0 +1,35 @@
...
+    @FeatureToggle
+    var enabled: Boolean = false
+
+    @Expose
+    @ConfigOption(
+        name = "Preset Values",
+        desc = "Comma-separated list of supercraft amounts to display."
+    )
+    @ConfigEditorText
+    var presets: String = "4,8,16,32,64,128,256,512"
```
NopoTheGamer: i would perfer if this was done in any other way. maybe a command?

HyperKids: How would you propose the command syntax to look like? In my opinion, having it in the configuration is the best for UX for a few reasons:
1. it allows users to view the current presets as they add/remove theirs
2. easy to add numbers in between other numbers (e.g. with a current config of `64,512`, you can change it to `64,256,512` without needing to specify a position or forcing low-to-high sorting)

I'm also not a fan of having comma-separated lists to manage the supercraft amounts, but I couldn't really think of a better solution given the constraints of MoulConfig.

NopoTheGamer: i really think its fine if its forced sorted low to high

HyperKids: I'm also fine with force sorting low-to-high, but still thinking about how to set up a command syntax if we were to go that route. I see two options:
1. one command like `/shsupercraftpreset 2,4,8,16,32,64,128,256,512` which would just overwrite all existing presets (could tab-complete the existing preset list so the user doesn't have to start from scratch)
2. two commands, `/shsupercraftpresetadd 1024`, `/shsupercraftpresetremove 1024` which would add/remove numbers one at a time, but the user is sort of doing this blind unless they manually reference the supercraft quantity GUI

Leaning towards option 2 but I still strongly prefer the config method since I think it's better UX and easier for discovery

src/main/java/at/hannibal2/skyhanni/features/inventory/SuperCraftPresets.kt
```diff
@@ -0,0 +1,110 @@
...
+        SignUtils.setTextIntoSign("$amount", 0)
+    }
+
+    private fun parsePresets(): List<Int> = config.presets
+        .split(",")
+        .mapNotNull { it.trim().toIntOrNull() }
+        .filter { it > 0 }
+        .distinct()
+}
+
```
hannibal002: too many empty lines

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftPresetsConfig.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    @Expose
+    @ConfigLink(owner = SuperCraftPresetsConfig::class, field = "enabled")
+    val signPosition: Position = Position(100, 100, false, true)
```
hannibal002: false and true are already the default values for the postion constructor. can be omitted.

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftPresetsConfig.kt
```diff
@@ -0,0 +1,35 @@
...
+}
+
```
hannibal002: only one empty line at the end of file is enough

src/main/java/at/hannibal2/skyhanni/config/features/inventory/SuperCraftPresetsConfig.kt
hannibal002: How about the command `/shsupercraftpreset <number>`.
Then, either if the number exists, remove it, or if it doesn't exist, add it.
After either of those two cases, send the full list of current presets in chat as a response.

hannibal002 (changes requested):

PR #5297, opened 2026-03-02
Improvement: Sky Mall and Lottery displays outside of mining/foraging islands

HyperKids:
## What
Modified configuration option for Sky Mall (mining) and Lottery (foraging) HUD, which allows them to be shown outside of mining/foraging islands. Config bumped to version 125.

Resolves https://discord.com/channels/997079228510117908/1474911522810757302

## Changelog Improvements
+ Added option to show Sky Mall and Lottery HUD outside of mining/foraging islands. - HyperKids

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/mining/HotxFeatures.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/mining/HotxFeatures.kt
@@ -2,6 +2,7 @@ package at.hannibal2.skyhanni.features.mining
 
 import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.api.event.HandleEvent
+import at.hannibal2.skyhanni.config.ConfigUpdaterMigrator
 import at.hannibal2.skyhanni.data.hotx.HotfData
 import at.hannibal2.skyhanni.data.hotx.HotmData
 import at.hannibal2.skyhanni.data.hotx.HotxHandler
@@ -11,6 +12,7 @@ import at.hannibal2.skyhanni.events.RenderItemTipEvent
 import at.hannibal2.skyhanni.events.chat.SkyHanniChatEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.test.command.ErrorManager
+import at.hannibal2.skyhanni.utils.ConfigUtils
 import at.hannibal2.skyhanni.utils.LorenzColor
 import at.hannibal2.skyhanni.utils.RenderUtils.highlight
 import at.hannibal2.skyhanni.utils.RenderUtils.renderRenderable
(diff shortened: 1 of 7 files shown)


src/main/java/at/hannibal2/skyhanni/features/mining/HotxFeatures.kt
```diff
@@ -26,8 +28,12 @@ object HotxFeatures {
     @HandleEvent
     fun onRenderOverlay(event: GuiRenderEvent.GuiOverlayRenderEvent) {
         val (handler, configPos) = when {
-            HotmData.inApplicableIsland && configHotm.skyMallDisplay -> Pair(HotmData, configHotm.skyMallPosition)
-            HotfData.inApplicableIsland && configHotf.lotteryDisplay -> Pair(HotfData, configHotf.lotteryPosition)
+            configHotm.skyMallDisplay.let { vis ->
+                vis != SkyMallDisplayVisibility.OFF && (vis == SkyMallDisplayVisibility.EVERYWHERE || HotmData.inApplicableIsland)
```
hannibal002: this code here looks complicated, and like code duplication. can be compacted with a when instead. and use a normal if instead of a when for the outer condition check. also please use a to b instead of Pair(a, b)

src/main/java/at/hannibal2/skyhanni/features/mining/HotxFeatures.kt
```diff
@@ -108,4 +114,29 @@ object HotxFeatures {
...
+    fun onConfigFix(event: ConfigUpdaterMigrator.ConfigFixEvent) {
+        event.transform(125, "mining.hotm.skyMallDisplay") {
+            ConfigUtils.migrateBooleanToEnum(it, SkyMallDisplayVisibility.MINING_ONLY, SkyMallDisplayVisibility.OFF)
+        }
+        event.transform(125, "foraging.hotf.lotteryDisplay") {
+            ConfigUtils.migrateBooleanToEnum(it, LotteryDisplayVisibility.FORAGING_ONLY, LotteryDisplayVisibility.OFF)
+        }
+    }
+
+    enum class SkyMallDisplayVisibility(val display: String) {
```
hannibal002: please move those enums in the config

hannibal002 (changes requested):

HyperKids:
Fixed an issue: when both should have been visible, only one would show up due to the when statement.

Additionally, HOTF Lottery is now y=120 by default so it doesn't sit directly on top of the HOTM Sky Mall one.

src/main/java/at/hannibal2/skyhanni/features/mining/HotxFeatures.kt
```diff
@@ -25,11 +30,16 @@ object HotxFeatures {
 
     @HandleEvent
     fun onRenderOverlay(event: GuiRenderEvent.GuiOverlayRenderEvent) {
-        val (handler, configPos) = when {
-            HotmData.inApplicableIsland && configHotm.skyMallDisplay -> Pair(HotmData, configHotm.skyMallPosition)
-            HotfData.inApplicableIsland && configHotf.lotteryDisplay -> Pair(HotfData, configHotf.lotteryPosition)
-            else -> return
-        }
+        if (configHotm.skyMallDisplay != SkyMallDisplayVisibility.OFF &&
```
HyperKids: Let me know if you see any way to make the new implementation more compact / readable.

hannibal002 (changes requested):
> Let me know if you see any way to make the new implementation more compact / readable.

Ideally, we would abstract the enum, but MoulConfig can't support that in an enum-like way in the easy approach I am looking for. So for now this is as compact as I can make it

What do you think of this approach?

HyperKids:
LGTM, definitely a lot more readable, thanks for the edits!

hannibal002 (approved):

PR #5304, opened 2026-03-02
Fix: Chronomatron helper misses first note of sequence

HyperKids:
## What
The chronomatron experiment table helper occasionally misses the first note of the sequence.

This PR initializes the table with `chronHasBeenEmpty = true` to ensure the first note is always captured.

My theorized explanation for why this bug happens is that on the initial run, `InventoryUpdatedEvent` sometimes doesn't fire until `activeColors.isEmpty()` is already false, which incorrectly leaves `chronHasBeenEmpty` as false due to the initialized state, which then triggers the return statement on line 273.

Getting the bug to work is very, very inconsistent across machines / installations - I personally initially had this issue on a Lunar client install, and it seemingly happened out of the blue after it was working perfectly well for months. I cannot replicate this issue on a fresh Windows or Mac install via Modrinth, but can replicate it 100% of the time by running the Minecraft client on MacOS from inside the IDE.

## Changelog Fixes
+ Fixed Chronomatron experiment table helper missing the first note. - HyperKids

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/experimentationtable/ExperimentsAddonsHelper.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/experimentationtable/ExperimentsAddonsHelper.kt
@@ -51,7 +51,7 @@ object ExperimentsAddonsHelper {
     private val userUltrasequencerProgress: MutableList<Int> = mutableListOf()
     private val ultrasequencerDyeMap: MutableMap<Int, ItemStack> = mutableMapOf()
 
-    private var chronHasBeenEmpty: Boolean = false
+    private var chronHasBeenEmpty: Boolean = true
     private var lastChronomatronSound: SimpleTimeMark = SimpleTimeMark.farPast()
     private var currentAddonPhase: HelperPhase? = null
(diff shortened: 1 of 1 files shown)


DavidArthurCole (changes requested):
I was able to replicate bug last week, on a singular private island server, and have been unable to since.
Your explanation make sense in theory, however I would like to see some debugging screenshots that point to this being the case, before we call this a fix.

There is a debug helper for experiment addons:
image

Please use this, and add debug line for `chronHasBeenEmpty`.
This should help to verify your explanation.

HyperKids:
YouTube: https://youtu.be/s4oCOt1mgaU

More or less confirms what I was thinking.

I was able to consistently replicate whenever I was not "picking up" the dye to start the experiment - for example, using middle click to select the experiment consistently triggered the bug. This is probably why it was highly noticeable on Lunar Client - they map left click to middle click in many menus.

hannibal002 (changes requested):
As David requested, please add a debug line for chronHasBeenEmpty to the existing debug panel.

hannibal002 (approved):

PR #5312, opened 2026-03-03
Fix: ff display showing ff wrong when bpc is disabled

NopoTheGamer:
## What
i pushed this 2 weeks ago and forgot to either test it or make a pr

## Changelog Fixes
+ Fixed FF Display being incorrect when bpc is disabled. - nopo


hannibal002:
stale

PR #5314, opened 2026-03-03
Fix: Derpy/Aura not stacking multiplicatively

Fazfoxy:
## What
Someone Should probably remember to do this

exclude_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/ElectionApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/ElectionApi.kt
@@ -307,9 +307,19 @@ object ElectionApi {
     val isDerpy get() = Perk.DOUBLE_MOBS_HP.isActive
     val isAura get() = Perk.WORK_HARDER.isActive
 
-    fun Int.derpy() = if (isDerpy) this / 2 else if (isAura) (this / 11 * 10) else this
+    fun Int.derpy(): Int {
+        var health = this
+        if (isDerpy) health /= 2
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/data/ElectionApi.kt
```diff
@@ -307,9 +307,19 @@ object ElectionApi {
     val isDerpy get() = Perk.DOUBLE_MOBS_HP.isActive
     val isAura get() = Perk.WORK_HARDER.isActive
 
-    fun Int.derpy() = if (isDerpy) this / 2 else if (isAura) (this / 11 * 10) else this
+    fun Int.derpy(): Int {
+        var health = this
+        if (isDerpy) health /= 2
+        if (isAura) (health / 11 * 10)
```
lunaynx:
```suggestion
        if (isAura) health = (health / 11 * 10)
```

lunaynx (changes requested):

hannibal002 (changes requested):
Someone Should probably remember to do this

?

also, this looks like a user facing impact

Fazfoxy:
> Someone Should probably remember to do this
> 
> ?
> 
> also, this looks like a user facing impact

I mean it's a user facing bug fix, but the next timer it impacts users is, unclear, 8.0.0 fixing it might not matter to a user if the next aura is in december again idk

hannibal002 (approved):
thanks for explaining your reasoning, makes sense to me

PR #5318, opened 2026-03-03
Backend + Improvement: Custom Modern Item Rendering

DavidArthurCole:
## What
Re-did both implementations of our item renderers, so that we now have an Atlas, which will store static textures (one per item - stores the "highest scale" one, and scales down as needed), and a PIP renderer, which will deal with animated items.

Now that these paths are efficient, it might be worth considering this logic:
```kt
        if (rotationVec == Vec3.ZERO && (totalItemScale <= 1 || !trackingState.usesBlockLight()))
            return item.normalRenderOnScreen(translateX, translateY, finalItemScale.toFloat())
```
Does it make sense to ever use the normal renderer?

## Changelog Improvements
+ Improved visual quality of items in displays. - Daveed

+ Re-enabled custom item rendering on modern Minecraft versions. - Daveed


src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniItemRenderCoordinator.kt
```diff
@@ -1,84 +1,135 @@
...
+            frameNumber,
+            resources.guiScale,
+        )
+        slot.render(renderContext, state, guiRenderState, projectionBuffer)
+    }
+
+    private fun handleEviction(frameNumber: Int) {
+        if (frameNumber == lastEvictFrame) return
+        lastEvictFrame = frameNumber
+        realtimeSlots.entries.removeIf { (id, slot) ->
```
hannibal002, line 127: should realtimeSlotLastSeen also get cleared here?

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniPipCoordinatorRenderer.kt
```diff
@@ -12,17 +12,15 @@ class SkyHanniPipCoordinatorRenderer(
...
     }
 
-    fun takePendingStates(): List<SkyHanniGuiItemRenderState> = synchronized(pendingStates) {
-        val result = ArrayList(pendingStates)
-        pendingStates.clear()
-        return result
-    }
-
-    override fun textureIsReadyToBlit(state: SkyHanniGuiItemRenderState): Boolean = synchronized(pendingStates) {
+    override fun textureIsReadyToBlit(state: SkyHanniGuiItemRenderState): Boolean {
```
hannibal002, line 15: synchronized(pendingStates) is gone now. what was it  used for earlier, and why dont we need it anymore here?

src/main/java/at/hannibal2/skyhanni/utils/render/item/atlas/SkyHanniItemAtlas.kt
```diff
@@ -1,280 +1,184 @@
...
+        allocateTextures(size, "SkyHanni item atlas", "SkyHanni item atlas depth", usage)
+        packer = SkyHanniAtlasBinPacker(size)
+        @Suppress("UnsafeCallOnNullableType")
+        renderer = SkyHanniItemAtlasRenderer(size, textureView!!, depthTextureView!!, texture!!, depthTexture!!)
+    }
 
     /**
      * Runs after rendering finishes.
      */
     private fun tryGrow() {
```
hannibal002: currently unused. previously, was called in SkyHanniItemAtlas.render()

src/main/java/at/hannibal2/skyhanni/utils/render/item/atlas/SkyHanniItemAtlas.kt
```diff
@@ -1,280 +1,184 @@
...
+    private var sizePixels = 0
+    private var packer: SkyHanniAtlasBinPacker? = null
+    private var renderer: SkyHanniItemAtlasRenderer? = null
+    private var needsGrowing = false
     private val positions = HashMap<SkyHanniAtlasKey, SkyHanniItemAtlasEntry>()
-    private val animatedFrames = HashMap<SkyHanniAnimatedAtlasKey, SkyHanniAnimatedItemAtlasEntry>()
 
-    private var cursorX = 0
-    private var cursorY = 0
+    val usage = GpuTexture.USAGE_RENDER_ATTACHMENT or
```
hannibal002: can be private

src/main/java/at/hannibal2/skyhanni/utils/render/item/atlas/SkyHanniItemAtlas.kt
```diff
@@ -1,280 +1,184 @@
...
+        if (entry is SkyHanniAnimatedItemAtlasEntry) {
+            positions[state.atlasKey] = SkyHanniAnimatedItemAtlasEntry(
+                entry.x, entry.y, entry.u, entry.v, entry.pixelSize, frameNumber
+            )
+        }
+        renderer?.submitBlitForState(state, guiRenderState, entry)
+        return true
     }
 
     fun invalidate() {
```
hannibal002, line 152: packer and renderer already get set to null in close(), no need to set them in invalidate() as well to null

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniPipCoordinatorRenderer.kt
DavidArthurCole: It was required when we were doing this as a "multi-threaded" thing. In essence this was caused by our mixin point being called in a loop by Minecraft's code.

Now that we have the dual-injection point(s), this is no longer needed, as all blitting happens on the render thread.

src/main/java/at/hannibal2/skyhanni/utils/render/item/atlas/SkyHanniItemAtlas.kt
```diff
@@ -1,280 +1,166 @@
...
+            val existing = positions[key]
+
+            if (existing != null && existing.pixelSize >= neededPixels) {
+                if (key is SkyHanniAnimatedAtlasKey) key.clearPreviousFrame()
+                // Cache hit, no render job needed, blit submitted later per-item
+                continue
+            }
+
+            val node = packer.insert(neededPixels) ?: run {
+                needsGrowing = true
```
hannibal002: you have deleted the tryGrow() function. now, the needsGrowing field  here gets written in but never read. if we truly dont need the field anymore, please delete it.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniGuiItemRenderState.kt
```diff
@@ -1,37 +1,62 @@
...
+        val itemStack = this@SkyHanniGuiItemRenderState.itemStack
+        itemStack.getSkullTexture()?.let { appendModelIdentityElement(it) }
+        itemStack.getSkullOwner()?.let { appendModelIdentityElement(it) }
+        appendModelIdentityElement(rotationVector)
+        // stableId intentionally NOT in modelIdentity. atlas key equality already excludes it,
+        // and including it here prevents cache hits for static items across frames.
+        frameNumber?.let { appendModelIdentityElement(it) }
+        if (rotationVector != Vec3.ZERO || frameNumber != null) setAnimated()
+    }
+    private val _atlasKey = run {
```
hannibal002: unnecessary backing field

hannibal002 (changes requested):
one  small thing, then the code is ready for merge imo.

i assume you have tested it extensively

DavidArthurCole:
> i assume you have tested it extensively

yeah. all of the issues with the old version (the weird resizing, scaling, etc.) is fixed, and it no longer used a bajillion gigs of ram to atlas, so, I would say it's good

hannibal002 (approved):

PR #5320, opened 2026-03-03
Fix: Rendering in Non-Render Events in Composter Overlay

DavidArthurCole:
Replaced calls to `update()` (which was causing rendering on non-render threads) with a dirty flag, that will cause it to be re-calced when needed.

## Changelog Fixes
+ Fixed error when first opening Composter with overlay enabled. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/composter/ComposterOverlay.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/composter/ComposterOverlay.kt
@@ -21,8 +21,6 @@ import at.hannibal2.skyhanni.events.InventoryFullyOpenedEvent
 import at.hannibal2.skyhanni.events.IslandChangeEvent
 import at.hannibal2.skyhanni.events.NeuRepositoryReloadEvent
 import at.hannibal2.skyhanni.events.RepositoryReloadEvent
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/composter/ComposterOverlay.kt
```diff
@@ -119,26 +122,27 @@ object ComposterOverlay {
...
     private val OIL_BARREL = "OIL_BARREL".toInternalName()
 
-    @HandleEvent(TabListUpdateEvent::class, priority = HandleEvent.LOW)
-    fun onTabListUpdate() {
-        if (inInventory) {
-            update()
+    @HandleEvent(onlyOnIsland = IslandType.GARDEN)
+    fun onWidgetUpdate(event: WidgetUpdateEvent) {
+        if (!isEnabled() || !event.isWidget(TabWidget.COMPOSTER)) return
+        val newComposterHash = event.widget.hashCode()
```
hannibal002: i dont think you need hash code comparisons here. 

the widget udpate event already only sends when the text in the widget changes or the widget dissapears/appreas.

also, the only value of event.widget that can happen at this code line is TabWidget.COMPOSTER. an enum entry has always the same hash code.

hannibal002 (changes requested):

PR #5328, opened 2026-03-05
Backend: Support for other RNG Meters in Copy RNG Meter Feature

jani270:
## What
Hello everyone, and welcome to this amazing PR. In this PR, I wanted to work on a feature that has been around for about 85,622,400 seconds or in other numbers 85,622,400,000,000,000 nanoseconds (≈ 8.56224 × 10¹⁶ ns). It has been a great feature for all that time. Unfortunately, it has become somewhat outdated since the addition of the Experiments RNG Meter, the Nucleus RNG Meter, and a few days ago, the Frozen Corpse RNG Meter.

For that reason, I had the wonderful idea of updating this feature to add support for these very creative RNG meters.

_While you are reading this i would like to point out the small little fact that Octopuses have three hearts._

I hope you had a great time reading this very useful text, and I wish you a wonderful rest of your day.

Greetings,
Jani

+ Added support for other RNG Meters in the Copy RNG Meter feature. - jani270

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/test/TestCopyRngMeterValues.kt
+++ b/src/main/java/at/hannibal2/skyhanni/test/TestCopyRngMeterValues.kt
@@ -8,7 +8,7 @@ import at.hannibal2.skyhanni.events.InventoryFullyOpenedEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.ItemUtils.getInternalName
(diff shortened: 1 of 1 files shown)


hannibal002 (changes requested):
Thank you for this fascinating PR. While reading it, I learned that octopuses have three hearts, that 85,622,400 seconds is approximately 2.7 years, and that this feature existed for that long. What I did not learn is what technical changes were actually made in this PR.
One more thing I noticed: Nucleus appears in the pattern but has no REGEX-TEST. Unlike octopus hearts, untested regex cases are not a fun fact — they are a bug waiting to happen.

hannibal002 (approved):

PR #5330, opened 2026-03-05
Backend: ClientShutdownEvent & RenderShutdownEvent

lunaynx:
## What
Got rid of shutdown hooks, and instead created two new events:
* `ClientShutdownEvent`: gets fired on the `ClientLifecycleEvents.CLIENT_STOPPING` Fabric event
* `RenderShutdownEvent`: gets fired on `Window#close` with an assurance that we're on the render thread
  * `ClientShutdownEvent` would technically work for this right now, but with Mojang's plans to split the render thread from the main client thread in the future, we might as well separate it now.

This is meant to achieve two things:
* Fixes "Rendersystem called from wrong thread" error due to atlas closing whenever the game is closed.
* May or may not reduce instances of people's configs getting corrupted.

+ Added `ClientShutdownEvent` and `RenderShutdownEvent` and migrated shutdown hooks to them as appropriate. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
+++ b/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
@@ -73,13 +73,6 @@ object SkyHanniMod {
         configManager.firstLoad()
         if (PlatformUtils.getRepoPatternDumpLocation() == null) EnoughUpdatesRepoManager.initRepo()
         MinecraftConsoleFilter.initLogging()
-        val runtime = Runtime.getRuntime()
-        runtime.addShutdownHook(
-            Thread { configManager.saveConfig(ConfigFileType.FEATURES, "shutdown-hook") },
-        )
-        runtime.addShutdownHook(
-            Thread { SkyHanniItemRenderCoordinator.closeAtlas() }
-        )
         try {
             if (PlatformUtils.getRepoPatternDumpLocation() == null) SkyHanniRepoManager.initRepo()
         } catch (e: Exception) {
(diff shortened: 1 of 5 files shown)


DavidArthurCole (approved):
Code looks fine in general, just wanna clarify, since it doesn't inherently look to be a render-safe event we're posting.

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
```diff
@@ -104,6 +97,13 @@ object SkyHanniMod {
         this.screenToOpen = null
     }
 
+    @HandleEvent
+    fun onClientShutdown() {
+        configManager.saveConfig(ConfigFileType.FEATURES, "shutdown-hook")
+
+        SkyHanniItemRenderCoordinator.closeAtlas()
```
DavidArthurCole, line 107: Is this actually going to fix the whole "not on a render thread" issue? It's still getting triggered by our event bus without an explicit check.

lunaynx: It should for now at least, based on reading of Minecraft and Fabric code, it should always get executed on the main (client/render) thread. It may have to be revisited if/when the render thread split happens.

DavidArthurCole: Would there be any harm in adding a caught RenderSystem thread execution then, just for good measure?

DavidArthurCole: As in a runCatching or equivalent to gracefully exit if it isn't on the render thread

DavidArthurCole (approved):

PR #5338, opened 2026-03-06
Backend: Fix VFE IDs on 1.21.11

DavidArthurCole:
1.21.10, Minecraft registers 0-5, on 1.21.11 they register 0-6, so load the last registered ID dynamically.

exclude_from_changelog


hannibal002:
https://github.com/hannibal002/SkyHanni-REPO/commit/a79f817dc9aa93683a75bf29cd4bdecd67809404

DavidArthurCole:
> [hannibal002/SkyHanni-REPO@a79f817](https://github.com/hannibal002/SkyHanni-REPO/commit/a79f817dc9aa93683a75bf29cd4bdecd67809404)

That really shouldn't ever show for users. I'm gong to add a PlatformUtils check to it, so it only complains in dev env. As I noted in contrib-chat, the "preferred ID" not being used is not at all a problem, and causes no issues even when it happens, it's just something for us to look for internally.

PR #5274, opened 2026-02-23
Backend: ItemRenderable Config Classes

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/5262

## What
Converts ItemStackRenderable & ItemStackRenderable to "config class" instantiations, where instead of providing direct values to the companion functions, configurations are now stored in a class object, and can be configured in an external lambda. This will become necessary with Pet Display, but at a more broad level, allows for on-the-fly editing of the configuration of a Renderable, which can help with re-use and modification without creating new instances.

I did my best to leave function calls intact and the same, where I could. The biggest change is changing from inline function params, to lamdba block setters instead, however, we were already specifying param names in most cases, and in others, we should have been, so this shouldn't be a massive change.

Also extracted `Bounce`, `Rotate`, and `Frame` functionalities from the `AnimatedItemStackRenderable` to broad decorators, so that they can be re-used, as the code for it was logical across the board.

## Changelog Technical Details
+ Extracted some re-usable functionalities from `AnimatedItemStackRenderable`. - Daveed
+ Refactored `ItemStackRenderable` and `AnimatedItemStackRenderable` to have structured configs. - Daveed

PR #5341, opened 2026-03-07
Backend: Add Alpha to Item Rendering

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/5274

## What
Adds a 0-1 alpha/transparency param to itemstack rendering.

+ Added alpha param for item stack rendering. - Daveed


DavidArthurCole:
@hannibal002 This should be another quick one if you wanna knock it out; lots of dependents.

hannibal002 (commented):
whats the point of this? i dont see the benefit. please explain.

also the mixin to addElementToMesh looks like it is for every item rendered, not only skyhanni. is this intended?

also, this behaviour does not have a toggle option, and doesnt have a user facing changelog.

big question marks for me

hannibal002 (changes requested):

DavidArthurCole:
> whats the point of this? i dont see the benefit. please explain.
> 
> also the mixin to addElementToMesh looks like it is for every item rendered, not only skyhanni. is this intended?
> 
> also, this behaviour does not have a toggle option, and doesnt have a user facing changelog.
> 
> big question marks for me


Alpha = opacity for the item. This allows us to render semi transparent items. Look at #5342:

image

That is an item that has a non-1 alpha value.

hannibal002:
so this pr is just backend because it doesnt change user facing stuff yet, it just prepares us for the future to allow this behaviour?

and whats about the scope of this behaviour? dont we accidentally change opacity of non skyhanni items as well? and we still want a "main toggle" option for this imo

DavidArthurCole:
> so this pr is just backend because it doesnt change user facing stuff yet, it just prepares us for the future to allow this behaviour?

Yes. It is just setting us up for the ability to do this.

> and whats about the scope of this behaviour? dont we accidentally change opacity of non skyhanni items as well?

No. It goes based on the alpha value that is passed during rendering, which is the always `-1` from Minecraft, so this only affects our own Blits.

>  and we still want a "main toggle" option for this imo

For a rendering option? This is passed in during construction of a single renderable state:

image

It's not like this affects every single item we render.

hannibal002 (approved):

PR #5343, opened 2026-03-07
Fix + Backend: AbstractRepo V2

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/5339

## What
Overhauled repo and integrated it with the new coroutine system from 5339, fixed a couple bugs causing slow and colliding IO ops.

## Changelog Fixes
+ Fixed issues with backup repo not loading. - Daveed

+ Added lots more suspension usage in repo for efficient IO. - Daveed


src/main/java/at/hannibal2/skyhanni/data/repo/RepoFileSystem.kt
```diff
@@ -37,46 +42,49 @@ sealed interface RepoFileSystem {
...
-
-                if (this@RepoFileSystem is DiskRepoFileSystem) {
-                    val outPath = root.toPath().resolve(relative).normalize()
-                    if (!outPath.startsWith(root.toPath())) {
-                        throw RuntimeException(
-                            "SkyHanni detected an invalid zip file. This is a potential security risk, " +
-                                "please report this on the SkyHanni discord.",
-                        )
+        coroutineConfig: CoroutineConfig
+    ): Boolean = coroutineConfig.withIOContext().async {
```
hannibal002: why do we call async here? this function is already suspend and only gets  called via withIOContext

src/main/java/at/hannibal2/skyhanni/data/repo/RepoFileSystem.kt
```diff
@@ -124,45 +132,56 @@ class MemoryRepoFileSystem(private val diskRoot: File) : RepoFileSystem, Disposa
...
         val entries = storage.entries.toList()
         progress.innerProgressStart(entries.size)
-        entries.parallelStream().forEach { (relativePath, bytes) ->
-            progress.innerProgressStep()
-            val out = base.resolve(relativePath)
-            Files.write(out, bytes)
-        }
+        progress.update("writing entries")
 
+        coroutineScope {
```
hannibal002: launchUnScoped and courutien scope: why do we need this inside the suspend function? this feels like causes of silent errors, improper task cancellation, etc.

hannibal002 (changes requested):
also please confirm/review my repoCoroutineConfig change

DavidArthurCole:
> also please confirm/review my repoCoroutineConfig change

your change looks good - I needed to do a holistic cleanup of the coroutines, which I've now done. It should make a lot more sense now.

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
```diff
@@ -215,28 +229,33 @@ abstract class AbstractRepoManager<E : AbstractRepoReloadEvent> {
...
+                    return@launch
+                }
             }
-            loaded.set(true)
-            reloadRepository(progress)
-        }
-        job.invokeOnCompletion {
-            if (!loaded.get()) {
-                progress.end("reached timeout")
+            if (!repoDirectoryHasContent()) {
```
hannibal002: i think this should be replaced to a else if, so that it only gets checked and run when repo auto update is false above

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
```diff
@@ -616,7 +617,8 @@ object ItemUtils {
     var bazaarOverrides = mapOf<String, String>()
         private set
 
-    private data class BazaarOverride(
+    @KSerializable
```
hannibal002: this change here looks unnecessary. can we revert it or does this actually do something?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
DavidArthurCole: It actually does something, yes - `KSerializable` allows the usage of `by lazy` in these data classes, both in the class itself, and GSON "generics".

Because we're loading a list of complex objects, and we don't have a separate type adapter, the Serialization annotation is needed.

hannibal002: i dont think thats correct. Why would `by lazy` work differntly if the `KSerializeable` annotation is added or not? have you tested it without?

DavidArthurCole: @hannibal002  Idk what you want me to say to that. The KSerializable annotation marks a class as being handled by the `KotlinTypeAdapterFactory` in `KSerializable.kt`, rather than deferring to base GSON methods. If GSON tries to unfurl something while it has lazy props, they will fail.

https://discord.com/channels/997079228510117908/1117409804382642296/1405514540342579350
> You can't use by lazy here. Since that is a GSON does not fill in the delegates. (At least it did do that, idk if [david] fixed that with KSerilzable). That is why I made the workaround of doing a lazy myself.

hannibal002 (changes requested):
i removed the annotation, and it still works just fine.

Do i need to change more code to properly test the "missing annotation" case here, or am i missing something entirely?

Or can it be that the annotation does not belong to the `BazaarOverride` class at all, since we don't access it with `by lazy`, and since we only read it as any other class via `event.getConstant`?

Image

Image

DavidArthurCole:
> Do i need to change more code to properly test the "missing annotation" case here, or am i missing something entirely?
> 
> Or can it be that the annotation does not belong to the `BazaarOverride` class at all, since we don't access it with `by lazy`, and since we only read it as any other class via `event.getConstant`?

This one was factored out in the last repo PR, and merged in here. So, it can be removed in this specific case. Larger context above regarding the `by lazy { }` still holds, however.

hannibal002 (changes requested):
please remove it then.

hannibal002 (approved):

PR #5348, opened 2026-03-07
Fix: Island Change Event not firing when it should've

Fazfoxy:
## What
yeah I did expect that one to break stuff

## Changelog Fixes
+ Fixed some features not working when changing from 1 island to the same island type. (e.g. Hub -> Hub) . - Fazfoxy


Fazfoxy:
This event is more cooked on beta than expected, converted to draft for a better temporary solution

PR #5349, opened 2026-03-07
Fix: Revert HypixelData Changes from 5270 due to event issues

Fazfoxy:
## What
This seems better than slowly working out all the backend issues with a better solution

## Changelog Fixes
+ Fixed some island features not working after server swapping. - Fazfoxy
    * This is a temporary fix so SHO AutoLoad & Rift Motes session will not work again.


Fazfoxy:
nopo decided to send it with 5348 so, fuck it we ball with less testing ig

PR #5352, opened 2026-03-07
Backend: Skull Textures in ItemUtils & Detekt Rule

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/632

## What
Fixes the detekt rule so it actually finds strings with skull textures in them, and fix a couple instances in ItemUtils.

<details>
<summary>Images</summary>

image

</details>

+ Fixed detekt rule that flags skull textures in strings. - Daveed


src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
```diff
@@ -896,34 +896,31 @@ object ItemUtils {
...
+
+    private val coinSkulls by lazy {
+        listOf(COIN_TEXTURE_1, COIN_TEXTURE_2, COIN_TEXTURE_3).mapIndexed { index, texture ->
+            texture to createSkull("<placeholder>", COIN_TEXTURE_UUIDS[index], texture)
+        }.toMap()
+    }
+
     // Taken from NEU
-    // TODO add cache
     fun getCoinItemStack(coinAmount: Number): ItemStack {
```
hannibal002: this function will cache the renamed entity, this causes the previous returned item stack to have their name changed to the new call.

e.g. when creating two coin item stacks in the same frame, one with 1k, one with 50k, in that order. that means that boht item stacks have the custom item name set to "50k coins", even for the 1k one.

as a solution, maybe we should copy the result of the when block, or add a proper cache per coin number.

hannibal002 (changes requested):
i hope this is clear what i mean. ask if thats unclear

src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
DavidArthurCole: I'm a bit weary about an un-flushed cache in this regard, since there can be so many distinct calls to this with different numbers, but I've added a time limited cache, and have updated the map cache to copy the skulls, so the conflict you reference should also be addressed.

hannibal002 (changes requested):
the cache impl doesn't work.

<details>
<summary>proof</summary>

Image

Image

```
[23:21:05] [DefaultDispatcher-worker-16/INFO] (Minecraft) [STDOUT]: cache miss: computing skull for 1000
[23:21:05] [DefaultDispatcher-worker-16/INFO] (Minecraft) [STDOUT]: cache miss: computing skull for 1000
[23:21:05] [DefaultDispatcher-worker-16/INFO] (Minecraft) [STDOUT]: cache miss: computing skull for 1000
```

</details>

<details>
<summary>why</summary>
The Kotlin stdlib's `getOrPut` is an inline function. Because it's inline, return inside the lambda is a non-local return: it jumps out of `getCoinItemStack` entirely, skipping the `put(key, answer)` line. The value is returned to the caller but never stored in the cache.
</details>

DavidArthurCole:
> the cache impl doesn't work.

good catch, thanks. fixed now

hannibal002 (approved):

PR #5364, opened 2026-03-09
added /shmobhighlight command.

HacktheTime:
fixed invisible base entities like rats etc not being highlightable via sh.

this is submitted as a temporary proof of concept and that wants the rendering fixed but i cant be fucked to make the rendering part stand up to the sh quality standard cus fuck rendering.(rendering fix for mobs like rats done via gpt)

exclude_from_changelog


hannibal002:
stale

PR #5366, opened 2026-03-10
Fix: Advert Blocking

DavidArthurCole:
## What
Changelog.

## Changelog Fixes
+ Fixed advertisement blocking in Custom Tab. - Daveed


hannibal002:
the old code returns `null` directly. then `mapNotNull` filters them out, leaving a `[]` map. when trying to deconstruct to `(header, footer)` this will throw an error.

now, we properly return an empty list per list, returning with a `[empty list, empty list]`, and this works without errors.

hannibal002 (approved):

PR #5368, opened 2026-03-10
Fix: Custom Scoreboard Cache

hannibal002:
## What
Currently, the custom scoreboard does two things every 250ms:
1. recalculating each line to a string
2. wrapping each string to a renderable

even though, most custom scoreboard elements only change once a second, many even updating way slower.

With this PR, we now cache the list of strings, and only run the "wrapping of strings to Renderables" if something changes.

to see this in action:

with the "show SkyBlock time" feature and its option "exact minutes" enabled, this updates around every 700ms.
with this patch,  there are two  cache hits per second.

without skyblcok time, without a timer active in the custom scoreboard, there is only one cache miss every 5ish seconds for my testings.

so overall, the amount of renderable creations saved is huge by this small change.

## Changelog Fixes
+ Improved Custom Scoreboard performance by caching rendered lines. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/gui/customscoreboard/CustomScoreboard.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/gui/customscoreboard/CustomScoreboard.kt
@@ -62,6 +62,8 @@ object CustomScoreboard {
 
     private var dirty = false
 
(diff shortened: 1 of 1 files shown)


jani270:
thank you for staying afk in game, so i could use you as a test object

PR #5370, opened 2026-03-10
Fix: Rendersystem error in SackApi

lunaynx:
## What
Not tested, but based on stack traces I've seen this should fix it.

## Changelog Fixes
+ Fixed sometimes getting an error in chat / disconnect when opening sacks. - Luna


hannibal002:
this is just a workaround. instead, we should investigate the proper bug to not cause confusing code. So, for now, at least please add a comment in the code why this exists and a todo that this needs to get properly fixed eventually.

hannibal002 (changes requested):
above

lunaynx:
I don't think ensuring we are running on the render thread is necessarily a "workaround". The event is fired from the network thread because it's in a packet handler, and last time we tried to make it not do that, it caused freezes for some reason.

We should probably put the `runOrNextTick` inside `SackDisplay.update` instead though, that should be fine since it won't delay the execution if we're already on the correct thread. I'll also add a comment to explain, though we don't on most call sites. Maybe we should give the `runOrNextTick` method a more descriptive name, but that would be a separate PR.

hannibal002 (approved):

PR #5373, opened 2026-03-11
Backend: hide last 0 in version string

hannibal002:
## What
remove the last `.0` from the version string in the main config GUI title.
old: `7.7.0`
new: `7.7`

does not change the version string for the internal logic, the debug commands or error messaging

why? because we dont use the last digit practially, and this is the first step in getting users to know that there is no 0.1

https://github.com/SkyHanniStudios/SkyHanniChangelogBuilder/commit/2d3eebb0d89ff30dc2192e6397ca268ec9b412e1

next steps are the chagnelog builder to change the discord/github release messages (already done, see link above), and the final build jar file name (and in changelog builder!).

internally, id like to keep the three digits, maybe later also update the error messages? like, everything user facing? not sure.

+ Removed the third digit from user facing version string if zero. - hannibal2


hannibal002:
need further proper pro/con discussions

PR #5378, opened 2026-03-11
Backend: Deleting PetChangeEvent.kt

hannibal002:
## What
Proof of concept!

This is how we could delete all event files by moving the class into the same file that mainly fires the event. this will be a data/api clas in most cases. That way, we can free up a lot of files in the project, and in the end we even have one fewer root level package.

+ Moved `PetChangeEvent` to `CurrentPetApi.kt`. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/hotx/HotfData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/hotx/HotfData.kt
@@ -1,4 +1,4 @@
-package at.hannibal2.skyhanni.data.hotx
+package at.hannibal2.skyhanni.data.hotxfeel
 
 import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.api.HotfApi
--- a/src/main/java/at/hannibal2/skyhanni/data/hotx/HotfData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/hotx/HotfData.kt
@@ -1,4 +1,4 @@
-package at.hannibal2.skyhanni.data.hotxfeel
+package at.hannibal2.skyhanni.data.hotx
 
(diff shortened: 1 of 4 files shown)


src/main/java/at/hannibal2/skyhanni/data/hotx/HotfData.kt
```diff
@@ -1,4 +1,4 @@
-package at.hannibal2.skyhanni.data.hotx
+package at.hannibal2.skyhanni.data.hotxfeel
```
DavidArthurCole: Intended?

DavidArthurCole (changes requested):
I like this as a standard for API/Data-driving events - having one class with all of them defined might be necessary, but centralizing is good.

src/main/java/at/hannibal2/skyhanni/data/hotx/HotfData.kt
hannibal002: no, was a typo. thanks

hannibal002:
> I like this as a standard for API/Data-driving events

glad you like the idea! you think there are events we should not migrate? like, i assume in the end the currnet events package will be empty.

> having one class with all of them defined might be necessary, but centralizing is good.
i like to not have the events inside other classes, just in the same file as the data/api classes that call them. otherwise the listeners would have to add the outer class as well, thats ugly.

also, for stuff liike fishing events, where we have 3-4 events that are all only one line and are all getting fired from the same fishing api, they just become 4 lines above the class fish api line. no real new file/class needed to house them.

once we have compelx features where many/multiline events are necessary, we can still think of extracting them into a local events file in that specifc package, right?

NopoTheGamer:
honestly not a fan personally

hannibal002:
note at self: stop coding at 4am

PR #5379, opened 2026-03-11
Backend: Deleting all fishing events

hannibal002:
## What
doing the same as of in  

- #5378

just with the fishing events.
As a test, i moved the events into their classes where they are also fired.
As alternative, we could also add a FishingEvents file that then holds all those events.

+ Moved all fishing events. - hannibal2


CalMWolfs:
whgy 👎

hannibal002:
note at self: stop coding at 4am

PR #5384, opened 2026-03-11
Backend: Unify Plugin + Library Versioning

DavidArthurCole:
## What
Our `.gradle.kts` files were all over the place with how they were doing dependency and plugin management. This PR extracts all version definitions to `lib.versions.toml` (where they should be).

+ Migrated all declared plugin + library versions to libs.version.toml. - Daveed
+ Unified separate gradle sub-projects to use dep resolution from libs.version.toml. - Daveed


sharedVariables/settings.gradle.kts
```diff
@@ -1,2 +1,10 @@
...
+dependencyResolutionManagement {
+    versionCatalogs {
+        create("libs") {
+            from(files("../gradle/libs.versions.toml"))
+        }
+    }
+}
+
+// Rest of this space has been intentionally left blank
 // It serves such that this folder is a new gradle workspace rather than being a subproject
```
NopoTheGamer: you dont need these comments anymore lol

NopoTheGamer (changes requested):

sharedVariables/settings.gradle.kts
DavidArthurCole: > you dont need these comments anymore lol

fixed

PR #5399, opened 2026-03-12
Fix: Island change event

hannibal002:
## What
the island change event got changed in the last few prs so that it wasnt getting sent properly, causing many unrelated looking issues.

related pr https://github.com/hannibal002/SkyHanni-REPO/pull/635

## Changelog Fixes
+ Fixed island change event not always being sent, causing many features to malfunction (incorrect scoreboard data, broken profit trackers, etc.). - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/mining/eventtracker/MiningEventTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/mining/eventtracker/MiningEventTracker.kt
@@ -197,6 +197,9 @@ object MiningEventTracker {
 
     @HandleEvent
     fun onIslandChange(event: IslandChangeEvent) {
+        val old = event.oldIsland
+        val new = event.newIsland
+        ChatUtils.chat("IslandChangeEvent: $old -> $new")
         if (apiError) canRequestAt = SimpleTimeMark.now()
     }
 
--- a/src/main/java/at/hannibal2/skyhanni/features/mining/eventtracker/MiningEventTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/mining/eventtracker/MiningEventTracker.kt
@@ -197,9 +197,6 @@ object MiningEventTracker {
 
     @HandleEvent
     fun onIslandChange(event: IslandChangeEvent) {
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/features/mining/eventtracker/MiningEventTracker.kt
```diff
@@ -197,6 +197,9 @@ object MiningEventTracker {
 
     @HandleEvent
     fun onIslandChange(event: IslandChangeEvent) {
+        val old = event.oldIsland
+        val new = event.newIsland
+        ChatUtils.chat("IslandChangeEvent: $old -> $new")
```
lunaynx: Why in `MiningEventTracker`? Also, should be `ChatUtils.debug`.

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/features/mining/eventtracker/MiningEventTracker.kt
hannibal002: forgot to remove this entirely. thanks!

Fazfoxy:
> Fixed many small error messages. - hannibal2

can we at least make a real-ish changelog more akin to 

>Fixed island Detection not working after server swapping. - Hannibal
>This Fixes many issues such as incorrect scoreboard lines, missing profit trackers or graph errors
>This is a temporary fix so SHO AutoLoad & Rift Motes session will not work again.

Since, for all I'm aware, your'e gonna re-introduce the previous island wiping thing that makes these not work, and it's gonna be noticeable to users & we should list some of the more obvious issues that people will have noticed in the last 2 or 3 betas.

hannibal002:
> Since, for all I'm aware, your'e gonna re-introduce the previous island wiping thing that makes these not work, and it's gonna be noticeable to users & we should list some of the more obvious issues that people will have noticed in the last 2 or 3 betas.

i tried to replicate either of those 2 bugs, i could not reproduce them.
let me know if you know how how to reproduce the bugs, then ill look into them again

PR #5401, opened 2026-03-12
Backend: ClickType Cleanup

lunaynx:
## What
Cleaned up the `ClickType` enums. Makes code less ambiguous and helps with 26.1 migration.

+ Renamed `data.ClickType` to `InteractClickType`. - Luna
+ Removed `GuiContainerEvent.ClickType` in favor of Minecraft's `ClickType` enum. - Luna


src/main/java/at/hannibal2/skyhanni/data/WorldClickType.kt
```diff
@@ -1,6 +1,6 @@
 package at.hannibal2.skyhanni.data
 
-enum class ClickType {
+enum class WorldClickType {
```
DavidArthurCole: I would float changing `WorldClickType` to `InteractClickType`. Most of the `World[...]` classes we have are aimed specifically at real World rendering, handling, etc., whereas our internal enum is also used for clicking an entity, a block, etc.

src/main/java/at/hannibal2/skyhanni/data/InteractClickType.kt
DavidArthurCole, line 1: IntelliJ hates open-source and won't let me view this PR in-IDE, so I'm not sure of the usage implications, but this being its own top-level class feels a bit silly for 2 enum entries.

DavidArthurCole (changes requested):
Changes look mostly good. Couple small things.

src/main/java/at/hannibal2/skyhanni/data/InteractClickType.kt
lunaynx: Yeah, it probably shouldn't be, but I didn't really know where to move it.

lunaynx:
Already got merged with the 26.1 port, closing.

PR #5400, opened 2026-03-12
Backend: Cleanup + Overhaul Tracker/Config Inheritance

DavidArthurCole:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/5388
- https://github.com/hannibal002/SkyHanni/pull/5343
- https://github.com/hannibal002/SkyHanni/pull/5380
- https://github.com/hannibal002/SkyHanni/pull/5403
- https://github.com/hannibal002/SkyHanni/pull/5385

## What
This "started" with the Bucketed trackers; the `object` for the feature class was made to extend the tracker, rather than delegating to a `tracker` val. This PR "finishes" that. All trackers now extend the tracker class, and use inherited config typing for others.

There's more to say, but still work to do.

## Changelog Technical Details
+ Overhauled Tracker classes to extend the tracker they implement. - Daveed

PR #5404, opened 2026-03-13
Backend: Improved render event documentation

rueblimaster:
This is the replacement of #4638 

+ Added and improved documentation of the render events. - rueblimaster

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/events/GuiRenderEvent.kt
+++ b/src/main/java/at/hannibal2/skyhanni/events/GuiRenderEvent.kt
@@ -6,7 +6,7 @@ import net.minecraft.client.gui.GuiGraphics
 sealed class GuiRenderEvent(context: GuiGraphics) : RenderingSkyHanniEvent(context) {
 
     /**
-     * Renders only while inside an inventory
+     * Renders only while inside an inventory.
      * This event does not render on signs.
      * Use ScreenDrawnEvent instead.
      * Also, ensure you do not render with this event while in a sign, as it will override ScreenDrawnEvent.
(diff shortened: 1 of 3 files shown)


DavidArthurCole (changes requested):
I'm not sure how this is a follow-up for #4638, since it seems like most of the actual changes from that one are gone here, but:
[#5400](https://github.com/hannibal002/SkyHanni/pull/5400)
Look at RenderDisplayConfig and the changes in turn made to RenderDisplayHelper.

On this PR:
I like the idea of adding more documentation to our events, especially ones that are ambiguous as to what they're targeting precisely, however I don't think the changes made here are "necessary" in their current state.

If you are going to document them (the events), please do so following KDocs standards:
https://kotlinlang.org/docs/kotlin-doc.html

Examples of this can be found across the repo.

rueblimaster:
@DavidArthurCole 
fair enough, it's more of a replacement

Also nice change you made with the RenderDisplayConfig!

I changed the documentation to be more aligned with the KDoc standards, is this how you imagined it?

DavidArthurCole (changes requested):
Better after the change, but I still think there is room for improvement, and potentially documenting more of the undocumented events that we have in the repo.

When I referenced KDocs, I was more intending an approach like this:

Image

Where, we are functionally documenting the parameters that are used, and any limitations or special cases that need to be handled.

Doing it this way allows IntelliJ (and other IDEs/processors) to digest the KDocs, for IDE hover tips, etc:

Image

rueblimaster:
@DavidArthurCole 
fair enough. Tho I don't think I know enough about the experiments (or the context being given around) to write it like that?

src/main/java/at/hannibal2/skyhanni/events/GuiRenderEvent.kt
```diff
@@ -6,20 +6,22 @@ import net.minecraft.client.gui.GuiGraphics
 sealed class GuiRenderEvent(context: GuiGraphics) : RenderingSkyHanniEvent(context) {
 
     /**
-     * Renders only while inside an inventory
-     * This event does not render on signs.
-     * Use ScreenDrawnEvent instead.
-     * Also, ensure you do not render with this event while in a sign, as it will override ScreenDrawnEvent.
```
NopoTheGamer, line 12: i dont get it, you rewrote the comment but i dont think its any clearly or explains anything differently

src/main/java/at/hannibal2/skyhanni/events/GuiRenderEvent.kt
```diff
@@ -6,20 +6,22 @@ import net.minecraft.client.gui.GuiGraphics
...
+     * Render event for inside inventories.
+     *
+     * Renders only while inside an inventory.
+     * Will override ScreenDrawnEvent but not render anything inside of a sign!
      */
     class ChestGuiOverlayRenderEvent(context: GuiGraphics) : GuiRenderEvent(context)
 
     /**
-     * Renders always, and while in an inventory it renders a bit darker, gray
+     * Render event for everywhere else.
```
NopoTheGamer, line 17: i think this is just worse

NopoTheGamer (changes requested):

hannibal002:
stale

PR #5407, opened 2026-03-13
Backend: Use proper KDoc tags for regex tests

lunaynx:
## What
We currently use raw strings in KDoc comments to define regex tests for patterns. This is not proper, and leads to it leaking into the documentation shown by IntelliJ and other tools – not even preserving the line breaks. This PR changes regex tests to use proper KDoc tags.

+ Changed syntax used to define regex tests. - Luna
    * `REGEX-TEST:` is now `@regexTest`.
    * `REGEX-FAIL:` is now `@regexFail`.
    * `WRAPPED-REGEX-TEST:` is now `@regexTestWrapped`.
    * Added `@regexFailWrapped`.
    * Leading and trailing spaces are now stripped from regex tests, therefore you must now always use `@regex{Test,Fail}Wrapped` if you are trying to intentionally include those in test strings.


DavidArthurCole (approved):
All looks good to me, but the raw diff size is a bit horrifying. Any chance we can split this into two branches? One for the logic changes, and another for changing every `REGEX-TEST` in the codebase 🥺

lunaynx:
Will split this up into two stages per the suggestion, at first supporting both formats.

lunaynx:
Closing – will make a separate PR to add support for the two syntaxes in parallel and then a follow-up PR to migrate everything once that is merged.

PR #5413, opened 2026-03-15
Fix: enter keybind around inventory open/close

hannibal002:
## What
When a screen (e.g. chat) was closed by pressing Enter, `isKeyClicked()` could immediately return true for Enter in the same tick since the lock was never set while the screen was open. Locks Enter in `lockedKeys` when any screen closes and Enter is still held.

## How this came up
i use "enter" key as keybind for the "exit key option in graph editor.
meaning, when i then run the command `/shgraphloadthisisland` there is a race condition that the exit action happens immediately after the graph editor opens, immediately gets disabled.

<details>
<summary>Images</summary>

image

</details>

## Changelog Fixes
+ Fixed Enter keybind firing incorrectly after closing chat or inventory. - hannibal2


NopoTheGamer:
this seems like a hack? wouldnt the same issue happen with any other keybind that closes guis?

hannibal002:
yes, this is a hack.
idk how a proper solution looks like tho.
like, what other keybinds are there that close guis? i only think of escape, and escape alr causes more issues when used with keybinds (and afaik isnt even possible to get set as an keybind).

so, as long as there is no "right" solution, i want this hack for the enter key.

NopoTheGamer:
> what other keybinds are there that close guis?
off the top of my head, esc, space, enter, inventory keybind

hannibal002:
@NopoTheGamer 
You're right, the same issue could theoretically happen with the inventory keybind.
However, locking all held keys on screen close would be risky (e.g. walking while a GUI closes server-side would lock movement keys).
A general solution needs more thought.
This targeted fix covers the main reported case for now.

PR #5414, opened 2026-03-15
Backend: Make SkyHanniEvent.post return itself

lunaynx:
## What
Currently, we have code like this:

```kotlin
if (MyEvent().post()) {
    // do something
}
```

This code is confusing and it is not clear at a glance that `.post()` returns a boolean indicating whether the event was cancelled.

Now, `.post()` returns the `SkyHanniEvent` itself – which conveniently already has an `isCancelled` field exposed. This makes code much clearer.

+ Changed SkyHanniEvent.post to now return the event itself instead of a boolean. - Luna
    * Callers now check cancellation via isCancelled.


src/main/java/at/hannibal2/skyhanni/api/event/EventHandler.kt
```diff
@@ -23,11 +23,11 @@ class EventHandler<T : SkyHanniEvent> private constructor(
         listeners.any { it.receiveCancelled },
     )
 
-    fun post(event: T, onError: ((Throwable) -> Unit)? = null): Boolean {
+    fun post(event: T, onError: ((Throwable) -> Unit)? = null) {
```
DavidArthurCole, line 26: To make this intention more clear
```suggestion
    fun post(event: T, onError: ((Throwable) -> Unit)? = null): T {
```

src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvent.kt
```diff
@@ -15,14 +15,15 @@ abstract class SkyHanniEvent protected constructor() {
...
-            val result = SkyHanniEvents.getEventHandler(javaClass).post(this, onError)
+            SkyHanniEvents.getEventHandler(javaClass).post(this, onError)
             DrawContextUtils.clearContext()
-            return result
+            return this
         }
-        return SkyHanniEvents.getEventHandler(javaClass).post(this, onError)
+        SkyHanniEvents.getEventHandler(javaClass).post(this, onError)
+        return this
     }
```
DavidArthurCole:
```suggestion
private fun prePost(onError: ((Throwable) -> Unit)?): SkyHanniEvent = apply {
    (this as? Rendering)?.let { DrawContextUtils.setContext(it.context) }
    SkyHanniEvents.getEventHandler(javaClass).post(this, onError)
    if (this is Rendering) DrawContextUtils.clearContext()
}
```

src/main/java/at/hannibal2/skyhanni/api/minecraftevents/RenderEvents.kt
```diff
@@ -55,61 +56,50 @@ object RenderEvents {
     // GameOverlayRenderPreEvent
     // todo need to post the rest of these, sadly fapi doesn't have the same layers as 1.8 does
     @JvmStatic
-    fun postHotbarLayerEventPre(context: GuiGraphics): Boolean {
-        return GameOverlayRenderPreEvent(context, RenderLayer.HOTBAR).post()
-    }
+    fun postHotbarLayerEventPre(context: GuiGraphics): SkyHanniEvent =
```
DavidArthurCole: If you add the signatures to these, please make them match the event type instead of generically, `SkyHanniEvent`

src/main/java/at/hannibal2/skyhanni/data/ChatManager.kt
```diff
@@ -133,7 +133,7 @@ object ChatManager {
                 trimmedMessage,
                 trimmedMessage.split(" "),
                 originatingModContainer,
-            ).post()
+            ).post().isCancelled
```
DavidArthurCole, line 143: This isn't specifically to this line, but logically, this does not read how I think it's intending to.

I read this as "is posting the event cancelled".
I think the long-term solution to this would be to rename `post()` to `postAndGet()`, or to make a `postAndIsCancelled()` function to use when that's the only property cared about.

Either of those changes would affect many hundreds of files, I'm sure, but maybe worth a todo on the event class?

src/main/java/at/hannibal2/skyhanni/features/misc/compacttablist/TabListRenderer.kt
```diff
@@ -44,8 +44,8 @@ object TabListRenderer {
     private var isPressed = false
     private var isTabToggled = false
 
-    @HandleEvent(onlyOnSkyblock = true, priority = HandleEvent.LOWEST)
-    fun onRenderOverlay(event: GuiRenderEvent.GuiOverlayRenderEvent) {
+    @HandleEvent(GuiRenderEvent.GuiOverlayRenderEvent::class, onlyOnSkyblock = true, priority = HandleEvent.LOWEST)
+    fun onRenderOverlay() {
```
DavidArthurCole: Please rename to `onGuiRenderOverlay`

DavidArthurCole (approved):
Looks fine. Just some minor changes.

src/main/java/at/hannibal2/skyhanni/api/event/EventHandler.kt
lunaynx: This wouldn't work because this particular method doesn't actually return anything.

src/main/java/at/hannibal2/skyhanni/data/ChatManager.kt
lunaynx: I guess, but I think it's better than the status quo.

I remember with the old `@SubscribeEvent` interface or something it used to be `.postAndCatch()`, though I'm not sure that was very clear either.

In theory, clarity at call sites could be maximized by doing `var event = SomeEvent(...)` and then `event.post()` and `event.isCancelled`, though that would introduce extra code lines. Maybe we can consider if it's worth doing later.

src/main/java/at/hannibal2/skyhanni/features/misc/compacttablist/TabListRenderer.kt
lunaynx: I swear I did that

src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvent.kt
```diff
@@ -15,14 +15,11 @@ abstract class SkyHanniEvent protected constructor() {
...
-            val result = SkyHanniEvents.getEventHandler(javaClass).post(this, onError)
-            DrawContextUtils.clearContext()
-            return result
-        }
-        return SkyHanniEvents.getEventHandler(javaClass).post(this, onError)
+    private fun prePost(onError: ((Throwable) -> Unit)?): SkyHanniEvent = apply {
+        (this as? Rendering)?.let { DrawContextUtils.setContext(it.context) }
+        SkyHanniEvents.getEventHandler(javaClass).post(this, onError)
+        if (this is Rendering) DrawContextUtils.clearContext()
+        return this
```
hannibal002: the line is unnecessary

src/main/java/at/hannibal2/skyhanni/api/minecraftevents/RenderEvents.kt
```diff
@@ -132,7 +121,5 @@ enum class RenderLayer {
     DEBUG,
     HELD_ITEM_TOOLTIP,
     ACTION_BAR,
-
-    // Not a real forge layer but is used on modern Minecraft versions
```
hannibal002, line 135: is this no longer the case?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/api/minecraftevents/RenderEvents.kt
lunaynx: We don't use Forge anymore, we're modern-only, so...

hannibal002 (changes requested):
dektekt:

* MessageStore.kt:1:1: File should not be empty [NoEmptyFile]
* HotfApi.kt:16, 17: Annotation target warnings for value parameter
* HotmApi.kt:112, 113, 153: Annotation target warnings for value parameter
* GuiOptionEditorBlocked.kt:38, 42: Overrides a deprecated member
* GuiOptionEditorHidden.kt:12, 16: Overrides a deprecated member
* GuiPositionEditor.kt:64: Supertype parameter naming conflicts for 'mouseX' and 'mouseY'
* Graph.kt:65:22: Overrides a deprecated member
* BetterContainers.kt:141:13: Condition is always 'true'
* ShiftClickEquipment.kt:24:13: Condition is always 'false'
* FeatureToggleProcessor.kt:61:103: Class is not recommended for use in Kotlin
* FeatureToggleProcessor.kt:62:50: Unchecked cast of 'Any!' to 'Property'
* CopyItemCommand.kt:33:108: Redundant call of conversion method
* SafeItemStack.kt:52:42: 'field item: Holder?' is deprecated
* GuiScreenUtils.kt:29:74: Redundant call of conversion method

src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvent.kt
hannibal002: the line is still unecessary

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvent.kt
hannibal002: still there

hannibal002 (changes requested):
bump

hannibal002 (approved):

PR #5380, opened 2026-03-11
Backend: Type Adapter Enums + Cleanup

DavidArthurCole:
## What
The Reforge parser was a bit overly complex, and could be flattened down into using gson-natives if it uses a LinkedHashMap instead of an Enum Map, since gson can't infer typing on the latter. The only other usage of a custom adapter was in `Graph`, which needed (and still needs) some more work.

End-goal of enumifying it is that by virtue of putting it in the adapters file, it will now be registered.

Cleaned up some of the statements by extracting more data into enum definitions.

## Changelog Technical Details
+ Refactored `SkyHanniTypeAdapters` into an enum class. - Daveed
+ Started cleanup of `Graph` and its related classes. - Daveed
+ Simplified Reforge data loading from NEU. - Daveed

PR #5417, opened 2026-03-15
Improvement: sligthly better pahfind render behaviour

hannibal002:
## Dependency
- #5380

## What
cleanup the pathfind render logic, in code and visually.

<details>
<summary>Images</summary>

image
image

</details>

## Changelog Improvements
+ Improved pathfind lines to look prettier. - hannibal2


DavidArthurCole (approved):
Code looks okay for the most part. I've got a lot of clean up to do in WorldRenderUtils, but nothing inherently wrong in this PR.

Just a couple visual artifacting things I noticed:

<details>
<summary>Strange "jumps" at elevation difference</summary>
Image
</details>

<details>
<summary>Over-curved paths at short distances</summary>
Image
</details>

hannibal002:
both are interesting findings. any chance you can recreate them again, because: for the first i assume in f5 view this looks "more correct", and for the second i think the nodes are just not that close to the house and are in fact at the at the npc

DavidArthurCole:
The first one, I noticed on and off while moving, and might not be something that even needs to be solved for unless there's an easy fix.

Second one, I'm good with calling it a node issue, however it did "shallow out" as I got further away from the node, so I'm not sure that's the _whole_ story. Can recreate if you need me to.

hannibal002:
1. cant tell how easy the fix is, it depends how the path looks like in the third person view here.

2. please explain what you mean with shalllow out. and you go further away in what direction? you go backwards in away from the direction the path points you towards? like you look directly at it then press S ?

PR #5418, opened 2026-03-16
Backend: Minor Cleanup

DavidArthurCole:
## What
In preparation for moving to the plugin and discontinuing use of the annotation param, deprecates them (I don't think it actually does anything but 🤷), and started the cleanup process around standardizing primary names for functions.

Config had some vars that really should not have been.

+ Started defining `PrimaryFunctions` for events missing them. - Daveed
+ Migrated all top-level configs in `SkyHanniConfig` from `var` to `val`. - Daveed
+ Renamed all `ChestGuiOverlay` render handlers to `onChestGuiRender`. - Daveed

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/events/GuiRenderEvent.kt
+++ b/src/main/java/at/hannibal2/skyhanni/events/GuiRenderEvent.kt
@@ -1,8 +1,10 @@
 package at.hannibal2.skyhanni.events
 
 import at.hannibal2.skyhanni.api.event.RenderingSkyHanniEvent
+import at.hannibal2.skyhanni.skyhannimodule.PrimaryFunction
 import net.minecraft.client.gui.GuiGraphics
 
+@PrimaryFunction("onGuiRender")
 sealed class GuiRenderEvent(context: GuiGraphics) : RenderingSkyHanniEvent(context) {
 
     /**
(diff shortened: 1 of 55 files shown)


src/main/java/at/hannibal2/skyhanni/events/GuiRenderEvent.kt
```diff
@@ -11,16 +13,21 @@ sealed class GuiRenderEvent(context: GuiGraphics) : RenderingSkyHanniEvent(conte
      * Use ScreenDrawnEvent instead.
      * Also, ensure you do not render with this event while in a sign, as it will override ScreenDrawnEvent.
      */
+    // TODO rename primary function to not be disambiguated with the event
+    //  rename to "onChestGuiRender" or similar
+    @PrimaryFunction("onBackgroundDraw")
     class ChestGuiOverlayRenderEvent(context: GuiGraphics) : GuiRenderEvent(context)
```
hannibal002, line 17: DrawBackgroundEvent exists. rn. this is a duple primary function case. also, should throw an error if two events have the same primaryfunction name in dev env

hannibal002 (changes requested):
lokks good! 
i can even confirm that `/shconfig reset config.garden` still works.
(now i need to reimport the old config in dev env again 🔥

src/main/java/at/hannibal2/skyhanni/events/GuiRenderEvent.kt
DavidArthurCole: The cleanup for this would be monumental - hence the TODO.
I can do it if you want, but it's going to balloon this PR.

DavidArthurCole: Sent in DMs, but will also send here, the plugin does this too 🙂, so there's that.
image

DavidArthurCole: 38 instances of it changed. `onBackgroundDraw` => `onChestGuiRender`

hannibal002 (approved):

PR #5424, opened 2026-03-16
Improvement + Backend: Removed patcher

hannibal002:
## What
patcher doesn't exist on 1.21, so no need to keep outdated code around.

Removed the entirety of the "patcher fixes" logic, and renamed the "patcher send coordinates" feature to drop the "patcher" part.

## Changelog Improvements
+ Renamed feature "Patcher Coords Waypoints" to "Coords Waypoints". - hannibal2

+ Removed leftover files from patcher support. - hannibal2


Fazfoxy:
> Removed the entirety of the "patcher fixes" logic, and renamed the "patcher send coordinates" feature to drop the "patcher" part.

idk if we don't have a different name for the format, they should still be referred to as patcher coordinates imo?

hannibal002:
we dont "have" to use the format `x: 1, y: 2, z: 3`, we could change that to anything else anytime, e.g. listen to  `x:1 y:2 z:3`. also in general the format is too generic to give credits to the format to patcher for it imo.

DavidArthurCole (approved):
Looks fine, rest in piss patcher 🙏

PR #5428, opened 2026-03-16
Backend: renderRenderable no longer nullable

hannibal002:
## What
renderRenderable accepted a nullable string so that we early return then inside the function. this is a bad design overall.
since there were only like 20-30 cases, i fixed them all here and did some close enough cleanup.

+ Made `renderRenderable` require non-nullable strings now. - hannibal2


DavidArthurCole (approved):
I fear the conflicts this will cause me, but yeah the guard on the receiver was just weird

PR #5388, opened 2026-03-11
Backend: Cleanup + Migration of Moveable HUDs

DavidArthurCole:
## What
This started as a general cleanup branch, but I got into the weeds on a couple things. Tried to split commits by "area" to make it easier to track. A lot of the earlier changes were semantic/removing code that was never reached. See changelog below.

## Changelog Technical Details
+ Migrated Moveable HUD elements to new `moveablehud` package. - Daveed
    * Config + impl. classes were almost entirely duplicated code.
+ Removed every instance of `pushMatrix()` and `popMatrix()` that can be removed. - Daveed
+ Cleaned up lots of deprecated usages of functions. - Daveed

PR #5429, opened 2026-03-16
Backend: Render Utils pushPop

hannibal002:
## What
was in render utils, noticed deprecated stuff that was easily fixable.

ignore_from_changelog


NopoTheGamer:
isnt there a pr that alr does this?

DavidArthurCole:
Yes, #5388

DavidArthurCole (changes requested):
As nopo pointed out already, every change made here is already encapsulated in #5388

hannibal002:
yield to #5388

PR #5437, opened 2026-03-18
Improvement: Better TPS Counter

lunaynx:
## What
Significantly improved the TPS counter, meaning it should no longer show unrealistically high TPS values when the server is clearly lagging, and then sometimes show low values later when the server has already recovered.

Instead of looking for any and all packets received each tick, we now look specifically for ClientboundPingPacket (which fires our already existing ServerTickEvent). And instead of counting how many we get each second, we measure the average distance between those packets, resulting in a much more accurate measurement.

We take the average of the last 5 seconds, but if the server sends zero ping packets for a whole second, we reset the displayed TPS to 0.0, because otherwise it would take multiple seconds for the TPS display to change at all when the server is completely unresponsive.

Note that as of recently, Hypixel no longer delays sending ping packets until 10 seconds after a world switch, but rather starts sending them immediately. But I still kept a 5 second timer on world switch before showing the TPS to make sure we've gathered enough samples for accurate measurement and avoid displaying a low TPS value when the server is actually working fine.

I've been running this side by side with the old TPS counter implementation for the past few days and it seems pretty accurate.

Of course, if your connection to Hypixel is dying, this will show low TPS. There's not really anything we can do about that.

## Changelog Improvements
+ Significantly improved accuracy of server TPS measurement. - Luna

+ Added `tick` and `timeMark` fields to `ServerTickEvent`. - Luna
+ Added `Duration.inPartialMilliseconds`. - Luna
+ Added `Duration.roundedUpSeconds`. - Luna
    * Useful for countdowns, because `inWholeSeconds` will round down, often leading to undesirable behavior where, for example, your 5-second countdown starts at 4 and shows 0 for a full second before finishing.
+ Removed unused `Duration.inPartialYears`. - Luna
    * The precise definition of a year is context-dependent anyway, so it's better not to have this.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/TpsCounter.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/TpsCounter.kt
@@ -50,7 +50,7 @@ object TpsCounter {
     @HandleEvent
     fun onSecondPassed() {
         if (lastServerTick.passedSince() >= 1.seconds) {
-            ChatUtils.debug("No ping packets received for 1 second, clearing TPS data")
+            ChatUtils.debug("No server ticks detected for 1 second, clearing TPS data")
             tpsList.clear()
         }
 
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/TpsCounter.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/TpsCounter.kt
@@ -12,11 +12,17 @@ import at.hannibal2.skyhanni.events.minecraft.ServerTickEvent
 import at.hannibal2.skyhanni.features.misc.limbo.LimboTimeTracker
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.test.command.ErrorManager
-import at.hannibal2.skyhanni.utils.*
+import at.hannibal2.skyhanni.utils.ChatUtils
+import at.hannibal2.skyhanni.utils.DelayedRun
 import at.hannibal2.skyhanni.utils.RenderUtils.renderRenderable
+import at.hannibal2.skyhanni.utils.SimpleTimeMark
+import at.hannibal2.skyhanni.utils.SkyBlockUtils
+import at.hannibal2.skyhanni.utils.TimeUtils
 import at.hannibal2.skyhanni.utils.collection.TimeAndSizeLimitedCache
+import at.hannibal2.skyhanni.utils.inPartialMilliseconds
 import at.hannibal2.skyhanni.utils.renderables.Renderable
 import at.hannibal2.skyhanni.utils.renderables.primitives.text
+import at.hannibal2.skyhanni.utils.roundedUpSeconds
 import kotlin.time.Duration.Companion.seconds
 
 @SkyHanniModule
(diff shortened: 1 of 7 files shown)


src/main/java/at/hannibal2/skyhanni/features/misc/TpsCounter.kt
```diff
@@ -6,57 +6,65 @@ import at.hannibal2.skyhanni.config.ConfigUpdaterMigrator
...
-    private val minimumSecondsDisplayDelay = 10.seconds
+    private val WORLD_SWITCH_DELAY = 5.seconds
 
-    private var packetsFromLastSecond = 0
-    private val tpsList = mutableListOf<Int>()
-    private var hasRemovedFirstSecond = false
+    private val config get() = SkyHanniMod.feature.gui
 
-    private var hasReceivedPacket = false
+    // MSPT (Milliseconds Per Tick) = 1000 / TPS
```
DavidArthurCole: Rename to `msPerTickList`, should be self explanatory w/o comment imo

src/main/java/at/hannibal2/skyhanni/features/misc/TpsCounter.kt
```diff
@@ -66,70 +74,52 @@ object TpsCounter {
...
+                if (TimeUtils.isAprilFoolsDay) currentTps /= 2
+                append("%s%.1f".format(getColor(currentTps), currentTps))
             }
         }
-        display = Renderable.text("§eTPS: $text")
     }
 
-    private fun fixTps(tps: Double): Double {
-        return if (TimeUtils.isAprilFoolsDay) tps / 2 else tps
+    private fun updateDisplay() {
```
DavidArthurCole: Only used once in the file, you can inline this assignment

src/main/java/at/hannibal2/skyhanni/features/misc/TpsCounter.kt
```diff
@@ -66,70 +74,52 @@ object TpsCounter {
...
 
-    @HandleEvent(priority = HandleEvent.HIGHEST, receiveCancelled = true)
-    fun onPacketReceive(event: PacketReceivedEvent) {
-        hasReceivedPacket = true
-    }
-
-    @HandleEvent
-    fun onRenderOverlay(event: GuiRenderEvent.GuiOverlayRenderEvent) {
+    @HandleEvent(GuiRenderEvent.GuiOverlayRenderEvent::class)
+    fun onRenderOverlay() {
```
DavidArthurCole: Please rename this to `onGuiRenderOverlay` in prep for the upcoming unfuckery of ambigious PrimaryFunction names 🙏

src/main/java/at/hannibal2/skyhanni/utils/ServerTimeMark.kt
```diff
@@ -12,7 +12,7 @@ import kotlin.time.Duration
  * the server's tps instead of real time, and therefore are affected by server lag.
  */
 @JvmInline
-value class ServerTimeMark private constructor(val ticks: Long) : Comparable<ServerTimeMark> {
+value class ServerTimeMark(val ticks: Long) : Comparable<ServerTimeMark> {
```
DavidArthurCole: Please make this constructor `internal` instead of being fully open

DavidArthurCole (changes requested):
Few small changes, overall lgtm. Tested in game and it worked well across a few hubs and islands.

PR #5438, opened 2026-03-18
Fix: Greenhouse Growth Stage alert sending multiple times

lunaynx:
## What
Fixed Greenhouse Growth Stage alert sometimes sending multiple times due to a race condition by setting `hasNotified` (renamed from `beep`) to `true` *before* sending the notification. Also while I'm at it, refactored the code so that we aren't sending notifications in render code.

## Changelog Fixes
+ Fixed Greenhouse Growth Stage alert sometimes sending multiple times. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/greenhouse/GrowthCycle.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/greenhouse/GrowthCycle.kt
@@ -6,7 +6,6 @@ import at.hannibal2.skyhanni.data.IslandType
 import at.hannibal2.skyhanni.data.ProfileStorageData
 import at.hannibal2.skyhanni.events.GuiRenderEvent
 import at.hannibal2.skyhanni.events.InventoryUpdatedEvent
-import at.hannibal2.skyhanni.events.SecondPassedEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.ItemUtils.getLoreComponent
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/greenhouse/GrowthCycle.kt
```diff
@@ -59,23 +58,36 @@ object GrowthCycle {
...
     }
 
     @HandleEvent
-    fun onSecondPassed(event: SecondPassedEvent) {
+    fun onSecondPassed() {
         if (!config.showDisplay) return
 
         updateDisplay()
+
+        val nextCycle = storage?.nextCycle ?: return
```
DavidArthurCole, line 72:
```suggestion
        val nextCycle = storage?.nextCycle?.takeIf { it.isInPast() } ?: return
```

src/main/java/at/hannibal2/skyhanni/features/garden/greenhouse/GrowthCycle.kt
```diff
@@ -59,23 +58,36 @@ object GrowthCycle {
...
-            beep = false
+            hasNotified = false
             updateDisplay()
         }
     }
 
     @HandleEvent
-    fun onSecondPassed(event: SecondPassedEvent) {
+    fun onSecondPassed() {
         if (!config.showDisplay) return
```
DavidArthurCole, line 68:
```suggestion
        if (!config.showDisplay || hasNotified) return
```

DavidArthurCole (approved):
Looks good, just a couple flow pattern suggestions.

src/main/java/at/hannibal2/skyhanni/features/garden/greenhouse/GrowthCycle.kt
lunaynx: This actually subtly changes the logic and I fear it could break things if `updateDisplay()` doesn't get called, but I changed it to be an early return.

src/main/java/at/hannibal2/skyhanni/features/garden/greenhouse/GrowthCycle.kt
lunaynx: Unnecessary now since I early return right below.

PR #5440, opened 2026-03-18
Backend: Resettable on Private Classes

DavidArthurCole:
## What
Fixes the reflection we perform in `Resettable` so that it can properly function on private classes, and private constructors.

+ Fixed Resettable not working on private classes. - Daveed


hannibal002:
(createInstance woudl check this as well, the current impl didnt

PR #5448, opened 2026-03-20
Fix + Backend: Discord RPC/IPC

DavidArthurCole:
## What
Every single IPC lib for java is horrendously outdated, so I just wrote our own.

## Changelog Fixes
+ Fixed Discord Rich Presence not connecting. - Daveed

+ Removed dependency on `kdiscordipc`. - Daveed
+ Added SHDiscordIPC, and reworked DiscordRPCManager. - Daveed


src/main/java/at/hannibal2/skyhanni/features/misc/discordrpc/SHDiscordIPC.kt
```diff
@@ -0,0 +1,277 @@
...
+
+/**
+ * A lightweight Discord IPC client implementing the Rich Presence protocol over Discord's local IPC pipe.
+ *
+ * Manages pipe discovery, handshake, presence updates, and clean disconnection.
+ * Uses named pipes on Windows and Unix domain sockets on Linux/macOS.
+ *
+ * @param clientId The Discord application client ID (from the Discord Developer Portal).
+ */
+class SHDiscordIPC(private val clientId: Long) : Closeable {
```
NopoTheGamer: really not a fan of classes with "sh" in the name of them

src/main/java/at/hannibal2/skyhanni/config/features/misc/DiscordRPCConfig.kt
```diff
@@ -104,4 +106,12 @@ class DiscordRPCConfig {
...
     @ConfigEditorInfoText
     val credits = ""
+
+    @SkyHanniModule
+    companion object {
+        @HandleEvent
+        fun onConfigFix(event: ConfigUpdaterMigrator.ConfigFixEvent) {
+            event.move(128, "gui.discordRPC.showEliteBotButton", "gui.discordRPC.showEliteSkyBlockButton")
+        }
+    }
```
NopoTheGamer, line 116: can we not just use the config fix event thats already in discord rpc manager

NopoTheGamer (changes requested):

src/main/java/at/hannibal2/skyhanni/config/features/misc/DiscordRPCConfig.kt
DavidArthurCole: image

:(

DavidArthurCole:
go to hell, bot

src/main/java/at/hannibal2/skyhanni/config/features/misc/DiscordRPCConfig.kt
DavidArthurCole: I'll say this here again, despite the fact that we keep talking about this, because I'm gonna mark this convo as resolved, again, but;

Especially as we go towards more modular configs, that don't necessarily have a static home, the migrations being within the config class make more sense than trying to force it into a random file. If you have any reason against this apart from the "I don't like it," that I've gotten in the past when I've asked, I am all ears and am willing to try to find a solution.

NopoTheGamer (changes requested):
doesnt work in intellij or prism

build.gradle.kts
```diff
@@ -125,10 +125,8 @@ dependencies {
         mappings(target.mappingDependency)
     }
 
-    // Discord RPC client
-    includeImplementation(libs.discord.ipc)
-    include(libs.junixsocket.common)
-    include(libs.junixsocket.native)
+    include("com.kohlschutter.junixsocket:junixsocket-common:2.6.2")
```
hannibal002: please use libs, as everywhere else

src/main/java/at/hannibal2/skyhanni/features/misc/discordrpc/DiscordRPCManager.kt
```diff
@@ -155,115 +134,97 @@ object DiscordRPCManager {
...
+                ).let { add(it) }
+            },
         )
-    }
 
+        try {
+            client?.setActivity(presence)
+        } catch (e: DiscordIPCException) {
+            updateDebugStatus("Discord RPC disconnected: ${e.message}")
+            client = null
```
hannibal002, line 208: should probably fire scheduleRetry here again

src/main/java/at/hannibal2/skyhanni/features/misc/discordrpc/DiscordRPCManager.kt
```diff
@@ -155,115 +134,97 @@ object DiscordRPCManager {
...
+                    updatePresenceProgress?.end("update presence loop finished first run, not logging further updates")
+                    updatePresenceProgress = null
+                    delay(5.seconds)
+                }
             }
         }
     }
 
-    private suspend fun updatePresence(progress: ChatProgressUpdates?) {
+    private val skyCryptUrl get(): String =
```
hannibal002: should be a function, not a field

src/main/java/at/hannibal2/skyhanni/features/misc/discordrpc/DiscordRPCManager.kt
```diff
@@ -155,115 +134,97 @@ object DiscordRPCManager {
...
+        "https://sky.shiiyu.moe/stats/${PlayerUtils.getName()}/${HypixelData.profileName.firstLetterUppercase()}".addSkyHanniUtm()
+
+    private val eliteSbUrl get(): String =
```
hannibal002: same here

hannibal002 (changes requested):

meowora (commented):
Wouldn't it be better to make this an external library rather then embedding it?

DavidArthurCole:
> Wouldn't it be better to make this an external library rather then embedding it?

Idk. All of the RPC issues we've had in the past have stemmed from relying on deployed IPC libraries. I don't see much difference in making my own IPC library and publishing it, versus embedding it into SH.

I'm sure there are many parts of our code that could define their own distinct modules, but to what end?

build.gradle.kts
DavidArthurCole: Oop. These are entirely un-necessary now. We declare our own sockets. I'll remove.

meowora:
> > Wouldn't it be better to make this an external library rather then embedding it?
> 
> Idk. All of the RPC issues we've had in the past have stemmed from relying on deployed IPC libraries. I don't see much difference in making my own IPC library and publishing it, versus embedding it into SH.
> 
> I'm sure there are many parts of our code that could define their own distinct modules, but to what end?

I just don't think that embedding a system like this into the mod directly makes any sense at all.

hannibal002 (approved):

PR #5454, opened 2026-03-20
Improvement: Custom Keybinds with Mousemat/Rod

lunaynx:
## What
Added Squeaky Mousemat and fishing rod toggles for Garden custom keybinds, plus some code cleanup.

## Changelog Improvements
+ Added options to use custom keybinds in the Garden while holding a Squeaky Mousemat or a fishing rod. - Luna

+ Renamed `BlockClickEvent.getBlockState` to `BlockClickEvent.blockState` to avoid implying it's a method. - Luna
+ Fixed `CropClickEvent` firing when right-clicking crops. - Luna
    * There are currently no crops (other than Rift crops, which aren't handled by this event) that are harvested using right-click, so these were just false positives.


src/main/java/at/hannibal2/skyhanni/config/features/garden/KeyBindConfig.kt
```diff
@@ -18,6 +18,24 @@ class KeyBindConfig {
...
+    var mousemat: Boolean = true
+
+    @Expose
+    @ConfigOption(
+        name = "Include Fishing Rod",
+        desc = "Also use custom keybinds while holding a fishing rod in your hand.\n" +
+            "§eRequires main toggle to be enabled!",
+    )
+    @ConfigEditorBoolean
+    var fishingRod: Boolean = true
```
NopoTheGamer: should not be on by defualt

NopoTheGamer (changes requested):

src/main/java/at/hannibal2/skyhanni/config/features/garden/KeyBindConfig.kt
lunaynx: Just the rod one? Or both?

lunaynx: Actually I'll make both default off to avoid surprising existing users.

hannibal002 (approved):

PR #5458, opened 2026-03-22
Feature: Add Spider Den Relic Pathfinder

legentpc:
## What
Adds a pathfinder for Spider's Den Relics, similar to the existing Fast Fairy Souls Pathfinder. Uses the existing `SPIDER_RELIC` graph nodes to calculate an optimal route to all relics in Spider's Den. Tracks found/unfound relics per profile.

<details>
<summary>Images</summary>

Relic1
Relic2

</details>

## Changelog New Features
+ Added Spider Den Relic Pathfinder. - LegentPc
    * Uses existing SPIDER_RELIC graph nodes.
    * Tracks found relics per profile.
    * Commands: /shrelicsreset, /shrelicsfoundall, /shrelicsreload.


Bloxigus:
Did you use AI to assist in making this PR?

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
```diff
@@ -0,0 +1,322 @@
...
+        fallback = "^\\+[\\d,]+ Coins! \\(\\d+/\\d+ Relics\\)$",
+    )
+
+    /**
+     * REGEX-TEST: You've already found this relic!
+     */
+    private val duplicatePattern by patternGroup.pattern(
+        key = "chat.duplicate.colorless",
+        fallback = "^You've already found this relic!$",
+    )
```
NopoTheGamer, line 60: you dont need colorless in the pattern

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
```diff
@@ -0,0 +1,322 @@
...
+import at.hannibal2.skyhanni.utils.SimpleTimeMark
+import at.hannibal2.skyhanni.utils.SkyBlockUtils
+import at.hannibal2.skyhanni.utils.TimeUtils.format
+import at.hannibal2.skyhanni.utils.chat.TextHelper.asComponent
+import at.hannibal2.skyhanni.utils.chat.TextHelper.send
+import at.hannibal2.skyhanni.utils.coroutines.CoroutineConfig
+import at.hannibal2.skyhanni.utils.navigation.NavigationUtils
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+
+@Suppress("MemberVisibilityCanBePrivate", "unused")
```
NopoTheGamer: dont suppress stuff

NopoTheGamer (changes requested):
rest looks fine, havent tested

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
CalMWolfs, line 60: You are already doing .matches so you dont need to assert start and end of the string

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
```diff
@@ -0,0 +1,322 @@
...
+@SkyHanniModule
+object SpiderDenRelicPathfinder {
+
+    val config get() = SkyHanniMod.feature.misc
```
CalMWolfs: can be private

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
```diff
@@ -0,0 +1,322 @@
...
+        }
+
+        fun checkHaveAll(): Boolean {
+            val haveAll = total > 0 && foundRelicsStore().size == total
+            if (haveAll) allFound("already found all relics in Spider's Den")
+            return haveAll
+        }
+    }
+
+    @HandleEvent(WorldChangeEvent::class)
```
CalMWolfs, line 136:
```suggestion
    @HandleEvent
```

Not needed as you already have the PrimaryFunction name

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
```diff
@@ -0,0 +1,322 @@
...
+    @HandleEvent
+    fun onTick(event: SkyHanniTickEvent) {
+        if (!isEnabled()) return
+        if (event.isMod(5) && calculating) {
+            val duration = calculatingStart.passedSince().format(showMilliSeconds = true)
+            "§e[SkyHanni] Calculating Relic route §b$duration".asComponent().send(calculatingMessageId)
+        }
+    }
+
+    @HandleEvent(SecondPassedEvent::class)
```
CalMWolfs:
```suggestion
    @HandleEvent
```

Not needed as you already have the PrimaryFunction name

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/config/features/misc/MiscConfig.kt
```diff
@@ -406,6 +406,14 @@ class MiscConfig {
     @ConfigEditorBoolean
     var fastFairySouls: Boolean = false
 
+    @Expose
+    @ConfigOption(
+        name = "Spider Den Relic Pathfinder",
+        desc = "Uses a fast pathfinder route to get to all Relics in Spider's Den.",
+    )
+    @ConfigEditorBoolean
+    var spiderRelicPathfinder: Boolean = false
```
hannibal002, line 408: this is a standalone feature, it needs a @FeatureToggle annotation

src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
```diff
@@ -980,6 +980,14 @@ class ProfileSpecificStorage(
         var found: MutableMap<IslandType, MutableSet<LorenzVec>> = mutableMapOf()
     }
 
+    @Expose
+    var spiderRelics: SpiderRelicsStorage = SpiderRelicsStorage()
```
hannibal002: how about separating this further into spider -> relics, so that we later can easier store spider island related data in the config without migrating?

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
```diff
@@ -0,0 +1,324 @@
...
+        val currentIsland = SkyBlockUtils.currentIsland
+        relicPathFindConfig.launch {
+            val route = NavigationUtils.getRoute(
+                missingRelics,
+                maxIterations = 300,
+                neighborhoodSize = 50,
+            ).toMutableList()
+
+            val duration = calculatingStart.passedSince()
+            "§e[SkyHanni] Calculated Relic route in §b${duration.format(showMilliSeconds = true)}"
```
hannibal002: those three lines here look odd to me. how about we extract the result of `format()`, then we can inline `duration` and remove the line break in front of `.asComponent`

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
```diff
@@ -0,0 +1,323 @@
...
+                    found = foundRelics.size,
+                    total = allRelics.size,
+                    route = route,
+                    allRelics = allRelics.map { it.position }.toSet(),
+                ).also { it.pathToNext() }
+            }
+        }
+    }
+
+    private fun createEmptyData(reason: String) =
```
DavidArthurCole, line 274: This is a local re-implementation of the `Resettable` interface.
Please see the docs for that, and wire it in here instead.

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
```diff
@@ -0,0 +1,323 @@
...
+            debugState = reason
+        }
+
+    private fun getRelicNodes(nodes: List<GraphNode>): List<GraphNode> =
+        nodes.filter { it.hasTag(GraphNodeTag.SPIDER_RELIC) }
+
+    private fun foundRelicsStore(): MutableSet<LorenzVec> =
+        ProfileStorageData.profileSpecific?.spider?.relics?.found ?: mutableSetOf()
+
+    private fun isEnabled() =
```
DavidArthurCole: Remove `inSkyBlock` check here, and use `onlyOnSkyblock` in the HandleEvent annotations.
Remove `isOnSpiderDen()` and use `onlyOnIsland` in the HandleEvent annotations.

All instance calls to this can be replaced with the config bool.

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
```diff
@@ -0,0 +1,323 @@
...
+import at.hannibal2.skyhanni.utils.coroutines.CoroutineConfig
+import at.hannibal2.skyhanni.utils.navigation.NavigationUtils
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+
+@SkyHanniModule
+object SpiderDenRelicPathfinder {
+
+    private val config get() = SkyHanniMod.feature.misc
+
+    private var data: Data? = null
```
DavidArthurCole, line 41: Data should never be `null`. It should be initialized, and with default values, but not nullish.

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
```diff
@@ -0,0 +1,323 @@
...
+    )
+
+    private class Data(
+        var found: Int,
+        val total: Int,
+        val route: MutableList<LorenzVec>,
+        val allRelics: Set<LorenzVec>,
+        var foundButNotClickedRelic: LorenzVec? = null,
+    ) {
+        var disabled = total > 0 && found == total
```
DavidArthurCole, line 69: Instead of making this a `var`, make it a `val` getter.

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
```diff
@@ -0,0 +1,323 @@
...
+        var debugState: String? = null
```
DavidArthurCole, line 70: Should not be a string, and should be a bounded enum.

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
```diff
@@ -0,0 +1,323 @@
...
+
+        fun allFound(state: String) {
+            disabled = true
+            foundRelicsStore().addAll(route)
+            found = foundRelicsStore().size
+            debugState = state
+        }
+
+        fun checkHaveAll(): Boolean {
+            val haveAll = total > 0 && foundRelicsStore().size == total
```
DavidArthurCole, line 130: Make this a getter on the data, not a local val

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
```diff
@@ -0,0 +1,323 @@
...
+            markFound(nearest)
+            pathToNext()
+        }
+
+        private fun getNearestRelic(): LorenzVec? {
+            val playerLocation = LocationUtils.playerLocation()
+            val nearest = allRelics.minByOrNull { it.distanceSq(playerLocation) } ?: return null
+            if (nearest.distanceToPlayer() < 10) return nearest
+            return null
+        }
```
DavidArthurCole, line 83: There's only one inline usage of this, so this can just be moved into foundNearby.
Please also do the distance filter before you calculate the distance between the player and all of the relics.

DavidArthurCole (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
hannibal002: this pr repllicates the behavior in FastFairySoulsPathfinder. there, we also use the identical createEmptyData logic.

Resettable feels like it is designed for storage data, stuff that actually can reset. here, i dont see the benefit of using it. spider den relic and fast fair souls both dont use the data class as storage, but as temp values data holder. thats different.

if you still think resettable can work here, please exaplain how.

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
hannibal002: good catch.

tho, the very same pattern exists in FastFairySoulsPathfinder. so, this should be changed in both files at once.
could be handelled for this prs scope sake by adding a todo comment over both isEnabled functions.

also, note that  the `condition = ` lines dont need the in skyblock check explicitly. the pathfind logic cancels on island change already.

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
hannibal002: i disagree. data here also represents the question "is this feature inizialised?"

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
hannibal002: there are two lines where we set disabled to true. this value is not a val getter, its a member of the data class with default state defined as `total > 0 && found == total`

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
hannibal002: good catch. the same change needs to be done in FastFairySoulsPathfinder

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
hannibal002: good catch. the same change needs to be done in FastFairySoulsPathfinder

src/main/java/at/hannibal2/skyhanni/features/misc/SpiderDenRelicPathfinder.kt
hannibal002: > There's only one inline usage of this, so this can just be moved into foundNearby.

i disagree, this just makes foundNearby larger for no reason. this is a clean separation of concerns.

> Please also do the distance filter before you calculate the distance between the player and all of the relics.

good catch.  maybe also check if getNearestSoul in FastFairySoulsPathfinder has the same issue

hannibal002 (changes requested):
overall this is a copy of FastFairySoulsPathfinder, so we should maybe create a generic helper class "find location, click thing", instead of all the code duplications. tho, not sure if this is out of scope for this pr. at least, we should add todo comments on each function that exists duplicate, and at the top of the class. we should be aware that fixing an issue in one of the two functions  is really bad if we dont also fix the sibling function in the other class.

DavidArthurCole (commented):
I can't resolve or dismiss outdated reviews by others. No idea what percentage, if any of them, are stale or not.

Image

hannibal002 (approved):

PR #5468, opened 2026-03-24
Fix: Compact Tab List being not so compact

lunaynx:
## What
Fixed issues with some text in Compact Tab List being too long, and some active effects being split into two lines.

## Changelog Fixes
+ Fixed Active Effects and Cookie Buff in Compact Tab List having really long text when inactive. - Luna
+ Fixed Active Effects in Compact Tab List being split into two lines when they shouldn't. - Luna

+ Fixed `Component.contains` and `Component.startsWith` adding legacy formatting codes where the original component string doesn't include them. - Luna


src/main/java/at/hannibal2/skyhanni/features/misc/compacttablist/TabListReader.kt
```diff
@@ -30,6 +29,8 @@ object TabListReader {
     private var lastTabComponents: List<Component>? = null
     private var lastFooterComponent: Component? = null
 
+    private var inUpgrades = false
```
hannibal002, line 32: should probably be reset at the start of parseFooterAsColumn

hannibal002 (changes requested):

hannibal002 (approved):

PR #5469, opened 2026-03-24
Improvement: Prevent breaking crops with Sprayonator

lunaynx:
## What
Turned "Prevent Breaking with Rod" into "Prevent Breaking Crops" and added a Sprayonator option to it. Also, now it only prevents breaking crops specifically – because preventing all block break attempts with Sprayonator would prevent safe spray change attempts when the player is looking at a non-crop block.

## Changelog Improvements
+ Added an option to prevent breaking crops with Sprayonator. - Luna
    * The existing feature that includes fishing rods will now only prevent breaking *crops*, not all blocks.


src/main/java/at/hannibal2/skyhanni/config/ConfigUpdaterMigrator.kt
```diff
@@ -13,7 +13,7 @@ import com.google.gson.JsonPrimitive
 object ConfigUpdaterMigrator {
 
     val logger = LorenzLogger("ConfigMigration")
-    const val CONFIG_VERSION = 127
+    const val CONFIG_VERSION = 128
```
NopoTheGamer: old config version

CalMWolfs (approved):
Seems cool, could also be expanded in future to prevent breaking crops with fist if you dont have suns grasp equip on as ive seen someone complain about always forgetting to re-equip it after doing pests and then their whole farm has no crops left as no replenish

PR #5482, opened 2026-03-26
Fix: correct inventory GUI scale overlay rendering

Akinsoft:
## What
Fixes SkyHanni overlays rendering incorrectly when another mod uses a different GUI scale for inventory screens. For example Profit overlays would resize when opening the inventory with a different GUI scale for inventory.

This keeps HUD-style overlays at the normal HUD scale while the player's inventory is open, and fixes Diana's nearest warp helper rendering a second oversized copy over the inventory.

Tested locally with Aaron Mod using a separate inventory GUI scale.

image
image

## Changelog Fixes
+ Fixed some overlays rendering at the wrong size with separate inventory GUI scales, and fixed Diana's nearest warp helper rendering twice over inventory. - akinsoft

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/event/diana/BurrowWarpHelper.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/event/diana/BurrowWarpHelper.kt
@@ -55,7 +55,7 @@ object BurrowWarpHelper {
         }
     }
 
-    @HandleEvent(GuiRenderEvent.GuiOverlayRenderEvent::class, onlyOnIsland = IslandType.HUB)
+    @HandleEvent(onlyOnIsland = IslandType.HUB)
     fun onRenderOverlay(event: GuiRenderEvent.GuiOverlayRenderEvent) {
         if (!config.burrowNearestWarp) return
         if (!DianaApi.isDoingDiana()) return
--- a/src/main/java/at/hannibal2/skyhanni/features/event/diana/BurrowWarpHelper.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/event/diana/BurrowWarpHelper.kt
@@ -55,8 +55,8 @@ object BurrowWarpHelper {
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/at/hannibal2/skyhanni/features/event/diana/BurrowWarpHelper.kt
```diff
@@ -55,8 +55,8 @@ object BurrowWarpHelper {
         }
     }
 
-    @HandleEvent(GuiRenderEvent::class, onlyOnIsland = IslandType.HUB)
-    fun onRenderOverlay() {
+    @HandleEvent(GuiRenderEvent.GuiOverlayRenderEvent::class, onlyOnIsland = IslandType.HUB)
```
hannibal002: remove the GuiRenderEvent.GuiOverlayRenderEvent::class part. we dont need both. having the event class in parameter is enough.

src/main/java/at/hannibal2/skyhanni/utils/compat/InventoryGuiScaleCompat.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import net.minecraft.client.Minecraft
+
+object InventoryGuiScaleCompat {
+
+    private val mc get() = Minecraft.getInstance()
+
+    fun <T> withOriginalHudScale(action: () -> T): T {
+        val window = mc.window
+        val currentScale = window.guiScale.toInt()
```
hannibal002: guiScale is int, no need to cast

hannibal002 (changes requested):
the approach looks good! found two small things that should get changed

AzureAaron (approved):
Looks good to me, will likely need changing for 26.1 though (as Mojang totally changed how Screen rendering worked/I had to change how the feature works too).

hannibal002 (approved):

PR #5483, opened 2026-03-26
Fix: Make "translate on click" feature work in modern versions

Rohan-Sangani:
## What

Addresses the TODOs that previously existed in the Translator file, and makes it so that the feature actually works again. The feature broke as part of migrating from 1.8.9 (and looking back at the code, I am confused as to how it worked in the first place).

- Split Translator.kt into GoogleTranslator and TranslatorCommand files
- Switched to named JSON as API response format, making the code easier to read
- Change chat component editing to use internal events and utils functions

## Changelog Fixes
+ Fixed Translate on Click feature. - NetheriteMiner


src/main/java/at/hannibal2/skyhanni/features/chat/translation/GoogleTranslator.kt
```diff
@@ -0,0 +1,86 @@
...
+            messageToSend.append(obj["trans"])
+        }
+        return@withContext arrayOf(messageToSend.toString(), language)
+    }
+
+    fun toNativeLanguage(args: Array<String>) {
+        val message = args.joinToString(" ").removeColor()
+
+        CoroutineConfig("translator toNativeLanguage", 10.seconds).launchCoroutine {
+            val translation = getTranslation(message, TranslatorCommand.getNativeLanguage())
```
hannibal002: toNativeLanguage gets called from TranslatorCommand, and then accesses TranslatorCommand here again. weird design. how about we just provide this via parameter at call site?

hannibal002 (changes requested):

PR #5484, opened 2026-03-26
Fix + Backend: Stop Tracking Commands on Disconnect

DavidArthurCole:
## What
https://discord.com/channels/997079228510117908/1486850417727111301

## Changelog Fixes
+ Fixed error when disconnecting while tracking sounds/particles. - Daveed

+ Extracted `TrackCommand` into base, and world class. - Daveed


src/main/java/at/hannibal2/skyhanni/test/command/track/TrackSoundsCommand.kt
```diff
@@ -49,10 +51,17 @@ object TrackSoundsCommand : TrackCommand<PlaySoundEvent, String>(
     override fun PlaySoundEvent.shouldAcceptTrackableEvent(): Boolean = when {
         soundName == "game.player.hurt" && pitch == 0f && volume == 0f -> false // remove random useless sound
         soundName.isEmpty() -> false // sound with empty name aren't useful
-        else -> {
+        else -> runCatching {
             distanceToPlayer // Need to call to initialize Lazy
```
hannibal002, line 56: this is a dev command, should show errors instead of hiding them

src/main/java/at/hannibal2/skyhanni/test/command/track/TrackCommand.kt
```diff
@@ -129,24 +120,6 @@ abstract class TrackCommand<T : CancellableWorldEvent, K>(
         isRecording = false
```
hannibal002, line 122: this is recording value does not get set to false on client disconnect event

hannibal002 (changes requested):
this fix looks more like a backend + fix to me

src/main/java/at/hannibal2/skyhanni/test/command/track/TrackSoundsCommand.kt
DavidArthurCole: The only error this is catching is the one that Luna reported, which happens when it is recording as world is changed - there are no errors we actually care about here, hence eating it.

DavidArthurCole:
> this fix looks more like a backend + fix to me

Yeah, probably. The V2 of the tracking command was an afterthought tbh. Updated the desc.

hannibal002:
> The only error this is catching is the one that Luna reported, which happens when it is recording as world is changed - there are no errors we actually care about here, hence eating it.

I assume you mean a NPE error bc player is not loaded?

ill fix this with localPlayerExists.

hannibal002 (approved):

PR #5504, opened 2026-03-30
Fix: Fix Farming Fortune Display displaying incorrect values with Bonus Pest Chance disabled

Nick-NCSU:
## What
The server's logic to increase the maximum number of pests before stat reduction by 1 for every 100 BPC you have only applies when BPC is enabled. The Farming Fortune Display now checks for BPC being enabled when calculating the FF reduction.

I've moved the Bonus Pest Chance Display's logic to the PestAPI so that it could be shared with the Farming Fortune Display.

## Changelog Fixes
+ Fixed Farming Fortune Display displaying incorrect values with Bonus Pest Chance disabled. - Nick-NCSU

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/FarmingFortuneDisplay.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/FarmingFortuneDisplay.kt
@@ -166,7 +166,6 @@ object FarmingFortuneDisplay {
                 pestBonusExpireTime = time
 
             }
-            update()
         }
     }
 
@@ -174,10 +173,7 @@ object FarmingFortuneDisplay {
         universalTabFortunePattern.firstMatcher(widget.lines.map { it.string }) {
             val fortune = group("fortune").toDouble()
             foundTabUniversalFortune = true
-            if (fortune != tabFortuneUniversal) {
-                tabFortuneUniversal = fortune
-                update()
-            }
+            tabFortuneUniversal = fortune
         }
         cropSpecificTabFortunePattern.firstMatcher(widget.lines.map { it.string }) {
             val crop = CropType.getByName(group("crop"))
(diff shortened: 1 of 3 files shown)


NopoTheGamer:
#5312

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestApi.kt
```diff
@@ -176,6 +180,17 @@ object PestApi {
         "PLAYING",
     )
 
+    private val bpcPatternGroup = RepoPattern.group("garden.bonuspestchance")
```
Nick-NCSU, line 183: I could add this regex to the `garden.pests-api.bonuspestchance.widget` so it could use the existing group in here but it would still also have to be in `garden.bonuspestchance.widget` as well for people who haven't updated right? Lmk how I should handle this.

src/main/java/at/hannibal2/skyhanni/features/garden/FarmingFortuneDisplay.kt
```diff
@@ -166,29 +166,22 @@ object FarmingFortuneDisplay {
...
 
     private fun checkStats(widget: TabWidget) {
         universalTabFortunePattern.firstMatcher(widget.lines.map { it.string }) {
             val fortune = group("fortune").toDouble()
             foundTabUniversalFortune = true
-            if (fortune != tabFortuneUniversal) {
-                tabFortuneUniversal = fortune
-                update()
-            }
+            tabFortuneUniversal = fortune
```
Nick-NCSU, line 176: I removed most of the times `update()` is called since it's already being called every other tick anyways which was causing it to be called multiple times every time the stats/pests widgets changed. I can definitely put this back if preferred or if there was some reason I missed for why the displayed text had to be updated in these functions.

PR #5510, opened 2026-03-30
Backend: Suppress "Mixin not found in any mixin config" warning

lunaynx:
## What
This warning is useless and misleading because we have automatic mixin registration.

exclude_from_changelog


hannibal002 (changes requested):
i dont like sharing this file on git.

for me, this is what this file looks like currently:

<details>
<summary>Details</summary>

```
<component name="InspectionProjectProfileManager">
  <profile version="1.0">
    <option name="myName" value="Project Default" />
    <inspection_tool class="KotlinInvalidBundleOrProperty" enabled="false" level="ERROR" enabled_by_default="false" />
    <inspection_tool class="MissingPrimaryFunction" enabled="false" level="WEAK WARNING" enabled_by_default="false" />
  </profile>
</component>
```

</details>

so when i want to hide some error messages in the future locally, i need to constantly deal with a uncommited file change.

any other solutions or ideas about what to do instead?

lunaynx:
Yeah, that's unfortunate. I'm pretty sure there's already an open issue on IntelliJ's issue tracker asking for more flexible configuration for this, but I can't find it right now.

There are a few potential workarounds, but every developer would have to one of do these (or just deal with unstaged changes and remember to not commit their local changes to this file):
- add your custom stuff to IDE-wide settings instead of per-project
- clone the default inspection profile and make your changes there, still per-project but may become out of sync with project config
- use `git update-index --skip-worktree` on the file

I'll just close this. We can probably do this via the IntelliJ plugin if we ever resurrect it.

PR #5511, opened 2026-03-30
Backend: Fix Detekt Processing Regex

DavidArthurCole:
## What
With 2.0, the `warning` is gone, so the processor regex needed updating.
This is why all PRs with detekt are getting "0 detekt failures".

exclude_from_changelog


DavidArthurCole:
(there are 2 detekt issues with beta rn, hopefully this will give us indication if the regex change will work as expected)

PR #5493, opened 2026-03-28
Backend: Get rid of LorenzVec

lunaynx:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/5460
- https://github.com/hannibal002/SkyHanni/pull/5512

## What
Removes `LorenzVec` in favor of Minecraft's native `Vec3` class with extension and utility methods in a new `VectorUtils` class.

## Changelog Technical Details
+ Removed `LorenzVec` in favor of `Vec3` and `VectorUtils`. - Luna

PR #5512, opened 2026-03-30
Backend: Add VectorUtils, deprecate LorenzVec

lunaynx:
## What
Precursor to #5493. Adds a new `VectorUtils` class and deprecates `LorenzVec`, but changes not much else (besides introducing hundreds of new deprecation warnings).

The idea is to eventually fully move away from `LorenzVec`, and use Minecraft's native `Vec3` class and the extension functions in `VectorUtils`.

+ Added `VectorUtils` class for `Vec3` extensions and marked `LorenzVec` as deprecated. - Luna
+ Added `Vec3ArgumentType` and removed `LorenzVecArgumentType`. - Luna


lunaynx:
I should probably pull in `Vec3ArgumentType` to this.

Edit: Done

lunaynx:
Will work on this a bit more, I think I can reduce the code duplication.

PR #5513, opened 2026-03-30
Fix: Update Reindrake spawn message

lunaynx:
## What
~~Currently on [alpha](https://hypixel.net/threads/march-23rd-jerry-island-qol-new-abiphone-contacts-reforge-anvil-removal-and-more.6075332/).~~ Released to main server.

## Changelog Fixes
+ Added support for new Reindrake spawn message. - Luna


lunaynx:
Marking this as a fix now since it's out on main.

PR #5517, opened 2026-03-31
Fix + Backend: JGit for Repo Downloading

DavidArthurCole:
## What
Following a conversation with Aaron, and generally in the goal of getting our repo system ready to possibly move to its own deployment, this PR adds JGit management for our AbstractRepoManagers, and is capable of statefully inheriting old configs (on the fly `.git` creation), cloning, and pulling.

In the case something goes wrong with this approach, the existing Zip approach is still in-place.
See changelog below for more, but also did some refactoring on the utils, now that they're less Github-centric, and more git-centric

## Changelog Fixes
+ Fixed issues trying to download NEU repo. - Daveed

+ Added JGit management for repo files. - Daveed
+ Reworked the `RepoLogger` to extend `SkyHanniLogger`. - Daveed
+ Moved repo logging to distinct files. - Daveed
   * Will be in `config/sh/repo/logs` and `neu/repo/logs` respectively.
   * Log dirs will not get nuked when repo fs clears.


hannibal002:
> now that they're less Github-centric, and more git-centric

:fire:

src/main/java/at/hannibal2/skyhanni/data/repo/GitRepo.kt
```diff
@@ -0,0 +1,179 @@
...
+                setNoCheckout(false)
+                if (source.isSshUri()) setTransportConfigCallback(sshConfigurer)
+                setProgressMonitor(RepoJGitMonitor(this@tryCloneRepo))
+            }.call().use { cloned ->
+                logger.debug("Cloned ${cloned.repository.directory.absolutePath} for $location via $source")
+            }
+            true
+        }.getOrElse { e ->
+            logger.error("Failed to clone $location from $source\n$e")
+            root.deleteRecursively()
```
hannibal002: arent logs in the root as well?

src/main/java/at/hannibal2/skyhanni/utils/SkyHanniLogger.kt
```diff
@@ -3,82 +3,49 @@ package at.hannibal2.skyhanni.utils
...
 
-        return logger
-    }
-
-    private fun createParent(file: File) {
-        val parent = file.parentFile
-        if (parent != null && !parent.isDirectory) {
-            parent.mkdirs()
+            if (!deletedExpired && SkyBlockUtils.onHypixel) {
+                deletedExpired = true
```
hannibal002, line 44: deletedExpired is companion. so, we only ever delete the first logger instance, never anything else.

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
```diff
@@ -82,8 +81,9 @@ abstract class AbstractRepoManager<E : AbstractRepoReloadEvent> {
     private val commonShortName by lazy { commonShortNameCased.lowercase() }
     private val successfulConstants = mutableSetOf<String>()
     private val unsuccessfulConstants = mutableSetOf<String>()
-    private val githubRepoLocation: GitHubUtils.RepoLocation
-        get() = GitHubUtils.RepoLocation(config.location, SkyHanniMod.feature.dev.debug.logRepoErrors)
+    private val gitRepo: GitRepo by lazy {
+        GitRepo(config.location, SkyHanniMod.feature.dev.debug.logRepoErrors)
```
hannibal002: logRepoErrors only gets checked once. this should be checked continuesly, not cached.

src/main/java/at/hannibal2/skyhanni/data/repo/GitRepo.kt
```diff
@@ -0,0 +1,179 @@
...
+import org.eclipse.jgit.transport.SshTransport
+import org.eclipse.jgit.util.FS
+import java.io.File
+
+/**
+ * Represents the location of a Git repository.
+ * @param config the [AbstractRepoLocationConfig] containing the information for this repository.
+ * @param shouldError If true, will throw an error if the latest commit SHA cannot be fetched, or if the download fails.
+ */
+data class GitRepo(
```
hannibal002: why is this a data class? this class is mainly logic, not data.

hannibal002: why is this a data class? this class is mainly logic, not data.

src/main/java/at/hannibal2/skyhanni/data/repo/GitRepo.kt
```diff
@@ -0,0 +1,179 @@
...
+
+    private fun String.isSshUri() = startsWith("git@") || startsWith("ssh://")
+
+    private fun RepoFileSystem.getAvailableSources(): List<String> {
+        val sources = mutableListOf("https://github.com/$user/$repo.git")
+
+        val userHome = FS.DETECTED.userHome() ?: return sources.also {
+            logger.debug("Skipping SSH fallback: Unable to determine user home directory.")
+        }
+        val keyPresent = File(userHome, ".ssh").listFiles()?.any { file ->
```
hannibal002, line 54: why would we want to offer auth access to github? that feels totally out of place of what the mc mod wants to do. pleasea remove, this looks confusing

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
```diff
@@ -82,8 +81,9 @@ abstract class AbstractRepoManager<E : AbstractRepoReloadEvent> {
...
+    }
     private val repoMutex = Mutex()
     val repoLocked get() = repoMutex.isLocked
```
hannibal002: unused

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
```diff
@@ -427,6 +434,17 @@ abstract class AbstractRepoManager<E : AbstractRepoReloadEvent> {
         forceReset: Boolean = false,
         switchToBackupOnFail: Boolean = true,
     ): FetchUnpackResult {
+        progress.update("try loading repo from jgit")
+        with(gitRepo) {
+            if (repoFileSystem.loadFromJGit()) {
+                progress.update("loaded from jgit")
+                return FetchUnpackResult.SUCCESS
```
hannibal002, line 442: we dont set isUsingBackup to false or update commitStorage. is this intentional?

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
```diff
@@ -461,22 +480,23 @@ abstract class AbstractRepoManager<E : AbstractRepoReloadEvent> {
...
-        return FetchUnpackResult.SUCCESS
     }
 
     private fun prepCleanRepoFileSystem(progress: ChatProgressUpdates) {
         progress.update("deleteRecursively")
-        repoDirectory.deleteRecursively()
+        repoDirectory.listFiles()?.forEach { if (it != logger.logsDir) it.deleteRecursively() }
 
         progress.update("createAndClean")
         repoFileSystem = repoDirectory.let { root ->
```
hannibal002, line 503: root is unused

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/repo/GitRepo.kt
DavidArthurCole: They are. Look at the function definition for `deleteRecursively`

src/main/java/at/hannibal2/skyhanni/data/repo/GitRepo.kt
DavidArthurCole: This is a fallback for instances where a user gets blocked from GitHub HTTP due to IP or LAN restrictions. It's just another fallback on top of the others.

I can remove it if you really want, but it will have no affect for 99% of users.

hannibal002: hmm not sure. just felt out of place considering everything else in this pr. we can keep it, but please explain the reason for this in the code as comment.

src/main/java/at/hannibal2/skyhanni/utils/SkyHanniLogger.kt
DavidArthurCole: That is the same functionality as it was previously, no? The previous iteration of this code also used a companion var, just in a less transparent way.

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
DavidArthurCole: `isUsingBackup` is a good call; but the commit storage I'm iffy on. It's not really "necessary" in this case, since the `.git` holds that information in a way that's more reusable than our JSON is.

src/main/java/at/hannibal2/skyhanni/data/repo/GitRepo.kt
hannibal002: public fun File.deleteRecursively(): Boolean = walkBottomUp().fold(true, { res, it -> (it.delete() || !it.exists()) && res })

this is in kotlin.io.Utils, has nothing to do with the functions that impl RepoFileSystem. any chance you access the wrong function? otherwise idk what you mean

hannibal002: ok i think the issue is you call the correct function on the wrong object. root is a file, not a repo manager instance

hannibal002 (approved):

PR #5521, opened 2026-04-01
Improvement: More April Fools stuff

lunaynx:
## What
- Halves displayed FPS during April Fools
- Fixes ping not being halved in Minecraft ping chart
- Fixes TPS not being halved in party commands

exclude_from_changelog


NopoTheGamer:
too silly

PR #5525, opened 2026-04-01
Fix + Improvement: Ctrl/Cmd confusion

lunaynx:
## What
Fixed messages so that they get the correct modifier key for the platform where applicable. Also cleaned up relevant code.

## Changelog Improvements
+ Added Rosewater Flask search tag to options related to evolving items. - Luna

## Changelog Fixes
+ Fixed the mod sometimes saying Control (Ctrl) even when the Command (Cmd) key is expected on macOS. - Luna

+ Changed `ChatUtils.deleteNextMessage` to use a `Component` in the predicate instead of a `String`. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/composter/ComposterOverlay.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/composter/ComposterOverlay.kt
@@ -597,7 +597,8 @@ object ComposterOverlay {
         return price
     }
 
-    fun onNeuRepositoryReload() {
+    @HandleEvent
+    fun onNeuRepoReload() {
         updateOrganicMatterFactors()
     }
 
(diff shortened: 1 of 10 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/composter/ComposterOverlay.kt
```diff
@@ -600,15 +597,14 @@ object ComposterOverlay {
         return price
     }
 
-    @HandleEvent(NeuRepositoryReloadEvent::class)
-    fun onNeuRepoReload() {
+    fun onNeuRepositoryReload() {
```
hannibal002: the event handler annotation has to be there. the primary function jsut means that you dont need to write down the event class anymore, you cant omit the annotation itself.

hannibal002 (changes requested):

hannibal002 (approved):

PR #5527, opened 2026-04-01
Feature: Mute Bugged Spade Sounds

lunaynx:
## What
The Ancestral/Archaic/Deific Spade can sometimes play a random "Easter Egg" sound instead of the normal sound. On 1.21 and above, this is bugged so that instead of playing a short-lived sound, sometimes it starts playing creepy low-pitched game music, which lasts about 10 minutes or until you switch lobbies. This feature attempts to prevent that sound from playing.

For reference, it sounds like this:
https://github.com/user-attachments/assets/eb18dd3e-2957-473f-8cec-2c346585f934

## Changelog New Features
+ Added a fix for a Hypixel bug where Diana spades will sometimes play low-pitched Minecraft music. - Luna


lunaynx:
Diana in 4 hours 🥺

TheDGOfficial:
Damn I was wondering if the vanilla game was acting up the first time i had this because I keep the vanilla music feature on and it would usually play chill noteblock music lol. Thank you for this!

Did you create a Hypixel bug report for it so that it can possibly be fixed server-side as well (probably in a year or so since their backlog and priorities ughh)?

lunaynx:
Yeah, I made a pretty detailed bug report.

PR #5539, opened 2026-04-03
Improvement: Add Shuriken to Sea Creature Health Overlay

Fazfoxy:
## What
Got asked for this, seems like a good idea to me.

## Changelog Improvements
+ Added Shuriken icon to Sea Creature Health Display . - Fazfoxy
    * This is Default Enabled.


Fazfoxy:
Does not work.

PR #5548, opened 2026-04-04
Backend: Deprecate unsafe MinecraftCompat methods

lunaynx:
## What
Following [SkyblockAPI's example](https://github.com/SkyblockAPI/SkyblockAPI/commit/010ec1c82d00f4980e5ba19c0ac49d6f2d8b1f45), deprecate `MinecraftCompat.localPlayer` and `MinecraftCompat.localWorld` in favor of their null-safe variants. These shortcuts may be convenient, but they obfuscate their inherent unsafety. It's best to replace them with explicit `?: return null` or `?: error(...)` at call sites.

This will probably be followed up by another PR that completely removes uses of the deprecated methods in the future.

+ Deprecated MinecraftCompat.localPlayer in favor of MinecraftCompat.localPlayerOrNull. - Luna
+ Deprecated MinecraftCompat.localWorld in favor of MinecraftCompat.localWorldOrNull. - Luna


hannibal002 (changes requested):
> It's best to replace them with explicit ?: return null or ?: error(...) at call sites.

i disagree. this is super annoyingto have to deal with nullable everywhere where we can savely assume that the player is not null (e.g. after a skyblcok/island check has passed). so this adds more code bloat overall.

slightly unrelated, we should/could prob move some of those callers in the future to use PlayerUtils

lunaynx:
We end up wrongly assuming them to exist in a lot of places which leads to errors. It's really not that noisy to put a few single-line sanity checks around the code, and it's worth it for the safety. SkyblockAPI used to have these unsafe methods using the same argument, but they deprecated them. Maybe @meowora @ThatGravyBoat can chime in.

Also, worth noting that if you have the `Player` object, you already have a non-nullable `Level` object you can get from it, so that saves some "noise".

hannibal002:
> Also, worth noting that if you have the Player object, you already have a non-nullable Level object you can get from it, so that saves some "noise".

Imo not relevant, its about getting the original player object.

yes, this sometimes thorws errors, but every time it does, its a warning for us that we run the code with missing checks, e.g. is on sb or is in world. those checks then get fixed properly. this is stuff we would never see if we just use ?: return. and thats the other issue i have with this. if the dev has the option between throwing an error or simply returning, the second choice feels most of the time more intuitive. but thats wrong and counterintuitive to what we actually want the feature to do most of the times.

we already have a dedicated "is the player exisitng" function, maybe we can even omit the nullable player function entirely? but thats either way antoher bigger discussion.

lunaynx:
It's true that silently returning can hide errors, but proper testing generally avoids that, and if it still slips through, we'll generally notice and be able to debug easily using breakpoints if a feature doesn't work anyway.

9 out of 10 times, when we get an NPE from these, it simply needs an early return, because someone didn't consider that their code could run when the world and/or player doesn't exist (e.g. on connect, world change, disconnect).

The general idea is:
- if we can easily get a non-nullable world (level) or player from the context, do that
- if we can't do that but we're in a condition where we're confident it has to exist, do `?: error(...)`
- if we're can't do that and we're not confident that it has to exist, do `?: return`

lunaynx:
Also, I strongly disagree with removing the nullable variants regardless. `localWorldExists` and `localPlayerExists` are not a substitute for proper null safety. It may work and be safe *most* of the time considering Minecraft is tick-based, but the compiler won't be happy (unless you're dealing with an object that lies to the compiler that it's non-nullable, like the helpers I'm trying to deprecate here), and it could technically be unsafe if you cross tick boundaries with `run(Or)NextTick` etc.

lunaynx:
#5887

PR #5549, opened 2026-04-04
Backend: Fix changelog command injection in Modrinth Version Upload

lunaynx:
## What
Fixes a command injection vulnerability in the Modrinth Version Upload action. Thankfully, to actually trigger it, a maintainer would have to actually publish a release with a malicious payload left intact in the changelog and then run the workflow. But this also fixes the workflow breaking on backticks and other characters.

exclude_from_changelog

### DIFF
--- a/.github/workflows/modrinth-version-upload.yml
+++ b/.github/workflows/modrinth-version-upload.yml
@@ -55,13 +55,17 @@ jobs:
                     echo "EOF" >> $GITHUB_OUTPUT
 
             -   name: Print environment variables
+                env:
+                    CHANGELOG: ${{ steps.get_changelog.outputs.changelog }}
                 run: |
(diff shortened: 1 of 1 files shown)


ItsEmpa:
looks bad to me, please dont merge

hannibal002 (changes requested):
empa

lunaynx:
That was obviously a joke...

ItsEmpa:
> empa

it was a joke???? 😭

hannibal002:
beta branch does no longer has the file at all

lunaynx:
Closing in favor of removal with #5612 for now. If we decide to bring back the workflow I may reopen this, but we'll probably just add a fixed version from the start if at all.

PR #5552, opened 2026-04-05
Fix: Fusion Keybinds not working via Abiphone

lunaynx:
## What
Made Fusion Keybinds work on any island, since Kysha has an Abiphone contact now, and changed the condition for checking duplicate keybinds to not be as spammy.

## Changelog Fixes
+ Fixed Fusion Keybinds not working if you call Kysha via Abiphone outside Galatea. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/hunting/FusionKeybinds.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/hunting/FusionKeybinds.kt
@@ -3,15 +3,13 @@ package at.hannibal2.skyhanni.features.hunting
 import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.events.GuiContainerEvent.ClickType
+import at.hannibal2.skyhanni.features.inventory.attribute.AttributeShardsData
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.ConditionalUtils.afterChange
-import at.hannibal2.skyhanni.utils.InventoryDetector
 import at.hannibal2.skyhanni.utils.InventoryUtils
 import at.hannibal2.skyhanni.utils.KeyboardManager.isKeyHeld
-import at.hannibal2.skyhanni.utils.RegexUtils.matches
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/features/hunting/FusionKeybinds.kt
```diff
@@ -2,53 +2,93 @@ package at.hannibal2.skyhanni.features.hunting
...
+    // <editor-fold desc="Patterns">
+    private val fusionBoxInventoryPattern by patternGroup.pattern(
+        "inventory.fusion-box",
+        "Fusion Box",
+    )
+
+    private val confirmFusionInventoryPattern by patternGroup.pattern(
+        "inventory.confirm-fusion",
+        "Confirm Fusion",
+    )
```
NopoTheGamer: dont these exist in attribute api?

NopoTheGamer (changes requested):

src/main/java/at/hannibal2/skyhanni/features/hunting/FusionKeybinds.kt
lunaynx: Yeah, you're right.

src/main/java/at/hannibal2/skyhanni/features/hunting/FusionKeybinds.kt
```diff
@@ -17,38 +16,52 @@ import kotlin.time.Duration.Companion.seconds
...
-        if (hasDuplicateKeybinds) return
-        when (InventoryUtils.openInventoryName()) {
-            "Fusion Box" -> {
-                if (!config.repeatFusionKeybind.isKeyHeld() || config.confirmFusionKeybind.isKeyHeld()) return
+
+    // No island check because Kysha Abiphone contact exists
+    @HandleEvent(onlyOnSkyblock = true)
+    fun onGuiKeyPress() {
+        val repeatHeld = config.repeatFusionKeybind.get().isKeyHeld()
+        val confirmHeld = config.confirmFusionKeybind.get().isKeyHeld()
```
DavidArthurCole, line 27: This is not specific to this PR, but I did want to note that it feels "odd" to be handling a key press event, that itself does not contain any info about the key that was pressed.

Long-term we'd probably want to add the keyCode to the event.

DavidArthurCole (approved):
Code looks good. Haven't tested in-game.

PR #5554, opened 2026-04-06
Fix: Reindrakes not counting as double hook

Fazfoxy:
## What
This is entirely based on 1 screnshot and pure guesswork that we're hiding the double hook message but being blocked from detection by the Reindrake hook message

image

ok got a new screenshot while making this PR, this is 100% fixable and exactly the bug I thought it was.

Untested Because Jerry just ended but, I think we ball

## Changelog Fixes
+ Fixed Double Hook Reindrakes not being picked up by SH. - Fazfoxy
    * This also fixes the Double Hook message being incorrectly removed without appending.
    * Also Sea Creature Tracker only tracking 1 Reindrake for double hooks.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
@@ -102,7 +102,6 @@ object SeaCreatureManager {
         doubleHook = false
     }
 
-
     /**
      * Autopet can be triggered via Sinkers as rod parts (Sponge, Prismarine, Icy) to trigger collection gain which goes between Double Hook! and the Catch message.
      * The Thunder sea Creature gives charge when hooked, which can cause thunder bottles to charge and send the full charge message between Double Hook! and Catch message.
@@ -110,8 +109,7 @@ object SeaCreatureManager {
     private fun isInterceptingColorCodeMessage(message: String): Boolean =
         (PetStorageApi.isAutopetMessage(message) || thunderBottleChargedPattern.matches(message))
 
-
-    // TODO Unify when both Clean. (I do not wanna touch that Pet Storage Autopet Regex, I will horrifically break something)
+    // TODO Unify when both use CleanMessage.
 
     /**
      * Reindrakes Send an empty line, the global message & another empty line Double Hook! and Catch message.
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
```diff
@@ -55,10 +56,11 @@ object SeaCreatureManager {
...
         }
-
-        if (thunderBottleChargedPattern.matches(event.message) ||
-            PetStorageApi.isAutopetMessage(event.message)
-        ) return
+        if (isInterceptingColorCodeMessage(event.message)) return
+        if (isInterceptingCleanMessage(event.cleanMessage)) return
+        /*
+        These
+         */
```
DavidArthurCole: ?

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
```diff
@@ -104,6 +105,31 @@ object SeaCreatureManager {
...
+    // TODO Unify when both Clean. (I do not wanna touch that Pet Storage Autopet Regex, I will horrifically break something)
+    private fun isInterceptingCleanMessage(message: String): Boolean {
+        var shouldIgnore = false
+        if (ReindrakeWarpHelper.spawnPattern.matches(message)) shouldIgnore = true
+        if (message.isEmpty()) shouldIgnore = true
+        /*
+         * Reindrakes Send an empty line, the global message & another empty line between double hook & Catch message.
+         */
+        return shouldIgnore
+    }
```
DavidArthurCole: These two can both be compacted into boolean conditions, with accompanying KDocs.

DavidArthurCole (changes requested):

hannibal002 (changes requested):
davids suggestion

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
Fazfoxy: I, think, this is correct but idk I'm quite sure the lines are too long and I'm not seeing a good split point so detekt will be being silenced

hannibal002 (approved):

PR #5560, opened 2026-04-06
Backend: Entity Features Cleanup

lunaynx:
## What
Various improvements and cleanup to entity-related features.

+ Split mob name and category display into a new Mob Display debug option. - Luna
+ Fixed Show Invisible option not working and restricted it to development environment only to prevent it from being used to cheat. - Luna
+ Fixed /shcopyentities showing the wrong amount of copied entities. - Luna
+ Fixed Vanquisher Waypoint Share keybind attempting to activate outside the Crimson Isle. - Luna
    * This is not really user-facing, as it returned early anyway and only logged a debug message.
+ Cleaned up EntityCompat methods and made them work universally on all LivingEntities. - Luna


AverageUser125:
I added "OnlyDebug" annotation, so does that like. conflict or replace your "OnlyDevEnv"?
(Or just rename/delete the OnlyDebug, your choice)

lunaynx:
We've been over this before, pretty sure they serve different purposes. Mine intentionally requires dev env specifically because it's a feature that can be used for debugging but can also be abused to cheat so it *must* be restricted to dev env.

lunaynx:
I'll split this up into more logical chunks soon.

PR #5584, opened 2026-04-11
Backend: Support private event handlers

lunaynx:
## What
Adds support for event handlers (functions annotated with `@HandleEvent`) to be private. Tested locally both with the existing public event handlers, as well as with all event handlers made private, both work fine.

+ Event handler functions are no longer required to be public. - Luna


CalMWolfs:
i dont see why this is needed

lunaynx:
Huh? It makes logical sense for event handlers to be private, no one should ever be calling them other than our event handling backend. The current limitation creates an awkward situation where we're forced to expose event handlers as part of the public API of helper classes like `ServerTime`. Sure, you could argue that public API and event handlers could be split into separate classes, but this makes that unnecessary (it can still be considered if the class is way too big and unwieldy otherwise).

hannibal002:
> no one should ever be calling them other than our event handling backend. 

I agree

CalMWolfs (approved):

PR #5587, opened 2026-04-13
Backend: Clean up verbose removeColor call chains

lunaynx:
## What
Replaced calls like `message.formattedTextCompat*.removeColor()` with `cleanMessage`, and `hoverName.{formattedTextCompat*,string}.removeColor()` with `cleanName()`.

exclude_from_changelog


CalMWolfs (commented):
code lgtm

PR #5596, opened 2026-04-15
Backend: More Lore Utils + Cache Cleanup

lunaynx:
## What
Added more extension methods for dealing with item lores, and cleaned up some nearby code and logic.

+ Item lore and NeuInternalName caches now use TimeLimitedCache instead of custom cache implementation. - Luna
+ Item lore cache is now keyed on DataComponentMap rather than ItemStack and stored as List\<Component> rather than List\<String>, allowing all methods dealing with it to benefit from it. - Luna
+ Added DataComponentMap.getLoreComponent and DataComponentMap.getCleanLore. - Luna
+ DataComponentMap.getLore no longer has a nullable receiver. - Luna
+ Added ItemStack.getCleanLore. - Luna
+ Adjusted ItemStack.getLoreComponent and ItemStack.getLore to use the underlying DataComponent methods. - Luna
+ Replaced ItemStack.getSingleLineLore with new List\<String>.toSingleLineLore method. - Luna


CalMWolfs (approved):
didnt seem to break anything

PR #5599, opened 2026-04-17
Backend: Fail at compile time for invalid HandleEvent annotations

lunaynx:
## What
Added a compile-time KSP check for invalid `@HandleEvent` annotations rather than only failing at runtime. I changed the runtime path to skip registering those event handlers instead of crashing if it somehow *does* compile anyway (it shouldn't, especially not in CI), which is the same as the current behavior with invalid event handlers that have too many arguments.

The `logger.warn()` calls that were printing informational messages were changed to `println()` because `logger.info()` doesn't show up unless I adjust the log level, which also brings in a bunch of other unwanted log messages.

+ Changed invalid @HandleEvent annotations to fail at compile time instead of only crashing at runtime. - Luna
    * At runtime, it now crashes in the development environment and otherwise skips event handler registration.


annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/VersionConstantsProcessor.kt
```diff
@@ -17,16 +17,23 @@ class VersionConstantsProcessor(
...
+                """
+                |package at.hannibal2.skyhanni.utils
+                |
+                |object VersionConstants {
+                |    const val MOD_VERSION = "$modVersion"
+                |    /**
+                |     * Do not use this Minecraft version as it is reflective of the compile-time version
+                |     * and might not be correct at runtime. We use it for the auto-updater only.
+                |     */
+                |    internal var MC_VERSION = "$mcVersion"
```
hannibal002: why changing const val to var here?

src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvents.kt
```diff
@@ -51,31 +52,52 @@ object SkyHanniEvents {
     val eventPrimaryFunctionNames: Map<String, Class<out SkyHanniEvent>> =
         GeneratedEventPrimaryFunctionNames.map
 
-    @Suppress("UNCHECKED_CAST")
+    // TODO refactor
+    @Suppress("ReturnCount")
```
hannibal002: why not refactoring the function here, in this pr?

annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/ModuleProcessor.kt
```diff
@@ -85,55 +110,97 @@ class ModuleProcessor(
...
     }
 
+    private fun KSAnnotation.hasExplicitEventSpec(): Boolean {
+        val annotationFilePath = (location as? FileLocation)?.filePath ?: return false
+        return arguments.any { argument ->
+            val value = argument.value
+            (argument.location as? FileLocation)?.filePath == annotationFilePath &&
+                when (value) {
+                    is KSType -> true
+                    is Iterable<*> -> value.all { it is KSType }
```
hannibal002: all() returns true if the list is empty. is this intended?

build.gradle.kts
```diff
@@ -453,13 +454,18 @@ afterEvaluate {
     )
 }
 
+fun SourceTask.excludeBuildDirectories() {
+    exclude { it.file.toPath().any { it.toString() == "build" } }
```
hannibal002: Implicit parameter 'it' of enclosing lambda is shadowed

build.gradle.kts
```diff
@@ -453,13 +454,18 @@ afterEvaluate {
...
+}
+
 tasks.withType<Detekt>().configureEach {
     source = source.matching {
         exclude { it.file.absolutePath.replace('\\', '/').contains("/build/generated/") }
```
hannibal002, line 492: whats the difference between this and the logic in excludeBuildDirectories?

hannibal002 (changes requested):

annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/VersionConstantsProcessor.kt
lunaynx: Old PR moment, this was a `var` at the time I made this PR.

src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvents.kt
lunaynx: Well, I could look into it, but I didn't want to make this PR too big – my change suddenly hit a threshold despite IMO not making the code noticeably harder to understand, and I considered that legacy code debt so I just added a suppression for now.

annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/ModuleProcessor.kt
lunaynx: Good catch, I'll add an `!isEmpty` check.

build.gradle.kts
lunaynx: Probably could reuse it.

lunaynx: Honestly, I'll remove the new exclude logic from this PR though, since it's not relevant. Will figure out what to do about it later and make a separate PR if needed.

lunaynx:
Added a new note to the PR description just in case you still question this later:

> The `logger.warn()` calls that were printing informational messages were changed to `println()` because `logger.info()` doesn't show up unless I adjust the log level, which also brings in a bunch of other unwanted log messages.

src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvents.kt
hannibal002: Fair, can be done in another pr after this gets merged

annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/ModuleProcessor.kt
hannibal002: Noo, add isNotEmpty instead!

annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/ModuleProcessor.kt
```diff
@@ -9,7 +9,9 @@ import com.google.devtools.ksp.processing.KSPLogger
 import com.google.devtools.ksp.processing.Resolver
 import com.google.devtools.ksp.symbol.ClassKind
 import com.google.devtools.ksp.symbol.KSAnnotated
+import com.google.devtools.ksp.symbol.KSAnnotation
```
hannibal002, line 13: wrong import order, reorder please

hannibal002 (changes requested):

hannibal002 (changes requested):
bump

annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/ModuleProcessor.kt
```diff
@@ -85,55 +110,97 @@ class ModuleProcessor(
...
         val sources = symbols.mapNotNull { it.containingFile }.toTypedArray()
-        val file = codeGenerator.createNewFile(Dependencies(true, *sources), "at.hannibal2.skyhanni.skyhannimodule", "LoadedModules")
+        val file = codeGenerator.createNewFile(
+            Dependencies(true, *sources),
+            "at.hannibal2.skyhanni.skyhannimodule",
+            "LoadedModules",
+        )
         OutputStreamWriter(file).use {
-            it.write("package at.hannibal2.skyhanni.skyhannimodule\n\n")
-            it.write("@Suppress(\"LargeClass\")\n")
```
hannibal002, line 123: why is this gone?

src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvents.kt
hannibal002: according to my intellij, return count does not flag here

hannibal002: i know we discussed this line earlier already, but i think the refaactor todo should either be more specifc (what is it we actaully want to change here) or just get removed. i dont know rihgt now of what to imporve here actually.

hannibal002 (changes requested):

annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/ModuleProcessor.kt
lunaynx: Generated files are excluded from Detekt, so this is redundant. I think they were originally made to be excluded in this PR, but then it already happened in a separate PR, so that part got removed from this PR.

hannibal002 (changes requested):
bump

src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvents.kt
```diff
@@ -54,33 +55,62 @@ object SkyHanniEvents {
...
+                    "SkyHanniEvent because it is annotated with @HandleEvent",
+            )
+            return null
+        }
+
+        @Suppress("UNCHECKED_CAST")
+        return options to listOf(eventType as Class<out SkyHanniEvent>)
+    }
+
+    private fun getEventData(method: Method): Pair<HandleEvent, List<Class<out SkyHanniEvent>>>? =
```
hannibal002: where do we check if the function has the HandleEvent annotation?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvents.kt
lunaynx: Actually, yeah, we don't check in the `else` branch. The other two short-circuit via `val options = method.options ?: return null`.

src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvents.kt
```diff
@@ -54,31 +55,64 @@ object SkyHanniEvents {
...
-                val primaryFunctionEventType = eventPrimaryFunctionNames[method.name]
-                if (primaryFunctionEventType != null) {
-                    return options to listOf(primaryFunctionEventType)
-                }
-                if (options.eventType != SkyHanniEvent::class) return options to listOf(options.eventType.java)
-                require(options.eventTypes.isNotEmpty()) {
-                    "Method ${method.name} must have at least one event type specified in @HandleEvent."
-                }
-                return options to options.eventTypes.map { it.java }
+    private val Method.options: HandleEvent? get() = getAnnotation(HandleEvent::class.java)
```
hannibal002: maybe inline this one timer used extended property

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvents.kt
lunaynx: Yeah, fair, it's only used once now.

hannibal002 (approved):

PR #5606, opened 2026-04-18
Backend: remove workflows

jani270:
exclude_from_changelog


Eveeifyeve:
Secrets still need to be changed, I have spent time to reverse engineer this.

TheDGOfficial:
I assume this happened because setup-gradle or wrapper-verification wasn't used before granting chmod +x and invoking ./gradlew in the workflow(s)? (For reference, by default files do not have execution permission. Running the chmod +x grants that permission if anyone is not familiar with unix commands.)

Edit: I wrote this before commit 6ed3f836ee143d169542ddbc58cdc2318514bc69 was committed/pushed so there was no information about this. I would suggest to use gradle/actions/wrapper-verification or [setup-gradle](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#why-use-the-setup-gradle-action) (latter also provides caching), along with [Immutable GitHub releases](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases), and make the auto updater use [Modrinth API](https://docs.modrinth.com/api/) instead of GitHub just in case.

Eveeifyeve:
@hannibal002 It would be good to publish a github security advisory.

eveeifyeve:
Also https://github.blog/changelog/2026-06-26-read-only-actions-cache-for-untrusted-triggers fixes the issue, I was contacting @JamieMagee regarding about this exact issue. Could someone test if this is still the case with the current workflows now?

PR #5611, opened 2026-04-18
Fix: Security incident faq times

Obsidianninja11:
## What
Fixes the UTC times in the security incident faq.

exclude_from_changelog


Obsidianninja11:
The times for the 7.x.x jars have been wrong. It seems like maybe the end time is actually the start time, and the end time is at 1:34utc when nopo sent the message about 7.0.0 and 7.16.1 (hopefully kaeso can confirm)

Obsidianninja11:
7.16.1 definitely existed by 00:42 utc, but it's possible (unlikely?) that it was changed as early as 8.0.0. (sourced from a screenshot by luna in contrib-chat)
7.0.0 start time is unknown, but assumed to be around the same time.
They might be tied to the prs at 00:32 and 00:35

Conversation: https://discord.com/channels/997079228510117908/1094190239532208228/1495065876460994777

PR #5615, opened 2026-04-20
Improvement: Wilted berberis solver v2

Mikecraft1224:
## What
Speeds up the Berberis Helper by looking at the spawn sequence of Berberis, since it is the same as the required breaking Sequence. This falls back to the old particle based system on:
- Initial break
- Invalid Sequence detected
- Partial Sequence exhausted

Especially invalid sequences are somewhat difficult to test requiring other players to break Berberis in the stored sequence.
Since this addition is a pure benefit and addition to the old system and is gated behind another toggle, it is set to true by default.

**Note:** I have added three ToDos, two are bugs/issues I was not able to properly fix or recreate, while one would be a not required quality of life improvement.
Especially the issue regarding skipping the particle system while rendering is very mysterious since in debugging the exact same condition copied evaluated to true, while stepping forward it set the condition to false.

## Changelog Improvements
+ Improved Berberis Helper by using the respawn sequence. - Mikecraft1224


src/main/java/at/hannibal2/skyhanni/features/rift/area/dreadfarm/RiftWiltedBerberisHelper.kt
```diff
@@ -40,13 +45,76 @@ object RiftWiltedBerberisHelper {
     private var isOnFarmland = false
     private var hasFarmingToolInHand = false
 
+    // Maps each field's center position to the number of Wilted Berberis in that field.
+    private val FIELD_CENTERS = mapOf(
```
lunaynx: These should probably be loaded from the repo instead.

src/main/java/at/hannibal2/skyhanni/features/rift/area/dreadfarm/RiftWiltedBerberisHelper.kt
```diff
@@ -152,6 +348,27 @@ object RiftWiltedBerberisHelper {
...
+            }
+        }
+    }
+
+    // ToDo: It might be possible to not instantly invalidate but instead check if there is a start point in the sequence which works and skip forward to that
+    private fun validateSequence(seq: BerberisSequence) {
+        for ((index, location) in seq.sequence.withIndex()) {
+            val block = location.getBlockAt()
+            val expectDeadBush = index >= seq.currentIndex
+            if (expectDeadBush && block != Blocks.DEAD_BUSH) { seq.invalidate(); return }
```
lunaynx: Use a multiline expression instead of semicolons.

src/main/java/at/hannibal2/skyhanni/features/rift/area/dreadfarm/RiftWiltedBerberisHelper.kt
```diff
@@ -152,6 +348,27 @@ object RiftWiltedBerberisHelper {
...
+            if (!expectDeadBush && block == Blocks.DEAD_BUSH) { seq.invalidate(); return }
```
lunaynx: Same as above.

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/features/rift/area/dreadfarm/RiftWiltedBerberisHelper.kt
Mikecraft1224: @lunaynx How would I do this? I never had to add something to the repo before.

lunaynx: You would add a file at e.g. `constants/rift/WiltedBerberisLocations.json` to [hannibal002/SkyHanni-REPO](https://github.com/hannibal002/SkyHanni-REPO), then in this repo define the schema in a data class like `WiltedBerberisLocationsJson` and load it in the feature class with `event.getConstantAsync`. Look for existing implementations in the mod for how to do these.

The data class would probably look something like this:
```kt
data class WiltedBerberisLocationsJson(
    @Expose @SerializedName("field_centers") val fieldCenters: Map<LorenzVec, Int>,
)
```

And then your JSON would look like this:
```json
{
  "field_centers": {
    "-29:71:-179": 12,
    ...more entries here...
  }
}
```

Mikecraft1224:
https://github.com/hannibal002/SkyHanni-REPO/pull/675
Also since fieldSequences is dependant on fieldCenters being present, I was unsure how to use the async function for getting constants

src/main/java/at/hannibal2/skyhanni/features/rift/area/dreadfarm/RiftWiltedBerberisHelper.kt
Mikecraft1224: changed in new commit

src/main/java/at/hannibal2/skyhanni/features/rift/area/dreadfarm/RiftWiltedBerberisHelper.kt
Mikecraft1224: Okay should be done correctly now I think

src/main/java/at/hannibal2/skyhanni/features/rift/area/dreadfarm/RiftWiltedBerberisHelper.kt
```diff
@@ -77,6 +165,9 @@ object RiftWiltedBerberisHelper {
             if (config.hideParticles && berberis != null) {
                 event.cancel()
             }
+            if (event.type == ParticleTypes.HAPPY_VILLAGER) {
+                handleRespawnParticle(location)
```
hannibal002: this should be called on main thread

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/WiltedBerberisLocationsJson.kt
```diff
@@ -0,0 +1,13 @@
+package at.hannibal2.skyhanni.data.jsonobjects.repo
+
+import at.hannibal2.skyhanni.utils.LorenzVec
+import com.google.gson.annotations.Expose
+
+data class WiltedBerberisLocationsJson(
+    @Expose val fieldCenters: List<FieldCenter>,
```
hannibal002: needs @SerializedName("field_centers")

src/main/java/at/hannibal2/skyhanni/features/rift/area/dreadfarm/RiftWiltedBerberisHelper.kt
```diff
@@ -119,13 +234,94 @@ object RiftWiltedBerberisHelper {
...
+            if (!seq.isRendering || seq.isAway) continue
+            if (seq.isAtCurrent(event.position)) {
+                seq.advance()
+                break
+            }
+        }
+    }
+
+    @HandleEvent
+    fun onBlockChange(event: ServerBlockChangeEvent) {
```
hannibal002, line 263: missing rift check

src/main/java/at/hannibal2/skyhanni/features/rift/area/dreadfarm/RiftWiltedBerberisHelper.kt
```diff
@@ -40,13 +47,74 @@ object RiftWiltedBerberisHelper {
...
+            isValid = false
+            isRendering = false
+            currentIndex = 0
+            previousTarget = null
+            lastParticleTime = SimpleTimeMark.farPast()
+        }
+
+        fun invalidate() {
+            reset()
+            ChatUtils.chat("Berberis sequence invalid, falling back to particle mode", prefixColor = "§c")
```
hannibal002, line 83: should maybe throw an user facing error with enouth debug data so we can investigate

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/rift/area/dreadfarm/RiftWiltedBerberisHelper.kt
Mikecraft1224: I understand where this is coming from but there are so many ways it could fail without the mod being at fault
Just listing some:
- you warping away and someone else breaking the crops
- you warping into another lobby
- you being too far away from the patch while respawning 
Etc.

So mostly this is just we assume something is the way we saw it last but check if it is still the case
This means this would happen regularly though mostly not very often since this is not the normal playing behavior someone would show while farming Berberis

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/WiltedBerberisLocationsJson.kt
Mikecraft1224: Does this even apply if the field in the repo is called the exact same?
Or does the naming convention for the repo require snake case for names?

src/main/java/at/hannibal2/skyhanni/features/rift/area/dreadfarm/RiftWiltedBerberisHelper.kt
Mikecraft1224: What do you mean by that exactly?
Do you mean the particle should just be queued or something and then worked through in like the on tick or smth?

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/WiltedBerberisLocationsJson.kt
lunaynx: Only if the repo field is actually using `snake_case`. But yes, that naming convention is preferred for JSONs.

(In the future we should probably add a custom Gson handler that automatically does this transformation, except for legacy files where this convention wasn't followed but can't be changed without breaking backwards compatibility.)

Mikecraft1224: Thanks for the clarification, will probably just change this then here and in the repo pr

hannibal002 (changes requested):
same as before

Mikecraft1224:
The breakdown into smaller functions looks good to me and is probably a lot more readable if you are looking at this without prior knowledge

src/main/java/at/hannibal2/skyhanni/features/rift/area/dreadfarm/RiftWiltedBerberisHelper.kt
lunaynx: Wrap it in `DelayedRun.runOrNextTick { ... }` presumably.

hannibal002 (approved):

PR #5618, opened 2026-04-21
Backend + Fix: Updater Rework

lunaynx:
## What
Reworked the updater so no more auto-update or queued installation, only link-outs. libautoupdate has been neutered as much as I could – it still runs a harmless no-op shutdown hook, we can update it later to add the ability to remove that. Also it now pulls update info and changelogs from Modrinth by default, but there is a debug option to switch it back to GitHub. While we're here, I also removed the ability to disable update checks, since we really want users to be aware of updates, especially since certain modpacks have tried to auto-disable them in the past, and it's a completely non-intrusive chat message that appears once per game start. Plus fixed a logic error in the GitHub update check that prevented updates from being found at all if the newest release doesn't have a JAR for the current Minecraft version.

## Changelog Fixes
+ Fixed the mod not finding updates if the newest SkyHanni version doesn't have a download available for your current Minecraft version. - Luna

+ Removed the ability to automatically download updates. - Luna
+ Added Modrinth as a new source for update checks and changelogs, replacing GitHub as default. - Luna
+ Changed the "Download update" button to open the browser at the relevant Modrinth or GitHub release instead of queuing the update to be automatically installed. - Luna
+ Removed the ability to disable update notifications. - Luna
    * These notifications only appear once per game start unless you manually check for updates.


src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangelogViewer.kt
```diff
@@ -62,39 +65,62 @@ object ChangelogViewer {
...
+                    while (true) {
+                        val pagedUrl = source.releaseApiUrl + mapOf(
+                            "per_page" to 100,
+                            "page" to pageNumber,
+                        ).toQueryString()
+                        val (_, jsonObject) = ApiUtils.getJsonResponse(pagedUrl, apiName = "github")
+                            .assertSuccessWithData()
+                            ?: error("Changelog Loading Failed")
+
+                        val page = Gson().fromJson<List<GithubRelease>>(jsonObject)
```
hannibal002: why do we use new gson here and not the existing instance from config manager?

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
```diff
@@ -121,6 +121,8 @@ object SkyHanniMod : CompatCoroutineManager by SkyHanniCoroutineManager(
     val isBetaVersion: Boolean
         get() = modVersion.isBeta
 
+    val userAgent: String = "SkyHanni/${SkyHanniMod.VERSION}-${PlatformUtils.MC_VERSION}"
```
hannibal002: unnecessary SkyHanniMod here

src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
```diff
@@ -93,92 +108,88 @@ object UpdateManager {
...
+                },
+                Minecraft.getInstance(),
+            )
     }
 
-    fun queueUpdate() {
-        if (updateState != UpdateState.AVAILABLE) {
-            logger.log("Trying to enqueue an update while another one is already downloaded or none is present")
+    fun openDownloadPage() {
+        val update = potentialUpdate ?: error("Attempted to open update download page with no known update")
```
hannibal002: the error call here is not protected by skyhanni error logic

src/main/java/at/hannibal2/skyhanni/features/misc/update/ModrinthUpdateSource.kt
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Locale
+import java.util.concurrent.CompletableFuture
+
+class ModrinthUpdateSource(private val projectId: String, private val slug: String) : JsonUpdateSource() {
+
+    override fun getGson(): Gson = ConfigManager.gson
+
+    fun getReleases(includeChangelog: Boolean = false): CompletableFuture<List<ModrinthRelease>?> =
+        getJsonFromURL(
+            String.format(Locale.ROOT, MODRINTH_API_URL, projectId) + mapOf(
```
hannibal002: the url that gets created here does not filter, it shows me forge versions and other mc versions.

src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangelogViewer.kt
```diff
@@ -124,12 +153,12 @@ object ChangelogViewer {
...
-        .replace("(\n[ \t]+)[+\\-*][^+\\-*]".toRegex(), "$1§7") // Formatting for sub points
+        .replace("(\n[ \t]+)[+\\-*][^+\\-*]".toRegex(), "$1§7") // Formatting for subpoints
         .replace("\n[+\\-*][^+\\-*]".toRegex(), "\n§a") // Formatting for points
         .replace("(- [^-\r\n]*(?:\r\n|$))".toRegex(), "§b§l$1") // Color contributors
-        .replace("\\[(.+?)\\]\\(.+?\\)".toRegex(), "$1") // Random Links
-        .replace("`", "\"") // Fix Code Blocks to look better
-        .replace("§l§9(?:Version|SkyHanni)[^\r\n]*\r\n".toRegex(), "") // Remove Version from Body
+        .replace("\\[(.+?)]\\(.+?\\)".toRegex(), "$1") // Random links
+        .replace("`", "\"") // Fix code blocks to look better
+        .replace("§l§9(?:Version|SkyHanni)[^\r\n]*\r\n".toRegex(), "") // Remove version from body
```
hannibal002, line 166: are you sure modrinth and github both use the same new lines logic?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangelogViewer.kt
lunaynx: Because our instance skips fields without `@Expose` annotations, which libautoupdate's `GithubRelease` class doesn't have.

src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangelogViewer.kt
lunaynx: It worked fine in my testing, just had to allow omitting `\r` for Modrinth.

src/main/java/at/hannibal2/skyhanni/features/misc/update/ModrinthUpdateSource.kt
lunaynx: Good catch, this looks like a potential bug in the Modrinth API. I'll ask them, and if we can't resolve it soon, I'll fall back to filtering the output on our side.

lunaynx: Never mind, apparently when the docs used `["fabric"]` as an example, they meant literally pass it like that.

src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangelogViewer.kt
hannibal002: please put that in the comment above the line

src/main/java/at/hannibal2/skyhanni/features/misc/update/ModrinthUpdateSource.kt
hannibal002: any chance we dont use the api properly? e.g. forgetting about a wrapper of the fields? like it expects a an array as value or someting?

src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
lunaynx: I think it should be? The documentation for `ErrorManager.skyHanniError` says to use `error()` if you don't need extra data, and that method does nothing more than cache extra data and throw `IllegalStateException`, which is the same exception as the one thrown by `error()`.

hannibal002: the issue is that we dont catch the error thrown here. there is no try catch block protecting this. if this error shows up, the game should crash right now.

src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
```diff
@@ -94,98 +107,87 @@ object UpdateManager {
...
+                Minecraft.getInstance(),
+            )
     }
 
-    private fun suggestModrinth() {
-        ChatUtils.clickableChat("§eClick here to manually download the version from Modrinth.", onClick = {
-            OSUtils.openBrowser("https://modrinth.com/mod/skyhanni")
-        })
+    fun getDownloadPage(): String {
+        val update = potentialUpdate ?: error("Attempted to open update download page with no known update")
```
hannibal002: still needs proper error handling

src/main/java/at/hannibal2/skyhanni/features/misc/update/ModrinthUpdateSource.kt
```diff
@@ -0,0 +1,40 @@
+package at.hannibal2.skyhanni.features.misc.update
+
+import at.hannibal2.skyhanni.config.ConfigManager
+import at.hannibal2.skyhanni.utils.StringUtils.toQueryString
+import at.hannibal2.skyhanni.utils.system.PlatformUtils
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
+import moe.nea.libautoupdate.JsonUpdateSource
+import moe.nea.libautoupdate.UpdateData
+import java.util.Locale
```
hannibal002: unused import

src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
```diff
@@ -86,6 +98,7 @@ object UpdateManager {
     }
 
     private val config get() = SkyHanniMod.feature.about
+    private val debugConfig get() = SkyHanniMod.feature.dev.debug
 
     fun reset() {
```
hannibal002, line 103: hasCheckedForUpdate doesnt get reset here

src/main/java/at/hannibal2/skyhanni/features/misc/update/ChangelogViewer.kt
```diff
@@ -62,39 +65,67 @@ object ChangelogViewer {
...
+                            "per_page" to 100,
+                            "page" to pageNumber,
+                        ).toQueryString()
+                        val (_, jsonObject) = ApiUtils.getJsonResponse(pagedUrl, apiName = "github")
+                            .assertSuccessWithData()
+                            ?: error("Changelog Loading Failed")
+
+                        // We cannot use ConfigManager.gson here because it excludes fields without
+                        // @Expose annotations, and GithubRelease comes from libautoupdate
+                        val page = Gson().fromJson<List<GithubRelease>>(jsonObject)
```
hannibal002: please create the gson object once in this file somewhere, not repeatedly in this function or even inside the loop

src/main/java/at/hannibal2/skyhanni/features/misc/update/ModrinthUpdateSource.kt
```diff
@@ -0,0 +1,40 @@
...
+                "game_versions" to gson.toJson(listOf(PlatformUtils.MC_VERSION)),
+                "include_changelog" to includeChangelog,
+            ).toQueryString(),
+            object : TypeToken<List<ModrinthRelease>>() {}.type,
+        )
+
+    override fun checkUpdate(updateStream: String): CompletableFuture<UpdateData?>? =
+        getReleases().thenApply { releases ->
+            releases?.asSequence()
+                ?.filter { it.versionType.updateStream == updateStream }
```
hannibal002: if update sttream is pre/beta, then the user will not see the probably newer full release

hannibal002 (changes requested):

hannibal002 (approved):

PR #5622, opened 2026-04-22
Backend: Remove deprecated eventTypes parameter from HandleEvent

lunaynx:
## What
Removes the deprecated `eventTypes` parameter from `@HandleEvent` and replaces existing usages with individual event handlers. This is just some early low-hanging fruit cleanup since it is a very small diff to review compared to removing `eventType`.

+ Removed deprecated eventTypes parameter from HandleEvent. - Luna


lunaynx:
Not that important, and bound to conflict with #6064.

PR #5624, opened 2026-04-23
Backend: Remove 1.21.10 support

lunaynx:
## What
Removes remaining traces of 1.21.10 support from the codebase and fixes newly appearing Detekt violations as a result of this. I decided to separate this out because the 26.1 support PR has stalled.

+ Fully removed 1.21.10 support. - Luna


NopoTheGamer (approved):
im like pretty sure this breaks the actions but like xd

PR #5629, opened 2026-04-24
Fix: Auto updater

Rainymei:
## What
Autoupdater was resetting to beta branch at any attempt to set it to Releases

## Changelog Fixes
+ Fixed Autoupdater Update Stream automatically changing to Beta when Full was selected. - Rain


Fazfoxy:
Should be user facing changelog probably

probably "Fixed Autoupdater Update Stream automatically changing to beta when Full was selected"

PR #5631, opened 2026-04-24
Fix: No config write when on alpha

Rainymei:
## What
Saving data to config now returns nothing if on alpha server.
This issue was causing data sync and overwriting errors when switching between the 2 servers.

## Changelog Fixes
+ Fixed config being written when on Alpha. - Rain
    * The mod won't write to config anymore when on Alpha, removing data sync/overwriting errors between Alpha and prod.


lunaynx:
This should probably be disabled in dev env?

Rainymei:
true, added it

PR #5633, opened 2026-04-24
Backend: Optimize repo downloads

lunaynx:
## What
Changed repo download code to remove JGit and use `.tar.gz` instead of `.zip`, both for backup repo and local download.

JGit did not help with repo issues, it only caused more headaches (remote not being updated, IntelliJ getting confused by the git subprojects, etc).

This uses Apache Commons Compress, which is already bundled with Minecraft, so no additional dependencies required. I've also changed it so that it deletes any old repo `.tar.gz` or `.zip` files after extracting for local downloads. Overall, this saves about 10 MB on the .jar size.

+ Changed backup repo from .zip to .tar.gz. - Luna
+ Changed repo download from JGit/.zip to .tar.gz. - Luna


hannibal002 (changes requested):
can stuff like the function internalGetZipResponse in ApiInternalUtils, the two functions getZipResponse in ApiUtils or the class ZipApiResponse get renamed now?

lunaynx:
> can stuff like the function internalGetZipResponse in ApiInternalUtils, the two functions getZipResponse in ApiUtils or the class ZipApiResponse get renamed now?

Done. Turns out those were misleadingly named – they have zero zip-specific logic, they're simply for fetching binary data.

hannibal002 (approved):

PR #5635, opened 2026-04-25
Backend: Update credits

lunaynx:
## What
We were still crediting NEU and Forge.

+ Updated credits for third party software. - Luna
    * Replaced NotEnoughUpdates with NotEnoughUpdates-REPO.
    * Replaced Forge with Fabric Loader and Fabric API.


src/main/java/at/hannibal2/skyhanni/config/features/About.kt
```diff
@@ -45,20 +45,24 @@ class About {
...
+        val fabricApi: Runnable = Runnable { openBrowser("https://github.com/FabricMC/fabric-api") }
 
         @ConfigOption(name = "LibAutoUpdate", desc = "LibAutoUpdate is available under the BSD 2 Clause License")
         @ConfigEditorButton(buttonText = "Source")
         val libAutoUpdate: Runnable = Runnable { openBrowser("https://github.com/nea89o/libautoupdate") }
 
         @ConfigOption(name = "Mixin", desc = "Mixin is available under the MIT License")
         @ConfigEditorButton(buttonText = "Source")
-        val mixin: Runnable = Runnable { openBrowser("https://github.com/SpongePowered/Mixin/") }
+        val mixin: Runnable = Runnable { openBrowser("https://github.com/SpongePowered/Mixin") }
```
AzureAaron: Might be better to credit Fabric's fork since that's what's used and its significantly diverged from its upstream: https://github.com/FabricMC/Mixin

Could potentially consider Mixin Extras too if y'all want to go the extra step

PR #5644, opened 2026-04-27
Feature: Dungeon Profit Tracker

cOnfusi0n1:
## What
Renamend DungeonFloor to DungeonBoss, added a new DungeonFloor enum that also stores the master mode state.
Copied pest profit tracker into dungeon and started with display logic.

<details>
<summary>Images</summary>

image

image

</details>

## Changelog New Features
+ Added Cool new feature. - your_name_here
    * Optional extra info.

+ Something technical you changed in the backend. - your_name_here
    * Optional extra info.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
@@ -140,7 +140,7 @@ class ProfileSpecificStorage(
 
         // we can not reuse runs data because it is inaccurate and gets autocorrected via opening the chest.
         @Expose
-        var availableCroesus: MutableMap<DungeonFloor, CroesusStorage> = mutableMapOf()
+        var availableCroesus: MutableMap<DungeonFloor, CroesusStorage> = enumMapOf()
 
         @Expose
         var profitTracker: DungeonProfitTracker.BucketData = DungeonProfitTracker.BucketData()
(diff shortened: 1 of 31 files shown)


src/main/java/at/hannibal2/skyhanni/config/storage/ProfileSpecificStorage.kt
```diff
@@ -133,7 +136,14 @@ class ProfileSpecificStorage(
 
     class DungeonStorage {
         @Expose
-        var bosses: MutableMap<DungeonFloor, Int> = enumMapOf()
+        var bosses: MutableMap<DungeonBoss, Int> = enumMapOf()
+
+        // we can not reuse runs data because it is inaccurate and gets autocorrected via opening the chest.
+        @Expose
+        var availableCroesus: MutableMap<DungeonFloor, CroesusStorage> = mutableMapOf()
```
Rainymei: prob use enumMapOf() instead of mutable

Killer560-OG:
NEU Used to have a feature on 1.8 that would show the most effective item to put rng meter on assuming your goal is simply profit. It essentially took the auction price or bazaar price of an item and compared that to the score/meter required to drop said item. So for instance if you have an item worth 1000 coins and it took 1000 meter, it would have a value of 1. For dungeons it worked slightly differently such that it would compare the cost of opening the chest as well into the equation. For instance if the chest cost of handle is 100m, the handle sells for 500m, and lets say for simple numbers it only took 100k meter, it would be 400m of profit divided on 100k meter, meaning each score was worth 4000. For something like slayers you could also implement this into the "bosses until meter" portion, which would also factor in boss spawn cost to show how profitable your rng meter item is. You could also have it highlight in the rng menu whatever item is the current best option for profit.

PR #5654, opened 2026-04-28
Fix: rare crop regexes

Rainymei:
## What
rare crop message got stripped of `Armor Set Bonus`

## Changelog Fixes
+ Fixed Armor Drop Tracker not tracking rare crop drops. - Rain


Rainymei:
this is gonna need to manually run the repo pattern update

PR #5655, opened 2026-04-28
Fix: Rarity Line v2

lunaynx:
## What
I thought it would be a good cleanup to change the `+` to a `*`, because it's misused in a lot of other places. Turns out, it was not.

## Changelog Fixes
+ Fixed error getting item category on items that have no category. - Luna


hannibal002 (changes requested):
I am a bit confused, the pr desc explains that changing the + to a * is wrong, still, thats what you do in the pr. am i missing something?

lunaynx:
It was changed in an earlier PR of mine as routine cleanup, because I *thought* there had to be at least one character to match there. But that isn't true, there are items with no category which will have zero characters there.

PR #5656, opened 2026-04-28
Fix: Custom Keybinds bugginess

lunaynx:
## What
Fixed breakages caused by my recent toggle support, such as keys sometimes not registering (primarily noticed this with multi-binds) and keys sometimes getting stuck in held state.

## Changelog Fixes
+ Fixed issues with keys not registering and keys getting stuck with Garden Custom Keybinds. - Luna

PR #5664, opened 2026-04-29
Fix: Flying with mouse keybind on Garden

lunaynx:
## Dependencies
- #5656

## What
Fixes a longstanding issue where having your Jump key bound to a mouse button in Custom Keybinds wouldn't let you start flying by double clicking.

## Changelog Fixes
+ Fixed not being able to start fly by double clicking if Jump is bound to a mouse button in Garden Custom Keybinds. - Luna


lunaynx:
Superseded by #5685.

PR #5666, opened 2026-04-30
Improvement: Multi Endcap Support

lunaynx:
## Dependencies
- #5655 (soft dependency, really just so that it can be tested without enchant parsing in Estimated Item Value showing everything as "Enchanted Book")
- https://github.com/hannibal002/SkyHanni-REPO/pull/689

## What
Adds support for enchantments to be upgraded multiple times by endcap items. Because apparently Hypixel doesn't know what the "end" in endcap means. This is needed for Turbo-Crop VI and VII enchantments.

You can test this by searching for Advanced Gardening Hoes on AH. (It is intentional that only one Turbo Gourd and one Enchanted Turbo Gourd is counted per item, because applying one of those upgrades all Turbo-Crop enchants by one level.)

## Changelog Improvements
+ Added support for Turbo-Crop VI and VII in Estimated Item Value. - Luna


hannibal002 (changes requested):
waiting on confirmation

PR #5672, opened 2026-05-03
Remove Ding on Lane Swap notification since it's overly abusable.

Fazfoxy:
## What
I've at least twice talked about how I think this oversteps the line of allowed features, I'm deleting it.

my entire justification for PRing this is just 
image
🤷‍♀️ 

## Changelog Removed Features
+ Removed Lane Switch Noise. - Fazfoxy
    * In my opinion this feature overstepped the boundary of permissable features..


hasbeenben:
This is just silly. If people want "admin detection" they are likely cheating, and their macro already has the feature. If not, they are just not paying attention to their game enough, and is their own fault if they miss any other macro checks. Removing this just negatively impacts legit players, and cheaters are unaffected.

Nealkitt:
i don't see the point in removing the dings, i treat it like the same way as a "solver" or just another accessibility feature, you already _have the ability_ to memorise said timings, especially because they're the same timings for the same crop every single time, all this does is make it 10% easier and have you do less memorisation, if we remove this for this reason there's certainly a lot of code that would fall under similar guidelines, and all things considered it isn't really cheaty, I believe the talk a year ago that said that it wouldn't ever show if you were interrupted farming is enough, all this really does is provide an audio queue for a set timing, it's the equivalent of setting a timer when farming, which is allowed, this timer is just calibrated within two rows of farming.

Fazfoxy:
—

PR #5673, opened 2026-05-03
Fix: crop milestone progress resetting and show Wrong data 

legentpc:
## What
Crop milestone progress display was showing an incorrect value until
/cropmilestones was opened.

An issue where crop milestone progress showed incorrect data on the first use of /cropmilestones, but displayed correctly on subsequent uses.

## Changelog Fixes
+ Fixed an issue where crop milestone progress showed incorrect data on the first use of /cropmilestones, but displayed correctly on subsequent uses. — legendpc


legentpc:
Maybe no need 
it is work fine now

PR #5674, opened 2026-05-03
Feature: Crash on Skyblock Join

Chissl:
## What
Unfortunately, players interacting with skyblock encourages actions that result in them getting banned or wiped. Skyhanni, in good conscience, should not sit idly by and allow this to happen. Anyone who disagrees with my idea is "fucking useless" and should try to not be a twit. 

## Changelog New Features
+ Added untogglable feature that prevents players from accessing skyblock. - Chissl

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/CrashOnJoin.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/CrashOnJoin.kt
@@ -14,7 +14,7 @@ object CrashOnJoin {
         Minecraft.getInstance().delayCrash(
             CrashReport(
                 "SkyHanni Crash on Join",
-                Throwable("Joining Skyblock may lead to bans!")
+                Throwable("Joining SkyBlock may lead to bans!")
             ),
         )
     }
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/CrashOnJoin.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/CrashOnJoin.kt
@@ -1,0 +1,21 @@
+package at.hannibal2.skyhanni.features.misc
+
+import at.hannibal2.skyhanni.SkyHanniMod
+import at.hannibal2.skyhanni.api.event.HandleEvent
+import at.hannibal2.skyhanni.events.ProfileJoinEvent
+import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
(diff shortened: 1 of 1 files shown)


lunaynx (approved):
lgtm

Nealkitt (approved):

Rainymei:
lgtm

Rainymei (approved):

Fazfoxy (approved):

src/main/java/at/hannibal2/skyhanni/features/misc/CrashOnJoin.kt
```diff
@@ -0,0 +1,21 @@
...
+import net.minecraft.client.Minecraft
+
+@SkyHanniModule
+object CrashOnJoin {
+    @HandleEvent
+    fun onProfileJoin(event: ProfileJoinEvent) {
+        Minecraft.getInstance().delayCrash(
+            CrashReport(
+                "SkyHanni Crash on Join",
+                Throwable("Joining Skyblock may lead to bans!")
```
Bloxigus: should be SkyBlock, otherwise lgtm

Bloxigus (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/CrashOnJoin.kt
Chissl: Fixed, thanks for letting me know. We averted a potentially critical issue that could have occurred when this gets merged.

Bloxigus (approved):

mariethepolarfox (approved):

Rainymei:
i suggest crashing the instance as soon as it launches considering you only use skyhanni for skyblock :pleading_face:

mariethepolarfox:
> i suggest crashing the instance as soon as it launches considering you only use skyhanni for skyblock 🥺

This is false. Skyhanni is essential when playing roblox

Chissl:
> i suggest crashing the instance as soon as it launches considering you only use skyhanni for skyblock :pleading_face: 

I was considering this, but figured that would be too intrusive and could potentially lead to players removing the mod. Ultimately, both implementations achieve the same goal but this leads to less unintended consequences

Chissl:
After some careful consideration, I have decided that it is not sufficient for any mod to implement this feature, and should be implemented by hypixel themselves.

PR #5685, opened 2026-05-05
Fix: Delegate Custom Keybinds handling to Minecraft

lunaynx:
## What
This fixes some more long-standing bugs with Custom Keybinds and makes the implementation safer by reducing the amount of custom handling and making it be more closely handled by Minecraft directly to the extent it is possible. A few special cases are still needed to avoid unexpected input stickiness.

The fixes listed below are the ones I know about as a result of this PR, but it's not necessarily exhaustive.

## Changelog Fixes
* Fixed not being able to double click to fly if Jump is bound to a mouse button with Custom Keybinds. - Luna
* Fixed some Custom Keybinds input behavior being inconsistent with vanilla Minecraft on macOS. - Luna


lunaynx:
Drafting while I investigate some potential platform-specific jank.

zumbiepig:
> * Fixed some Custom Keybinds input behavior being inconsistent with vanilla Minecraft on macOS. - Luna

does this fix the bug where if you are holding a hoe, and you type something like /chips then press enter quickly then you will move backwards a little? and the same thing with /wd moving right

lunaynx:
Yes, but only on macOS, because on Windows/Linux vanilla keybinds do this too, so it's kinda out of scope for us to fix it.

lunaynx:
Making it a draft again for now because of a small issue I noticed.

PR #5686, opened 2026-05-05
Fix: Misleading tool XP gain message

lunaynx:
## What
It turns out, you *do* gain tool XP when farming a mismatching crop, but Hypixel doesn't update the NBT data, only the lore. I really don't want to implement lore parsing for this, so I'll just fix our misleading warning message instead.

## Changelog Fixes
+ Fixed Hoe Levels Display incorrectly claiming that you are not gaining XP if you are farming the wrong crop. - Luna
    * You do gain XP, but we cannot track it in real time because Hypixel doesn't update the NBT data.


lunaynx:
Never mind, this is a Hypixel bug causing XP to be gained from Seeds.

PR #5692, opened 2026-05-07
Fix: Make Personal Sea Creature cap 10 Everywhere

AverageUser125:
## What
Since the new update:
https://hypixel.net/threads/hypixel-skyblock-0-24-5-assorted-qol-changes.6094244/
The personal sea creature cap is 10 in all islands.

## Changelog Fixes
+ Fixed Personal Sea Creature Cap. - AverageUser125


src/main/java/at/hannibal2/skyhanni/features/fishing/BarnFishingTimer.kt
```diff
@@ -34,23 +34,10 @@ object BarnFishingTimer {
 
     private val config get() = SkyHanniMod.feature.fishing.barnTimer
     private const val GLOBAL_CAP = 60
+    private const val PERSONAL_CAP = 10;
```
lunaynx:
```suggestion
    private const val PERSONAL_CAP = 10
```

lunaynx (changes requested):

lunaynx (approved):
Looks good for the most part, though I think there's still a 30 global cap on Galatea, you should probably add that.

(Ideally it should be repo-based in case there are further changes/exceptions in the future, but we should get this in for 7.21.0 either way.)

AverageUser125:
I feel like changing the global cap is outside the scope of this PR atleast.

lunaynx:
I'll just do that + repo-ify it in a follow-up PR then.

hannibal002 (approved):

PR #5696, opened 2026-05-08
Fix: Bandaid fix all possible ESP's with skull wearing entities

Fazfoxy:
## What
We've had this issue since 1.21.10, this should fix all examples
You can complain this is a bandaid but a bandaid is better than letting people latch onto this if they heavily notice.

exclude_from_changelog


Fazfoxy:
superceded by https://github.com/hannibal002/SkyHanni/pull/5700

PR #5702, opened 2026-05-09
Fix: Don't suppress discontinued version notification if extraInfo is missing

lunaynx:
## What
UpdateManager was skipping sending the discontinued version notification if `extraInfo` was missing, even though there is already a perfectly functional notification that can be sent even without `extraInfo`.

If necessary, I can add a field to disable the notification, but this subtle behavior is not a good way to go about it.

## Changelog Fixes
+ Fixed discontinued Minecraft version notification not being sent in some cases. - Luna


NopoTheGamer:
this was on purpose

lunaynx:
I'll figure something out to keep a way to disable the notification then, but the current way is pretty unintuitive and easy to accidentally not send a notification when you meant to.

lunaynx:
I'll just open a new PR with something later.

PR #5706, opened 2026-05-09
Fix: Fishing tracker 

Rainymei:
## What
Fishing tracker now ignores hotbar slot 9 so the baits don't get added to profit
bandage fix, optimal would be ignoring hotbar slot 9 globally

## Changelog Fixes
+ Fixed the Fishing Profit Tracker detecting the baits in hotbar slot 9. - Rain

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/FishingProfitTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/FishingProfitTracker.kt
@@ -195,6 +195,7 @@ object FishingProfitTracker {
 
     private var slot9Item: NeuInternalName? = null
 
+    @HandleEvent
     fun onInventoryUpdate(event: OwnInventoryItemUpdateEvent) {
         if (event.slot == 44) {
             slot9Item = event.itemStack.getInternalNameOrNull()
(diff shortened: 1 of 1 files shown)


AverageUser125:
I think what I did in #5698 with the bait detection can help with this problem.
It would still be a bandage fix, but a more proper bandage fix I suppose.

NopoTheGamer (changes requested):
item add event shouldnt send anything that gets changed in slot 9

src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/FishingProfitTracker.kt
```diff
@@ -191,9 +193,18 @@ object FishingProfitTracker {
         }
     }
 
+    private var slot9Item: NeuInternalName? = null
+
+    fun onInventoryUpdate(event: OwnInventoryItemUpdateEvent) {
```
hannibal002, line 199: missing annotation

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/FishingProfitTracker.kt
```diff
@@ -191,9 +193,19 @@ object FishingProfitTracker {
         }
     }
 
+    private var slot9Item: NeuInternalName? = null
+
+    @HandleEvent
+    fun onInventoryUpdate(event: OwnInventoryItemUpdateEvent) {
+        if (event.slot == 44) {
```
hannibal002, line 200: please use a constant variable for the hotbar slot id and give it a good name, so that its obvious what we check against here.

hannibal002 (changes requested):

hannibal002 (changes requested):
bump

PR #5708, opened 2026-05-09
Fix: isFishingDetection now uses Ownership/lootshare range

Fazfoxy:
## What
In theory this might fail detect if someone is just, chilling next to someone fishing without LSing but like, why.

## Changelog Improvements
+ Improved Fishing Tracker to appear when within lootshare range of other players' mobs. - Fazfoxy

## Changelog Fixes
+ Fixed Fishing Tracker disappearing while killing mobs occasionally. - Fazfoxy

+ Moved lootshare-related rendering and range to LootshareUtils. - Fazfoxy


src/main/java/at/hannibal2/skyhanni/utils/LootshareUtils.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    private val existingCircles = mutableMapOf<LorenzVec, Color>()
+
+    @Suppress("HandleEventInspection")
+    fun renderLootshareSphere(event: SkyHanniRenderWorldEvent) {
+        var tooCloseSpheres = 0
+        for (sphere in existingCircles) {
+            if (sphere.key.distance(sphere.key) < 10) tooCloseSpheres++
+            if (tooCloseSpheres > 2) continue
+            event.drawSphereWireframeInWorld(sphere.value, sphere.key, RANGE)
```
hannibal002: this function gets called not only once, but once per sea creature (loop in loop)

src/main/java/at/hannibal2/skyhanni/utils/LootshareUtils.kt
```diff
@@ -0,0 +1,30 @@
...
+import at.hannibal2.skyhanni.events.minecraft.SkyHanniRenderWorldEvent
+import at.hannibal2.skyhanni.utils.LocationUtils.distanceToPlayer
+import at.hannibal2.skyhanni.utils.render.WorldRenderUtils.drawSphereWireframeInWorld
+import java.awt.Color
+
+object LootshareUtils {
+
+    const val RANGE = 30.0f
+
+    private val existingCircles = mutableMapOf<LorenzVec, Color>()
```
hannibal002: map can grow indefinetly. should have a clear/reset logic

src/main/java/at/hannibal2/skyhanni/utils/LootshareUtils.kt
```diff
@@ -0,0 +1,30 @@
+package at.hannibal2.skyhanni.utils
+
+import at.hannibal2.skyhanni.events.minecraft.SkyHanniRenderWorldEvent
+import at.hannibal2.skyhanni.utils.LocationUtils.distanceToPlayer
+import at.hannibal2.skyhanni.utils.render.WorldRenderUtils.drawSphereWireframeInWorld
+import java.awt.Color
+
+object LootshareUtils {
```
hannibal002, line 5: class name is too broad. how about LootshareRenderer?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/LootshareUtils.kt
Fazfoxy: Removed the rendering changes (they're kinda pointless refactoring as it stands and confuses the purpose of the LootshareUtils (to store useful information in relation to Lootshare)

src/main/java/at/hannibal2/skyhanni/features/fishing/IsFishingDetection.kt
```diff
@@ -6,8 +6,10 @@ import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.EntityUtils
 import at.hannibal2.skyhanni.utils.LocationUtils
 import at.hannibal2.skyhanni.utils.LocationUtils.distanceToPlayer
+import at.hannibal2.skyhanni.utils.LootshareUtils.isInLootshareRange
```
hannibal002: do not use function imports (except for extended functions)

src/main/java/at/hannibal2/skyhanni/utils/LootshareUtils.kt
```diff
@@ -0,0 +1,10 @@
+package at.hannibal2.skyhanni.utils
+
+import at.hannibal2.skyhanni.utils.LocationUtils.distanceToPlayer
+
+object LootshareUtils {
+
+    const val RANGE = 30.0f
+
+    fun isInLootshareRange(pos: LorenzVec): Boolean = pos.distanceToPlayer() < RANGE
```
hannibal002: since this is in the lootshare class, the function can be renamed to "isInRange"

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
```diff
@@ -15,12 +16,9 @@ import at.hannibal2.skyhanni.utils.render.WorldRenderUtils.drawSphereWireframeIn
 object SeaCreatureLootshareSphere {
     private val config get() = SkyHanniMod.feature.fishing
 
-    private const val RANGE = 30.0f
-
     private val seaCreatures = mutableSetOf<LivingSeaCreatureData>()
-    private val existingCircles = mutableSetOf<LorenzVec>()
 
-    fun isInRange(pos: LorenzVec): Boolean = pos.distanceToPlayer() < RANGE
+    private val existingCircles = mutableSetOf<LorenzVec>()
```
hannibal002: given we clear the set each time before we start using it, why do we need this outside the function? can be a inside function variable

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
```diff
@@ -32,16 +30,16 @@ object SeaCreatureLootshareSphere {
     fun onRenderWorld(event: SkyHanniRenderWorldEvent) {
         if (!config.lootshareRange) return
         existingCircles.clear()
+        var circleCount = 0
```
hannibal002: moving circle count outside the loop, this will stop render some stuff. intended?

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
```diff
@@ -32,16 +30,16 @@ object SeaCreatureLootshareSphere {
...
         for (seaCreature in seaCreatures) {
             if (!seaCreature.exists()) continue
             val pos = seaCreature.pos ?: continue
-            var circleCount = 0
+            val color = if (seaCreature.isOwn || isInLootshareRange(pos)) LorenzColor.GREEN else LorenzColor.WHITE
```
hannibal002: the color variable can be defined after the circleCount > 2 check below.

hannibal002 (changes requested):

hannibal002 (changes requested):
my review above

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
Fazfoxy: I can't tell if I'm really stupid for not understanding this but, the point is that circleCount is meant to stop stuff like Fiery Scuttlers rendering 7+ Boxes in a short area dropping framerates (In theory we want to just detect them as 1 Sea Creature but idk how that really works out)

hannibal002: the concept of not rendering everything is good. but we need to do this properly. the current impl will just avoid rendering some stuff, and has the logic for that filter in the render thread. that should be a tick event thingy

Fazfoxy: This should be what you meant (I think I've also accidentally done like half the genericification necessary to make this happen somewhere else tbh)

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
```diff
@@ -15,12 +15,14 @@ import at.hannibal2.skyhanni.utils.render.WorldRenderUtils.drawSphereWireframeIn
...
 
-    private const val RANGE = 30.0f
+    private val seaCreatures = mutableListOf<LivingSeaCreatureData>()
 
-    private val seaCreatures = mutableSetOf<LivingSeaCreatureData>()
-    private val existingCircles = mutableSetOf<LorenzVec>()
+    private val spherePositions = mutableListOf<LootshareSphere>()
 
-    fun isInRange(pos: LorenzVec): Boolean = pos.distanceToPlayer() < RANGE
+    data class LootshareSphere(
```
hannibal002: please put the term lootshare in the shared dictionary

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
```diff
@@ -31,18 +33,26 @@ object SeaCreatureLootshareSphere {
     @HandleEvent(onlyOnSkyblock = true)
     fun onRenderWorld(event: SkyHanniRenderWorldEvent) {
         if (!config.lootshareRange) return
-        existingCircles.clear()
+        scLoop@ for (lootshareSphere in spherePositions) {
+            if (lootshareSphere.position == null) return
```
hannibal002: why return here? continue looks better

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
```diff
@@ -31,18 +33,26 @@ object SeaCreatureLootshareSphere {
...
+            event.drawSphereWireframeInWorld(lootshareSphere.color.toColor(), lootshareSphere.position, LootshareUtils.RANGE)
+            spherePositions.remove(lootshareSphere)
```
hannibal002: concurrent modificatino exxception

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
```diff
@@ -31,18 +33,26 @@ object SeaCreatureLootshareSphere {
...
+        val seaCreaturesToRender = mutableSetOf<LorenzVec>()
         for (seaCreature in seaCreatures) {
             if (!seaCreature.exists()) continue
+            var otherNearbySpheres = 0
             val pos = seaCreature.pos ?: continue
-            var circleCount = 0
-            existingCircles.forEach {
-                if (it.distance(pos) < 10) circleCount++
+            val color = if (seaCreature.isOwn || LootshareUtils.isInRange(pos)) LorenzColor.GREEN else LorenzColor.WHITE
+            seaCreaturesToRender.forEach {
```
hannibal002: this is an empty list. the whole forEach here is dead code

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
```diff
@@ -31,18 +33,26 @@ object SeaCreatureLootshareSphere {
...
+                if (it.distance(pos) < 10) otherNearbySpheres++
+                if (otherNearbySpheres < 2) spherePositions.add(LootshareSphere(pos, color))
```
hannibal002: what do we check here? if we only need the int to block too many entires in the sphere position, we can use break instead, or use a list take() function.

src/main/java/at/hannibal2/skyhanni/features/fishing/IsFishingDetection.kt
```diff
@@ -6,8 +6,10 @@ import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.EntityUtils
 import at.hannibal2.skyhanni.utils.LocationUtils
 import at.hannibal2.skyhanni.utils.LocationUtils.distanceToPlayer
+import at.hannibal2.skyhanni.utils.LootshareUtils.isInRange
```
hannibal002: do not import function calls (except  extended functions)

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
```diff
@@ -15,12 +15,14 @@ import at.hannibal2.skyhanni.utils.render.WorldRenderUtils.drawSphereWireframeIn
...
-    private const val RANGE = 30.0f
+    private val seaCreatures = mutableListOf<LivingSeaCreatureData>()
 
-    private val seaCreatures = mutableSetOf<LivingSeaCreatureData>()
-    private val existingCircles = mutableSetOf<LorenzVec>()
+    private val spherePositions = mutableListOf<LootshareSphere>()
 
-    fun isInRange(pos: LorenzVec): Boolean = pos.distanceToPlayer() < RANGE
+    data class LootshareSphere(
+        val position: LorenzVec? = null,
```
hannibal002: what is this? this is always null, can never change?

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
```diff
@@ -15,12 +15,14 @@ import at.hannibal2.skyhanni.utils.render.WorldRenderUtils.drawSphereWireframeIn
 object SeaCreatureLootshareSphere {
     private val config get() = SkyHanniMod.feature.fishing
 
-    private const val RANGE = 30.0f
+    private val seaCreatures = mutableListOf<LivingSeaCreatureData>()
```
hannibal002: why making this from set to list? is the order relevant? do we no longer care about duplicates? can duplicates exist at all, regardless of the impl of the list?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
Fazfoxy: I have literally zero recollection why this happened

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
Fazfoxy: Yeah this is dead code from, something, past me tried but I can't recollect why or what.

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
Fazfoxy: I can't fully tell what you meant by this but I put this into LootshareUtils for now, probably will make sense in whatever file ends up doing the generic Lootshare Sphere rendering in the future

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
Fazfoxy: This is also pointless due to position nullability being pointless but 100% was an error in how it was written

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
Fazfoxy: Moved to an iterator

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
Fazfoxy: This should've been checking the spherePositions list in the new system, fixed to ensure it works properly

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
Fazfoxy: The point is Primarily to avoid when multi entity Sea Creatures overwhelm the list since they're all handled as seperate instances by Mob Detection currently without sacrificing unrelated SC's ability to be rendered if they're far enough away Heuristically.

src/main/java/at/hannibal2/skyhanni/features/fishing/IsFishingDetection.kt
Fazfoxy: Missed this one when doing the removals last time, fixed.

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
```diff
@@ -6,21 +6,17 @@ import at.hannibal2.skyhanni.events.fishing.SeaCreatureEvent
...
 object SeaCreatureLootshareSphere {
     private val config get() = SkyHanniMod.feature.fishing
 
-    private const val RANGE = 30.0f
-
     private val seaCreatures = mutableSetOf<LivingSeaCreatureData>()
-    private val existingCircles = mutableSetOf<LorenzVec>()
 
-    fun isInRange(pos: LorenzVec): Boolean = pos.distanceToPlayer() < RANGE
+    private var spherePositions = mutableSetOf<LootshareUtils.Sphere>()
```
hannibal002: can be val

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureLootshareSphere.kt
```diff
@@ -31,18 +27,30 @@ object SeaCreatureLootshareSphere {
     @HandleEvent(onlyOnSkyblock = true)
     fun onRenderWorld(event: SkyHanniRenderWorldEvent) {
         if (!config.lootshareRange) return
-        existingCircles.clear()
+        val sphereIterator = spherePositions.iterator()
```
hannibal002: all entries are rendered, then removed. iterator is unnecessary. use a normal for loop, then clear the list once.

hannibal002 (changes requested):

hannibal002 (approved):

PR #5709, opened 2026-05-10
Improvement: Split Chest Value own inventory toggle into separate tog…

legentpc:
## What
There are separate on/off options for Inventory, Backpack, and Ender Chest.
Previously, the "Enable in Own Inventory" toggle was the same for all three, but now it's separated like this:
- Own Inventory
- Ender Chest
- Backpack

https://discord.com/channels/997079228510117908/1499178081485852682

## Changelog Improvements
+ Split Chest Value "Enabled in own Inventory" into separate toggles for Own Inventory, Ender Chest, and Backpack. - legentpc


CalMWolfs (changes requested):
This option needs a config fix and if the user had this feature on before it should add all 3 to the list of enabled places

PR #5710, opened 2026-05-11
Fix: make Sensitivity reducer less prone to disabling and re-enabling when falling off small ledges

Fazfoxy:
## What
This isn't just a pull request — it's a merge request

I'm genuinely sorry for this joke description I have made a continuous and serious...

## Changelog Fixes
+ Added config option to modify number of ticks off Ground until sensitivity reducer toggles on/off. - Fazfoxy


zumbiepig:
#5711 makes this behavior based on a distance check to the nearest block below. and i made it only raycast if the player is not onGround and not flying, so it shouldn't cause performance issues.

hannibal002:
Closing in favor of #5711, which solves the same problem using a distance-based raycast check instead of a tick delay. The raycast approach is more accurate since it directly checks whether a block is within a configurable distance below the player, rather than applying a fixed time window.

PR #5716, opened 2026-05-12
Fix: Allow Hide Base Pest Drops without Pest Profit Tracker enabled

BonkersTurnip:
## What
Removed requirement for Pest Profit Tracker to be enabled for the player to hide basic pest drops. I wasn't sure if the option was best suited to stay in this config category, so I've left it in place for now but I'm not sure it makes a whole lot of sense for the category it's in.

## Changelog Fixes
+ Fixed hiding regular pest drop messages not working if you didn't have the Pest Profit Tracker enabled. - BonkersTurnip


Rainymei:
:pleading_face:

PR #5718, opened 2026-05-12
Feature: Common Sea Creature Hider

AverageUser125:
## What
Make common sea creatures have lower opacity when holding a fishing rod, can be overwritten to always reduce opacity.

## Changelog New Features
+ Added Hide common sea creatures - AverageUser125


CalMWolfs (changes requested):
Only while holding rod does not work for me, probably because you are always holding a rod when you fish up a sea creature

AverageUser125:
> Only while holding rod does not work for me, probably because you are always holding a rod when you fish up a sea creature

Fixed

CalMWolfs (approved):

PR #5661, opened 2026-04-29
Improvement: Ignore any Visitor in Shopping List

f4bii:
## What
Adds a way to ignore and unignore any Visitor in the Shopping List

<details>
<summary>Images</summary>

image

</details>

## Changelog Improvements
+ Improved Garden Shopping List to ignore any Visitor. - FabiHBBBT

## Changelog Removed Features
+ Removed the toggle of ignoring Spaceman in the Shopping List. - FabiHBBBT
    * Can now be done by clicking on the Visitor name in the Shopping List.

PR #5733, opened 2026-05-13
Improvement: Add ignoreTaylor Option to GardenVisitorShoppingList

Alumara:
## What
Added `ignoreTaylor` boolean config option to `ShoppingListConfig`, mirroring the existing `ignoreSpaceman` option.
Refactored visitor ignore logic in `prepareDrawingData` by extracting a `getIgnoredVisitors()` helper that builds a set of ignored visitor names from config, replacing the individual if blocks with a single set membership check.

## Changelog Improvements
+ Added the option to exclude crops requested by Taylor from the visitor's shopping list. - Ambrosy


Fazfoxy:
Wouldn't it be easier if you instead made ignoredVisitors an Enum since you already add them to a list and config fixed with event.transform or whatever from the boolean into the Enum?

NopoTheGamer:
#5661

Alumara:
Oops! Definitely should have looked harder at other PRs

PR #5714, opened 2026-05-12
Fix + Backend: Cocoon Message support for Sea Creature Tracker

Fazfoxy:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/5737

## What
This PR is horrid why did I do this

I'm gonna actually work out wherever the hell all this goes in terms of changelog *after* it works maybe

## Changelog Improvements
+ Improved cool feature. - your_name_here
    * Optional extra info.

PR #5737, opened 2026-05-14
Fix: Move existing Sea Creature Cocoon Features to Chat Message

Fazfoxy:
## What
This is extracted changes from https://github.com/hannibal002/SkyHanni/pull/5714 since that PR is gonna require a lot more work and this makes a lot more sense as its own changes

## Changelog Fixes
+ Fixed all rare instances of Sea Creature Cocoon Warning falsely activating by detecting the Hypixel cocooning chat message instead of cocoon entity spawns. - Fazfoxy


src/main/java/at/hannibal2/skyhanni/features/combat/cocoon/CocoonAPI.kt
```diff
@@ -128,6 +142,14 @@ object CocoonAPI {
         existingCocoons.removeIf { it.cocoonID == event.entity.id }
     }
 
+    @HandleEvent(onlyOnSkyblock = true)
+    fun onChatMessageRecieved(event: SkyHanniChatEvent.Allow) {
```
hannibal002: typo. schould be onChatMessageReceived

src/main/java/at/hannibal2/skyhanni/events/combat/CocoonChatMessageEvent.kt
```diff
@@ -0,0 +1,5 @@
+package at.hannibal2.skyhanni.events.combat
+
+import at.hannibal2.skyhanni.api.event.SkyHanniEvent
+
+class CocoonChatMessageEvent(val mobName: String): SkyHanniEvent()
```
hannibal002: add docs, a short explanation of when the event gets fired

src/main/java/at/hannibal2/skyhanni/features/combat/cocoon/CocoonAPI.kt
```diff
@@ -37,17 +41,27 @@ object CocoonAPI {
     private val COCOON_SKULL_TEXTURE by lazy { SkullTextureHolder.getTexture("RIFT_LARVA") }
 
     val expectedLifetime = 6.4.seconds
+    /*
+    roughly where cocoon times landed across a few hundred logged cocoons.
```
hannibal002: wrong comment, on wrong spot

src/main/java/at/hannibal2/skyhanni/features/combat/cocoon/CocoonAPI.kt
```diff
@@ -128,6 +142,14 @@ object CocoonAPI {
         existingCocoons.removeIf { it.cocoonID == event.entity.id }
     }
 
+    @HandleEvent(onlyOnSkyblock = true)
+    fun onChatMessageRecieved(event: SkyHanniChatEvent.Allow) {
+        cocoonChatMessage.matchMatcher(event.cleanMessage) {
+            CocoonChatMessageEvent(group("name")).post()
+        }
+    }
+
```
hannibal002, line 161: double empty lines

src/main/java/at/hannibal2/skyhanni/features/combat/cocoon/CocoonAPI.kt
```diff
@@ -37,17 +41,27 @@ object CocoonAPI {
...
+    /**
+     * REGEX-TEST: CAUGHT! You cocooned a Crypt Ghoul!
+     * REGEX-TEST: CAUGHT! You cocooned a Enderman!
+     */
+    private val cocoonChatMessage by patternGroup.pattern(
+        "spawn",
+        "CAUGHT! You cocooned an? (?<name>(\\w+| )+)!"
+    )
+    // Hypixel don't use an in this message but that's a spelling mistake on their end that might get fixed
```
hannibal002: grammar: don't -> doesn't

src/main/java/at/hannibal2/skyhanni/features/combat/cocoon/CocoonAPI.kt
```diff
@@ -37,17 +41,27 @@ object CocoonAPI {
...
-     roughly where cocoon times landed for me across a few hundred cocoons
-     Might require some sort of ping based tweaking?
-     */
     private val existingCocoons: TimeLimitedSet<CocoonMob> = TimeLimitedSet(8.seconds)
     private val logger: SkyHanniLogger = SkyHanniLogger("Combat/Cocoon")
+    private val patternGroup = RepoPattern.group("combat.cocoon")
+
+    /**
+     * REGEX-TEST: CAUGHT! You cocooned a Crypt Ghoul!
+     * REGEX-TEST: CAUGHT! You cocooned a Enderman!
```
hannibal002, line 68: missing regex test for the " an " case

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/combat/cocoon/CocoonAPI.kt
Fazfoxy: This would be faking a regex test which, while expected in the future I think is against the spirit of what a Regex test is *meant* to be

hannibal002: oh, so why do we have the n? check then?

Fazfoxy: To quote the comment you critiqued my Grammer in
"Hypixel don't use an in this message but that's a spelling mistake on their end that might get fixed"
They don't *yet* use An, but as it stands I don't think I really should need to fake a regex test to tell you that an? should match a and an 🥺

src/main/java/at/hannibal2/skyhanni/features/combat/cocoon/CocoonAPI.kt
Fazfoxy: Ok for some reason the primary function name is actually just "onChat" which I think is wrong but, is the correct wording currently.

src/main/java/at/hannibal2/skyhanni/features/combat/cocoon/CocoonAPI.kt
hannibal002: >  I don't think I really should need to fake a regex test to tell you that an? should match a and an 🥺

well, thats the point of the regex tests everywhere. to test what we know will pass the test. if you dont add the an to the a in the test, the regex looks like it machtes more than we expect it to match and someone might flag this error with tooling (that cant understand the comment below)

src/main/java/at/hannibal2/skyhanni/features/combat/cocoon/CocoonAPI.kt
```diff
@@ -21,33 +23,45 @@ import at.hannibal2.skyhanni.utils.ItemUtils.getInternalName
...
 import net.minecraft.world.entity.decoration.ArmorStand
-import at.hannibal2.skyhanni.utils.SafeItemStack
 import kotlin.time.Duration.Companion.seconds
 
 @SkyHanniModule
 object CocoonAPI {
     private val COCOON_SKULL_TEXTURE by lazy { SkullTextureHolder.getTexture("RIFT_LARVA") }
 
+    /*
+    roughly where cocoon times landed across a few hundred logged cocoons.
```
hannibal002: "where cocoon times landed" thats not very clear. what is it the time represent? the time until something becomes cocooned? the time from start to end of the cocoon process? the duration something is cocooned?

hannibal002 (changes requested):
and the missing a -> an regex test above

src/main/java/at/hannibal2/skyhanni/features/combat/cocoon/CocoonAPI.kt
Fazfoxy: Added the faked regex test asked for.

src/main/java/at/hannibal2/skyhanni/features/combat/cocoon/CocoonAPI.kt
Fazfoxy: This should be as much explanation as necessary for someone to understand what the val is for.

src/main/java/at/hannibal2/skyhanni/features/combat/cocoon/CocoonAPI.kt
```diff
@@ -21,33 +23,49 @@ import at.hannibal2.skyhanni.utils.ItemUtils.getInternalName
...
+    /**
+     * REGEX-TEST: CAUGHT! You cocooned a Crypt Ghoul!
+     * REGEX-TEST: CAUGHT! You cocooned a Enderman!
+     * REGEX-TEST: CAUGHT! You cocooned an Enderman!
+     */
+    private val cocoonChatMessage by patternGroup.pattern(
+        "spawn",
+        "CAUGHT! You cocooned an? (?<name>[\\w ]+)!",
+    )
+    // Hypixel doesn't currently use an in this message but grammatical errors on their part should be accounted for in advance.
```
lunaynx: You can probably put these comment lines *above* the regex test block comment and Detekt won't complain?

hannibal002 (approved):

PR #5740, opened 2026-05-14
Fix: Chat Edit/Delete Lag with large history

lunaynx:
## What
Refactors chat handling to avoid calling `refreshTrimmedMessages` or Minecraft methods that call it internally, which is costly because it asks Minecraft to redraw every single message in the history. Popular mods like Chat Patches increase the history size to large amounts like 16,384, making micro-stutters a frequent complaint.

## Changelog Fixes
+ Fixed lagspikes when SkyHanni edits or deletes chat messages with mods that increase chat history limit such as Chat Patches. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/ChatManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/ChatManager.kt
@@ -10,16 +10,18 @@ import at.hannibal2.skyhanni.events.chat.SkyHanniChatEvent
 import at.hannibal2.skyhanni.events.minecraft.packet.PacketSentEvent
 import at.hannibal2.skyhanni.features.chat.ChatHistoryGui
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
+import at.hannibal2.skyhanni.test.command.ErrorManager
 import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.DelayedRun
 import at.hannibal2.skyhanni.utils.IdentityCharacteristics
-import at.hannibal2.skyhanni.utils.SkyHanniLogger
 import at.hannibal2.skyhanni.utils.ReflectionUtils.getClassInstance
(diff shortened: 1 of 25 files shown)


src/main/java/at/hannibal2/skyhanni/data/ChatManager.kt
```diff
@@ -191,24 +194,23 @@ object ChatManager {
...
             modified = true
             loggerModified.log(" ")
-            loggerModified.log("[original] " + component.formattedTextCompat())
+            loggerModified.log("[original] " + original.formattedTextCompat())
             loggerModified.log("[modified] " + modifiedComponent.formattedTextCompat())
-            messageHistory[key] = MessageFilteringResult(component, ActionKind.MODIFIED, null, modifiedComponent, reason)
+            messageHistory[key] = MessageFilteringResult(original, ActionKind.MODIFIED, null, modifiedComponent, reason)
         } else {
-            messageHistory[key] = MessageFilteringResult(component, ActionKind.ALLOWED, null, null, null)
+            messageHistory[key] = MessageFilteringResult(original, ActionKind.ALLOWED, null, null, null)
```
NopoTheGamer, line 213: i did it like this because i was scared of the event mutating the original component somehow and bricking this

lunaynx: IntelliJ flagged the assignment as redundant. It shouldn't be able to mutate it, since it's typed as `Component` and not `MutableComponent`, unless it's explicitly casting somewhere down the line, which I don't think we are? Either way, I don't think a simple reassignment would stop it.

NopoTheGamer: im not saying either way is correct, im just stating why it was like that for extra context

lunaynx:
I removed the dependency on the "drop 1.21.10" PR by gating the broken code to 1.21.11+ instead, in case someone wants to merge this before 9.0.0.

PR #5742, opened 2026-05-14
Backend: pest regex

Rainymei:
## What
Handles overbloom in pest drop as of new patch.

+ Added handling for overbloom stat in pest Rare Drops. - Rain


src/main/java/at/hannibal2/skyhanni/features/garden/tracker/PestProfitTracker.kt
zumbiepig: do you need to update these tests

PR #5743, opened 2026-05-14
Backend: Cleaner render state data storage

lunaynx:
## What
Currently, we store full entity references on `EntityRenderState`s. While this isn't inherently problematic, as it's a reference that's unlikely to persist long-term beyond the entity's lifetime, there's still a risk, and Fabric's `RenderStateDataKey` API is made for exactly this purpose. This PR switches to using that API to store lightweight key/value pairs for entity render state-related data.

(I was asked to make this PR by @NopoTheGamer.)

+ Refactored render state data storage to use Fabric API's RenderStateDataKey with lightweight key/value pairs instead of storing full entity references on render states. - Luna


lunaynx:
~~I'm gonna try to refactor this to avoid storing entity references too.~~ Done

src/main/java/at/hannibal2/skyhanni/mixins/hooks/SkyHanniRenderStateData.kt
```diff
@@ -0,0 +1,63 @@
...
+import net.minecraft.world.entity.Entity
+import net.minecraft.world.entity.LivingEntity
+
+object SkyHanniRenderStateData {
+
+    private const val NO_ENTITY_ID = -1
+    private const val NO_ENTITY_TRANSPARENCY = -1
+
+    private val ENTITY_ID: RenderStateDataKey<Int> =
+        RenderStateDataKey.create { "skyhanni:entity_id" }
```
AverageUser125, line 12: Shouldn’t this use Identifier and `SkyHanniMod.id`?
(I don’t know much about this but still)

src/main/java/at/hannibal2/skyhanni/mixins/hooks/EntityRenderDispatcherHook.kt
```diff
@@ -2,27 +2,20 @@ package at.hannibal2.skyhanni.mixins.hooks
...
-import java.util.Collections
-import java.util.IdentityHashMap
 
-private var savedEntity: Entity? = null
 private var savedEntityRenderState: EntityRenderState? = null
 
-val activeHolographicEntities: MutableSet<LivingEntity> =
-    Collections.newSetFromMap(IdentityHashMap())
-
 fun setEntity(state: EntityRenderState?) {
```
AverageUser125: I would say “put this inside an object”.
But I don’t wanna repeat of last time with compat block.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/SkyHanniRenderStateData.kt
```diff
@@ -0,0 +1,63 @@
...
+import at.hannibal2.skyhanni.data.entity.EntityTransparencyManager
+import at.hannibal2.skyhanni.utils.EntityUtils
+import net.fabricmc.fabric.api.client.rendering.v1.RenderStateDataKey
+import net.minecraft.client.renderer.entity.state.EntityRenderState
+import net.minecraft.world.entity.Entity
+import net.minecraft.world.entity.LivingEntity
+
+object SkyHanniRenderStateData {
+
+    private const val NO_ENTITY_ID = -1
```
AverageUser125: You use -1 for fake players. (Like fake player renderable)
Are you sure this doesn’t conflict?

src/main/java/at/hannibal2/skyhanni/mixins/hooks/SkyHanniRenderStateData.kt
lunaynx: It only takes a string as argument, called "debug name". Naming it similar to an identifier is just a convention used by some, and creating an `Identifier` object only to immediately convert it to a string would be pointless.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/SkyHanniRenderStateData.kt
lunaynx: Fair, it probably would. This was made before that was a thing. I double checked and `null` works fine without even having to declare our types as nullable, because the Fabric render state data is inherently nullable – it builds fine and entity glow works fine without crashing.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/EntityRenderDispatcherHook.kt
lunaynx: Nah, this one should be safe to refactor.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/SkyHanniRenderStateData.kt
AverageUser125: Feels weird having 3 states (missing, null, value).

but if the system is designed around this idea (or missing = null) then it’s probably fine

lunaynx: It does seem to essentially treat missing and null the same, though the underlying map *may* technically differentiate between them.

jani270:
I dont think any user will understand this changelog

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinItemFeatureRenderer.java
```diff
@@ -34,7 +34,7 @@ public abstract class MixinItemFeatureRenderer {
         method = "prepareOutlineSubmit",
         at = @At(value = "INVOKE", target = "Lcom/mojang/blaze3d/vertex/QuadInstance;setColor(I)V"), index = 0)
     private int modifyAlpha(int originalColor) {
-        if (EntityRenderDispatcherHookKt.getEntity() instanceof LivingEntity livingEntity) {
+        if (EntityRenderDispatcherHook.getEntity() instanceof LivingEntity livingEntity) {
```
hannibal002, line 37: Use getEntityTransparency() here too, holographic entities have no entity id so their transparency never reaches this path.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinItemFeatureRenderer.java
```diff
@@ -48,7 +48,7 @@ private int modifyAlpha(int originalColor) {
         method = "prepareOutlineSubmit",
         at = @At(value = "INVOKE", target = "Lnet/minecraft/client/resources/model/geometry/BakedQuad$MaterialInfo;itemRenderType()Lnet/minecraft/client/renderer/rendertype/RenderType;"))
     private RenderType modifyRenderLayer(RenderType layer) {
-        if (EntityRenderDispatcherHookKt.getEntity() instanceof LivingEntity livingEntity) {
+        if (EntityRenderDispatcherHook.getEntity() instanceof LivingEntity livingEntity) {
```
hannibal002, line 51: Same here

src/main/java/at/hannibal2/skyhanni/mixins/hooks/SkyHanniRenderStateData.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    @JvmStatic
+    fun clearEntityId(state: EntityRenderState) {
+        state.setData(ENTITY_ID, null)
+    }
+
+    @JvmStatic
+    fun getEntity(state: EntityRenderState): Entity? {
+        val entityId = state.getData(ENTITY_ID) ?: return null
+        return EntityUtils.getEntityByID(entityId)
```
hannibal002, line 36: This used to be a field read, now every getEntity() call is a level lookup

src/main/java/at/hannibal2/skyhanni/mixins/hooks/SkyHanniRenderStateData.kt
```diff
@@ -0,0 +1,54 @@
...
+    private val ENTITY_TRANSPARENCY: RenderStateDataKey<Int> =
+        RenderStateDataKey.create { "skyhanni:entity_transparency" }
+
+    private val USING_CUSTOM_OUTLINE: RenderStateDataKey<Boolean> =
+        RenderStateDataKey.create { "skyhanni:using_custom_outline" }
+
+    @JvmStatic
+    fun setFromEntity(state: EntityRenderState, entity: Entity) {
+        state.setData(ENTITY_ID, entity.id)
+        setUsingCustomOutline(state, RenderLivingEntityHelper.getEntityGlowColor(entity) != null)
```
hannibal002, line 23: MixinEntityRenderer.shouldAlsoGlow already computed this in the same extractRenderState call, so getEntityGlowColor now runs twice per entity per frame.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinEntityRenderDispatcher.java
```diff
@@ -19,28 +19,27 @@
...
-
     @Inject(method = "submit", at = @At(value = "HEAD"), cancellable = true)
     public void onRenderPre(EntityRenderState renderState, CameraRenderState cameraRenderState, double d, double e, double f, PoseStack poseStack, SubmitNodeCollector submitNodeCollector, CallbackInfo ci) {
-        EntityRenderDispatcherHookKt.setEntity(renderState);
-        Entity entity = EntityRenderDispatcherHookKt.getEntity();
-        if (entity instanceof LivingEntity livingEntity
-            && !EntityRenderDispatcherHookKt.getActiveHolographicEntities().contains(livingEntity)) {
+        EntityRenderDispatcherHook.setEntity(renderState);
+        Entity entity = EntityRenderDispatcherHook.getEntity();
+        if (entity instanceof LivingEntity livingEntity) {
```
hannibal002, line 26: Post and the two Specials events used to fire for holographic entities and now don't, intended?

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinEntityRenderDispatcher.java
```diff
@@ -19,28 +19,27 @@
...
             //noinspection deprecation
             if (new SkyHanniRenderEntityEvent.Pre<>(livingEntity, d, e, f).post().isCancelled()) {
                 ci.cancel();
+                EntityRenderDispatcherHook.clearEntity();
```
hannibal002, line 30: Did you hit an actual symptom here, or is it defensive?

hannibal002 (changes requested):

lunaynx:
> I dont think any user will understand this changelog

Fair point. I could simplify it further but I think I'll just change it back to "Backend", I don't want to overpromise on theoretical memory usage improvements when the PR that depends on this already claims similar and is more likely to actually do something.

lunaynx:
Probably gonna just redo this later.

PR #5749, opened 2026-05-15
Feature: hide non-boss entity nametags during slayer boss

AverageUser125:
## What
Add a toggle to hide non-boss entity nametags (e.g., Primordial Bats) once a Slayer Boss spawns.
This adds ontop of the "Active Boss Transparency" Feature
To be honest, I am unsure if this should be part of "Hide Irrelevant Mobs" or "Active Boss Transparency", but I think the latter makes more sense.

## Changelog New Features
+ Add a toggle to hide non-boss entity nametags once a Slayer Boss spawns - AverageUser125

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/slayer/ActiveBossTransparencyConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/slayer/ActiveBossTransparencyConfig.kt
@@ -25,4 +25,9 @@ class ActiveBossTransparencyConfig {
     @ConfigOption(name = "Other Players", desc = "Also change the transparency for other players.")
     @ConfigEditorBoolean
     var applyToPlayers: Boolean = false
+
+    @Expose
+    @ConfigOption(name = "Hide Name Tags", desc = "Hide custom names of mobs when they are transparent.")
+    @ConfigEditorBoolean
+    val hideNameTags: Boolean = false
 }
--- a/src/main/java/at/hannibal2/skyhanni/features/slayer/ActiveBossTransparency.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/slayer/ActiveBossTransparency.kt
@@ -81,6 +81,8 @@ object ActiveBossTransparency {
                 if (!config.applyToPlayers) return
             }
             if (category == MobCategory.PLAYER && !config.applyToPlayers) return
+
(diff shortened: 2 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/config/features/slayer/ActiveBossTransparencyConfig.kt
```diff
@@ -25,4 +25,9 @@ class ActiveBossTransparencyConfig {
     @ConfigOption(name = "Other Players", desc = "Also change the transparency for other players.")
     @ConfigEditorBoolean
     var applyToPlayers: Boolean = false
+
+    @Expose
+    @ConfigOption(name = "Hide Name Tags", desc = "Hide custom names of mobs when they are transparent.")
```
zumbiepig, line 30: is this ai? who would call this "hide custom names of mobs"

src/main/java/at/hannibal2/skyhanni/features/slayer/ActiveBossTransparency.kt
```diff
@@ -81,6 +81,8 @@ object ActiveBossTransparency {
                 if (!config.applyToPlayers) return
             }
             if (category == MobCategory.PLAYER && !config.applyToPlayers) return
+
+            if (config.hideNameTags) mob.armorStand?.isCustomNameVisible = false
```
zumbiepig, line 85: are we fr

did we just try setting a function to false

Image

Image

zumbiepig (changes requested):

src/main/java/at/hannibal2/skyhanni/features/slayer/ActiveBossTransparency.kt
NopoTheGamer: im not sure you should be reviewing prs if you dont understand how the code works

CalMWolfs: > are we fr
> 
> did we just try setting a function to false
> Image Image

it is a setter. It would not compile if it was actually an error
image

zumbiepig: well anyways i dont think you can set entity data without causing problems since the data comes from the server

wont that data get overridden every tick when the entity is updated by hypixel

PR #5751, opened 2026-05-15
Improvement: remove useless vertical spacer

Rainymei:
## What
Removes the vertical spacer on top of Elite Leaderboards HUDs.

<details>
<summary>Images</summary>
before
image
after
image

</details>

## Changelog Improvements
+ Improved the display of Elite Leaderboard HUDs. - Rain


Chissl:
The spacer is intentional—it is there to prevent the entire element from shifting down when opening an inventory, as additional lines are appended to the top of the element. It additionally lets people know where the clickables will appear when in /sh gui, preventing overlap with other elements.

zumbiepig:
> The spacer is intentional—it is there to prevent the entire element from shifting down when opening an inventory, as additional lines are appended to the top of the element. It additionally lets people know where the clickables will appear when in /sh gui, preventing overlap with other elements.

yeah

but it also makes it so i cant move the gui any higher on my screen

its stuck 10 pixels down

PR #5762, opened 2026-05-17
Backend + Fix: Use Hypixel Mod API for location info

lunaynx:
## What
Migrates from unreliable locraw/tablist/etc. based detection of on-Hypixel, on-alpha, and current island to Hypixel Mod API. Seemed to work fine for me based on limited testing.

## Changelog Fixes
+ Fixed the mod sometimes not correctly detecting the island you're on. - Luna
    * The "Hypixel Mod API" mod (https://modrinth.com/mod/hypixel-mod-api) is now required to use SkyHanni. You might already have this installed because of other mods.

+ Migrated to Hypixel Mod API for server and location detection. - Luna

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,7 +2,7 @@
 libautoupdate = "1.3.1"
 moulconfig = "4.5.0"
 jbAnnotations = "24.1.0"
-hypixelmodapi = "1.0.2"
+hypixelmodapi = "1.0.1"
 hypixelmodapi-fabric = "1.0.1+build.1+mc1.21"
 fabricLanguageKotlin = "1.13.9+kotlin.2.3.10"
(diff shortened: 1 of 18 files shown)


gradle/libs.versions.toml
```diff
@@ -2,7 +2,7 @@
 libautoupdate = "1.3.1"
 moulconfig = "4.5.0"
 jbAnnotations = "24.1.0"
-hypixelmodapi = "1.0.2"
+hypixelmodapi = "1.0.1"
```
NopoTheGamer, line 5: ?

lunaynx: 1.0.2 seems to be 26.1 only, it was crashing because of the incompatibility once I fixed it to actually load the mod properly in dev env.

NopoTheGamer (commented):
might work might not no way to know

PR #5767, opened 2026-05-19
Backend: Fix old PRs

hannibal002:
## What
I've looked through some of the PRs that got merged since last beta and noticed small issues.

ignore_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCustomKeybinds.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCustomKeybinds.kt
@@ -150,7 +150,7 @@ object GardenCustomKeybinds {
     private fun refreshState(keyBindings: Iterable<KeyMapping>) {
         for (keyBinding in keyBindings) {
             if (keyBinding.isToggle()) continue
-            keyBinding.setDown(keyBinding.key.isDown())
+            keyBinding.isDown = keyBinding.key.isDown()
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCustomKeybinds.kt
```diff
@@ -176,6 +174,7 @@ object GardenCustomKeybinds {
         return internalName.isFarmingTool() ||
             (config.mousemat && internalName == SQUEAKY_MOUSEMAT) ||
             (config.fishingRod && internalName.isFishingRod()) ||
+            // TODO confirm why we check for item air here. getItemInHand should return null if there is no item there.
```
Rainymei, line 184: afaik getItemInHand returns EMPTY if hand is empty.
But it's true we could use `getItemInHand.isEmpty()` instead of comparing heldItem

hannibal002: you mean Item.AIR, right? im working on a new pr that addresses this

lunaynx: I confirmed AIR works correctly when making this, maybe it should be fixed to return empty/null or whatever though.

hannibal002: good, this means my changes in https://github.com/hannibal002/SkyHanni/pull/5768 will be fine.

PR #5771, opened 2026-05-19
Fix: Anita Extra FF showing integer limit

Rainymei:
## What
Anita Extra FF feature shows integer limit when at tier 0/15.

PSA: This is untested cause my dev env does not want to render it idk why, but this fix breaking the feature would make no sense to me.

<details>
image

</details>

## Changelog Fixes
+ Fixed Anita's Extra FF cost to max out showing integer limit. - Rain


hannibal002 (changes requested):
this PR is a workaorund. we need a todo comment added to verify and properly fix the underlying bug

hannibal002 (approved):

PR #5780, opened 2026-05-21
Fix: /shcopyentites entity count does not reset

AverageUser125:
## What
doing /shcopyentities did not reset the count each time it was called, meaning it just kept accumulating.

exclude_from_changelog


NopoTheGamer:
https://github.com/hannibal002/SkyHanni/pull/5689

PR #5783, opened 2026-05-21
Fix: Harvest Feast Data Handling

lunaynx:
## What
Fixes Harvest Feast timer expecting milliseconds instead of seconds, and also fixes the timer for when all crops in `next` are `null` because it's the last rotation of the feast.

## Changelog Fixes
+ Fixed Harvest Feast in-season crop display sometimes showing incorrect information. - Luna


src/main/java/at/hannibal2/skyhanni/data/jsonobjects/elitedev/EliteFeastJson.kt
```diff
@@ -67,4 +91,27 @@ data class EliteFeastData(
     }
 
     private fun List<String>.toCropTypes(): List<CropType> = map { CropType.getByName(it) }
+
+    companion object {
+        private const val HARVEST_FEAST_END_MONTH = 10
+        private const val HARVEST_FEAST_END_DAY = 1
+        private const val GRAND_FEAST_END_MONTH = 3
+        private const val GRAND_FEAST_END_DAY = 27
```
NopoTheGamer, line 99: cant wait for jerry :)

PR #5784, opened 2026-05-22
Feature: Draw Optimal Diana Burrow Path

MidnightJerry:
## What
Added a feature that draws the shortest path connecting diana burrow entries

<details>
<summary>Images</summary>
burrow_showcase_image

</details>

## Changelog New Features
+ Added Optimal Mythological Ritual Burrow Path Finder - Midnight_Jerry

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/event/diana/DianaConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/event/diana/DianaConfig.kt
@@ -68,6 +68,16 @@ class DianaConfig {
 
     @Expose
     @ConfigOption(
+        name = "Optimal Path",
+        desc = "Draw the shortest path connecting burrows and guess locations.\n" +
+        "It is recommended to disable Line To Next to avoid conflicts.",
+    )
+    @ConfigEditorBoolean
+    @FeatureToggle
+    var drawOptimalPath: Boolean = false
+
+    @Expose
+    @ConfigOption(
         name = "Nearby Detection",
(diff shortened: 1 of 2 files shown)


NopoTheGamer:
this seems like a cool idea but the whole "more burrows will spawn when you dig burrows" thing seems to break the flow of this (at least in my mind?)

MidnightJerry:
> this seems like a cool idea but the whole "more burrows will spawn when you dig burrows" thing seems to break the flow of this (at least in my mind?)

It might be useful in some case? I tried it out for a bit and the general path doesn't change frequently. It might be helpful with deciding the path.

src/main/java/at/hannibal2/skyhanni/config/features/event/diana/DianaConfig.kt
```diff
@@ -66,6 +66,16 @@ class DianaConfig {
     @ConfigEditorBoolean
     var multiGuesses: Boolean = true
 
+    @Expose
+    @ConfigOption(
+        name = "Optimal Path",
+        desc = "Draw the shortest path connecting burrows and guess locations.\n" +
+        "It is recommended to disable Line To Next to avoid conflicts.",
```
Fazfoxy: Shouldn't this probably just use the same line for line from first burrow --> Crosshair instead of telling the user to turn it off?

MidnightJerry: The first burrow of the path may not be the nearest burrow sometimes, so it may have two different burrows drawn to crosshair if both are turned on

PR #5786, opened 2026-05-22
Feature: Wrong Farming Tool Alert

BonkersTurnip:
## What
Adds an alert for farming crops on the Garden with the wrong Specialized Farming Tool.

## Changelog New Features
+ Added an alert for farming with the wrong Farming Tool. - BonkersTurnip


hannibal002:
this needs a delay so that this warning doesnt show immediately, but only after lets say 10 seconds

BonkersTurnip:
forgive me if I'm missing something obvious, but... why? i can't think of a situation where I'd want to farm for ten seconds before switching to the correct tool instead of just using it from the start

hannibal002:
> forgive me if I'm missing something obvious, but... why? i can't think of a situation where I'd want to farm for ten seconds before switching to the correct tool instead of just using it from the start

if the feature warns you instantly, then this could be used to flag macro checks/anti afk detections where your item in hand gets changed/items gets moved out of the hotbar/switched with other items or tools in your inventory.

We do not want to help cheaters. delaying the detection by couple seconds would give watchdog/staff more time to detect and punish.

Another solution would be to flag it instantly if this is the first block broken in some time.

lunaynx:
Yeah, we could delay it only if the player was farming in the last X seconds.

PR #5792, opened 2026-05-25
Fix: Extend Barn Bounds to full plot

BonkersTurnip:
## What
Extends the bounds of the barn plot in the code to cover the entire barn plot. Fixes any issues with custom keybinds not being disabled in certain parts of the barn or visitors not being found if below the accepted y-level.

## Changelog Fixes
+ Fixed Visitors not being detected below Y=70. - BonkersTurnip
+ Fixed Custom Keybinds not being disabled in the majority of the Barn plot even when Exclude Barn was enabled. - BonkersTurnip

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/ConfigUpdaterMigrator.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/ConfigUpdaterMigrator.kt
@@ -13,7 +13,7 @@ import com.google.gson.JsonPrimitive
 object ConfigUpdaterMigrator {
 
     val logger = SkyHanniLogger("ConfigMigration")
-    const val CONFIG_VERSION = 133
+    const val CONFIG_VERSION = 134
     fun JsonElement.at(chain: List<String>, init: Boolean): JsonElement? {
         if (chain.isEmpty()) return this
         if (this !is JsonObject) return null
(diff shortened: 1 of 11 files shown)


zumbiepig:
no

zumbiepig:
some people have farms built in the barn. its intentionally like that

lunaynx:
Yeah, but that's quite rare, and those people can turn the option off. But maybe we can add a separate option to bring back the old behavior.

zumbiepig:
> Yeah, but that's quite rare, and those people can turn the option off. But maybe we can add a separate option to bring back the old behavior.

but why should custom keybinds be enabled in the rest of the barn
wouldnt that also cause issues with bordering plots

BonkersTurnip:
I've never heard of a farm being built in the barn and I can't imagine why you would want to, but I do think that that problem is trivially solved by turning off the Exclude Barn option... why is this a problem?

BonkersTurnip:
> > Yeah, but that's quite rare, and those people can turn the option off. But maybe we can add a separate option to bring back the old behavior.
> 
> but why should custom keybinds be enabled in the rest of the barn wouldnt that also cause issues with bordering plots

Custom keybinds ideally *shouldn't* be enabled in the rest of the barn, if people are specifically choosing to exclude the barn then the keybinds should be disabled in the barn, because that's what the config says it does. I'm assuming that when you say bordering plots you're referring to /tptoplot farms which would be on the border of the barn, and if that's the case then I still don't believe this should be an issue since the bounds are at the centres of the corner blocks (+-0.5) and the tptoplot puts you at the exact border (+-0.0) so you wouldn't be considered "in the barn" at that point. I'll test and get back to you on that though.

BonkersTurnip:
Can confirm that this change does not cause any issues for /tptoplot for any of plots 1,2,3,4 which are the only ones bordering the barn, when running /tptoplot you're teleported outside of the extended barn bounds and custom keybinds work as expected from the moment you teleport.

Fazfoxy:
> I've never heard of a farm being built in the barn and I can't imagine why you would want to, but I do think that that problem is trivially solved by turning off the Exclude Barn option... why is this a problem?

In theory this "isn't a problem" but if you have a farm that exists in the edges of the barn and still want them off when in the real barn this is an active downgrade from the previous system, it would be preferable to have some like "only disable in Unbuildable Barn Zone" that this extended plot area is not used for that the current option becomes

PR #5794, opened 2026-05-26
Improvement: Add Harvest Feast rare crops to Pest Profit Tracker

lunaynx:
## What
Adds Harvest Feast rare crop drops to the Pest Profit Tracker if the player is holding a vacuum or a lasso at the time they gain the drop.

## Changelog Improvements
+ Harvest Feast rare crops are now added to the Pest Profit Tracker if you get them from pests. - Luna
    * Please let us know if it fails to track, e.g. if you switch off your vacuum/lasso too early.


NopoTheGamer (changes requested):
field mouse ?

lunaynx:
Yeah uhh, is there a reliable way to detect that the player killed a Field Mouse when they might kill multiple pests at once? It will probably just get attributed to the wrong pest right now.

PR #5796, opened 2026-05-26
Improvement + Fix: Sensitivity Reducer Rewrite

zumbiepig:
## What
Huge rewrite of the sensitivity reducer.

## Changelog Improvements
+ Improved Mouse Sensitivity Reducer with mouse lock support. - zumbiepig
    * Mouse rotation can now be fully locked instead of just reducing sensitivity.
    * Auto modes for Vacuum and Sprayonator.
    * Sensitivity reduction now uses a percentage instead of a division factor.
    * Ground tolerance option to prevent unlocking on small height drops.

## Changelog Fixes
+ Fixed `/shsensreduce` not working. - zumbiepig
+ Fixed garden warp keybinds having an unnecessary cooldown. - zumbiepig

+ Merged all mouse sensitivity features into `MouseSensitivityReducer`. - zumbiepig
    * `MouseMixin` now hooks into `player.turn(x, y)` rather than `options.sensitivity().get()`.


src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
```diff
@@ -0,0 +1,214 @@
...
+
+@SkyHanniModule
+object SensitivityReducer {
+
+    private val config get() = GardenApi.config.sensitivityReducer
+
+    private val commandMessageId = ChatUtils.getUniqueMessageId()
+    private val SQUEAKY_MOUSEMAT = "SQUEAKY_MOUSEMAT".toInternalName()
+
+    private var state: SensitivityState = SensitivityState.UNCHANGED
```
hannibal002: what is the diff betweeen state and manual state?

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    @HandleEvent
+    fun onWorldChange() {
+        manualState = null
+    }
+
+    @HandleEvent
+    fun onTick() {
+        manualState?.setActive()
```
hannibal002: why do we set the state every tick here? cant we do this once when the state changes?

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
```diff
@@ -0,0 +1,214 @@
...
+            ?: if (!shouldAutoReduce()) SensitivityState.UNCHANGED.setActive()
```
hannibal002: the three if blocks here can be wrapped in a when block, and isntead of calling setState, use state = when {

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
```diff
@@ -0,0 +1,214 @@
...
+            else if (!config.lockMouse) SensitivityState.REDUCED.setActive()
+            else SensitivityState.LOCKED.setActive()
+    }
+
+    private fun shouldAutoReduce(): Boolean {
+        if (!config.enabled) return false
+
+        if (!GardenApi.inGarden()) return false
+
+        if (config.mode.none {
```
hannibal002: this if block should be a standalone function

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
```diff
@@ -0,0 +1,214 @@
...
+                    SensitivityReducerConfig.Mode.KEYBIND -> config.keybind.isKeyHeld() && Minecraft.getInstance().screen == null
+                    SensitivityReducerConfig.Mode.MOUSEMAT -> GardenApi.itemInHand?.getInternalName() == SQUEAKY_MOUSEMAT
+                    SensitivityReducerConfig.Mode.VACUUM -> PestApi.hasVacuumInHand()
+                    SensitivityReducerConfig.Mode.SPRAYONATOR -> PestApi.hasSprayonatorInHand()
+                }
+            }) return false
+
+        if (config.onlyPlot && GardenApi.onUnfarmablePlot) return false
+
+        if (config.onGround) {
```
hannibal002: this if block should be a standalone function

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
```diff
@@ -0,0 +1,214 @@
...
+                }
+            }
+        }
+    }
+
+    @HandleEvent
+    fun onGuiRenderOverlay() {
+        if (!config.showGui) return
+
+        if (SensitivityState.REDUCED.isActive()) {
```
hannibal002: this can avoid code duplication by returning the text via the if, then doing the render call only once below (if the text is not null)

src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt
```diff
@@ -0,0 +1,214 @@
...
+        } else if (SensitivityState.LOCKED.isActive()) {
+            config.position.renderRenderable(
+                Renderable.text("§eMouse Locked"),
+                posLabel = "Sensitivity Reducer",
+            )
+        }
+    }
+
+    @HandleEvent
+    fun onDebugDataCollect(event: DebugDataCollectEvent) {
```
hannibal002, line 161: reducingPercent and other similar relevant config values should also be shown. also, this should not be shown via addData but via addIrrelevant if the feature is disabled/not in garden.

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
```diff
@@ -0,0 +1,214 @@
...
+            }
+            val newList = JsonArray()
+            when (element.asString) {
+                "OFF" -> newList.add("TOOL")
+                else -> newList.add(element.asString)
+            }
+            newList
+        }
+        event.transform(134, "$base.reducingFactor") {
+            event.add(134, "$base.reducingPercent") {
```
hannibal002: have you tested that this "calling a modify inside a modify" actuallly works in the config migration logic?

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    @JvmStatic
+    fun remapSensitivity(original: Double): Double = state.transform(original)
+
+    @HandleEvent
+    fun onChat(event: SkyHanniChatEvent.Allow) {
+        if (!config.disableOnTeleport) return
+
+        manualState?.let {
```
hannibal002: could use early return instead

hannibal002 (changes requested):
first look, might have missed something.

also the changelog is way too much. combine or omit parts of it. stuff in fix might be actually improvements, and the technical details shound be way waay shorter or even removed entirely if the relevant code is already mentioned in the user facing changelog. also make sure that the user facing changelog (improvements + fix) is actually readable for non technical users.

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
zumbiepig: I copied the existing block above for `$base.enabled` and that seems to work. So I'm going on the assumption that this should work the same way.

The alternative way to implement this is a `event.move` then an `event.transform`.

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
zumbiepig: `state` is the actual state the mixin will pull from anytime the mouse is moved.

`manualState` is an override when /shmouselock or /shsensreduce is used. It was easiest to set these separately because of race conditions. It's also a lot more readable than the old `isAutoActive`, `isActive`, `shouldAutoReduce` and everything else that was used to track it before.

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
zumbiepig: I did it this way because I needed to store the `manualState` locally. Otherwise it would break and give the wrong message if you spam the commands really quickly.

I can change this to a `val state = manualState ?: return` if you would like.

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
zumbiepig: I might honestly put these checks in the enum. Then check each thing in the enum. It would make it a lot easier to add more triggers later in the future.

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
zumbiepig: I could have done `if (state != manualState) manualState?.setActive()`, but then I would have to add a `manualState != null` check too. Also, there is minimal performance impact of re-setting the field every tick.

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
zumbiepig: Okay, sure. Is like this good? (I can also do it without the `with` block if that's preferred.)

```kotlin
with (SensitivityState) {
    state = when {
        manualState != null -> manualState
        !shouldAutoReduce() -> UNCHANGED
        !config.lockMouse -> REDUCED
        else -> LOCKED
    }
}

File: src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt (hunk 5 above)
Comment (zumbiepig) [reply src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt]: Sure. I was originally trying to avoid using so many helper functions but I can move this out.

File: src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt (hunk 3 above)
Comment (lunaynx) [reply src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt]: I think you *should* be able to omit the `with` here if you add `-Xcontext-sensitive-resolution` to `freeCompilerArgs` in `build.gradle.kts`. Just gotta make sure Kotlin knows `state` is a `SensitivityState`.

Comment (zumbiepig) [reply src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt]: I don't need the `with`. I did that for readability. Is it more readable without the `with`?

Comment (lunaynx) [reply src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt]: I'd generally say avoid it for cases like this. It can be fine sometimes if you're doing a bunch of operations on the same object.

My suggestion would let you drop the enum class prefixes without using `with`, but you can also just drop it and add the prefixes if you don't want to be the one responsible for enabling context-sensitive resolution (something I've been planning on doing anyway).

File: src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt (hunk 7 above)
Comment (zumbiepig) [reply src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt:161]: Ok, I added this.

File: src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt (hunk 9 above)
Comment (zumbiepig) [reply src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt]: I made it an early return.

File: src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt (hunk 2 above)
Comment (zumbiepig) [reply src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt]: Changed it to a `state = when {}`. The performance impact of setting this every tick should not be noticeable.

File: src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt (hunk 3 above)
Comment (zumbiepig) [reply src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt]: I made the change and just used prefixes, and I added a `// TODO` for when you enable this.

File: src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt (hunk 4 above)
Comment (zumbiepig) [reply src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt]: Moved these checks into the enum.

File: src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt (hunk 5 above)
Comment (zumbiepig) [reply src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt]: Made this a standalone function.

File: src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt (hunk 6 above)
Comment (zumbiepig) [reply src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt]: Implemented this change.

**Comment by zumbiepig** [comment]
@hannibal002 can you take a look at this again, please?

File: src/main/java/at/hannibal2/skyhanni/config/features/garden/SensitivityReducerConfig.kt [hunk 10]
```diff
@@ -1,63 +1,94 @@
...
+
+    @Expose
+    @ConfigOption(name = "Disable in Barn or Greenhouse", desc = "Disable reduced sensitivity in barn and greenhouse plots.")
+    @ConfigEditorBoolean
+    var onlyPlot: Boolean = true
 
     @Expose
     @ConfigLink(owner = SensitivityReducerConfig::class, field = "showGui")
-    val position: Position = Position(400, 400, 0.8f)
+    val display: Position = Position(400, 200)
```
hannibal002: do not rename position to display. it is internally a position object, and we everywhere use the name position, the change is confusing (and would break the config since no migration), so please revert.

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
```diff
@@ -0,0 +1,224 @@
...
+        "(?<plot>Warping\\.\\.\\.)",
+    )
+
+    @JvmStatic
+    fun remapSensitivity(original: Double): Double = state.transform(original)
+
+    @HandleEvent
+    fun onChat(event: SkyHanniChatEvent.Allow) {
+        if (config.unlockOnTeleport == SensitivityReducerConfig.UnlockOnTeleport.NEVER) return
+        val state = manualState ?: return
```
hannibal002: shadow global variable name. rename

src/main/java/at/hannibal2/skyhanni/config/features/garden/SensitivityReducerConfig.kt
```diff
@@ -1,63 +1,94 @@
...
 
     @Expose
-    @ConfigOption(name = "Only in Ground", desc = "Lower sensitivity when standing on the ground.")
+    @ConfigOption(name = "Chat Message", desc = "Show a message in chat when toggling mouse lock.")
     @ConfigEditorBoolean
-    val onGround: Property<Boolean> = Property.of(false)
+    var chatMessage: Boolean = true
+
+    @Expose
+    @ConfigOption(name = "Reducing percent", desc = "Change how much the sensitivity is lowered by.")
```
hannibal002: confusing desc. the variable is the percentage reducation TO, not reduction BY that value. e.g. the default value 10% is not a reducation by 10% (90% remaining), but a reduction to 10% (a reducttion of 90%).

How about changing the wording to "Set sensitivity to this percentage of your normal sensitivity" ?

hannibal002 (changes requested):
> Added the ability to automatically lock your mouse while farming. - zumbiepig

This ability is an extension of the existing Sensitivity Reducer, not a standalone new feature. Change New Feature -> Improvements

> Huge rewrite of the sensitivity reducer.

would love to read more on the why. what was the reason to do the pr, what was bad/insufficient with the original design? a small technical background.

src/main/java/at/hannibal2/skyhanni/config/features/garden/SensitivityReducerConfig.kt
zumbiepig: fixed

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
zumbiepig: fixed

src/main/java/at/hannibal2/skyhanni/config/features/garden/SensitivityReducerConfig.kt
zumbiepig: i updated the description but i couldnt think of a better name for the variable

zumbiepig:
> > Added the ability to automatically lock your mouse while farming. - zumbiepig
> 
> This ability is an extension of the existing Sensitivity Reducer, not a standalone new feature. Change New Feature -> Improvements

fixed

zumbiepig:
> > Huge rewrite of the sensitivity reducer.
> 
> would love to read more on the why. what was the reason to do the pr, what was bad/insufficient with the original design? a small technical background.

@hannibal002 

it was badly written, theres no way to lock mouse except 10k reducing factor which had some drifting problems even after the fix and even if i dont touch my mouse at all

the original file had way too many functions that either literally didnt do anything or were hard to understand. same with the variables

the actual reducing implementation also sucked and to see what i mean, compare the old one side by side with the new one and move your mouse a bit.

also some features were conflicting and like /shsensreduce didnt even work

then there were just some features i wanted to add but it was really hard to do with the current state of it

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
```diff
@@ -0,0 +1,222 @@
...
+            }
+        }
+        event.registerBrigadier("shmouselock") {
+            description = "Lock/Unlock the mouse so it will no longer rotate the player (for farming)"
+            category = CommandCategory.USERS_ACTIVE
+            aliases = listOf("shlockmouse")
+            simpleCallback {
+                if (manualState != SensitivityState.LOCKED) {
+                    manualState = SensitivityState.LOCKED
+                    ChatUtils.chat("Mouse rotation is now locked. Type /shlockmouse to unlock your mouse.", messageId = commandMessageId)
```
hannibal002: dont show the alias, show the main command. also on click on the message run the command

src/main/java/at/hannibal2/skyhanni/config/features/garden/SensitivityReducerConfig.kt
```diff
@@ -1,63 +1,94 @@
...
 
     @Expose
-    @ConfigOption(name = "Only in Ground", desc = "Lower sensitivity when standing on the ground.")
+    @ConfigOption(name = "Chat Message", desc = "Show a message in chat when toggling mouse lock.")
     @ConfigEditorBoolean
-    val onGround: Property<Boolean> = Property.of(false)
+    var chatMessage: Boolean = true
+
+    @Expose
+    @ConfigOption(name = "Reducing percent", desc = "Set sensitivity to this percentage of your normal sensitivity")
```
hannibal002: percent in name should be upper case (name of feature, a title)

src/main/java/at/hannibal2/skyhanni/config/features/garden/SensitivityReducerConfig.kt
```diff
@@ -1,63 +1,94 @@
...
-    @ConfigOption(name = "Disable in Barn", desc = "Disable reduced sensitivity in barn plot.")
+    @ConfigOption(name = "Lock mouse", desc = "Lock the mouse instead of reducing sensitivity.")
     @ConfigEditorBoolean
-    val onlyPlot: Property<Boolean> = Property.of(true)
+    var lockMouse: Boolean = false
+
+    @Expose
+    @ConfigOption(name = "Unlock on Teleport", desc = "Choose whether teleporting to a plot should unlock your mouse rotation.")
+    @ConfigEditorDropdown
+    val unlockOnTeleport: UnlockOnTeleport = UnlockOnTeleport.ALWAYS
```
hannibal002: should be var, not val

src/main/java/at/hannibal2/skyhanni/features/garden/SensitivityReducer.kt
```diff
@@ -0,0 +1,222 @@
...
+import at.hannibal2.skyhanni.utils.RenderUtils.renderRenderable
+import at.hannibal2.skyhanni.utils.renderables.Renderable
+import at.hannibal2.skyhanni.utils.renderables.primitives.text
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+import com.google.gson.JsonArray
+import com.google.gson.JsonPrimitive
+import net.minecraft.client.Minecraft
+
+@SkyHanniModule
+object SensitivityReducer {
```
hannibal002: class name should contain the name "mouse" somewhere

src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt
```diff
@@ -1,5 +1,6 @@
 package at.hannibal2.skyhanni.utils
 
+import at.hannibal2.skyhanni.utils.LocationUtils.canSeeFace
```
hannibal002: unnecessary import

src/main/java/at/hannibal2/skyhanni/utils/BlockUtils.kt
```diff
@@ -62,7 +62,7 @@ object BlockUtils {
         LocationUtils.playerEyeLocation(),
         MinecraftCompat.localPlayer.lookAngle.toLorenzVec(),
         distance,
-    )?.roundToBlock()
+    ).roundToBlock()
```
hannibal002, line 65: usages of getTargetedBlockAtDistance are now wrong. e.g. in SkyHanniDebugsAndTests.

src/main/java/at/hannibal2/skyhanni/config/features/garden/MouseSensitivityReducerConfig.kt
```diff
@@ -1,63 +1,94 @@
...
+
+    @Expose
+    @ConfigOption(name = "Disable in Barn or Greenhouse", desc = "Disable reduced sensitivity in barn and greenhouse plots.")
+    @ConfigEditorBoolean
+    var onlyPlot: Boolean = true
 
     @Expose
     @ConfigLink(owner = SensitivityReducerConfig::class, field = "showGui")
-    val position: Position = Position(400, 400, 0.8f)
+    val position: Position = Position(400, 200)
```
hannibal002, line 84: position has no migration (from garden.mouseLock.display to garden.sensitivityReducer.display)

src/main/java/at/hannibal2/skyhanni/config/features/garden/SensitivityReducerConfig.kt
```diff
@@ -1,63 +1,94 @@
...
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorDropdown
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorInfoText
 import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorKeybind
 import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorSlider
 import io.github.notenoughupdates.moulconfig.annotations.ConfigLink
 import io.github.notenoughupdates.moulconfig.annotations.ConfigOption
-import io.github.notenoughupdates.moulconfig.observer.Property
 import org.lwjgl.glfw.GLFW
 
 class SensitivityReducerConfig {
```
hannibal002: the class name should have the mouse word somewhere. (the sensitivity of "what" gets reduced?)

src/main/java/at/hannibal2/skyhanni/config/features/garden/GardenConfig.kt
```diff
@@ -132,11 +132,6 @@ class GardenConfig {
     @Accordion
     val yawPitchDisplay: YawPitchDisplayConfig = YawPitchDisplayConfig()
 
-    @Expose
-    @ConfigOption(name = "Mouse Lock", desc = "")
-    @Accordion
-    val mouseLock: MouseLockConfig = MouseLockConfig()
-
     @Expose
     @ConfigOption(name = "Sensitivity Reducer", desc = "")
```
hannibal002: missing name "mouse"

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt
hannibal002: this is still unnecessary

src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt
```diff
@@ -0,0 +1,224 @@
...
+import com.google.gson.JsonArray
+import com.google.gson.JsonPrimitive
+import net.minecraft.client.Minecraft
+
+@SkyHanniModule
+object MouseSensitivityReducer {
+
+    private val config get() = GardenApi.config.sensitivityReducer
+
+    private val commandMessageId = ChatUtils.getUniqueMessageId()
```
hannibal002: the message id is shared between sens reduce and mouse lock. intended?

src/main/java/at/hannibal2/skyhanni/config/features/garden/MouseSensitivityReducerConfig.kt
```diff
@@ -0,0 +1,94 @@
...
+        name = "Note",
+        desc = "You can type §e/shmouselock §rto lock your mouse rotation, and §e/shsensreduce §rto reduce your sensitivity.",
+    )
+    @ConfigEditorInfoText
+    var notice: String = ""
+
+    @Expose
+    @ConfigOption(name = "Auto Enable", desc = "Automatically lower mouse sensitivity while in the garden.")
+    @ConfigEditorBoolean
+    var enabled: Boolean = false
```
hannibal002: use a better variable name. this suggests it is the main toggle, it isnt.

src/test/java/at/hannibal2/skyhanni/test/garden/MouseSensitivityHookTest.kt
```diff
@@ -1,20 +0,0 @@
-package at.hannibal2.skyhanni.test.garden
-
-import at.hannibal2.skyhanni.features.garden.sensitivity.MouseSensitivityManager
-import at.hannibal2.skyhanni.mixins.hooks.MouseSensitivityHook
-import org.junit.jupiter.api.Assertions.assertEquals
-import org.junit.jupiter.api.Test
-
-class MouseSensitivityHookTest {
```
hannibal002, line 8: why is this gone without replacement?

hannibal002 (changes requested):
also i have rewrote the changelog lines entirely. werer way too long/wordy

src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt
zumbiepig: i think my ide is putting it in
i swear i deleted that

src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt
zumbiepig: yes. because the mouse lock/reduce can only be in one state at any point in time, so only the latest is relevant

src/test/java/at/hannibal2/skyhanni/test/garden/MouseSensitivityHookTest.kt
zumbiepig: it was added when we had the floating bug from something like `(1 / 0.333) * 1.333` or smth. the test was used to make sure the added manual override would be fine. the test is not needed now since its just `x * 0`.

src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt
hannibal002: please write this explanation  in the code as comment above the variable

src/test/java/at/hannibal2/skyhanni/test/garden/MouseSensitivityHookTest.kt
hannibal002: fair

src/main/java/at/hannibal2/skyhanni/config/features/garden/MouseSensitivityReducerConfig.kt
zumbiepig: this is the old name. i would have to add a config migration

hannibal002: was the logic behind this toggle the same prior to your pr?

src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt
zumbiepig: @hannibal002 yeah i just deleted it and intellij immediately put it back.

src/main/java/at/hannibal2/skyhanni/config/features/garden/MouseSensitivityReducerConfig.kt
zumbiepig: yes. i renamed it anyways since i needed to rename the whole config to include the word mouse

src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt
```diff
@@ -0,0 +1,204 @@
...
+                else -> newList.add(element.asString)
+            }
+            newList
+        }
+        // old migrations from separate mouse lock feature
+        event.move(135, "misc.lockMouseLookChatMessage", "garden.mouseLock.chatMessage")
+        event.move(135, "misc.lockedMouseDisplay", "garden.mouseLock.display")
+        // variable renames
+        event.move(137, oldBase, base)
+        event.move(137, "$base.enable", "$base.autoEnable")
```
hannibal002: the old name was enabled, not enable

src/main/java/at/hannibal2/skyhanni/config/features/garden/MouseSensitivityReducerConfig.kt
```diff
@@ -0,0 +1,111 @@
...
+    @Expose
+    @ConfigOption(name = "Disable in Barn or Greenhouse", desc = "Disable reduced sensitivity in barn and greenhouse plots.")
+    @ConfigEditorBoolean
+    var onlyPlot: Boolean = true
+
+    @Expose
+    @ConfigLink(owner = MouseSensitivityReducerConfig::class, field = "showGui")
+    val position: Position = Position(400, 200)
+
+    enum class AutoMode(private val displayName: String, val condition: () -> Boolean) {
```
hannibal002: the enum belongs in a feature class, not a config. especially because the config now imports from many different feature classes, that should not be the case.

src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt
```diff
@@ -0,0 +1,204 @@
...
+    fun onChat(event: SkyHanniChatEvent.Allow) {
+        if (manualState == null || config.unlockOnTeleport == MouseSensitivityReducerConfig.UnlockOnTeleport.NEVER) return
+
+        teleportPattern.matchMatchers(event.cleanMessage) {
+            if (config.unlockOnTeleport.condition(group("plot"))) {
+                val message = if (manualState == SensitivityState.REDUCED) "Mouse sensitivity has been restored because you teleported."
+                else "Mouse rotation has been unlocked because you teleported."
+
+                manualState = null
+                update()
```
hannibal002, line 64: in the old code, this was delayed by one tick. now, its instantaneous. do we need that more save delay back, or do you know why we do no longer need it?

src/main/java/at/hannibal2/skyhanni/config/features/garden/MouseSensitivityReducerConfig.kt
```diff
@@ -0,0 +1,111 @@
...
+    @ConfigEditorKeybind(defaultKey = GLFW.GLFW_KEY_N)
+    var keybind: Int = GLFW.GLFW_KEY_N
+
+    @Expose
+    @ConfigOption(name = "Show GUI", desc = "Show the GUI element while the feature is enabled.")
+    @ConfigEditorBoolean
+    var showGui: Boolean = true
+
+    @Expose
+    @ConfigOption(name = "Chat Message", desc = "Show a message in chat when toggling mouse lock.")
```
hannibal002: desc is a bit inaccurate. how about "Show a message in chat when toggling the mouse sensitivity or lock state."?

src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt
```diff
@@ -0,0 +1,204 @@
...
+    private var activeState: SensitivityState = SensitivityState.UNCHANGED
+    private var manualState: SensitivityState? = null
+
+    /**
+     * REGEX-TEST: Teleported you to Plot - 1!
+     * REGEX-TEST: Teleported you to Plot - 20!
+     * REGEX-TEST: Teleported you to The Barn!
+     * REGEX-TEST: Warping...
+     */
+    private val teleportPattern by RepoPattern.list(
```
hannibal002, line 47: regex fail test for "Teleported you to the spawn location!" is gone.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt
zumbiepig: ok so

theres a fake chat message by skyhanni saying ChatUtils.chat("Teleported you to the spawn location!", prefix = false) for some reason

1. that doesnt even include the word "plot"
2. i dont think that even ever makes it into @HandleEvent onChat

lunaynx: I don't know why we send a message there, but I think that's safe to ignore – you pretty much *never* want to unlock the mouse when you just do `/warp garden`, as that's usually the rewarp command. If you do add a toggle for it anyway, it should be entirely separate and disabled by default.

src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt
zumbiepig: oh i maybe probably put the delay back, but idk if it really matters

its based on the assumption that i need to give GardenApi an extra tick to realize that we changed plots

but then, this depends on hypixel updating the scoreboard, which i dont think happens every 1 or 2 ticks anyways

lunaynx: As far as I could tell, GardenApi determines what plot you are in based on your position, since I had to make a fix for edge cases with Disco Destination – so a delay shouldn't be necessary in theory.

zumbiepig: i figured that

1. most people using /shmouselock manually will NOT have "auto enable" on, so we dont need the delay
2. or if they are using auto enable, its fine since onChat only matters for manual enable with the command

src/main/java/at/hannibal2/skyhanni/config/features/garden/MouseSensitivityReducerConfig.kt
zumbiepig: ok

src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt
zumbiepig: ok

src/main/java/at/hannibal2/skyhanni/config/features/garden/MouseSensitivityReducerConfig.kt
zumbiepig: ok

src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt
zumbiepig: hmm i can make this a dropdown

zumbiepig: wait, it already has this

image

this seems sufficient

lunaynx: Yeah, but `/warp garden` (and its aliases like `/home`) should still be excluded from this unless it's made a separate option, and I'm not quite sure at a glance what a new good wording would be for it – I would just not add it unless someone specifically requests it with a strong use case.

Currently:
- Always is good for people who *don't* have Disco Destination and teleport & fly around multiple plots to kill pests
- Never or Barn Only is good for people who *do* have Disco Destination

src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt
zumbiepig: > As far as I could tell, GardenApi determines what plot you are in based on your position, since I had to make a fix for edge cases with Disco Destination – so a delay shouldn't be necessary in theory.

This might be a separate issue, but this is currently how the check is implemented. btw, why dont we just check this every tick or every player movement?

image

image

src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt
zumbiepig: i explicitly made /warp garden not unlock it for now. i added a TODO so we can decide what to do with this later, if someone requests the feature

src/main/java/at/hannibal2/skyhanni/features/garden/MouseSensitivityReducer.kt
lunaynx: Yeah, I don't think we need to delay it that aggressively, ~25 bounding box contains checks can probably be done every tick with an early return for "still in the same plot as last tick". I'll try that and then make a PR if it works well.

src/main/java/at/hannibal2/skyhanni/config/features/garden/MouseSensitivityReducerConfig.kt
```diff
@@ -0,0 +1,94 @@
...
+    @ConfigOption(name = "Disable in Barn or Greenhouse", desc = "Disable reduced sensitivity in barn and greenhouse plots.")
+    @ConfigEditorBoolean
+    var onlyPlot: Boolean = true
+
+    @Expose
+    @ConfigLink(owner = MouseSensitivityReducerConfig::class, field = "showGui")
+    val position: Position = Position(400, 200)
+
+    enum class UnlockOnTeleport(private val displayName: String, val condition: (String) -> Boolean) {
+        ALWAYS("Always", { it != "Warping..." }), // TODO: decide what to do with /warp garden
```
hannibal002, line 87: so always means always except with /warp garden? what?

src/main/java/at/hannibal2/skyhanni/utils/HypixelCommands.kt
```diff
@@ -71,8 +71,8 @@ object HypixelCommands {
         send("gardenlevels")
     }
 
-    fun setHome() {
-        send("sethome")
+    fun setSpawn() {
```
hannibal002, line 74: why do we change sethome to setspawn? is there a difference?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/HypixelCommands.kt
lunaynx: > why do we change sethome to setspawn? is there a difference?

They're the same thing, it's entirely subjective.

src/main/java/at/hannibal2/skyhanni/config/features/garden/MouseSensitivityReducerConfig.kt
lunaynx: Yes, this is on purpose. 99% of the time people do not want their mouse to be auto-unlocked when doing `/warp garden` since that's used as a "rewarp to last farming position or beginning of farm" action. Auto-unlocking when teleporting to a plot via `/tptoplot` makes sense for pest farming if you don't have the Disco Destination "suck-box" setup that collects all pests in one convenient box with no mouse movement required whatsoever; and auto-unlocking when teleporting to the barn also generally makes sense to take care of visitors, composter, etc.

Maybe the wording can be clarified here because technically `/warp garden` can also be interpreted as "teleporting to a plot", I just don't know how. Possibly make it some sort of draggable list for deciding when to unlock, and then if we really want we can also add the `/warp garden` case, although almost no one is going to use that.

hannibal002 (approved):

src/main/java/at/hannibal2/skyhanni/utils/HypixelCommands.kt
zumbiepig: i changed it because it was the preferred name in the official hypixel wiki. also nobody calls it /sethome, ppl usually refer to it as /setspawn. i just didnt rename the feature internal variable names cuz i was avoiding config migrations

hannibal002: yeah that makes sense!

lunaynx:
> * Fixed `/shmouselock` not unlocking when warping to the garden. - zumbiepig

This line should be removed from the changelog, because this "fix" was reverted.

PR #5799, opened 2026-05-26
Fix: Tia the Fairy 9f Relay positions

f4bii:
## What
Hypixel changed location of the Relay on top of the Auction House

## Changelog Fixes
+ Fixed 9f Relay positions being wrong. - FabiHBBBT


lunaynx:
Yeah, I was gonna fix this, but I was like "this should be in repo", and thus procrastination ensues... better fixed than not.

PR #5806, opened 2026-05-28
Fix: Skip bait update while in GUI

lunaynx:
## What
When you open a GUI other than your own inventory, Hypixel changes the bait slot back to the SkyBlock Menu, which was resulting in false "Bait changed: [bait] -> None" warnings.

## Changelog Fixes
+ Fixed false fishing bait change warnings when opening a GUI while holding a fishing rod. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/fishing/FishingApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/fishing/FishingApi.kt
@@ -42,6 +42,9 @@ import at.hannibal2.skyhanni.utils.compat.getCompoundOrDefault
 import at.hannibal2.skyhanni.utils.compat.getStringOrDefault
 import at.hannibal2.skyhanni.utils.getLorenzVec
 import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/fishing/FishingApi.kt
```diff
@@ -232,6 +236,11 @@ object FishingApi {
         postBaitUpdate(null, 0, ItemStack.EMPTY)
     }
 
+    private fun hasGuiOpen(): Boolean {
```
zumbiepig, line 239: @lunaynx can you move this to a common utils class? (then replace every usage of it in the code)
also i think you can just do this:
Image

hannibal002: The function name is a bit unclear: what is a gui? a custom mod gui, a mc config gui (esc), a custom inventory, a normal chest invetnroy, the player inventory? please be more descriptive in the name or add a comment.

hannibal002 (changes requested):

PR #5807, opened 2026-05-28
Backend: Fix config migrator not adding values to all profiles

f4bii:
## What
Fix config migrator not adding values to all profiles

+ Fixed config migrator not adding values to all profiles. - FabiHBBBT


NopoTheGamer:
it should still return right ?

f4bii:
I was thinking this just rn. Its 2am gonna add a return tomorrow after the for loop 🥺

PR #5811, opened 2026-05-29
Carry tracker

zumbiepig:
improve the carry tracker and include stuff like:
- coins per hour
- bosses per hour
- multiple prices
- trade detect and add carries
- party chat counter
- spawn notification
- some more

image


zumbiepig:
wrong branch

PR #5812, opened 2026-05-29
Carry tracker

zumbiepig:
## What

improve the carry tracker and include stuff like:
- coins per hour
- bosses per hour
- multiple prices
- trade detect and add carries
- manually update total, completed, and coins traded on the counter
- overflow carries and coins (eg 11/10)
- bulk price detection (10+)
- party chat counter
- spawn notification
- some more

image
image


hannibal002:
Couple questions:

coins per hour: how do you detekt while the player is doing the carry/waiting on the customer to collect xp and not being idle? Manually or automatic? If automatic, how? Or are you just counting the money you get versus the time when the boss got killed? That's then inaccurate around the beginning/end of the session.
bosses per hour: that's useful, but also relevant to differentiate between downtime and no dt
multiple prices: for the same tier and boss? Why would we want that? You sell the same carry to multiple users for different prices?
trade detect and add carries: what do you mean exactly? Auto add new carries when someone pays you? How do you know what type of slwyer? Because of the area the player is at?
manually update total, completed, and coins  traded on the counter: somewhat useful if data gor bugged yes, as long as it's intuitive and doable without commands. 
overflow carries and coins (eg 11/10): what do you want to do in that case? The carry feature already supports that as is
bulk price detection (10+): what do you mean?
party chat counter: this does already exist
spawn notification: in party chat? Or to the user doing the carry? That's useful 
some more: please don't do too much stuff in one pr. Pick couple small useful things or one big thing and leave the rest for after the first pr is merged. Otherwise we end up with a huge pr that takes ages to review and merge.

zumbiepig:
@hannibal002

> Couple questions:
>
> coins per hour: how do you detect while the player is doing the carry/waiting on the customer to collect xp and not being idle? Manually or automatic? If automatic, how? Or are you just counting the money you get versus the time when the boss got killed? That's then inaccurate around the beginning/end of the session.

I don't detect if the client is idle, it's not relevant. I'm counting the coins/hr based on `bosses/hr * coins per carry`.

> bosses per hour: that's useful, but also relevant to differentiate between downtime and no dt

I'm not going to detect downtime, because the clock still ticks while your client goes to call maddox. And if they take a long time to spawn a boss, the coins per hour you are making from that client is much lower. This is the exact use case of the bosses/hr information, so that you can be informed which clients are slower.

> multiple prices: for the same tier and boss? Why would we want that? You sell the same carry to multiple users for different prices?

So that you can understand, here are some of the carry prices for various popular carry servers, just for eman carries in the Void Sepulture alone (carries in the Bruiser Hideout are more expensive, as spawns take longer). I'm not even going to list blaze carry prices, but my point is that the current system is not very flexible for handling these different prices, and requires me to manually calculate prices for many clients.

For eman carries in the sepulture (t3 / t4 / t4 bulk 10+):
SBZ: 0 (free to unlock t4) / 1.3m / 1m
SBM: 700k / 1.8m / 1.5m
Dungeon Hub: 800k / 1.6m / 1.4m
F10: 900k / 1.6m / 1.4m (10+) / 1.2m (25+)

> trade detect and add carries: what do you mean exactly? Auto add new carries when someone pays you? How do you know what type of slwyer? Because of the area the player is at?

Zen and Athen currently have this feature. How it works is the carrier configures a list of carry prices for e.g. eman t4. They can set the value to something like `1,1.3,1.5,1.8`. Anytime coins are traded to you that is a multiple of those amounts, it gives a chat prompt with one of these two messages:
`[ModName] Received payment for 50x Void T4 carries from Player_Name. Click to add`
`[ModName] Found multiple matches, add for: [600x Void T4] [300x Blaze T2] [500x Blaze T2] [200x Blaze T3] [240x Blaze T3] [100x Blaze T4] [120x Blaze T4]`

> manually update total, completed, and coins  traded on the counter: somewhat useful if data got bugged yes, as long as it's intuitive and doable without commands.

It's going to be done through commands. I might add a GUI later, but commands are perfectly fine, as long as we add command auto-complete (which I'm planning to do).

> overflow carries and coins (eg 11/10): what do you want to do in that case? The carry feature already supports that as is

Oh cool. I didn't test it out and I just assumed it stopped counting.

> bulk price detection (10+): what do you mean?

Tbh, I was just listing ideas and we shouldn't need to detect anything if we just do the system I mentioned above where the carrier clicks an option.

> party chat counter: this does already exist

Currently you must manually click on the player's name on the carry counter hud to send it out. What I mean is when I kill a boss for PlayerA, a message is sent out in party chat saying `PlayerA: 6/10`.

> spawn notification: in party chat? Or to the user doing the carry? That's useful

It would be helpful to both the carrier and the client (as some clients don't realize that their boss spawned, then dies to it and wastes my time). The notification would be sent out in party chat, as well as a big title + ding on the carrier's screen. Another scenario this is useful for is if I have several clients spawning their bosses at the same time, then if one client spawns their boss other clients know not to spawn theirs at the same time (hopefully).

This feature does not actually exist in any other mods right now.

> some more: please don't do too much stuff in one pr. Pick couple small useful things or one big thing and leave the rest for after the first pr is merged. Otherwise we end up with a huge pr that takes ages to review and merge.

Don't worry, this PR won't be that big (as I shouldn't need to touch too many different files). A lot of the framework that's needed for these features already exist, I just need to implement the detection and logic.

hannibal002:
> Don't worry, this PR won't be that big (as I shouldn't need to touch too many different files). A lot of the framework that's needed for these features already exist, I just need to implement the detection and logic.

I'm not worried about the amount of files you change, I'm worried abou the amount of stuff you add that can be tested per pr. Currently, this prs description reads like a wish list of different features, but every single one should be ideally one independent pr. Then this whole list and the discussion around the ideas should be on discord in one suggestion channel instead.

hannibal002:
Also I just realised that this pr doesn't contain any code. That's not the point of a pr. Use discord for planning/discussion phases. A pull request, even when as a draft, is for showing off code.

PR #5819, opened 2026-06-01
Fix: Don't fire ItemAddEvent for armor slots

lunaynx:
## What
Fixes some annoying bugs with profit trackers where certain armor pieces (e.g. Tiki Mask) and bait (if in repo profit items) would get falsely tracked. The latter is due to a recent Hypixel change of showing bait in slot 9.

## Changelog Fixes
+ Fixed profit trackers falsely reporting gaining an item when equipping it in your wardrobe (e.g. Tiki Mask) (hopefully). - Luna
+ Fixed fishing profit tracker falsely reporting gaining bait when using it. - Luna

+ OwnInventoryItemUpdateEvent is now only posted for actual inventory slots, excluding armor slots and the last hotbar slot (SkyBlock Menu). - Luna
+ Added OwnInventoryArmorUpdateEvent for updates to the player's armor slots. - Luna
+ Added OwnInventoryMenuUpdateEvent for updates to the last hotbar slot (where the SkyBlock Menu generally resides). - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/OwnInventoryData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/OwnInventoryData.kt
@@ -35,7 +35,7 @@ object OwnInventoryData {
 
     private var itemAmounts = mapOf<NeuInternalName, Int>()
     private var dirty = false
-    private var lastWardrobeClose = SimpleTimeMark.farPast()
+    private val armorSlots = 36..39
 
     /**
      * REGEX-TEST: §aMoved §r§e10 Wheat§r§a from your Sacks to your inventory.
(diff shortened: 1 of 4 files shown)


NopoTheGamer:
btw he make this pr to fix this issue, so reverting it also doesnt fix the issue

if you also add my closed pr into the mix it will fix it but only for tiki masks and not like cog

lunaynx:
Look more closely, I added a check to ignore armor slots. I did also remove the wardrobe check though, because it's redundant now.

NopoTheGamer:
doesnt fix the old problem, just brings it back

lunaynx:
How does it not fix it? I tested it and it seemed to work. Or is there another thing my change breaks?

lunaynx:
It should be fine now, I actually had the slots wrong. Taking it off *shouldn't* trigger it because the amount doesn't update for ItemAddEvent, I tested both in normal inventory and `/eq`. But if it somehow does, then it probably would have done it already, so this is a step up from the status quo.

PR #5826, opened 2026-06-03
Fix: Invalid ModelPartFeatureRenderer mixin on 26.1

lunaynx:
## What
This didn't get caught by the mixin force-load. It doesn't crash the game, but it probably breaks some glow-related features.

## Changelog Fixes
+ Fixed a potential issue with glowing entities on 26.1. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/mixins/transformers/renderer/MixinModelPartFeatureRenderer.java
+++ b/src/main/java/at/hannibal2/skyhanni/mixins/transformers/renderer/MixinModelPartFeatureRenderer.java
@@ -25,8 +25,7 @@ public abstract class MixinModelPartFeatureRenderer {
         OutlineBufferSource outlineConsumer,
         int color,
         Operation<Void> original,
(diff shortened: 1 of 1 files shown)


NopoTheGamer:
whats different?

lunaynx:
I removed the `argsOnly` because apparently it was wrong to include that even on 26.1, it wasn't picking up the `ModelPartSubmit` with it. Haven't tested what actually changes in game yet, I'll probably do it after Moonflower is out of season. But the broken mixin probably makes some model parts not glow or use the wrong color.

NopoTheGamer:
the 26.1 code hasnt changed ?

lunaynx:
Yes it has, the replacement condition was `< 26.1`.

NopoTheGamer:
is the syntax really backwards

lunaynx:
The committed source files are 26.1 now, not 1.21.11. But if you prefer, we can flip all those annotations from `if < 26.1 'A' -> 'B'` to `if >= 26.1 'B' -> 'A'` and it will still work correctly when generating 1.21.11 sources.

PR #5838, opened 2026-06-05
Fix: Slayer Boss Warning title never showing

NicolasRieu:
## What

The original regex pattern was:
`" \\(?(?<progress>[0-9.,k]+)/(?<total>[0-9.,k]+)\\)?.*"`

The pattern already correctly made parentheses optional (`\\(?` / `\\)?`) to handle both progress formats:
- Kill-based quests: `12/40 Kills` (no parentheses)
- XP-based quests: `(0/4.8k) Combat XP` (with parentheses)

The actual bug was a mandatory leading space at the start of the pattern.
`SlayerApi.getSlayerLines()` applies `.trim()` to every scoreboard and tab widget line before storing them, so the progress string never starts with a space. 
This caused the pattern to never match, meaning the completion ratio was never computed and the title was never sent.

The fix removes the leading space and tightens the number format from `[0-9.,k]+` to `[0-9,.]+k?`, making the `k` suffix explicitly optional at the end rather than mixed into the character class.

Also adds a chat message alongside the title to follow project conventions.

## Changelog Fixes
+ Fixed Slayer Boss warning title never appearing. - iLLuMiiNaTy


src/main/java/at/hannibal2/skyhanni/features/slayer/SlayerBossSpawnSoon.kt
```diff
@@ -17,9 +18,13 @@ object SlayerBossSpawnSoon {
 
     private val config get() = SlayerApi.config.slayerBossWarning
 
+    /**
+     * REGEX-TEST: (120/500) Atomic Slayer
```
lunaynx: Regex tests should be real examples from the game, not strings made up to fit the pattern.

lunaynx (changes requested):

Rainymei:
please do not use ai to make your reviews

lunaynx:
Please turn-off Copilot auto-review, we do not want our pull request threads to be cluttered with AI slop. On your own computer, you're free to do whatever you want, as long as you understand and test the code.

NicolasRieu:
Changes done! 
Also sorry for the noise caused by GitHub Copilot auto-reviews, I wasn't aware it was enabled, I disabled it.

hannibal002 (approved):

PR #5839, opened 2026-06-05
Fix: False pest spawn detection on player messages

lunaynx:
## What
Fixes pest spawns being falsely detected if someone sends you a message containing the right thing, because of the overly broad regex. Also switched it to `SystemMessageEvent` to further harden against this issue in the future.

## Changelog Fixes
+ Fixed some player messages being falsely detected as pest spawns. - Luna


CalMWolfs (approved):

hannibal002 (changes requested):
regex:

PestSpawn.kt:53:37: Repo pattern offlinePestsPattern passed regex test even though it was set to fail [RepoPatternRegexTestFailed]

hannibal002 (approved):

PR #5840, opened 2026-06-05
Backend: Fix generateRepoPatterns not working

lunaynx:
## What
There may be a better way to do this, but this will do for now. (Can't do `mods.from(project.configurations.getByName("implementation")`.)

+ Fixed generateRepoPatterns Gradle task not working. - Luna


CalMWolfs (approved):
looks like good to  me

NopoTheGamer:
drunk cal said it was ok

PR #5842, opened 2026-06-06
Fix: Config button clicks registering multiple times

lunaynx:
## What
Fixes config buttons triggering on press, release, *and* mouse movement, often causing the "Manually download" button to open Modrinth multiple times.

## Changelog Fixes
+ Fixed "Manually Download" button opening Modrinth multiple times. - Luna


hannibal002 (changes requested):
detekt:

GuiOptionEditorUpdateCheck.kt:8:1: Unused import [NoUnusedImports]

src/main/java/at/hannibal2/skyhanni/features/misc/update/GuiOptionEditorUpdateCheck.kt
```diff
@@ -86,17 +114,21 @@ class GuiOptionEditorUpdateCheck(option: ProcessedOption) : GuiOptionEditor(opti
...
-        if (!isInside(getChangelogPosition(width - 20), height = 30, changelog)) return false
 
-        if (UpdateManager.updateState != UpdateManager.UpdateState.NONE)
-            UpdateManager.getNextVersion()?.let { ChangelogViewer.showChangelog(currentVersion, it) }
-                ?: ErrorManager.logErrorStateWithData(
-                    "Can't get Changelog because of internal error",
-                    "UpdateManager.getNextVersion is null even though updateState is != NONE",
-                    "state" to UpdateManager.updateState,
-                )
+        if (changelogClicked) {
```
hannibal002: early return

src/main/java/at/hannibal2/skyhanni/features/misc/update/GuiOptionEditorUpdateCheck.kt
```diff
@@ -63,20 +62,49 @@ class GuiOptionEditorUpdateCheck(option: ProcessedOption) : GuiOptionEditor(opti
...
+
         fun isInside(width: Int, height: Int, def: GuiElementButton): Boolean {
             val inX = (mouseX - width - x) in (0..def.width)
             val inY = (mouseY - height - y) in (0..def.height)
-            return MouseCompat.getEventButtonState() && inX && inY
+            return inX && inY
         }
 
-        if (isInside(getButtonPosition(width - 20), height = 10, button)) {
+        val isMouseButtonDown = event.mouseState
```
hannibal002, line 99: the new code from here until to ` wasInsideChangelog = isInsideChangelog` feels like it belongs into a new function, instead of bloating the 50+ lines function

src/main/java/at/hannibal2/skyhanni/features/misc/update/GuiOptionEditorUpdateCheck.kt
```diff
@@ -16,7 +15,7 @@ import kotlin.math.max
 
 class GuiOptionEditorUpdateCheck(option: ProcessedOption) : GuiOptionEditor(option) {
 
-    val button = GuiElementButton()
+    val download = GuiElementButton()
```
hannibal002: those three variables should be private

src/main/java/at/hannibal2/skyhanni/features/misc/update/GuiOptionEditorUpdateCheck.kt
```diff
@@ -63,20 +62,49 @@ class GuiOptionEditorUpdateCheck(option: ProcessedOption) : GuiOptionEditor(opti
...
 
-    private fun getButtonPosition(width: Int) = width - button.width
+    private fun getDownloadPosition(width: Int) = width - download.width
     private fun getChangelogPosition(width: Int) = width - changelog.width
     override fun getHeight(): Int {
         return 55
     }
 
-    override fun mouseInput(x: Int, y: Int, width: Int, mouseX: Int, mouseY: Int, mouseEvent: MouseEvent): Boolean {
+    private var wasMouseButtonDown = false
```
hannibal002, line 72: those 3 variables should be reset when exiting the config to avoid edge cases

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/update/GuiOptionEditorUpdateCheck.kt
lunaynx: I changed it to use a `when` expression instead for some increased cleanliness, because it felt weird to only early-return for `changelogClicked` just because it happens to be the last one but is otherwise the same pattern as `downloadClicked`.

src/main/java/at/hannibal2/skyhanni/features/misc/update/GuiOptionEditorUpdateCheck.kt
lunaynx: Maybe, but it's not that many lines, and changing this would require passing a bunch of state around.

src/main/java/at/hannibal2/skyhanni/features/misc/update/GuiOptionEditorUpdateCheck.kt
lunaynx: Seems a little annoying to do with MoulConfig, as I don't see an easily accessible "on close" listener, but maybe I can figure something out.

hannibal002: > Seems a little annoying to do with MoulConfig, as I don't see an easily accessible "on close" listener, but maybe I can figure something out.

simplest solution would be a in house skyhanni thingy: a tick based "is the moulconfig open" check, then fireing a "sh config close event"

hannibal002 (changes requested):
Without a reset on config close, all three state variables survive between sessions. The critical scenario: user presses the download button, closes the config without releasing the mouse (e.g. presses Escape), reopens it, and moves their cursor onto the button. At that point `wasMouseButtonDown = true` and `wasInsideDownload = true` are still set from the previous session. The next mouse-up event while hovering the button fires the download action without the user ever intentionally clicking it.

A SkyHanni config close event (driven by a tick-based MoulConfig-open check) would let this class reset all three variables on close, eliminating the stale state entirely.

src/main/java/at/hannibal2/skyhanni/features/misc/update/GuiOptionEditorUpdateCheck.kt
lunaynx: I managed to get something done, probably better than the tick event workaround. There was in fact a usable "on close" listener, it just required a bunch of bridging code.

hannibal002 (approved):

PR #5848, opened 2026-06-07
Fix: ISE in CropMilestones Widget on IslandChange

Rainymei:
## What
Since the ISE is exclusive to IslandChange while holding a tool, lazying the rendering should be everything needed

## Changelog Fixes
+ Fixed a Crop Milestones error being thrown on Island Changes. - Rain


hannibal002 (changes requested):
changelog missing

hannibal002 (changes requested):
- VerticalContainerRenderable has the same eager width/height computation but does not receive the same fix.

- The "What" section doesn't explain what the ISE was, where in the stack trace it originated, or why lazy evaluation prevents it. Could you add a brief explanation or include the relevant stack trace excerpt?

hannibal002:
this is a new one for me
image

PR #5850, opened 2026-06-08
Improvement: Repofiy Garden Visitor chat showing

AverageUser125:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/741

## What
Show Tel Kar dialoge when he is visitor, cause he likes to walk around the entire garden.

## Changelog Improvements
+ Added Tel Kar Garden Visitor to Dialoge Showing Exception list. - AverageUser125


f4bii:
Can we add this to the repo maybe?

AverageUser125:
> Can we add this to the repo maybe?

I ain't that experienced with the repo, but I can try.

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/GardenJson.kt
```diff
@@ -32,6 +32,7 @@ data class GardenVisitor(
     @Expose val mode: String,
     @Expose @SerializedName("need_items") val needItems: List<String>,
     @Expose @SerializedName("unknown_rewards") val unknownRewards: Boolean?,
+    @Expose val showChatMessage: Boolean?,
```
lunaynx: Cleaner if we just ~~flip it and~~ let it default to false.
```suggestion
    @Expose val showChatMessage: Boolean,
```

src/main/java/at/hannibal2/skyhanni/config/features/garden/visitor/VisitorConfig.kt
```diff
@@ -84,7 +84,7 @@ class VisitorConfig {
     @Expose
     @ConfigOption(
         name = "Hide Chat",
-        desc = "Hide chat messages from the visitors in the garden. (Except Beth, Maeve, and Spaceman)"
+        desc = "Hide chat messages from the visitors in the garden. (Except Beth, Maeve, Tel Kar, and Spaceman)"
```
lunaynx: It's better to not specifically list them here anymore since it's going to be remotely updated.
```suggestion
        desc = "Hide chat messages from the visitors in the garden (except those with long dialogues)."
```

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/GardenJson.kt
lunaynx: No wait I'm stupid, show is correct, but let it default to false instead of making it nullable.

AverageUser125: if you don't do `= false` does it default to false?

lunaynx: Yes, it should. It probably doesn't hurt to add it, but IIRC it may be misleading, because sometimes it just silently ignores Kotlin default values, e.g. if you were to try to default it to true instead.

AverageUser125: It’s probably fine, so I just explicitly set it to false.

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/GardenJson.kt
```diff
@@ -32,6 +32,7 @@ data class GardenVisitor(
     @Expose val mode: String,
     @Expose @SerializedName("need_items") val needItems: List<String>,
     @Expose @SerializedName("unknown_rewards") val unknownRewards: Boolean?,
+    @Expose val showChatMessage: Boolean = false,
```
lunaynx: Actually, this should probably be like this. Some legacy fields don't follow this convention, but others in this file already do, and we can't really rename the outliers without breaking the repo for older versions.

Of course, adjust the repo PR too after applying this.
```suggestion
    @Expose @SerializedName("show_chat_message") val showChatMessage: Boolean = false,
```

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/GardenJson.kt
AverageUser125: `skinOrType` doesn't follow that convention, but sure.

lunaynx: Yeah, unfortunately, read my second sentence.

(We could probably soft-migrate those with Gson's alternate names and eventually change it in the repo, but that's out of scope for this PR.)

lunaynx (approved):

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/GardenJson.kt
AverageUser125: My bad for missing that sentence.

(Okay cool, I don't get why we whispering though)

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorChat.kt
```diff
@@ -55,6 +59,16 @@ object GardenVisitorChat {
         "§aYou gave some of the required items!",
     )
 
+    var allowedChatMessageVisitors: Set<String> = emptySet()
```
hannibal002: can be private, move above the regex pattern

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/GardenJson.kt
```diff
@@ -32,6 +32,7 @@ data class GardenVisitor(
     @Expose val mode: String,
     @Expose @SerializedName("need_items") val needItems: List<String>,
     @Expose @SerializedName("unknown_rewards") val unknownRewards: Boolean?,
+    @Expose @SerializedName("show_chat_message") val showChatMessage: Boolean = false,
```
hannibal002, line 35: why is this false, and the filed above nullable? consistency

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/GardenJson.kt
AverageUser125: cause luna asked it to not be nullable and just be false.
So idk.

AverageUser125: Now they both aren't nullable.
consistency fixed.

lunaynx: Yeah, nullability is unnecessary for both actually. I didn't think too deep about `unknown_rewards`, I assumed it really needed to be a tri-state boolean, but it doesn't. And I'm not convinced Gson would even serialize that correctly, unlike default false which should be fine.

AverageUser125: > Yeah, nullability is unnecessary for both actually. I didn't think too deep about `unknown_rewards`, I assumed it really needed to be a tri-state boolean, but it doesn't. And I'm not convinced Gson would even serialize that correctly, unlike default false which should be fine.

It was using 1.5 bits. and just like hypixel we need to save 0.5 bits of bandwidth. /s

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorChat.kt
```diff
@@ -55,6 +60,14 @@ object GardenVisitorChat {
         "§aYou gave some of the required items!",
     )
 
+    private val repoReloadCoroutine = CoroutineSettings("allowed chat message visitors repo reload")
+
+    @HandleEvent
+    fun onRepoReload(event: RepositoryReloadEvent) = repoReloadCoroutine.launch {
+        val data = event.getConstantAsync<GardenJson>("Garden")
+        allowedChatMessageVisitors = data.visitors.filterValues { visitorData -> visitorData.showChatMessage }.keys
```
hannibal002: add a .toSet() at the end to create a clean copy.

hannibal002 (changes requested):

hannibal002 (approved):

PR #5851, opened 2026-06-08
Fix: Inventory overlay GUI scale handling

Akinsoft:
## What
Some HUD overlays were moving or changing size when opening inventories/chests if another mod changed the inventory GUI scale.

This fixes the overlays I found having that issue, including the ones that need to stay clickable while an inventory is open.
This also fixes the position editor being all messed up with a separate inventory GUI scale.

Tested in-game with:
- Item Pickup Log
- Garden Crop Milestone Display
- Bazaar Daily Limit
- Custom Scoreboard
- Visitor Shopping List

Before:
image
image
image
image

After:
image
image
image
image

## Changelog Fixes
+ Fixed some HUD overlays moving or resizing when opening inventories with a separate inventory GUI scale. - akinsoft


hannibal002 (changes requested):
detekt:

* **HotfApi.kt** Line 16, 17: This annotation is currently applied to the value parameter only but in the future it will also be applied to field.
* **HotmApi.kt** Line 112, 113, 153: This annotation is currently applied to the value parameter only but in the future it will also be applied to field.
* **GuiOptionEditorBlocked.kt** Line 38, 42: This declaration overrides a deprecated member but is not marked as deprecated itself.
* **GuiOptionEditorHidden.kt** Line 12, 16: This declaration overrides a deprecated member but is not marked as deprecated itself.
* **Graph.kt** Line 65: This declaration overrides a deprecated member but is not marked as deprecated itself.
* **BetterContainers.kt** Line 141: Condition is always 'true'.
* **ShiftClickEquipment.kt** Line 24: Condition is always 'false'.
* **FeatureToggleProcessor.kt** Line 61: This class is not recommended for use in Kotlin so use 'kotlin.Boolean' instead.
* **FeatureToggleProcessor.kt** Line 62: Unchecked cast of 'Any!' to 'Property'.
* **CopyItemCommand.kt** Line 33: Redundant call of conversion method.
* **SafeItemStack.kt** Line 52: 'field item: Holder?' is deprecated in Java.
* **GuiScreenUtils.kt** Line 29: Redundant call of conversion method.

Nealkitt:
> detekt:
> 
>     * **HotfApi.kt** Line 16, 17: This annotation is currently applied to the value parameter only but in the future it will also be applied to field.
> 
>     * **HotmApi.kt** Line 112, 113, 153: This annotation is currently applied to the value parameter only but in the future it will also be applied to field.
> 
>     * **GuiOptionEditorBlocked.kt** Line 38, 42: This declaration overrides a deprecated member but is not marked as deprecated itself.
> 
>     * **GuiOptionEditorHidden.kt** Line 12, 16: This declaration overrides a deprecated member but is not marked as deprecated itself.
> 
>     * **Graph.kt** Line 65: This declaration overrides a deprecated member but is not marked as deprecated itself.
> 
>     * **BetterContainers.kt** Line 141: Condition is always 'true'.
> 
>     * **ShiftClickEquipment.kt** Line 24: Condition is always 'false'.
> 
>     * **FeatureToggleProcessor.kt** Line 61: This class is not recommended for use in Kotlin so use 'kotlin.Boolean' instead.
> 
>     * **FeatureToggleProcessor.kt** Line 62: Unchecked cast of 'Any!' to 'Property'.
> 
>     * **CopyItemCommand.kt** Line 33: Redundant call of conversion method.
> 
>     * **SafeItemStack.kt** Line 52: 'field item: Holder?' is deprecated in Java.
> 
>     * **GuiScreenUtils.kt** Line 29: Redundant call of conversion method.

how is a pr that doesn't touch hotfapi or hotmapi responsible for this

hannibal002:
> how is a pr that doesn't touch hotfapi or hotmapi responsible for this

i just tested to run detekt on beta again, none of those errors show up there. so idk whats going on exactly. i just  know that detekt flags those errors in this pr, and not in other prs. so its the prs task to fix it.

Akinsoft:
I ran detekt earlier and pretty much only GuiPositionEditor showed up for this commit, will be pushing soon

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorShoppingList.kt
```diff
@@ -269,12 +270,17 @@ object GardenVisitorShoppingList {
         visitorMissingItemsWarnTime.removeIf { it.value.passedSince() > 10.minutes }
     }
 
-    @HandleEvent(GuiRenderEvent::class)
-    fun onGuiRender() {
+    @HandleEvent
+    fun onGuiRenderTop() {
```
hannibal002, line 274: whas was the SignEditScreen check removed? does gui on top renderer assume "ina sign" counts as "in an inventory"? if not, then the check is wrong. have you tested this with inside an sign?

src/main/java/at/hannibal2/skyhanni/config/core/config/gui/GuiPositionEditor.kt
```diff
@@ -278,3 +302,134 @@ class GuiPositionEditor(
...
+        return GuiScreenUtils.withScreenMetricsOverride(
+            oldScreen.width,
+            oldScreen.height,
+            oldScreenScaleFactor(oldScreen),
+            action,
+        )
+    }
+
+    fun <T> withRenderTransform(action: () -> T): T = withMetrics {
+        val oldScreen = oldScreen ?: return@withMetrics action()
```
hannibal002: the check if oldScreen is null to then run action already happens inside withMetrics, so the line here can never fail.

src/main/java/at/hannibal2/skyhanni/config/core/config/gui/GuiPositionEditor.kt
```diff
@@ -122,14 +138,11 @@ class GuiPositionEditor(
 
     private fun renderRectangles(): Int {
         var hoveredPos = -1
-        DrawContextUtils.pushPop {
-            width = getScaledWidth()
-            height = getScaledHeight()
+        var alreadyHadHover = false
```
hannibal002, line 141: why was DrawContextUtils.pushPop removed here? not saying this is definitely wrong, just curious on your reasoning. if there is none, i assume its save to keep it there, since it protects against small render issues breaking everything.

src/main/java/at/hannibal2/skyhanni/features/skillprogress/SkillProgress.kt
```diff
@@ -84,16 +87,26 @@ object SkillProgress {
         }
     }
 
-    @HandleEvent
-    fun onGuiRender() {
+    @HandleEvent(GuiRenderEvent.GuiOnTopRenderEvent::class)
+    fun onGuiRenderAllSkills() {
```
hannibal002: please rename the function to onGuiRenderTop, so it matches the primary function of GuiOnTopRenderEvent. also, then the explicit class parameter can go away (in the handle event annotation above)

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/visitor/GardenVisitorShoppingList.kt
Akinsoft: Tested it before with bz/supercraft signs, seemed fine on my end and I tested again now and it's all good. I can add the SignEditScreen check back if you'd rather keep it extra safe though.

src/main/java/at/hannibal2/skyhanni/config/core/config/gui/GuiPositionEditor.kt
Akinsoft: I’ll clean that up.

src/main/java/at/hannibal2/skyhanni/config/core/config/gui/GuiPositionEditor.kt
Akinsoft: Mostly because the new per-position transform already wraps the transformed renders with pushPopResult. That only applies to the chest GUI positions though, so I can add the outer pushPop back too if you prefer

up 2 u boss

src/main/java/at/hannibal2/skyhanni/features/skillprogress/SkillProgress.kt
Akinsoft: Gotcha

src/main/java/at/hannibal2/skyhanni/config/core/config/gui/GuiPositionEditor.kt
hannibal002: you have read the code more recently than me, so if you think its fine to leave it out, its fine by me.

Akinsoft: Sounds good

hannibal002 (approved):

PR #5852, opened 2026-06-08
Backend: Merge ChatComponent mixins

lunaynx:
## What
For some reason, we had three mixin files targeting `ChatComponent`. This PR merges them into a single file.

exclude_from_changelog


lunaynx:
This will be part of the 26.2 PR or a bigger cleanup PR, since it turns out it's not in fact required for my other linked PR.

PR #5843, opened 2026-06-06
Fix: Occasional chat edit/delete errors

lunaynx:
## What
Fixes the occasional "Error while editing/deleting message" errors, hopefully. I've been running this for a few days with a 16k chat history size. No errors, lagspikes, or memory leaks.

This is basically copying the exact same thing vanilla 26.1 does: storing a reference to the parent `GuiMessage` on each `GuiMessage.Line` instead of generating IDs and referencing them that way.

## Changelog Fixes
+ Fixed occasional error while editing/deleting a message. - Luna

PR #5854, opened 2026-06-08
Fix: Copy Chat sometimes copying the wrong message

lunaynx:
## Dependencies
- #5843

## What
Fixes various issues with Copy Chat sometimes copying the wrong message, in cases such as:
- multiple messages being received in the same tick
- chat being shifted by mods like Chat Patches
- mods like Chat Patches injecting messages into chat (persistent chat history)

## Changelog Fixes
+ Fixed Copy Chat sometimes copying the wrong message. - Luna


hannibal002 (changes requested):
detekt errors in the dependency pr, pls request a review when the dep is merged (until the workflows thingies are finally fully working)

hannibal002 (approved):

PR #5816, opened 2026-05-31
Improvement + Fix: Wormhole finder disappearing and option for Pathfinding

f4bii:
## What
Added an option to remove the path finding to wormholes and Improved the Wormhole locator to being less disappearing

## Changelog Improvements
+ Improved Wormhole finder to have an option to remove Pathfinding. - FabiHBBBT

## Changelog Fixes
+ Fixed Wormhole finder not disappearing as much. - FabiHBBBT

PR #5855, opened 2026-06-09
Improvement: Wormhole Finder v2

Akinsoft:
## What

Improves the Wormhole Finder so it tracks active wormholes more reliably.

It now uses the wormhole arrows together with the wormhole particles to show possible wormholes as `Wormhole?` and confirmed ones as `Wormhole!`. Confirmed wormholes stay locked even when their particles stop rendering from distance, and closing a wormhole removes the closest tracked one instead of wiping all active markers.

Also adds a Guide Mode setting:
- Off
- Direct
- Navigation

Line mode draws one direct line at a time, prioritizing confirmed wormholes first, then the closest possible one. Pretty much exactly like the Diana Burrows one works.

See it in action: [bazinga](https://youtu.be/wtPqmAQoBOM)

## Changelog Improvements
+ Improved Wormhole Finder. - akinsoft
    * Added confirmed/possible wormhole markers and guide modes.


f4bii:
#5855

f4bii:
took me way too long to see... i meant #5816

PR #5857, opened 2026-06-09
Fix: Pets menu detection

lunaynx:
## What
Fixes Pets menu not being detected after a recent Hypixel issue.

The search query seems to be no longer shown in the title, and the capturing groups `currentpage`/`maxpage` appear to never have been used at all, so I removed those.

## Changelog Fixes
+ Fixed Pets menu no longer being detected after a recent Hypixel update. - Luna


Akinsoft:
Noticed this while working on pet display v3 as well

hannibal002 (approved):

PR #5858, opened 2026-06-10
Fix: Wrong location for Mineshaft entrance waypoints

Piggered:
## What
Fixes the Glacite Mineshaft entrance/ladder waypoints being added to the wrong location. Currently, either `x: 0, y: -1, z: 0` or `x: -182, y: ?9, z: -192` is set as the entrance location >90% of the time.

This is done by listening for incoming player position packets until the client receives a ClientboundLevelChunkWithLightPacket after entering the Mineshaft and using the location fields of the latest one with absolute rotation values, treating it as the Mineshaft spawnpoint.

## Changelog Fixes
+ Fixed Glacite Mineshaft entrance/ladder waypoints being set to wrong location. - Piggered


Piggered:
> this hasn't happened to me, and such cases would simply result in entrance/ladder waypoints not being added at worst

Well it just happened now...

Piggered:
Instead of relying on a 1-second threshold to stop listening for player position packets, it now stops listening whenever the client receives a ClientboundLevelChunkWithLightPacket. From my admittedly limited testing, this is much more consistent, even with significant lag (this was tested on my terrible laptop which runs Minecraft at ~20 fps).

src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/MineshaftWaypoints.kt
```diff
@@ -28,32 +35,38 @@ object MineshaftWaypoints {
...
-                .add(y = -15)
-            waypoints.add(MineshaftWaypoint(waypointType = MineshaftWaypointType.LADDER, location = location))
+    @HandleEvent(onlyOnIsland = IslandType.MINESHAFT)
+    fun onPacketReceived(event: PacketReceivedEvent) {
+        if (isWorldLoaded) return
+
+        when (event.packet) {
+            is ClientboundLevelChunkWithLightPacket -> isWorldLoaded = true
+            is ClientboundPlayerPositionPacket -> {
+                if (!event.packet.relatives.isEmpty()) return
```
hannibal002: use isNotEmpty()

hannibal002 (changes requested):
also please confirm that `IslandChangeEvent(MINESHAFT)` gets fired before `ClientboundPlayerPositionPacket` gets fired/that you have tested the fixxx

src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/MineshaftWaypoints.kt
```diff
@@ -28,32 +35,38 @@ object MineshaftWaypoints {
...
+    private var isWorldLoaded = false
 
     @HandleEvent
     fun onWorldChange() {
         waypoints.clear()
+        isWorldLoaded = false
     }
 
     @HandleEvent
     fun onIslandChange(event: IslandChangeEvent) {
```
lunaynx: While you're at it, might as well update this to use the non-deprecated `IslandJoinEvent`.

(I think deprecations aren't currently showing up in IntelliJ because of an oopsie made by @Rainymei, I'll make a PR to revert that in a bit.)

Piggered:
> also please confirm that `IslandChangeEvent(MINESHAFT)` gets fired before `ClientboundPlayerPositionPacket` gets fired/that you have tested the fixxx

Yeah I can confirm it does. I have done close to triple digits of Glacite Mineshafts with this PR on, both on my terrible school laptop at ~20 FPS and my gaming desktop. :)

Now I have to verify it works the same with `IslandJoinEvent` instead, but I doubt that'd break things.

hannibal002 (approved):

PR #5859, opened 2026-06-10
Fix: Lava Replacement chunk refresh on wrong thread

lunaynx:
## What
Fixes Lava Replacement running chunk refresh on the network thread instead of the render thread when switching islands, causing an error. Also migrates it from `IslandChangeEvent` to the non-deprecated `IslandJoinEvent`.

<details>
<summary>Images</summary>

</details>

## Changelog Fixes
+ Fixed error with Lava Replacement when switching lobbies. - Luna


lunaynx:
Rolled this into #5878.

PR #5869, opened 2026-06-12
Backend: Add More Bridgiar Argument Types

AverageUser125:
## What
There was no built-in way for command suggestion for player names, and uuids. so I added them.
I think the API is good enough.
Example usage:
```kotlin
 literal("add") {
                argCallback(
                    "username",
                    PlayerArgumentType.player(),
                    PlayerSuggestions.build {
                        include(PlayerCategory.ISLAND_PLAYERS, PlayerCategory.SELF)
                        filterNot { it in contributorNames }
                    }
                ) { player ->
                    val gameProfile = player.gameProfile
                    addTestContributor(gameProfile.id, player.gameProfile.name)
                }
```
## Changelog Fixes
+ Fixed putting quotes as arguments in skyhanni command chopping off. - AverageUser125

+ Add Component, UUID, and Player Bridgiar Argument types. - AverageUser125
+ Added the option for skyhanni commands to suggest player names. - AverageUser125

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/PlayerSuggestions.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/PlayerSuggestions.kt
@@ -1,5 +1,6 @@
 package at.hannibal2.skyhanni.config.commands.brigadier
 
+import at.hannibal2.skyhanni.config.commands.brigadier.BrigadierUtils.dynamicSuggestionProvider
 import at.hannibal2.skyhanni.config.commands.brigadier.BrigadierUtils.toSuggestionProvider
 import at.hannibal2.skyhanni.features.commands.tabcomplete.PlayerCategory
 import com.mojang.brigadier.suggestion.SuggestionProvider
@@ -39,12 +40,12 @@ class PlayerSuggestions private constructor(
         }
     }
 
-    private val cached: List<String> by lazy {
+    private fun getPlayers(): List<String>{
         val excludedNames = exclude
             .flatMap(PlayerCategory::usernames)
             .toSet()
 
-        include
+        return include
             .flatMap(PlayerCategory::usernames)
             .filterNot(excludedNames::contains)
             .filter(filter)
@@ -52,7 +53,7 @@ class PlayerSuggestions private constructor(
     }
 
     fun toBrigadier(): SuggestionProvider<FabricClientCommandSource> {
-        return cached.toSuggestionProvider()
+        return dynamicSuggestionProvider { getPlayers() }
     }
 
     companion object {
--- a/src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/PlayerSuggestions.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/PlayerSuggestions.kt
@@ -28,11 +28,13 @@ class PlayerSuggestions private constructor(
         }
 
         fun exclude(vararg categories: PlayerCategory) = apply {
-            val excluded = categories.asSequence()
+            val excludedSeq = categories
+                .asSequence()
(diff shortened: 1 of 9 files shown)


lunaynx:
[`GameProfileArgument`](https://mcsrc.dev/1/26.1.2/net/minecraft/commands/arguments/GameProfileArgument) should be able to do this?

AverageUser125:
> [`GameProfileArgument`](https://mcsrc.dev/1/26.1.2/net/minecraft/commands/arguments/GameProfileArgument) should be able to do this?

Didn't know that existed, but the point was more to also suggest party members and carry customers and also do the include and exclude logic.
I could also add guild I suppose.

This PR is honestly not super usefull but just a nice to have.

lunaynx:
Yeah, it's not super important, but you could probably look into extending that to support whatever you need somehow. If it's too complicated, don't bother.

AverageUser125:
> Yeah, it's not super important, but you could probably look into extending that to support whatever you need somehow. If it's too complicated, don't bother.

on the contrary, it's quite simple.
the API usage is like this:
```
argCallback(
    "username",
    PlayerArgumentType.player(
    include = setOf(PlayerSource.WORLD, PlayerSource.SELF), 
    exclude = setOf(PlayerSource.GUILD)
) { player ->
    val gameProfile = player.gameProfile
    addTestContributor(gameProfile.id, player.gameProfile.name)
}
```
It can also be suggestions on a regular string argument.

lunaynx:
I meant having it extend `GameProfileArgument`. But like I said, it's not too important if it doesn't seem easy.

AverageUser125:
> I meant having it extend `GameProfileArgument`. But like I said, it's not too important if it doesn't seem easy.

I could, but `GameProfileArgument` also supports writing `@a` and doing stuff like
 `@a[name="hannibal02",limit=2, sort=nearest]`.
While I just wanted it to always return a single player and not a collection.

lunaynx:
Ah, fair.

src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/PlayerSuggestions.kt
```diff
@@ -0,0 +1,94 @@
...
+        override fun usernames(): Sequence<String> =
+            GuildApi.getAllMembers().asSequence()
+    },
+
+    CARRY_CUSTOMER {
+        override fun usernames(): Sequence<String> =
+            CarryTracker.customers.asSequence().map { it.name }
+    };
+
+    abstract fun usernames(): Sequence<String>
```
ItsEmpa: this would probably look better as a kotlin property getter instead of just overriding the function for each one, and its generally how things like this are done in skyhanni

so something like this
```kt
enum class PlayerSource(private val usernamesGetter: () -> Sequence<String>) {
   WORLD({ EntityUtils.getPlayerEntities().asSequence().map { it.gameProfile.name } }),
   SELF({ sequenceOf(PlayerUtils.getName()) }),
   PARTY({ PartyApi.partyMembers.asSequence() }),
   GUILD({ GuildApi.getAllMembers().asSequence() }),
   CARRY_COSTUMER({ CarryTracker.customers.asSequence().map { it.name } }),
   ;

   fun usernames(): Sequence<String> = usernamesGetter()
}
```

ItsEmpa (changes requested):

src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/PlayerSuggestions.kt
AverageUser125: didn't know that existed even, so sure.

AverageUser125:
imma just close this, as this is gonna be merged at once with ContributorManager v2.

PR #5872, opened 2026-06-13
Fix + Improvement: Properly update toolkit crop icon cache and color enchants

GrowlingGrizzly:
## What
Fixes issues with the previous cache not updating when adding a tool back/replacing with a different tool and changing profiles. Also copies over custom data to allow proper coloring of enchantments

<details>
<summary>Images</summary>

image

</details>

## Changelog Improvements
+ Added enchantment color support to Toolkit Crop Icons. - Growling_Grizzly

## Changelog Fixes
+ Fixed Toolkit Crop Icons cache not updating. - Growling_Grizzly


src/main/java/at/hannibal2/skyhanni/features/garden/farming/ToolkitCropReplacer.kt
```diff
@@ -25,16 +43,40 @@ object ToolkitCropReplacer {
         if (event.slot !in 10..16 && event.slot !in 20..24) return
```
hannibal002: make event.slot into a local variable. 9 usages!

src/main/java/at/hannibal2/skyhanni/features/garden/farming/ToolkitCropReplacer.kt
```diff
@@ -25,16 +43,40 @@ object ToolkitCropReplacer {
...
 
         val item = event.originalItem
-        val cropType = item.getCropType() ?: return
+        if (item.isEmpty) {
+            if (!fullyLoaded) return
+            iconCache.remove(event.slot)
```
hannibal002: whats the plan here? why do we remove the empty item? is a return missing here?

src/main/java/at/hannibal2/skyhanni/features/garden/farming/ToolkitCropReplacer.kt
```diff
@@ -25,16 +43,40 @@ object ToolkitCropReplacer {
...
         val iconId = "toolkit_crop_replacer:${cropType.name}"
 
         val replacementStack = iconCache.getOrPut(event.slot) {
-            cropType.getItemStackCopy(iconId).apply {
+            val stack = cropType.getItemStackCopy(iconId).apply {
                 setLore(item.getLoreComponent())
                 setCustomItemName(item.hoverName)
             }
+            stack.set(DataComponents.CUSTOM_DATA, item.getExtraAttributes()?.let { CustomData.of(it) })
+            return@getOrPut stack
```
hannibal002: can omit `return@getOrPut`

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/farming/ToolkitCropReplacer.kt
GrowlingGrizzly: without it i get `Return type mismatch: expected 'ItemStack', actual 'CustomData?'.` and says casting will always fail

lunaynx: Yeah no, it is needed.

hannibal002: > without it i get Return type mismatch: expected 'ItemStack', actual 'CustomData?'. and says casting will always fail

You get this type mismatch because removing the return makes the `set` function the last statement. The lambda then tries to return `CustomData?` instead of the `ItemStack`. We can still omit the explicit return. We just need to place `stack` alone on the last line of the lambda. Alternatively we can move the `set` call directly into the `apply` block.

> Yeah no, it is needed.

no it is not. we cant remove the whole line, this is correct. but we surely can remove the `return@getOrPut` part, leaving `stack` in the last line in the getOrPut block.

image

lunaynx: You're right, we were both assuming you meant "just remove literally only the `@getOrPut` part", but that makes sense.

lunaynx: Actually looking back now I don't know why I went with that interpretation too when you *did* say "remove `return@getOrPut`" and not "remove `@getOrPut`". 🤷‍♀️

src/main/java/at/hannibal2/skyhanni/features/garden/farming/ToolkitCropReplacer.kt
```diff
@@ -2,39 +2,80 @@ package at.hannibal2.skyhanni.features.garden.farming
...
 
 @SkyHanniModule
 object ToolkitCropReplacer {
 
     private val config get() = GardenApi.config.farmingToolkit
 
     private val iconCache: MutableMap<Int, SafeItemStack> = mutableMapOf()
+    private val storedTools: MutableMap<Int, CropType> = mutableMapOf()
+
+    private var fullyLoaded = false
```
hannibal002: unused

src/main/java/at/hannibal2/skyhanni/features/garden/farming/ToolkitCropReplacer.kt
```diff
@@ -2,39 +2,80 @@ package at.hannibal2.skyhanni.features.garden.farming
...
+        storedTools[slot] = cropType
+
         val iconId = "toolkit_crop_replacer:${cropType.name}"
 
-        val replacementStack = iconCache.getOrPut(event.slot) {
+        val replacementStack = iconCache.getOrPut(slot) {
             cropType.getItemStackCopy(iconId).apply {
                 setLore(item.getLoreComponent())
                 setCustomItemName(item.hoverName)
+                    .set(DataComponents.CUSTOM_DATA, item.getExtraAttributes()?.let { CustomData.of(it) })
```
hannibal002: this line and chain now looks confusing. we have getOrPut, we have apply, and we have let. can we please make this a bit less complicated to read thanks

src/main/java/at/hannibal2/skyhanni/features/garden/farming/ToolkitCropReplacer.kt
```diff
@@ -2,39 +2,80 @@ package at.hannibal2.skyhanni.features.garden.farming
...
         if (!config.replaceMenuIcons) return
         if (!GardenApi.toolkitInventory.isInside()) return
-        if (event.slot !in 10..16 && event.slot !in 20..24) return
+
+        val slot = event.slot
+        if (slot !in 10..16 && slot !in 20..24) return
 
         val item = event.originalItem
-        val cropType = item.getCropType() ?: return
+        if (item.isEmpty) return
```
hannibal002: can use event.hasItem instead

hannibal002 (changes requested):

hannibal002 (approved):

PR #5874, opened 2026-06-13
Feature: Move `Insert Item Into Sack` button back to original place

AverageUser125:
## What
Move `Insert Item Into Sack` button back to original place.

<details>
<summary>Images</summary>
image
</details>

Not added as of PR making: the config

## Changelog New Features
+ Added Option To Move `Insert Item Into Sack` button back to original placement. - AverageUser125


lunaynx:
I seem to remember the position of the button isn't always consistent with every sack GUI, check stuff like Gemstone Sack and Rune Sack to be sure.

AverageUser125:
> I seem to remember the position of the button isn't always consistent with every sack GUI, check stuff like Gemstone Sack and Rune Sack to be sure.

Since the width is always 9 slots, I just made it compute it using the size.
Should probably work with any inventory size

NopoTheGamer:
i dont like the idea

lunaynx:
Why not? This isn't really much different from slot keybinds, and it's an optional feature.

lunaynx:
Honestly, I would keep this default disabled and probably even remove the FeatureToggle so it doesn't get enabled without an explicit action. As much as it is probably muscle memory for 95% of existing players, the general principle of moving inventory slots around is a bit invasive if not explicitly requested by the user.

AverageUser125:
> Honestly, I would keep this default disabled and probably even remove the FeatureToggle so it doesn't get enabled without an explicit action. As much as it is probably muscle memory for 95% of existing players, the general principle of moving inventory slots around is a bit invasive if not explicitly requested by the user.

I personally agree, but a couple people in discord asked for it to be default enabled.

lunaynx:
SkyHanni devs or random people?

AverageUser125:
> SkyHanni devs or random people?

Uh, I think Shana counts.

lunaynx:
I more meant some higher ups, but I doubt they would suggest that. But yeah, IMO it's best to keep it default disabled and just go around advertising the feature once it's in a beta.

AverageUser125:
Alright. Will do.

src/main/java/at/hannibal2/skyhanni/features/misc/MoveInsertIntoSackButton.kt
```diff
@@ -0,0 +1,80 @@
...
+        if (item == null || item.getItem() != Items.CHEST) return
+
+        inSackMenu = true
+        chestSlot = computedChestSlot
+        swapTargetSlot = computedTargetSlot
+        insertIntoSackItem = item
+        originalSlotItem = items[swapTargetSlot]
+    }
+
+    @HandleEvent
```
CalMWolfs: should have skyblock check

src/main/java/at/hannibal2/skyhanni/features/misc/MoveInsertIntoSackButton.kt
```diff
@@ -0,0 +1,80 @@
...
+        val newSlotItem = when (event.slot) {
+            swapTargetSlot -> insertIntoSackItem
+            chestSlot -> originalSlotItem
+            else -> return
+        } ?: return
+
+        event.replace(newSlotItem)
+    }
+
+    @HandleEvent
```
CalMWolfs: should have skyblock check

src/main/java/at/hannibal2/skyhanni/features/misc/MoveInsertIntoSackButton.kt
```diff
@@ -0,0 +1,80 @@
...
+        val items = openEvent.inventoryItems
+        val lastSlotIndex = openEvent.inventorySize - 1
+
+        val computedChestSlot = lastSlotIndex - 2
+        val computedTargetSlot = lastSlotIndex - 5
+
+        if (computedTargetSlot < 0) return
+
+        val item = items[computedChestSlot]
+        if (item == null || item.getItem() != Items.CHEST) return
```
CalMWolfs:
```suggestion
        if (item?.getItem() != Items.CHEST) return
```
can be simplified

CalMWolfs (changes requested):
just reviewing code as the feature doesnt work for me as i dont have the button without the feature so the feature cannot move it

CalMWolfs (approved):
turns out i was in the wrong sack 🤦 feature works

PR #5877, opened 2026-06-13
Backend: Use build matrix for workflow

lunaynx:
## What
Updates the build workflow to use a build matrix to compile SkyHanni for all supported Minecraft versions in parallel, and upload the output JAR without zipping. Also removes `MultiVersionStage` as there is no reason to disable it anymore – instead, all relevant tasks have been wired to run on the primary target (currently 26.1) only.

+ Improved GitHub Actions build process. - Luna
    * All versions are now built in parallel and the uploaded JARs are no longer zipped.


AverageUser125:
Maybe the jars should be in a `tar.gz`so that they are slightly compressed?
I get why not zip, since a jar is already just a zip in disguise.

lunaynx:
No point, the whole idea is to avoid an unnecessary extraction step for users. Also, compressing a compressed archive is generally not going to achieve much.

#5633 exists which will reduce JAR sizes by ~10 MB.

hannibal002 (approved):

Issue #209, opened 2023-06-04
Add Tab complete to /shtrackcollection, /shcroptime and /sh <categories>


hannibal002:

A-N-07:
Hi! I'm looking to make my first contribution to SkyHanni and this issue caught my eye. Is this still relevant and open to work on? If so, I'd love to take a shot at it!

Rainymei:
sure, shtrackcollection got removed so don't bother doing that one

A-N-07:
Noted. I will look for another issue. Suggestions are welcome.

Rainymei:
i meant shtrackcollection, the other 2 are stgill here

A-N-07:
I seem to have read it wrong. My apologies. I will get on it.

A-N-07:
Ready for review

https://github.com/hannibal002/SkyHanni/pull/5883

Zelmari:
#6332 Ready. Changed to conform to what a contributor asked of it.

PR #5883, opened 2026-06-17
Improvement: Tab completion for /shcroptime

A-N-07:
## What
Reworked `/shcroptime` to use proper Brigadier argument parsing instead of `legacyCallbackArgs`.

- Split the command into two typed arguments (`amount`, `item`) instead of parsing a raw `args` array.
- Added a live `SuggestionProvider` on the `item` argument that tab-completes crop names from `CropMoneyDisplay.multipliers`, cleaned via `repoItemName.removeColor().lowercase()`.
- Dropped the earlier `/sh <category>` tab-complete from this PR — as pointed out in review, command categories aren't a meaningful thing to suggest here, so that part was removed entirely rather than reworked.

> [!NOTE]
> I have not implemented /sh to this PR. Reasoning in `Why /sh tab completion isn't implemented in this PR`

## Why
The previous `legacyCallbackArgs` implementation had no tab-completion at all for `/shcroptime`, forcing users to type full crop names manually. This PR adds tab-completion specifically for that command, scoped down per review feedback.

### Result Command with no arguments
shcroptime 1

### Result Command with the suggestions
Screenshot 2026-06-20 175219

### Result Command with one letter filled in
image

### Result Command
Screenshot 2026-06-20 175659

> [!NOTE]
> It says "No speed data!" because I don't have the wheat farming tool which is required to actually calculate the crop time. But regardless it gives out the right messages after using the tab to auto complete and then the press on ENTER.

## Why `/sh` tab completion isn't implemented in this PR

`/sh <query>` works as a search function, not a navigation command. It filters config entries by matching `<query>` against text inside SkyHanni's sub-menus. That makes command-category suggestions a poor fit for tab completion.

If `/sh garden` tab-completed "garden" as a suggested category, a user would reasonably expect it to open the Garden menu directly. In reality, it still runs a text search for "garden" across all sub-menus, which isn't the same thing and would likely confuse more than help.

There's also no clean source for what should be suggested. Unlike `/shcroptime`, where the valid arguments are a finite, well-defined list (`CropType` entries), `/sh` has no equivalent list of valid searches since it searches arbitrary config text. I don't have enough experience with the mod's full config tree to confidently hand-pick a useful 5-10 keyword shortlist that represents real usage patterns.

**Two ways forward, as I see it:**

1. **Curated keyword list.** If a maintainer can provide a short list of the most commonly searched terms, such as menu or category names users search for most, I can wire those up as static suggestions. This is straightforward to add.
2. **Live prefix search.** A more sophisticated version that dynamically suggests matching words from inside the sub-menus as the user types. This is a meaningfully bigger feature, since it essentially means building a searchable index of config text, and feels out of scope for issue #209.

**My recommendation:** either provide a keyword list for option 1, or scope `/sh` out of this issue entirely and track it separately if a maintainer wants the dynamic version.

## Changelog Improvements
+ Added tab completion for the `/shcroptime` command, suggesting crop names as you type. - A-N-07


lunaynx:
I'm confused what the point of tab completing command categories is here. What are you gonna do with `/sh users_active`?

A-N-07:
I tried to handle this issue

https://github.com/hannibal002/SkyHanni/issues/209

There was limited information on the implementation of the auto-complete for /sh.

From what I interpreted it should work like a search filter. to quick access the menu.

I was looking for the enum class that has the correct autocomplete options for the /sh command and found CommandCategory.kt and thought that was the correct file with all the auto-complete options. 

But I might be wrong as I see from you reaction.

Rainymei:
commandcategory is a category enum, not commands. did you even test if /sh developer_test etc worked

A-N-07:
I am new to this mod and new to contributing to open source so please bear with me.

I made a mistake by suggesting the command categories. 

I thought I tested them correctly. But after retesting I see that I was wrong and jumped the gun on making a pull request.

I think I understand now which commands I need to suggest, which are just the player commands.

I converted this pull request to a draft and I am working on suggesting the right commands.

My apologies.

src/main/java/at/hannibal2/skyhanni/config/features/commands/TabCompleteConfig.kt
```diff
@@ -76,4 +76,16 @@ class TabCompleteConfig {
...
+    @ConfigOption(name = "Crop Time", desc = "Tab-complete crops in the SkyHanni command §e/shcroptime§7.",)
+    @ConfigEditorBoolean
+    @FeatureToggle
+    var shCropTime: Boolean = true
+
+    @Expose
+    @ConfigOption(name = "SkyHanni Categories", desc = "Tab-complete command categories in the main SkyHanni command §e/sh§7.",)
+    @ConfigEditorBoolean
+    @FeatureToggle
+    var shCategories: Boolean = true
```
NopoTheGamer: these shouldnt be options

src/main/java/at/hannibal2/skyhanni/features/commands/tabcomplete/TabComplete.kt
```diff
@@ -27,6 +27,8 @@ object TabComplete {
         PlayerTabComplete.handleTabComplete(fullCommand)?.let { return it }
         PartyCommands.customTabComplete(command)?.let { return it }
         ViewRecipeCommand.customTabComplete(command)?.let { return it }
+        CropTimeTabComplete.handleTabComplete(command)?.let { return it }
+        ShTabComplete.handleTabComplete(command)?.let { return it }
```
NopoTheGamer: this shouldnt be done like this, you should modify the brigadier command registration for these commands

src/main/java/at/hannibal2/skyhanni/features/commands/tabcomplete/ShTabComplete.kt
```diff
@@ -0,0 +1,19 @@
...
+
+object ShTabComplete {
+
+    private val config get() = SkyHanniMod.feature.misc.commands.tabComplete
+
+    fun handleTabComplete(command: String): List<String>? {
+        if (!isEnabled()) return null
+        if (command != "sh" && command != "skyhanni") return null
+
+        return CommandCategory.entries.map { it.name.lowercase() }
```
NopoTheGamer: this also just isnt close to correct as stated before

NopoTheGamer (changes requested):

AverageUser125:
I think for `shcroptime` you should just use `BridgiarUtils.integer()` & `EnumArgumentType<CropType>`.
so bassically something like:
```kotlin
        event.registerBrigadier("shcroptime") {
            description =
                "Calculates with your current crop per second speed how long you need to farm a crop to collect this amount of items"
            category = CommandCategory.USERS_ACTIVE
            arg("amount", BrigadierArguments.integer()) { amountRef ->
                argCallback(
                    "cropType",
                    EnumArgumentType.create(
                        CropType::class.java,
                        isGreedy = true,
                        CropType::cropName
                    )
                ) { cropType ->
                    onCommand(getArg(amountRef), cropType)
                }
            }
        }
```

About the `/sh` command, I don't really know. I think remove it since it's for searching option in config. but idk.
Listen to nopo, they are very smart.

AverageUser125:
1. too many redundant comments
2. do not use `com.mojang.brigadier.suggestion.SuggestionProvider` use `dynamicSuggestions` instead and just provide the supplier
3. `internalName.repoItem` is already without color I believe.
4. use `argCallback` in the last one argument.
5. `onCommand(arrayOf(amount, item))` very lazy, make the onCommand actually take them as parameters.
6. why is `amount` a `string`? use `BridigiarArguments.integer(1)`
7. doing `name = "item"` is probably not needed, just write `"item"`

AverageUser125:
1. my mistake for saying `BrigadierArguments.integer`, 
just use `BrigadierArguments.long` to let the user type very big numbers.
And also so you don't need to do the cast to long.

2. Maybe I am missing something, but why change the message when the map is empty?

AverageUser125:
Btw, you need to use the [pull request template](https://github.com/hannibal002/SkyHanni/blob/beta/pull_request_template.md).
Which includes changing the title to start with `Improvement: `
And also merge the current `beta` branch into your branch since workflows have been re-enabled.

A-N-07:
@AverageUser125 Thank you for your patience and providing the needed feedback. I have implemented all your suggestions. I hope the code is now worthy for the codebase. 🥲

AverageUser125:
> @AverageUser125 Thank you for your patience and providing the needed feedback. I have implemented all your suggestions. I hope the code is now worthy for the codebase. 🥲

It's a pleasure.
And using Brigadier is a pain, so you doing alright!

AverageUser125:
@A-N-07 
Though, just read the messages I sent before you sent the thank you.
1. merging the `beta` branch into yours.
2. abiding to the pull request template.
3. using `BridgiarArguments.long` instead of `integer` cause I oopsie.

A-N-07:
> @A-N-07 Though, just read the messages I sent before you sent the thank you.
> 
> 1. merging the `beta` branch into yours.
> 2. abiding to the pull request template.
> 3. using `BridgiarArguments.long` instead of `integer` cause I oopsie.

Done with all three:
1. Merged `beta` into my branch.
2. Updated the description to follow the PR template (added the Changelog section, fixed the title prefix).
3. Switched `amount` to `BrigadierArguments.long(1)` instead of `integer`.

Also re: your earlier comment on `repoItemName` already being color-free. I tested this more carefully and it turns out `repoItemName` does still carry color codes for some items (e.g. Pumpkin showed a stray `§f`). I kept `.removeColor()` on all three usages since I can't predict in advance which crops will have color codes attached.

AverageUser125:
@A-N-07 
```
/home/runner/work/SkyHanni/SkyHanni/src/main/java/at/hannibal2/skyhanni/features/garden/
GardenCropTimeCommand.kt:15:1: Unused import [NoUnusedImports]
```

Well, it seems you got one final fix left.
Basically just remove the unused imports.

hannibal002 (changes requested):
detekt:

GardenCropTimeCommand.kt:15:1: Unused import [NoUnusedImports]

AverageUser125:
> detekt:
> 
> GardenCropTimeCommand.kt:15:1: Unused import [NoUnusedImports]

touche

A-N-07:
I removed the unused import and pushed it.

hannibal002:
good! now please rerequest a review from me

image

src/main/java/at/hannibal2/skyhanni/features/garden/GardenCropTimeCommand.kt
```diff
@@ -3,14 +3,15 @@ package at.hannibal2.skyhanni.features.garden
 import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.config.commands.CommandCategory
 import at.hannibal2.skyhanni.config.commands.CommandRegistrationEvent
+import at.hannibal2.skyhanni.config.commands.brigadier.BrigadierArguments
+import at.hannibal2.skyhanni.config.commands.brigadier.BrigadierUtils.dynamicSuggestionProvider
```
hannibal002: do not use function imports (except extension functions)

src/main/java/at/hannibal2/skyhanni/features/garden/GardenCropTimeCommand.kt
```diff
@@ -83,7 +77,18 @@ object GardenCropTimeCommand {
...
                 "Calculates with your current crop per second speed how long you need to farm a crop to collect this amount of items"
             category = CommandCategory.USERS_ACTIVE
-            legacyCallbackArgs { onCommand(it) }
+
+            arg("amount", BrigadierArguments.long(1)) { amountArg ->
+                argCallback(
+                    "item",
+                    BrigadierArguments.greedyString(),
+                    suggestions = dynamicSuggestionProvider {
+                        CropMoneyDisplay.multipliers.keys.map { it.repoItemName.removeColor().lowercase() }
```
hannibal002: any downside in making this not lowercase?

src/main/java/at/hannibal2/skyhanni/features/garden/GardenCropTimeCommand.kt
```diff
@@ -83,7 +77,18 @@ object GardenCropTimeCommand {
...
+                    }
```
hannibal002: trailing comma missing

src/main/java/at/hannibal2/skyhanni/features/garden/GardenCropTimeCommand.kt
```diff
@@ -21,35 +22,28 @@ object GardenCropTimeCommand {
...
-        val searchName = rawSearchName.lowercase()
+        val searchName = item.lowercase()
 
         val map = mutableMapOf<String, Long>()
         for (entry in multipliers) {
             val internalName = entry.key
-            val itemName = internalName.repoItemName
-            if (itemName.removeColor().lowercase().contains(searchName)) {
+            val itemName = internalName.repoItemName.removeColor()
+            if (itemName.lowercase().contains(searchName)) {
```
hannibal002: instead of lowercase both, use contains with ignore case arg

hannibal002: the whole body of this if block can get extracted in a new function. to avoid too long functions.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/GardenCropTimeCommand.kt
NopoTheGamer: i dont believe there is any reason to make it lower case

src/main/java/at/hannibal2/skyhanni/features/garden/GardenCropTimeCommand.kt
```diff
@@ -21,56 +23,50 @@ object GardenCropTimeCommand {
 
     private val config get() = GardenApi.config.moneyPerHours
 
-    private fun onCommand(args: Array<String>) {
-        if (!config.display) {
-            ChatUtils.userError("Command /shcroptime requires 'Show money per Hour' feature to be enabled to work!")
-            return
+    private fun processCropEntry(internalName: NeuInternalName, amount: Long, searchName: String): Pair<String, Long>? {
+        val itemName = internalName.repoItemName.removeColor()
```
hannibal002: do not remove the color code from the item name here, just remove it for the contains check. otherwise, the final text shown to the user has no item color anymore.

src/main/java/at/hannibal2/skyhanni/features/garden/GardenCropTimeCommand.kt
```diff
@@ -21,56 +23,50 @@ object GardenCropTimeCommand {
...
+        if (!itemName.contains(searchName, ignoreCase = true)) return null
```
hannibal002: why do we use contains and not starts with here?

src/main/java/at/hannibal2/skyhanni/features/garden/GardenCropTimeCommand.kt
```diff
@@ -83,7 +79,18 @@ object GardenCropTimeCommand {
             description =
                 "Calculates with your current crop per second speed how long you need to farm a crop to collect this amount of items"
             category = CommandCategory.USERS_ACTIVE
-            legacyCallbackArgs { onCommand(it) }
+
+            arg("amount", BrigadierArguments.long(1)) { amountArg ->
```
hannibal002, line 87: what happens if i type in fewer than 2 args? in the old code, we would show /shcroptime <amount> <item>, we dont anymore, right? this is a downgrade

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/GardenCropTimeCommand.kt
A-N-07: Hadn't considered it. But 'startswith' is better. So Implemented it.

src/main/java/at/hannibal2/skyhanni/features/garden/GardenCropTimeCommand.kt
A-N-07: removed it.

src/main/java/at/hannibal2/skyhanni/features/garden/GardenCropTimeCommand.kt
A-N-07: I now made it in a way where it gives the usage message if the command has no or only one argument in the command.

src/main/java/at/hannibal2/skyhanni/features/garden/GardenCropTimeCommand.kt
```diff
@@ -21,56 +23,50 @@ object GardenCropTimeCommand {
...
-    private fun onCommand(args: Array<String>) {
-        if (!config.display) {
-            ChatUtils.userError("Command /shcroptime requires 'Show money per Hour' feature to be enabled to work!")
-            return
+    private fun processCropEntry(internalName: NeuInternalName, amount: Long, searchName: String): Pair<String, Long>? {
+        val itemName = internalName.repoItemName
+        if (!itemName.removeColor().startsWith(searchName, ignoreCase = true)) return null
+
+        val (baseId, baseAmount) = NeuItems.getPrimitiveMultiplier(internalName)
+        val baseName = baseId.repoItemName.removeColor()
```
hannibal002: baseName also still loses its color when shown to the user

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/GardenCropTimeCommand.kt
A-N-07: My apologies I didn't understand what you meant. 

So the list should be colorless like this:
image


And the message like this should then with color like this?
image

hannibal002: Sorry for not being more clear above.

Remove color codes from baseName only inline for CropType.getByName, not for the display text.

hannibal002 (approved):

A-N-07:
LETSGOOOOOOO

PR #5885, opened 2026-06-17
Fix: Experiment Addons Helper ConcurrentModificationException 

AverageUser125:
## What
a `ConcurrentModificationException` was possible while iterating them.
this only happened in chronomatron for me, but I did ultrasequencer too.

## Changelog Fixes
+ Fixed Rare Experiment Addons Helper Error in Chromotron. - AverageUser125


AverageUser125:
Doesn't fix the root problem.

PR #5890, opened 2026-06-18
Fix: Trapper Cooldown ready too early

lunaynx:
## What
Fixes Trapper Cooldown sometimes being shown as ready up to a second too early, because of reliance on `SecondPassedEvent` to decrement the timer which may not perfectly sync up with the quest start time.

I also made it updated on `SkyHanniTickEvent` instead because a second is too long and would often end up triggering too late instead with the new logic.

## Changelog Fixes
+ Fixed Trapper Cooldown often showing as ready up to a second too early. - Luna


src/main/java/at/hannibal2/skyhanni/features/misc/trevor/TrevorFeatures.kt
```diff
@@ -248,14 +254,14 @@ object TrevorFeatures {
     }
 
     private fun updateTrapper() {
```
hannibal002, line 255: remove the call in the on chat event. (no longer needed, dead code)

hannibal002 (changes requested):

hannibal002 (approved):

PR #5892, opened 2026-06-18
Fix: fix enchant parsing for fatal tempo and stacked enchants

zumbiepig:
## What
fixed enchant parsing for fatal tempo
i changed the regex to include a lookbehind for `, ` or `^` (start of string)

and also fixed stacked enchants
i made §r optional before the number since hypixel doesnt include it anymore

## Changelog Fixes
+ Fixed enchant parsing for Fatal Tempo and stacked enchants. - zumbiepig


lunaynx:
I'm pretty sure `§7` is used when you don't have the required Enchanting level for the enchantment.

Alex33856:
> I'm pretty sure `§7` is used when you don't have the required Enchanting level for the enchantment.

image

zumbiepig:
> I'm pretty sure `§7` is used when you don't have the required Enchanting level for the enchantment.

Okay, I added it back. Good thing I did the lookbehind then. That was intended as a backup but it will be the main method to check now.

PR #5785, opened 2026-05-22
Backend: Detekt Warnings

AverageUser125:
## What
Fix those darn Deket warnings.

## Changelog Technical Details
+ Fixed all the Detekt warnings. - AverageUser125

PR #5898, opened 2026-06-20
Backend: Safe Map.filterNotNull* methods

lunaynx:
## What
Makes `Map.filterNotNull{Keys,Values}` actually type-safe, and adds `Map.filterNotNull` to `CollectionUtils`, as well as fixing existing unsafe usages + removing one redundant `takeIf`.

(`X & Any` is essentially Kotlin syntax for "`X` but not null", because the nullable any type is `Any?`.)

+ Added safe Map.filterNotNull, Map.filterNotNullKeys, and Map.filterNotNullKeys extension methods to CollectionUtils. - Luna


AverageUser125:
The change in `HarvestFeastManager` will actually conflict with my Detekt resolve PR (#5785).
Since I also changed it there.
Just thought to mention it, even though your solution is more elegant.

lunaynx:
Yeah, it's fine, will deal with it if mine is merged later.

CalMWolfs (approved):

PR #5900, opened 2026-06-21
Fix: Make Sea Creature Tracker migration atomic + repo based

lunaynx:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/753

## What
Refactored Sea Creature Tracker migration logic:
- Moved migration logic out of `drawDisplay` (makes no sense to be there)
- Now uses an `AtomicBoolean` to ensure we don't try to display the tracker while it's being migrated
- Old names are now read from the repo instead of being hardcoded in the code

I did some small refactors on nearby code as well, including removing `SkyHanniTracker.modifyEachMode` because it's literally duplicating the same thing the single-argument `SkyHanniTracker.modify` does in a different way for seemingly no reason.

## Changelog Fixes
+ Fixed Sea Creature Tracker error when attempting to migrate outdated sea creature names to new ones. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
@@ -7,7 +7,6 @@ import at.hannibal2.skyhanni.config.commands.CommandRegistrationEvent
 import at.hannibal2.skyhanni.data.IslandType
 import at.hannibal2.skyhanni.data.jsonobjects.repo.ExcludedSeaCreatureAreasJson
 import at.hannibal2.skyhanni.events.RepositoryReloadEvent
-import at.hannibal2.skyhanni.events.fishing.FishingBobberCastEvent
 import at.hannibal2.skyhanni.events.fishing.SeaCreatureFishEvent
 import at.hannibal2.skyhanni.features.fishing.FishingApi
 import at.hannibal2.skyhanni.features.fishing.SeaCreatureManager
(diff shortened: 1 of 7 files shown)


src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
```diff
@@ -28,10 +27,12 @@ import at.hannibal2.skyhanni.utils.tracker.SessionUptime
 import at.hannibal2.skyhanni.utils.tracker.SkyHanniTracker
 import at.hannibal2.skyhanni.utils.tracker.TrackerData
 import com.google.gson.annotations.Expose
+import kotlin.concurrent.atomics.AtomicBoolean
 
 @SkyHanniModule
 object SeaCreatureTracker {
-    private var needMigration = true
+
+    private val isMigrating = AtomicBoolean(false)
```
hannibal002, line 35: what is the usage of atomic here trying to fix?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/fishing/tracker/SeaCreatureTracker.kt
lunaynx: It's to make sure we don't try to migrate while we're already migrating. Considering e.g. config load and repo reload happen on different threads, I think it's reasonable to use an atomic boolean to be extra sure. It shouldn't have any noticeable performance impact.

PR #5902, opened 2026-06-21
Backend: Improved achievement constructor calls

hannibal002:
## What
This PR improves the Achievement class and all constructor calls:

- They now use named arguments to improve readability.
- The value for `userLuckAmount` is now strictly required. 
	- Spots where `userLuckAmount` was not defined were fixed by setting it to `0`.
- Added secondary constructor to avoid duplicate `.asComponent()` calls.
- Removed redundant `secret = false` constructor lines (secret is false by default already).
- Missing trailing commas were added to parameter lists across all modified files.

+ Refactored Achievement constructors and enforced named parameters. - hannibal2


NopoTheGamer:
> The value for userLuckAmount is now strictly required.

>    Spots where userLuckAmount was not defined were fixed by setting it to 1.

? this isnt a fix, you just decided to change my default of 0 to 1?

hannibal002:
> ? this isnt a fix, you just decided to change my default of 0 to 1?

sure, then we change all those spots back to 0f, i just want to read that default to be seen explicitly, not that someone forgets to add a number altogether.

PR #5903, opened 2026-06-21
Backend: Chat events documentations

hannibal002:
## What
Added extensive documentations to all chat events to clearly distinguish between the `Allow` (read-only) and `Modify` (editing) phases of the chat processing pipeline. Cleaned up the backend by removing the `blockedReason` parameter from all Modify classes since blocking should exclusively be handled during the Allow phase.

+ Added comprehensive docs to the chat event classes. - hannibal2
+ Removed the unused `blockedReason` parameter from all Modify chat event classes. - hannibal2


NopoTheGamer:
doesnt mention that you should only use modify to modify message, never for data collection of any type

PR #5905, opened 2026-06-21
Backend: Add automated Detekt review comments on pull requests

hannibal002:
## What

A `workflow_run` two-workflow split is used so the review can be posted with write permissions even for fork PRs, without executing any fork-controlled code in the privileged context. `detekt.yml` saves the SARIF output as an artifact. `detekt-review.yml` triggers on `workflow_run` completion, downloads the artifact, resolves the PR number via the GitHub API, and runs a Kotlin script from the base branch to post the review.

The script posts the first 20 findings as inline PR review comments. If there are more, the overflow appears in a `<details>` spoiler in the review body. If the GitHub API rejects inline comments with HTTP 422 (line not in diff), it falls back to a body-only review listing all findings. A "Detekt" label is added to the PR when violations are found and removed when the run is clean.

+ Added `detekt-review.yml` workflow to post Detekt findings as inline PR review comments. - hannibal2
    * Used a workflow_run two-workflow split so fork PRs are handled securely without running fork-controlled code with write permissions.
    * Managed a "Detekt" label on the PR based on whether violations are found.
+ Added `post_detekt_review.main.kts` Kotlin script for parsing SARIF output and calling the GitHub review API. - hannibal2
+ Updated `detekt.yml` to save the SARIF output as a workflow artifact. - hannibal2


AverageUser125:
1. I think `gh api` only shows 30 PRs maximum.
2. what if 2 PRs use the same branch name? (example: patch-1)
3. Maybe read the PR number from the artifact metadata? (very unsure)
4. Can the workflow delete previous detekt review comments so it doesn't clutter the reviews?
5. Detekt can give "startLine=0", which might be problematic. (probably not)

hannibal002:
> Can the workflow delete previous detekt review comments so it doesn't clutter the reviews?

not easily, sadly.

the other 4 points are implemented now.

hannibal002:
@AverageUser125 ok now it should work without spamming old comments

AverageUser125:
> @AverageUser125 ok now it should work without spamming old comments

cool

PR #5911, opened 2026-06-22
Backend: REGEX-TEST group enforcement

AverageUser125:
## Dependencies
- #5913

## What
Adds a check that every single group inside a regex pattern must be tested atleast once.
Also adds a cache for parsing REGEX-TEST comments since this is becoming quite slow

+ Enforced REGEX-TEST to test all regex groups atleast once. - AverageUser125
+ Added caching to REGEX-TEST verification. - AverageUser125


lunaynx:
This just seems useless the way you implemented it?

What we'd actually want is something like `REGEX-TEST: " 4sn_: Archer (29)", level="29"` (provisional syntax, not final).

AverageUser125:
> This just seems useless the way you implemented it?
> 
> What we'd actually want is something like `REGEX-TEST: " 4sn_: Archer (29)", level="29"` (provisional syntax, not final).

The point was more a generic "this must be a number".
But tbh a more controlled version with a value for each one works too.

AverageUser125:
Just thought having a generic "REGEX-GROUP" would make adding it everywhere alot easier.
instead of doing it manually for each REGEX-TEST.

well, I guess automating adding those would not be hard, but still.

lunaynx:
What you've done right now achieves nothing at all, you're simply restating the original regex or adding an additional regex condition upon it.

AverageUser125:
> What you've done right now achieves nothing at all, you're simply restating the original regex or adding an additional regex condition upon it.

It's a regex on the group value, so like ensuring level would actually be a number.
Even though it passes the overall REGEX-TEST, for the specific group it would fail.

lunaynx:
I don't see how the group match would fail in your example.
Screenshot of all tests passing on regex101

AverageUser125:
Oh, I used the fixed regex in my example.
In original one where the `level` group contained the color codes.

(I just fixed the example in the description, sorry about that)

lunaynx:
My question is under what circumstances would you ever want to add `REGEX-GROUP: level=\d+` and *not* change the group itself to `\d+`?

AverageUser125:
None tbh, So I guess your idea with explicit group values works better.
Though they are functionally equivalent.

AverageUser125:
There are now 2 Detekt rules for parsing all comments.
I should probably make them use some shared database or smth, or cache. But I got no clue how.
I just wish the `reportIssue` function would let me say the warning name.

but till then, I guess this is fine?

AverageUser125:
Oh wow, 8 minute detekt builds instead of 10-11 minutes.
I guess the cache _did_ work.

AverageUser125:
I think Imma just remove the part where you can specify explicit group values.
And keep:
1. Ensure all groups are accounted in tests
2. The cache, which seems to work remarkably well.
3. Fixed Regex tests that failed at point 1.

(Tomorrow)

PR #5913, opened 2026-06-22
Backend: Enforce WRAPPED-REGEX-TEST for surrounding whitespace

lunaynx:
## What
Makes it illegal to create plain `REGEX-TEST` or `REGEX-FAIL` with leading or trailing whitespace, requiring the usage of `WRAPPED-REGEX-TEST` or the newly added `WRAPPED-REGEX-FAIL` instead.

Why? Because it's too easy to accidentally add extra leading/trailing whitespace, and even when it is intentional, it's difficult to tell at a glance what's a separator and what's actually part of the pattern. Trailing whitespace is especially bad, because it's invisible in a lot of places (though Git tends to warn you about it).

+ Added WRAPPED-REGEX-FAIL type for regex tests. - Luna
+ Changed regex tests that contain leading or trailing whitespace to require WRAPPED-REGEX-TEST or WRAPPED-REGEX-FAIL. - Luna

### DIFF
--- a/detekt/src/main/kotlin/RepoPatternElement.kt
+++ b/detekt/src/main/kotlin/RepoPatternElement.kt
@@ -62,10 +62,20 @@ class RepoPatternElement private constructor(
 
             kDoc.getDefaultSection().getContent().lines().forEach { line ->
                 if (line.contains("REGEX-TEST: ")) {
-                    regexTests.add(line.substringAfter("REGEX-TEST: "))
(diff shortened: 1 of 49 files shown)


AverageUser125:
"REGEX-TEST" is contained inside "WRAPPED-REGEX-TEST".
Which causes that to also run.
it should use startsWith probably,
With left trim.

CalMWolfs (approved):

PR #5918, opened 2026-06-23
Backend: Document KSP and mixin conventions in CONTRIBUTING.md

hannibal002:
## What
Restructuring CONTRIBUTING.md to separate coding rules from technical implementation details for `@SkyHanniModule` and mixin registration. Adds a dedicated KSP section under "Software Used in SkyHanni".

+ Documented KSP annotation processing and mixin coding conventions in CONTRIBUTING.md. - hannibal2

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -215,12 +215,13 @@ Make sure such pull requests have a good explanation in the **What** section.
     - If you can improve the existing feature in a meaningful way.
 - All new classes should be written in Kotlin, with a few exceptions:
     - Mixin classes in `at.hannibal2.skyhanni.mixins.transformers`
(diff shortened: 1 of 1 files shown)


AverageUser125:
Maybe also suggest disabling the gradle daemon?

hannibal002:
why should someone disable the gradle daemon?

AverageUser125:
Cause the cache for a lot of things doesn’t like it.
Like if you change the Detekt code, it sometime doesn’t  use the new one till you restart the daemon.

hannibal002:
> Cause the cache for a lot of things doesn’t like it. Like if you change the Detekt code, it sometime doesn’t use the new one till you restart the daemon.

Permanently disabling the daemon would significantly slow down all builds because the JVM would have to be restarted for each build.

PR #5921, opened 2026-06-23
Improvement: Use smooth mode for lowest BIN prices

lunaynx:
## What
Switches Auction House price fetching to use `mode=smooth`, which removes clear outliers and makes it more resilient to manipulation, which is a frequent cause of false "Tracker Drop" notifications.

## Changelog Improvements
+ Reduced cases of inflated prices being reported due to Auction House manipulation. - Luna


lunaynx:
Put on hold while I look into making this configurable.

AverageUser125:
Maybe add bazzar to this too?

lunaynx:
We use Hypixel API directly for Bazaar data, which doesn't have any sort of average. Elite API does, but it just says "average", it's unclear what the period is.

If we change that, it should probably be its own PR. Bazaar is less inherently susceptible to the same kind of manipulation anyway, it's more rare there.

lunaynx:
Changed back to lowest BIN but with smooth mode, per discussion on Discord.

CalMWolfs (approved):

PR #5927, opened 2026-06-24
Backend: Consolidate & deprecate formattedTextCompat methods

lunaynx:
## What
This PR does a few things to consolidate `formattedTextCompat*` methods: first, it renames the `noExtraResets` parameter of `formattedTextCompat` to `extraResets`, flipping all call sites. Then, it flips the default from `extraResets = true, leadingWhite = false` to `extraResets = false, leadingWhite = true`. Finally, it performs the following replacements:
- `formattedTextCompat()` (75 uses) → `formattedTextCompat(extraResets = true, leadingWhite = false)`
- `formattedTextCompatLeadingWhite()` (5 uses) → `formattedTextCompat(extraResets = true)`
- `formattedTextCompatLeadingWhiteLessResets()` (193 uses) → `formattedTextCompat()`
- `formattedTextCompatLessResets()` (95 uses) → `formattedTextCompat(leadingWhite = false)`

I got confirmation from @NopoTheGamer on Discord, the "has to be separate for pattern mappings" concern is no longer relevant.

It also marks `formattedTextCompat` and related public methods as deprecated.

The goal of this PR is to generally leave all functionality as-is to be on the safe side, and just improve code readability a bit and clearly indicate deprecation – cases of "this could easily just use the clean name or `formattedTextCompat` with no arguments" will be dealt with in a separate PR.

A few lines that got newly flagged by Detekt due to these changes had to be reformatted.

+ Consolidated and deprecated formattedTextCompat methods. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/DataWatcherApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/DataWatcherApi.kt
@@ -7,7 +7,7 @@ import at.hannibal2.skyhanni.events.entity.EntityCustomNameUpdateEvent
 import at.hannibal2.skyhanni.events.entity.EntityHealthUpdateEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.EntityUtils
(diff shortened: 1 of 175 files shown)


lunaynx:
Probably too ambitious.

PR #5916, opened 2026-06-23
Improvement + Backend: Rework Mineshaft corpse finder logic

Piggered:
## What
(Co-authored by @hannibal002.)

Some notable changes made in this PR:

* Added `CorpseFoundEvent`, which fires whenever a corpse is in the player's line of sight.
* Use `EntityEquipmentChangeEvent` to store all corpse entities instead of polling all entities every second.
* Perform `canBeSeen` checks for corpses on EntityMoveEvent<LocalPlayer> instead of `SecondPassedEvent`.
* Moved corpse identification data from `MineshaftWaypoint` into `CorpseType`.

## Changelog Improvements
+ Improved Glacite Mineshaft corpse detection to be instantaneous. - Piggered

## Changelog Technical Details
+ Reworked Glacite Mineshaft corpse finding logic.  - Piggered & hannibal2

PR #5917, opened 2026-06-23
Feature: Corpse Waypoints and All Found Alert

GrowlingGrizzly:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/759

## What
Adds corpse waypoints to mineshafts. I believe this to be a decent improvement over skyocean's due to a) all mineshafts being up to date (thats missing littlefoot) and b) sight checks remove potential waypoints after looking at them if theres no corpse there. Also, if you find all corpses, all remaining potential waypoints dissapear. This PR also adds a (toggleable) alert when finding all corpses. Video shows one shaft with alert off and one with it on.

<details>
<summary>Video</summary>

https://github.com/user-attachments/assets/38edde9f-0a3d-48f5-b67d-18d2435e542d

</details>

## Changelog New Features
+ Added Corpse Waypoints. - Growling_Grizzly
+ Added All Frozen Corpses Found Alert. - Growling_Grizzly

## Changelog Technical Details
+ Added the ability to set a max alpha in filled waypoints. - Growling_Grizzly

PR #5929, opened 2026-06-24
Feature: Mute Organ Donor when all Corpses are found

Piggered:
## Dependencies
- #5916

## What
This PR adds a feature to mute the Dings made by the Organ Donor accessory when all corpses in the Glacite Mineshaft are found (i.e., seen by the player).

## Changelog New Features
+ Added Mute when all Mineshaft Corpses are found. - Piggered
    * Mutes the sounds made from the Organ Donor accessory once all Frozen Corpses in the Mineshaft have been found.
    * Requires Corpse Locator to be enabled.


src/main/java/at/hannibal2/skyhanni/features/mining/glacitemineshaft/corpse/MuteOrganDonor.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    @HandleEvent(onlyOnIsland = IslandType.MINESHAFT)
+    fun onPlaySound(event: PlaySoundEvent) {
+        if (!config.muteWhenAllFound || !allCorpsesFound) return
+        if (event.soundName == "block.note_block.harp") {
+            event.cancel()
+        }
+    }
+
+    @HandleEvent(priority = 1)
```
CalMWolfs: priority is not necessary here

CalMWolfs (changes requested):
I feel this can be default on plus doesnt need to send the chat message but others can let me know if that shouldnt be the case

Piggered:
> doesnt need to send the chat message

I think you're right, especially considering #5917 also has an "All Corpses Found" alert, which makes the message kinda redundant (though it has yet to be merged).

CalMWolfs (approved):

PR #5930, opened 2026-06-24
Fix: Blank Stack Detection

lunaynx:
## What
Fixes Improved SkyBlock Menus incorrectly treating all black stained glass panes as filler items by adding more checks.

## Changelog Improvements
+ Added search tags to "Improved SkyBlock Menus" option to make it easier to find. - Luna

## Changelog Fixes
+ Fixed Improved SkyBlock Menus treating all black stained glass pane items as filler menu items. - Luna


AverageUser125:
> Task :26.1:compileKotlin FAILED
e: file:///home/runner/work/SkyHanni/SkyHanni/src/main/java/at/hannibal2/skyhanni/features/inventory/BetterContainers.kt:219:30 No value passed for parameter 'context'.
e: file:///home/runner/work/SkyHanni/SkyHanni/src/main/java/at/hannibal2/skyhanni/features/inventory/BetterContainers.kt:219:30 No value passed for parameter 'player'.
e: file:///home/runner/work/SkyHanni/SkyHanni/src/main/java/at/hannibal2/skyhanni/features/inventory/BetterContainers.kt:219:30 No value passed for parameter 'tooltipFlag'.

lunaynx:
Oh what, let me see.

src/main/java/at/hannibal2/skyhanni/features/inventory/BetterContainers.kt
```diff
@@ -211,9 +213,11 @@ object BetterContainers {
...
     private fun getClickedSlot(): Int = if (clickedSlotAt.passedSince() <= 500.milliseconds) clickedSlot else -1
 
-    private fun isBlankStack(
-        stack: SafeItemStack,
-    ): Boolean = stack.isStainedGlassPane(ColoredBlockCompat.BLACK)
+    private fun isBlankStack(stack: SafeItemStack): Boolean = stack.isStainedGlassPane(ColoredBlockCompat.BLACK) &&
+        stack.count == 1 &&
+        stack.hoverName.string.isEmpty() &&
+        stack.getLoreComponent().isEmpty() &&
+        stack.getTooltip().isEmpty()
```
NopoTheGamer, line 217: cant you just check if the tooltip is hidden....

lunaynx: Yeah, I can probably remove the hoverName and lore checks (might keep the count just in case). The underlying Minecraft method called by `getTooltip` already returns an empty list if the tooltip is hidden, so that should be fine.

hannibal002: so if we already have this check here, why having the lore component check above?

hannibal002: or as alternative, we leave getTooltip check out entirely. but having both is a perf regression

hannibal002 (changes requested):
Needs in-game verification. Can getTooltip().isEmpty() actually return true for a real SkyBlock filler pane?

lunaynx:
Will do once Mojang auth is back up. It should work in theory, because Hypixel makes the tooltip hidden, which should result in it returning an empty list.

hannibal002 (changes requested):
bump

lunaynx:
Yep, seems to work. Added in a tiny search QoL change because it was pissing me off.

hannibal002 (approved):

PR #5933, opened 2026-06-25
Fix: Prepare for equipment/wardrobe update on alpha

AverageUser125:
## What
basically just preparing the regex and detection.
No point in changelog really but still.
Idk how to add to the changelog that it also works on alpha, but it is implied so it's probably fine.

## Changelog Fixes
+ Fixed Custom wardrobe not working with new hypixel update. - AverageUser125


AverageUser125:
This is probably gonna get deleted with the new custom layout menu Fabi is working on.

it technically works and is ready for review just probably is more of a temporary change

lunaynx:
This can now be promoted to a "Fix". 🔥

AverageUser125:
image
Okay, something is wrong lol.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/WardrobeSlot.kt
```diff
@@ -9,6 +9,11 @@ class WardrobeSlot(
     val leggingsSlot: Int,
     val bootsSlot: Int,
 ) {
+    val necklaceSlot get() = helmetSlot
```
hannibal002: whats going on here? can we just use the wardrobe names instead of the armor names and then make those 4 val get() unnecessary?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/WardrobeSlot.kt
AverageUser125: there were wardrobe names? I didn't see that lol.
I just added this for if an equipement wardrobe is even added.

But if there are proper names like "top", "top middle", then sure.
Just tell me what they are.

AverageUser125: I can just remove this, since nothing uses it really.

hannibal002: no there arent. i mean the names necklace, etc.. i assumes there is no point in having the names helmet, leging, boots etc when the context is wardrobe and not armor

AverageUser125: hannibal, I do not understand your point.
Imma just remove those for now.
Explain in discord.

AverageUser125:
I mean, I know something is still broken.
with the armor extraction.
But I ain't got MVP+ so I can't fix it.

hannibal002:
lets user beta testers to find out!

hannibal002 (approved):

PR #5934, opened 2026-06-25
Fix: Prevent memory leaks from cached objects

AverageUser125:
## What
Use weak key refrences where possible.

Even though these are temporary memory leaks that flushed after a while, for extreme cases like the 2 minutes `CachedItemData` it was causing skyhanni to take like 1GB of RAM when scrolling the auction house.

And while I was here I made a similar fixes to the other TimeLimitedCache usage.

## Changelog Fixes
+ Fixed memory leaks caused by caches keeping unused objects alive. - AverageUser125


lunaynx:
Can't you just add a parameter to `TimeLimitedCache`? Something like this:
```kt
@Suppress("UnstableApiUsage")
class TimeLimitedCache<K : Any, V : Any>(
    expireAfterWrite: Duration,
    useWeakKeys: Boolean = false,
    removalListener: ((K?, V?, RemovalCause) -> Unit)? = null,
) : CacheMap<K, V>() {

    override val cache: Cache<K, V> = buildCache {
        if (useWeakKeys) weakKeys()
        expireAfterWrite(expireAfterWrite.inWholeNanoseconds, TimeUnit.NANOSECONDS)
        setRemovalListener(removalListener)
    }
}
```

AverageUser125:
> Can't you just add a parameter to `TimeLimitedCache`? Something like this:
> 
> ```kotlin
> @Suppress("UnstableApiUsage")
> class TimeLimitedCache<K : Any, V : Any>(
>     expireAfterWrite: Duration,
>     useWeakKeys: Boolean = false,
>     removalListener: ((K?, V?, RemovalCause) -> Unit)? = null,
> ) : CacheMap<K, V>() {
> 
>     override val cache: Cache<K, V> = buildCache {
>         if (useWeakKeys) weakKeys()
>         expireAfterWrite(expireAfterWrite.inWholeNanoseconds, TimeUnit.NANOSECONDS)
>         setRemovalListener(removalListener)
>     }
> }
> ```

sure

AverageUser125:
I just hope the tests failing is due to them not holding any strong refrence to the ItemStack.

lunaynx:
Ah my argument order was bad, probably put the new one at the end – though as we discussed on Discord this may not work well due to it being wrapped in IdentityCharacteristics anyway.

AverageUser125:
I  generalized the idea to use a weak hash map anywhere it would make sense instead of just CachedItemData.
So that's good.

src/main/java/at/hannibal2/skyhanni/api/VanquisherApi.kt
```diff
@@ -64,7 +64,7 @@ object VanquisherApi {
     private var lastSoundPos: LorenzVec? = null
     private var lastSoundTime = SimpleTimeMark.farPast()
 
-    private val vanquishers = TimeLimitedCache<Mob, VanquisherData>(6.minutes) { mob, data, _ ->
+    private val vanquishers = TimeLimitedCache<Mob, VanquisherData>(6.minutes, useWeakKeys = true) { mob, data, _ ->
```
hannibal002: `useWeakKeys = true` means Guava calls the removal listener with null for the key when the Mob is garbage collected, so the existing if (mob != null) guard will always skip postDespawn().

src/main/java/at/hannibal2/skyhanni/utils/collection/TimeLimitedSet.kt
```diff
@@ -5,11 +5,18 @@ import kotlin.time.Duration
 
 class TimeLimitedSet<T : Any>(
     expireAfterWrite: Duration,
+    weak: Boolean = false,
```
hannibal002: in TimeLimitedCache the variable name is useWeakKeys. pleas estay consistent over both files

hannibal002 (changes requested):
i dont see the memory leak. TimeLimitedCache already drops objects after a while. if there is one still, please explain more in the what block.

src/main/java/at/hannibal2/skyhanni/utils/collection/TimeLimitedSet.kt
AverageUser125: This is a set, using `useWeakKeys` when there are none feels wrong.
especially if `useWeakValues` is later added

AverageUser125: I'll do it anyways, though I disagree.

hannibal002 (approved):

PR #5936, opened 2026-06-25
Fix: BazaarCancelledBuyOrderClipboard regex fix

AverageUser125:
## What
Made it be colorless too.
Does mean it doesn't fix for all people backwards compatible, but it's fineeeeeeeee.

## Changelog Fixes
+ Fixed Bazaar cancelled Buy Order clipboard not working. - AverageUser125


src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/BazaarCancelledBuyOrderClipboard.kt
```diff
@@ -52,14 +53,14 @@ object BazaarCancelledBuyOrderClipboard {
         val stack = event.inventoryItems[11] ?: return
         if (!stack.hoverName.string.contains("Cancel Order")) return
 
-        val lore = stack.getLore()
+        val lore = stack.getLore().map { it.removeColor() }
```
lunaynx:
```suggestion
        val lore = stack.getLoreComponent().map { it.string.removeColor() }
```

lunaynx (changes requested):

PR #5937, opened 2026-06-25
Feature: Compacted Item Form Recipe Keybind

Piggered:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/761

## What
This feature adds an inventory keybind to open the recipe of the hovered item for its compacted form. This is useful for quickly Supercrafting items from sacks into its compacted form.

The current way to do what this feature does is to use a mod like SkyBlock Item List or REI with a compatibility mod, press the "View Usages" keybind on that item, and click on the "Show SkyBlock Craft" button for the compacted form's recipe. However, that recipe sometimes isn't among the first ones in the list, so you have to go through multiple pages of recipes to find it.

## Changelog New Features
+ Added keybind to view the recipe for the item's compacted form. - Piggered
    * Example: Pressing the keybind on a Diamond will open the recipe for Enchanted Diamond.


Piggered:
Drafting this PR as there are missing recipes and I want to make some changes:

* Make a Python script in SkyHanni-REPO to generate the recipes JSON file (it's starting to become annoying to maintain).
* Actually use components for the feature's messages.

PR #5940, opened 2026-06-26
Backend: Fix compilation warnings

AverageUser125:
## What
Fixes the miscellaneous build warnings in the codebase, there really weren't much.
and also disables deprecation warnings from building, InteliJ still shows the method as deprecated so doesn't change much except cleanup the workflow logs.

+ Resolved all compilation warnings. - AverageUser125
+ Suppressed deprecation warnings in build process. - AverageUser125
+ Fixed REI transitive dependencies not being properly excluded, leading to a bunch of log spam at compile time from Loom remap warnings. - Luna


lunaynx (approved):
Looks good to me, didn't test.

AverageUser125:
the build deprecation flag didn't work.
So Imma remove that for now.

AverageUser125:
I found the build flag in the end, apperantly the flag only exists for new kotlin versions so it was abit hard to find.
The only thing I wish I could remove was:
```
Cannot remap method_10764 because it does not exist in any of the targets [net/minecraft/class_2535]
```
But the fabric loom got no way to mute this warning.
Wonder if it has something to do with mixin parameter names not being accurate.

lunaynx:
Yeah, I saw that too. Codex told me it was just a stale local cache but it was wrong, it happens in a clean CI environment too and also on 26.1 which doesn't have mappings at all IIRC. I'll try again sometime.

AverageUser125:
Well, now there are only 1-off warnings like:
```
:jar: No valid plugin descriptors were found in META-INF/gradle-plugins
Cannot remap modifiers because it does not exist in any of the targets [] or their parents.
Not publishing sources jar as it was not created by the java plugin. Use java.withSourcesJar() to fix.
```

But the workflow logs are already excellent so imma keep it as is.

AverageUser125:
should I also disable gradle daemon for workflows in this PR?

lunaynx:
It should basically make no difference whether the daemon is enabled in CI or not, as it's a short-lived isolated runner that gets destroyed at the end of the run regardless. If anything, I'd keep it on to gain a potential slight speedup in case we ever have multiple Gradle invocations within the same workflow run.

AverageUser125:
Okay, I will fix the new warnings first.

AverageUser125:
Okay, new warnings have been fixed.
Though I am unsure about the “isAllowedFromServer” one.

lunaynx:
I'd probably just remove those lines entirely, if it can't be nullable in the first place the check is redundant. If we end up figuring that we need `isAllowedFromServer` in the future we can add it, but I wouldn't risk potentially accidentally breaking things for now (though it probably won't).

src/main/java/at/hannibal2/skyhanni/features/inventory/BetterContainers.kt
```diff
@@ -289,7 +290,7 @@ object BetterContainers {
         val isSuperpairs = unformattedLower.startsWith("Superpairs") && !containsStakes
 
         for (index in 0..<size) {
-            val stack: SafeItemStack = handlerInventory.getItem(index) ?: continue
+            val stack: SafeItemStack = handlerInventory.getItem(index)
```
hannibal002: why

src/main/java/at/hannibal2/skyhanni/features/misc/massconfiguration/FeatureToggleProcessor.kt
```diff
@@ -43,6 +43,7 @@ class FeatureToggleProcessor : ConfigStructureReader {
         pathStack.pop()
     }
 
+    @Suppress("UNCHECKED_CAST", "PLATFORM_CLASS_MAPPED_TO_KOTLIN")
```
hannibal002, line 47: suppress is correct here, but please add a short comment explaining why it is needed. (e.g. `java.lang.Boolean required for boxed generic type check` or `cast is safe after require checks`)

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/BetterContainers.kt
AverageUser125: cause kotlin says it's a redundant elvis operator.

AverageUser125: also, since it's already index for the 0 to size, it's never gonna be null/invalid anyways.

src/main/java/at/hannibal2/skyhanni/features/misc/massconfiguration/FeatureToggleProcessor.kt
AverageUser125: feels redundant. but alright.

src/main/java/at/hannibal2/skyhanni/features/inventory/BetterContainers.kt
hannibal002: The concern isn't whether the index is valid.

It's that the behaviour changed for empty slots. Previously, `getItem()` returned null for empty slots, so `?: continue` skipped them. Now it returns `ItemStack.EMPTY`, and without the skip, empty slots fall through with `isBlankStack(EMPTY) = false`, which sets `isSlot = true`. Those slots then get the slot texture rendered in the second loop, which never happened before.

Is that intentional?

src/main/java/at/hannibal2/skyhanni/features/misc/massconfiguration/FeatureToggleProcessor.kt
hannibal002: "Unbox" is the wrong term here. The reason `java.lang.Boolean` is needed is that generic type arguments are always boxed, so `actualTypeArguments[0]` on a `Property<Boolean>` returns `java.lang.Boolean::class.java`, not `kotlin.Boolean::class.java`.

A more accurate comment would be:
`// java.lang.Boolean required: generic type arguments are always boxed, so kotlin.Boolean would not match here`

hannibal002 (changes requested):
bump

src/main/java/at/hannibal2/skyhanni/features/inventory/BetterContainers.kt
lunaynx: That was probably already changed earlier (probably in the 1.21 port). The elvis operator isn't going to fall back on empty ItemStacks anyway, so it was already "broken".

We should probably standardize ItemStack handling code so that all SkyHanni utils return a nullable `stack.takeUnless { it.isEmpty() }`, and then make a general extension method like `.orNull()` or `.takeUnlessEmpty()` for vanilla return paths we can't control.

For this PR, this change seems fine, but I'd add TODO comments to ItemStack-related redundant elvis removals to investigate them later.

AverageUser125: Imma just add a check for isEmpty then continue, just incase.

lunaynx: That works too, might as well do that right now for cases that still had a redundant elvis. Probably just use `takeIf` like I said to keep it a one-liner.

lunaynx: Actually `.takeUnless { it.isEmpty() }`, but close enough.

AverageUser125: Imma just add an "takeIfNotEmpty", where do you want it defined?

lunaynx: Probably in the closest equivalent to an ItemStack utils class that exists.

Also, I just edited my earlier comment, probably better to do it like this to sound more natural:
```kt
fun SafeItemStack.takeUnlessEmpty(): SafeItemStack? = takeUnless { it.isEmpty() }
```

(`isNotEmpty` may not even exist for ItemStacks unless we already defined it.)

AverageUser125:
I added `SafeItemStack.takeUnlessEmpty` and `Style.takeUnlessEmpty`.

hannibal002 (approved):
> We should probably standardize ItemStack handling code so that all SkyHanni utils return a nullable `stack.takeUnless { it.isEmpty() }`, and then make a general extension method like `.orNull()` or `.takeUnlessEmpty()` for vanilla return paths we can't control.

I agree.

PR #5941, opened 2026-06-26
Backend: Fix Loom remap warnings from REI during build

lunaynx:
## What
Fixes a bunch of "Cannot remap ..." warnings emitted by Loom when configuring the 1.21.11 project during build, which were coming from REI transitive dependencies that weren't being properly excluded.

There is a single remaining warning in the log, but this appears to be from Fabric API itself, and I don't want to mess with it:
```
Cannot remap modifiers because it does not exist in any of the targets [] or their parents.
```

+ Fixed REI transitive dependencies not being properly excluded, leading to a bunch of log spam at compile time from Loom remap warnings.


AverageUser125:
Uh, should I just copy this and put this in my fix compilation warning PR?

lunaynx:
You can just merge it into your branch and add it to the "Dependencies" section, I guess. Or I can actually just close this PR if you add both of our names to your changelog entry, since there's no real need for it to be separate.

AverageUser125:
I copied and added you to the changelog.
So I guess close this PR

PR #5943, opened 2026-06-26
Fix: Prevent stack overflow in ErrorManager

lunaynx:
## What
Potentially fixes this (or at least, stops the real error from being obscured):
https://discord.com/channels/997079228510117908/1520172858729304074
https://bytebin.lucko.me/GDGFQ1npP6

## Changelog Fixes
+ Fixed a rare case with error handling sometimes infinitely recursing and causing the game to freeze or crash. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/test/command/ErrorManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/test/command/ErrorManager.kt
@@ -213,9 +213,14 @@ object ErrorManager {
 
     // This is intentionally not an enum, because unnecessary object allocation can be problematic
     // if we're dealing with a stack overflow.
-    private const val ERROR_STATE_LOGGED = 0
-    private const val ERROR_STATE_BLOCKED_NOT_NEEDED = 1
-    private const val ERROR_STATE_BLOCKED_CAN_NOT_SHOW = 2
+    @JvmInline
+    private value class ErrorState(val value: Int) {
+        companion object {
+            val LOGGED = ErrorState(0)
+            val BLOCKED_NOT_NEEDED = ErrorState(1)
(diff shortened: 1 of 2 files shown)


AverageUser125:
I probably misunderstood, but should there be detection for the error logging itself failing not causing an infinite recursion of it trying to print the error?
Or is that was this PR does?

I just thought it would be an AtomicBoolean “errorInProgress” tbh.

src/main/java/at/hannibal2/skyhanni/test/command/ErrorManager.kt
```diff
@@ -209,10 +211,15 @@ object ErrorManager {
 
     data class CachedError(val className: String, val lineNumber: Int, val errorMessage: String)
 
-    enum class ErrorState {
-        LOGGED,
-        BLOCKED_NOT_NEEDED,
-        BLOCKED_CAN_NOT_SHOW,
+    // This is intentionally not an enum, because unnecessary object allocation can be problematic
```
hannibal002, line 214: I'm very confused. An enum can't create more objects/instances of the actual class than there are fields, a class like this below here can though. This allows the opposite of what the comment suggests

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/test/command/ErrorManager.kt
lunaynx: You're right that an enum won't create multiple copies of itself, but it was struggling to initialize *the enum itself*. But actually, I did some testing, and while this value class construct is more lightweight and has a higher chance of working, it isn't foolproof either. I'm going back to the true compile-time constant solution.

hannibal002 (approved):

PR #5945, opened 2026-06-27
Fix: fix dragon tracker regex

zumbiepig:
## What
fix dragon tracker, long broken, someone put color codes in the regex, then changed it to event.cleanMessage

## Changelog Fixes
+ Fixed Dragon Profit Tracker. - zumbiepig

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/combat/end/DragonFightAPI.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/combat/end/DragonFightAPI.kt
@@ -2,8 +2,8 @@ package at.hannibal2.skyhanni.features.combat.end
 
 import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.data.IslandType
-import at.hannibal2.skyhanni.data.hypixel.chat.event.SystemMessageEvent
 import at.hannibal2.skyhanni.events.ScoreboardUpdateEvent
+import at.hannibal2.skyhanni.events.chat.SkyHanniChatEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.NumberUtil.formatInt
 import at.hannibal2.skyhanni.utils.RegexUtils.matchMatcher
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/combat/end/DragonFightAPI.kt
```diff
@@ -22,27 +22,27 @@ object DragonFightAPI {
...
     )
 
     /**
-     * REGEX-TEST: §r§f                           §r§6§lOLD DRAGON DOWN!§r
+     * WRAPPED-REGEX-TEST: "                           OLD DRAGON DOWN!"
      */
     private val chatDeath by group.pattern(
         "chat.death",
-        "§r§f {27}§r§6§l(?<type>.*) DOWN!§r",
+        " +(?<type>.+) DOWN!",
```
lunaynx, line 37: This will match "END STONE PROTECTOR DOWN!". Also, in meantime, I checked, and the capturing group is not and was never used for the death pattern, so it can be safely removed.
```suggestion
        " +\\w+ DRAGON DOWN!",
```

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/features/combat/end/DragonFightAPI.kt
lunaynx: Also add:
```
WRAPPED-REGEX-FAIL: "                    END STONE PROTECTOR DOWN!"
```

jani270 (changes requested):
Note: This has not been tested ingame.

Image

lunaynx:
Just noticed some of this logic is duplicated in DragonFeatures, so someone might want to look into it. But it's okay if that's a separate PR.

src/main/java/at/hannibal2/skyhanni/features/combat/end/DragonFightAPI.kt
```diff
@@ -58,12 +58,14 @@ object DragonFightAPI {
     fun inNestArea() = IslandType.THE_END.isInIsland() && nestAreaPattern.matches(SkyBlockUtils.graphArea)
 
     @HandleEvent
-    fun onChat(event: SystemMessageEvent.Allow) {
+    fun onChat(event: SkyHanniChatEvent.Allow) {
```
hannibal002, line 61: why got SystemMessageEvent changed? the chat message that we check for is not a user generated message, so the system message event was more precise

hannibal002 (changes requested):
> Fixed Dragon Profit Tracker. 

that is too broad

src/main/java/at/hannibal2/skyhanni/features/combat/end/DragonFightAPI.kt
zumbiepig: because system message event is bugged and fires 2 times per message. i will make this pr a draft until i get that fixed in another pr

PR #5946, opened 2026-06-27
Feature: Added Solo mode for living metal snakes

Icetrixwastaken:
## What
Adds a Solo mode to the Living Metal Snake Helper that highlights both the head and tail of the snakes for players who are both breaking and calming the snake with a toggle in the config.
Idea from https://discord.com/channels/997079228510117908/1136434428047020152/threads/1518273021725511690

<details>
<summary>Images</summary>

Screenshot (209)
The snakes with solo mode toggled off
Screenshot (208)
A not touching air snake with solo mode turned on
Screenshot (207)
A spawning snake with solo mode turned on
Screenshot 2026-06-26 170317
the config

</details>

## Changelog New Features
+ Added solo mode for rift snake features. - Icetrix


AverageUser125:
I don't think this should be default enabled.

lunaynx:
Yeah, it also shouldn't be a FeatureToggle, at least with the current implementation. If it was like "auto-detect based on hotbar items", it could be.

Icetrixwastaken:
so I should make it detect if both tools are in the hotbar and if so show both?

CalMWolfs (approved):

PR #5947, opened 2026-06-27
Fix: Minor contributor manager fixes

AverageUser125:
## What
Fixes a typo in the achievement, and makes it not give an error in singleplayer.
Sometimes.
Also should reduce any error by calling `PlayerUtils.getName() & PlayerUtils.getUuid()`

## Changelog Fixes
+ Fixed "Notice Me Senpai" Achievement description. - AverageUser125

+ Changed `PlayerUtils` to use `localUser` instead of `localPlayer` for player name and UUID lookup. - AverageUser125

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
@@ -51,8 +51,9 @@ object ContributorAchievement {
             Achievement(
                 "Notice Me Senpai".asComponent(),
                 componentBuilder {
-                    append("Have your friend request ignored by ")
+                    append("Have your friend request ignored by a")
                     appendWithColor("SkyHanni", TextHelper.chromaStyle)
+                    append(" contributor")
                 }
             ),
             CONTRIBUTOR_NOBODY_ACHIEVEMENT
--- a/src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
@@ -51,7 +51,7 @@ object ContributorAchievement {
             Achievement(
                 "Notice Me Senpai".asComponent(),
                 componentBuilder {
-                    append("Have your friend request ignored by a")
+                    append("Have your friend request ignored by a ")
                     appendWithColor("SkyHanni", TextHelper.chromaStyle)
                     append(" contributor")
                 }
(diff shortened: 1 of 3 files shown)


AverageUser125:
Feels odd that only `localUser` has documentation in the `MinecraftCompat` class, but oh well.

lunaynx:
#5887 will fix that partially :pray:

src/main/java/at/hannibal2/skyhanni/utils/compat/MinecraftCompat.kt
```diff
@@ -21,6 +22,10 @@ object MinecraftCompat {
 
     val localPlayerOrNull get(): LocalPlayer? = Minecraft.getInstance().player
 
+    // This class contains the local user information, such as the username and UUID.
+    // It will always be non-null, even if the player is not in a world / singleplayer.
```
lunaynx: Make it a proper KDoc.
```kt
    /**
     * This class contains the local user information, such as the username and UUID.
     * It will always be non-null, even if the player is not in a world / singleplayer.
     */
```

lunaynx (changes requested):

AverageUser125:
I would prefer PlayerUtils.getName not to be called PlayersUtils.getNameOrThrow.
So this change still makes sense.

But yeah if the naming was more explicit this probably wouldn’t have happened.

lunaynx:
It's mainly necessary for stuff that explicitly uses `!!` or similar, hiding unsafety. Not every single method everywhere that can throw needs to be called `*OrThrow`.

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
```diff
@@ -51,8 +51,9 @@ object ContributorAchievement {
             Achievement(
                 "Notice Me Senpai".asComponent(),
                 componentBuilder {
-                    append("Have your friend request ignored by ")
+                    append("Have your friend request ignored by a")
```
hannibal002: missing space at end

src/main/java/at/hannibal2/skyhanni/utils/compat/MinecraftCompat.kt
```diff
@@ -21,6 +22,12 @@ object MinecraftCompat {
 
     val localPlayerOrNull get(): LocalPlayer? = Minecraft.getInstance().player
 
+    /**
+     * This class contains the local user information, such as the username and UUID.
```
hannibal002: wording is a bit confusing. we describe a member of the class, not the whole class with this line.

hannibal002 (changes requested):

hannibal002 (approved):

PR #5948, opened 2026-06-27
Backend: replace all usages of \\s with literal space

zumbiepig:
## What
Replaced all usages of '\\w' with space in regexes.

We can save a lot of processing power by doing this since space is the only allowed whitespace character in Minecraft

chatgpt says this can speed up regexes by 3-4 times, and when we have 300 regexes running for every chat message, that could add up

+ Replaced all usages of '\\w' with space in regexes. - zumbiepig
    * We can save a lot of processing power by doing this since space is the only allowed whitespace character in Minecraft.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/ExperimentationTableApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/ExperimentationTableApi.kt
@@ -195,7 +195,7 @@ object ExperimentationTableApi {
      */
     private val expOverRewardsLorePattern by patternGroup.pattern(
         "inventory.experiment-over.rewards",
-        "§8 \\+(?:§.| )*(?:\\[Lvl \\d+] )?(?<reward>.*?)(?=\\s\\((?:Stakes|Pairs)\\)|\$)(?:\\s\\((?:Stakes|Pairs)\\))?",
+        "§8 \\+(?:§.| )*(?:\\[Lvl \\d+] )?(?<reward>.*?)(?= \\((?:Stakes|Pairs)\\)|\$)(?: \\((?:Stakes|Pairs)\\))?",
     )
 
     /**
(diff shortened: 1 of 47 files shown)


AverageUser125:
I think new line character also counts as white space.
So double check that nothing relied on it.

zumbiepig:
> I think new line character also counts as white space. So double check that nothing relied on it.

I dont think we ever parse newlines in chat messages, ill check for regexes that arent for chat messages

AverageUser125:
1. Your “What” section mentions “\w” while the PR changes “\s”
2.
home/runner/work/SkyHanni/SkyHanni/src/main/java/at/hannibal2/skyhanni/features/garden/contest/FarmingContestApi.kt:40:36: Repo pattern `sidebarCropPattern` must have a regex test. [RepoPatternRegexTestMissing]
/home/runner/work/SkyHanni/SkyHanni/src/main/java/at/hannibal2/skyhanni/features/nether/kuudra/KuudraApi.kt:31:33: Repo pattern `completePattern` must have a regex test. [RepoPatternRegexTestMissing]

Odd that your changes caused this, but still.

AverageUser125:
Also, what about `\S` which is the inverse of `\s`?

lunaynx:
This is probably fine, but I highly doubt it's a "3–4x performance improvement". 

For me, the reasoning model says:

> In your case, **100+ regexes per message is definitely worth caring about**, but `\s` versus literal space is probably a micro-optimization, not the main bottleneck.

---

> Also, what about `\S` which is the inverse of `\s`?

Could use `[^ ]` for that.

lunaynx:
All the `[ ]` can be changed to just ` `, by the way. (I know the original regexes had `[\s]`, that was pointless too.)

hannibal002:
Closing, this is a correctness regression not a perf win.

`\s` matches six chars, not just space (space, tab, LF, VT, FF, CR), so swapping it for a literal space silently changes behavior wherever the input can contain those. Concrete break: `[\s\S]*` → `[ \S]*` in CarnivalQuickStart.kt. `[\s\S]` is the "match anything incl. newlines" idiom (the comment right above says exactly that), and `[ \S]` no longer matches newlines. The Snow Particles filter in ChatFilter.kt looks like the same.

Perf-wise, ASCII `\s` is already a cheap six-char check and any gap is noise after JIT warmup.

PR #5949, opened 2026-06-27
Backend: Fix SystemMessageEvent being fucked

zumbiepig:
## What
So, as you can see from this screenshot, SystemMessageEvent is fucked, and does not filter out anything. I replaced all usages in the code of SystemMessageEvent with SkyHanniChatEvent, and deprecated SystemMessageEvent (while also fixing SystemMessageEvent).

SystemMessageEvent will now work properly (and only get posted for "system messages"), but I deprecated it since we really shouldn't be using it anyway.

Also: I fixed the dragon tracker regex having color codes despite using event.cleanMessage

image

## Changelog Fixes
+ Fixed Dragon Profit Tracker regex. - zumbiepig

+ Fixed SystemMessageEvent being posted multiple times per chat message. - zumbiepig

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/EntityMovementData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/EntityMovementData.kt
@@ -9,7 +9,7 @@ import at.hannibal2.skyhanni.events.entity.EntityMoveEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.DelayedRun
 import at.hannibal2.skyhanni.utils.LorenzVec
-import at.hannibal2.skyhanni.utils.RegexUtils.matches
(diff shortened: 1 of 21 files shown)


hannibal002:
Closing this too. It bundles several unrelated changes and the reasoning doesn't line up, so it's not mergeable as-is.

The description says SystemMessageEvent "doesn't filter anything", but the screenshot and changelog show the real issue is it firing multiple times per message, which is a different bug. And it fixes, deprecates, and replaces all usages of SystemMessageEvent in one go: if we're dropping every caller and deprecating it, fixing its behavior is wasted work, pick one. The dragon tracker regex is unrelated, and "color codes despite cleanMessage" points to a bug in cleanMessage (or reading the wrong field), not the regex, so patching the regex just hides it.

If you want the useful parts in: the micro-optimization as its own PR, and the repo-pattern change as its own PR.

PR #5950, opened 2026-06-27
Fix: Hoppity Rabbits Found Regex

Alex33856:
## What
Fixes the Hoppity Rabbits Found regex pattern.
Should be made colorless, but that can be done later 😄.

## Changelog Fixes
+ Fixed Found Hoppity Rabbits not being detected. - Alex

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityCollectionStats.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityCollectionStats.kt
@@ -102,11 +102,11 @@ object HoppityCollectionStats {
     )
 
     /**
-     * REGEX-TEST: §2§l§m                      §f§l§m   §r §e395§6/§e457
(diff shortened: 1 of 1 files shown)


lunaynx:
If you wanted to do a "double fix", you could make it colorless under a `.colorless` pattern key and then leave the legacy one with a TODO to remove it later.

AverageUser125:
> If you wanted to do a "double fix", you could make it colorless under a `.colorless` pattern key and then leave the legacy one with a TODO to remove it later.

That TODO is never gonna happen lol.
A better solution would be to make it colorless here and make a repo PR for the non colored.

Or just don’t do a repo PR and tell people to update their mod.
(Hint: Option 2 is easier)

lunaynx:
It's fine, it doesn't really do much harm if we have a little dead code. Regardless of what you choose though, **do not** make a manual repo PR, it will be overwritten by the workflow.

AverageUser125:
Superseded #5960

PR #5951, opened 2026-06-27
Improvement: Reuse event cache better

AverageUser125:
## What

And `SkyBlockLocationData` is unused since `HypixelLocationApi` was made the default.
As Such I just made everything call `HypixelLocationApi` directly where possible.
And also skipped the ‘HypixelData` step.

Also made the event island check cache last until island change instead of tick end.
Or till some one invalidates the cache.

## Changelog Improvements
+ Improved island detection for better performance. - AverageUser125


lunaynx:
Storing a boolean on each IslandType with no guarantees that only one of them will be true at a time sounds like a recipe for disaster. There are other ways to shorten it without resorting to hacks like this.

AverageUser125:
The guarantee is via there only being a single source that modifies it.
Also just returning a boolean is faster then a comparing and then returning. (probably)
I could make it access `HypixelLocationApi` directly I guess if you perfer that.

lunaynx:
That's not what I meant by a "guarantee". It may be guaranteed today, but it makes it significantly easier for anyone editing the code later to make a mistake.

AverageUser125:
Now it's a more proper solution.

AverageUser125:
/home/runner/work/SkyHanni/SkyHanni/src/main/java/at/hannibal2/skyhanni/features/minion/MinionFeatures.kt:51:1: Unused import [NoUnusedImports]

I go sleep.
I will fix Detekt tommorow. Otherwise can review.
(Why did I make it draft? Idk myself)

AverageUser125:
I removed the check for if a handler is disable inside `shouldInvoke` since its caller checks it already.

Though `blockedInvoker` thing is unused, wonder what’s it meant for.

hannibal002 (changes requested):
Detekt

AverageUser125:
> Detekt

woops

CalMWolfs (approved):

hannibal002 (changes requested):
> I removed the check for if a handler is disable inside shouldInvoke since its caller checks it already.

The caller checks `isDisabledHandler`, which disables all listeners for an event type.
`isDisabledInvoker` is a separate check that disables specific listener methods by name, loaded from the repo via `DisabledEventsJson`.
After this change, `disabledHandlerInvokers` is still populated on repo reload but never queried. This means the per-invoker kill switch is broken.

AverageUser125:
Oh, so that’s why. Imma fix it in like 30 minutes max.
Thanks for telling me.

AverageUser125:
Should I cache the disabled check too btw?
And onRepoReload invalidate the cache?

hannibal002 (changes requested):
> Should I cache the disabled check too btw?

im not sure what you mean

> And onRepoReload invalidate the cache?

yes

AverageUser125:
> > Should I cache the disabled check too btw?
> 
> im not sure what you mean
> 
> > And onRepoReload invalidate the cache?
> 
> yes

Then that’s done already.

AverageUser125:
I added the `isDisabledInvoker` check into `cachedPredicates`.
Is what I meant

src/main/java/at/hannibal2/skyhanni/api/event/EventListeners.kt
```diff
@@ -118,6 +118,7 @@ class EventListeners private constructor(val name: String, private val isGeneric
                 options.onlyOnIslandTypeTag.takeIfNotEmpty()?.let { tags ->
                     add { _ -> tags.any { it.isInIsland() } }
                 }
+                add { _ -> SkyHanniEvents.isDisabledInvoker(name) }
```
hannibal002: i think this is inverted

src/main/java/at/hannibal2/skyhanni/data/IslandTypeTag.kt
```diff
@@ -77,7 +77,7 @@ enum class IslandTypeTag(vararg types: SkyHanniIslandType) : SkyHanniIslandType
         newValues.mapNotNullTo(types) { EnumUtils.enumValueOfOrNull<IslandType>(it.uppercase()) }
     }
 
-    override fun isInIsland(): Boolean = SkyBlockUtils.inSkyBlock && contains(SkyBlockUtils.currentIsland)
+    fun isInIsland(): Boolean = contains(SkyBlockUtils.currentIsland)
```
hannibal002: why is the is in sb check gone now? while outside of sb, island type is none, leaving isInIsland checks true for BUSY and NO_FIXED_NPC_LOCATIONS

src/main/java/at/hannibal2/skyhanni/data/IslandTypeTag.kt
```diff
@@ -11,9 +11,9 @@ import java.util.EnumSet
 
 // TODO maybe rename this class to IslandTypeGroup
 /**
- * Each [IslandTypeTag] consists of one or more [SkyHanniIslandType]
+ * Each [IslandTypeTag] consists of one or more [IslandType] or [IslandTypeTag]
  */
-enum class IslandTypeTag(vararg types: SkyHanniIslandType) : SkyHanniIslandType {
+enum class IslandTypeTag(vararg types: Any) {
```
hannibal002, line 16: why was this changed? previously, there was more type safety

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/IslandTypeTag.kt
AverageUser125: The loop inside already cycles all the types and doesn’t use the interface.
And it would still crash at startup immedialty if someone made a flaw.

it just seemed like a good idea to not make a function virtual when it doesn’t have to be

src/main/java/at/hannibal2/skyhanni/data/IslandTypeTag.kt
AverageUser125: Could have sworn I reverted that change…

src/main/java/at/hannibal2/skyhanni/data/IslandTypeTag.kt
AverageUser125: I don’t mind reverting this change.
It just felt odd using the interface when it was unused and could have just been an empty interface with no functions and it would still work

hannibal002: i dont mind it not being an interface, i was just curious why you removed it

AverageUser125: Cool

hannibal002 (approved):

PR #5953, opened 2026-06-28
Backend: Deprecate formattedTextCompat methods

lunaynx:
## What
Way less ambitious PR. Just marking them as deprecated and a little reordering, that's pretty much it.

+ Marked all TextCompat.formattedTextCompat\* methods as deprecated. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/compat/TextCompat.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/compat/TextCompat.kt
@@ -60,6 +60,7 @@ private enum class FormattedTextSettings {
 
 private data class TextCacheKey(val settings: FormattedTextSettings, val component: Component)
 
+@Deprecated("Use string unless you really need color codes", ReplaceWith("this.string"))
 fun Component.unformattedTextForChatCompat(): String {
     return unformattedTextCache.getOrPut(this) {
         computeUnformattedTextCompat()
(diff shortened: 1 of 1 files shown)


AverageUser125:
Didn’t you want to add
`Component.stripped` extension since hypixel and color codes?

lunaynx:
Yes, that will be a separate PR.

AverageUser125:
Doesn’t it make more sense for the deprecation warning to point at `stripped` instead of `string` though?

lunaynx:
No shit, you don't need to scrutinize everything to death. I can easily just make that PR depend on this one and change it there.

AverageUser125:
Fair enough, sorry for annoying you. 🥺🥲.
Have a good day.

src/main/java/at/hannibal2/skyhanni/utils/compat/TextCompat.kt
```diff
@@ -73,17 +74,12 @@ private fun Component.computeUnformattedTextCompat(): String {
     return (this.contents as? PlainTextContents)?.text().orEmpty()
 }
 
+@Deprecated("Use string unless you really need color codes", ReplaceWith("this.string"))
```
hannibal002: replace with: this is not ideal. if someone presses the replace with button in intellij they break behaviour. just leave the replace with field empty.

src/main/java/at/hannibal2/skyhanni/utils/compat/TextCompat.kt
```diff
@@ -73,17 +74,12 @@ private fun Component.computeUnformattedTextCompat(): String {
...
 fun Component.unformattedTextCompat(): String =
     iterator().joinToString(separator = "") { it.unformattedTextForChatCompat() }
 
-// has to be a separate function for pattern mappings
```
hannibal002, line 79: why is this comment gone?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/compat/TextCompat.kt
lunaynx: It's no longer true on modern per @NopoTheGamer, now we're just keeping them around because lots of code uses them.

src/main/java/at/hannibal2/skyhanni/utils/compat/TextCompat.kt
lunaynx: Yeah, fair.

hannibal002 (approved):

PR #5956, opened 2026-06-28
Improvement: Hide gui during F3

AverageUser125:
## What
Adds a toggle to hide all UI elements while F3 (debug menu) is active.

I got no clue what is a good way to handle this breaking `SkyhanniDebugAndTests` techincally, since that relies on debug menu being active. but I guess whoever is using it can disable the setting.

I also did really minor cleanup on the way.
like using `org.joml.Vector2i` instead of a custom made `Vector2i`,
And moving `hideGui` to `MinecraftCompat`.

I worry that this could lower performance, as this check is on a very hot path.
But it's probably fine.

## Changelog Improvements
+ Added option to hide all SkyHanni GUI elements while the F3 menu is open.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/minecraftevents/RenderEvents.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/minecraftevents/RenderEvents.kt
@@ -47,7 +47,7 @@ object RenderEvents {
 
     private fun postGui(context: GuiGraphicsExtractor, tick: DeltaTracker) {
         if (MinecraftCompat.hideGui) return
-        if (MinecraftCompat.showDebugHud && config.hideGuiInDebugMenu) return
+        if (config.hideGuiInDebugMenu && MinecraftCompat.showDebugHud) return
         RenderData.postRenderOverlay(context)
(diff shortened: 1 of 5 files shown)


src/main/java/at/hannibal2/skyhanni/api/minecraftevents/RenderEvents.kt
```diff
@@ -44,7 +46,8 @@ object RenderEvents {
     }
 
     private fun postGui(context: GuiGraphicsExtractor, tick: DeltaTracker) {
-        if (Minecraft.getInstance().options.hideGui) return
+        if (MinecraftCompat.hideGui) return
+        if (MinecraftCompat.showDebugHud && config.hideGuiInDebugMenu) return
```
hannibal002: please change the order of the two boolean checks: always first check the fast config state, then the "relatively" heavy MinecraftCompat.showDebugHud call

hannibal002 (changes requested):
everything else looks fine

hannibal002 (approved):

PR #5959, opened 2026-06-29
Backend: Bump kotlin to 2.4.0

AverageUser125:
## What
If the workflows gonna be 2.4.0 might aswell make the mod itself too.
https://kotlinlang.org/docs/whatsnew24.html#gradle

the `-Xintrinsic-const-evaluation` makes it it do stuff like `Enum.name` at compile time which seems like a nice bonus.
Could also enable `-Xcollection-literals` which would let you do stuff like:
```kotlin
    val fruit = ["apple", "banana", "cherry"]
    val identityMatrix: DoubleMatrix = [
        [1.0, 0.0, 0.0],
        [0.0, 1.0, 0.0],
        [0.0, 0.0, 1.0],
    ]
```
But that feels a bit more weird.

Also enabled `-Xcontext-sensitive-resolution` which lets you NOT specify the enum name in a `when` statement.

+ Bumped Kotlin version to 2.4.0. - AverageUser125


lunaynx:
You should also bump fabric-language-kotlin to 1.13.12+kotlin.2.4.0.

lunaynx:
Also maybe bump the language version from 2.2 to 2.4 (would allow removing `-Xnested-type-aliases`).

AverageUser125:
```
Provided Metadata instance has version 2.4.0, while maximum supported version is 2.3.0. To support newer versions, update the kotlin-metadata-jvm library.
```
I got no clue what is this, and updating the `shadow` package did not help.

lunaynx:
It might be Detekt that you need to update.

lunaynx:
Not a big deal if we can't get language version 2.4 though, no rush for that.

AverageUser125:
> It might be Detekt that you need to update.

```org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':1.21.11:shadowJar'.```
Maybe also detekt, but shadow most certainly.

But yeah you're right it's probably fine.

**Comment by lunaynx** [comment]
IIRC language version 2.3 should already let us get rid of `-Xnested-type-aliases`.

**Comment by AverageUser125** [comment]
> IIRC language version 2.3 should already let us get rid of `-Xnested-type-aliases`.

good to know.

**Comment by hannibal002** [comment]
> Also enabled -Xcontext-sensitive-resolution which lets you NOT specify the enum name in a when statement.

please explain further. also, why?

**Comment by lunaynx** [comment]
We've already discussed this before as something that would be nice to enable, I just never made the PR to enable it because I was going to refactor existing code to take advantage of it, but it's probably better to take it slowly.

Basically, this lets the Kotlin compiler infer the enum name from the context where unambiguous. So instead of having to write `MySuperLongEnumName.FOO` every single time or using an import alias / importing each enum entry individually, you can just write `FOO`. IntelliJ and Detekt are already aware and fully handle this last I checked. At worst, IntelliJ might sometimes (not always) still add imports that are made redundant by this if that hasn't been fixed since.

File: gradle/libs.versions.toml [hunk 1]
```diff
@@ -13,7 +13,7 @@ methanol = "1.8.3"
 
 shadow = "9.3.1"
 loom = "1.15-SNAPSHOT"
-kotlin = "2.3.20"
+kotlin = "2.4.0"
 ksp = "2.3.6"
```
hannibal002: bump to 2.3.9

hannibal002 (changes requested):

gradle/libs.versions.toml
AverageUser125: Sure, but any particular reason not to do 2.4.0?

also isn’t 2.3.9 less then the current 2.3.20?

lunaynx: 2.3.9 appears to be the latest KSP version, I don't see a 2.4.0. Pretty sure they stopped matching it to the Kotlin version.

https://github.com/google/ksp/releases

AverageUser125: Oh, you were talking about the ksp version. That makes a lot more sense

AverageUser125:
KSP 2.3.9 (or any version really) with kotlin 2.4.0 has a bug where it fails to detect all functions inside a module if the module is marked as internal.

I doubt there are any objects annotated with SkyhanniModule which are also internal.
But just seemed worth mentioning.

AverageUser125:
Or maybe I misunderstood something and it’s also for functions. Here is the issue:
https://github.com/google/ksp/issues/2964

I am unsure if it affects this PR since I made the language version 2.3 and not 2.4.

AverageUser125:
I mean, this PR doesn’t break anything. Just means doing internal modules/functions annotated won’t work.
Which isn’t done anyways. I hope.

hannibal002 (approved):
`-Xcontext-sensitive-resolution` is fine

PR #5960, opened 2026-06-29
Fix: Fix HoppityCollectionStats rabbits found regex

AverageUser125:
## What
Made the regex colorless, also added a failcase for this error in the future.
I tested it in game, and it works.
(Even with debugger to ensure the line was hit and got the correct number)

I ain't sure about the changelog message, but eh.

## Changelog Fixes
+ Fixed Hoppity rabbit count not being detected in the collection menu. - AverageUser125

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityCollectionStats.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityCollectionStats.kt
@@ -645,15 +645,15 @@ object HoppityCollectionStats {
         addResidentRabbitsInformationToHud(newList)
         addHotspotRabbitsInformationToHud(newList)
 
-        newList.addString("")
-
         val loggedRabbitCount = loggedRabbits.size
         val foundRabbitCount = getFoundRabbitsFromHypixel(event) ?: run {
+            newList.addString("")
             newList.addString("§cError: Failed to read rabbit progress from the inventory.")
             return newList
         }
 
         if (loggedRabbitCount < foundRabbitCount) {
+            newList.addString("")
             newList.add(
                 Renderable.wrappedText(
                     "§cPlease Scroll through \n" + "§call pages!",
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityCollectionStats.kt
```diff
@@ -102,11 +103,11 @@ object HoppityCollectionStats {
...
 
     /**
-     * REGEX-TEST: §2§l§m                      §f§l§m   §r §e395§6/§e457
+     * WRAPPED-REGEX-TEST: "                          395/457"
      */
     private val rabbitsFoundPattern by patternGroup.pattern(
-        "rabbits.found",
-        "§.§l§m[ §a-z]+§r §.(?<current>[0-9]+)§./§.(?<total>[0-9]+)",
+        "rabbits.found.colorless",
+        "\\s+(?<current>\\d+)/(?<total>\\d+)",
```
hannibal002, line 110: this regex does not check for the amount of spaces in front of the numbers. so either the regex test is useless or the pattern is broken.

src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityCollectionStats.kt
```diff
@@ -644,11 +645,15 @@ object HoppityCollectionStats {
         addResidentRabbitsInformationToHud(newList)
         addHotspotRabbitsInformationToHud(newList)
 
+        newList.addString("")
```
hannibal002: the empty line being here, and not in the if anymore, means we can have the gui with an empty line at the end. intended?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/event/hoppity/HoppityCollectionStats.kt
AverageUser125: Should I put in the regex the exact amount of spaces then?
I just did "\\s+" for some generic number of spaces.

Also the Regex Test is a real example of an input, not a made up one. So I think it's fine.

lunaynx: I would say no, it's a bad idea to depend on any exact amount or range of spaces unless absolutely necessary. The regex test is fine. Maybe the coffee hasn't hit for Hannibal yet.

hannibal002 (approved):

PR #5965, opened 2026-06-30
Backend: Add Explicit Return Type Detekt Rule

AverageUser125:
## What
Public functions should have explicit return type.
That's it.
Could make it also be property. but honestly just functions is good and has alot less false positives.

exclude_from_changelog

### DIFF
--- a/detekt/src/main/kotlin/style/PublicFunctionMustSpecifyReturnType.kt
+++ b/detekt/src/main/kotlin/style/PublicFunctionMustSpecifyReturnType.kt
@@ -1,0 +1,59 @@
+package style
+
+import SkyHanniRule
+import dev.detekt.api.Config
+import dev.detekt.api.RequiresAnalysisApi
+import org.jetbrains.kotlin.analysis.api.analyze
+import org.jetbrains.kotlin.analysis.api.symbols.KaSymbolVisibility
+import org.jetbrains.kotlin.analysis.decompiler.psi.text.getAllModifierLists
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
+import utils.DetektUtils.hasAnnotation
+
+class PublicFunctionMustSpecifyReturnType(config: Config) :
+    SkyHanniRule(
+        config,
+        "Public functions should have an explicit return type. " +
+                "Inferred return types can easily be changed by mistake which may lead to breaking changes.",
+    ), RequiresAnalysisApi {
+
+    private val ignoredModifiers = listOf(
+        KtTokens.OVERRIDE_KEYWORD,
+        KtTokens.OPERATOR_KEYWORD,
+        KtTokens.PRIVATE_KEYWORD,
+        KtTokens.INTERNAL_KEYWORD,
+        KtTokens.PROTECTED_KEYWORD,
+    )
+
+    override fun visitNamedFunction(function: KtNamedFunction) {
+        function.check()
+        super.visitNamedFunction(function)
+    }
+
+    private fun KtNamedFunction.check() {
+        modifierList?.let {
+            ignoredModifiers.forEach { modifier ->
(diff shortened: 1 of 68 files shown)


detekt/src/main/kotlin/style/PublicFunctionMustSpecifyReturnType.kt
```diff
@@ -0,0 +1,40 @@
...
+        if (this is KtNamedFunction) {
+            if (isLocal || hasBlockBody() || bodyExpression?.text == "Unit") return
+            analyze(this) {
+                if (symbol.returnType.isUnitType) return
+                if (!isPublicApi(symbol)) return
+            }
+        }
+
+        reportIssue(
+            "Library function '${this.nameAsSafeName}' without explicit return type.",
```
lunaynx: We're not really a library so this can be adjusted.
```suggestion
            "Public function '${this.nameAsSafeName}' without explicit return type.",
```

AverageUser125:
okay, this is suspicious.
I expected detekt to fail.

zumbiepig:
I think public properties in enums, classes, and objects should also have explicit types. Also prefer `val list: List<String> = listOf()` over `val list = listOf<String>()`

AverageUser125:
> I think public properties in enums, classes, and objects should also have explicit types. Also prefer `val list: List<String> = listOf()` over `val list = listOf<String>()`

The major reason stopping that is RepoPattern tbh.
But I want to Atleast start with the functions, as this PR already had to make so many exceptions so the diff ain’t huge.

lunaynx:
> I think public properties in enums, classes, and objects should also have explicit types.

In theory a good idea, but this would massively blow up the scope of this PR. There's a _lot_ of public properties that could be private and would currently get flagged by this.

> Also prefer `val list: List<String> = listOf()` over `val list = listOf<String>()`

This really doesn't matter, except as part of satisfying "explicit return type for public properties". Otherwise, it just makes code more verbose. In fact, Kotlin's official style guide uses the latter in its examples (but does not say anything about preferred syntax).

hannibal002:
> I think public properties in enums, classes, and objects should also have explicit types. Also prefer val list: List<String> = listOf() over val list = listOf<String>()

do you also prefer `val b: Boolean = true` over `val b = true`?

AverageUser125:
> > I think public properties in enums, classes, and objects should also have explicit types. Also prefer val list: List = listOf() over val list = listOf()
> 
> do you also prefer `val b: Boolean = true` over `val b = true`?

Personally I prefer it only for data classes that got @Expose

hannibal002:
the reasons i want explicit return types on public functions is that when the impl of the utils class changes, the error does not show up in the feature class but inside the feature class directly, and that when reading the function, the return type is not easily understandable, especially true for lambda/when block heavy logic. for members/properties, this is almost never the case. you see the type in the same line directly, as the type is obvious bc the value is e.g. 3, true, "hello", etc. So this would mostly add bloat/verbosity

zumbiepig:
personally, i prefer val b: Boolean = true

zumbiepig:
> > I think public properties in enums, classes, and objects should also have explicit types.
> 
> In theory a good idea, but this would massively blow up the scope of this PR. There's a _lot_ of public properties that could be private and would currently get flagged by this.
> 
> > Also prefer `val list: List<String> = listOf()` over `val list = listOf<String>()`
> 
> This really doesn't matter, except as part of satisfying "explicit return type for public properties". Otherwise, it just makes code more verbose. In fact, Kotlin's official style guide uses the latter in its examples (but does not say anything about preferred syntax).

I would use `val list: List<String> = listOf()` in the class/`this` object context. and I would use `val list = listOf<String>()` (or just not specify string at all) for local variables.

AverageUser125:
```
e: file:///home/user/Projects/SkyHanni/src/main/java/at/hannibal2/skyhanni/features/event/carnival/CarnivalFruitDigging.kt:126:43 Argument type mismatch: actual type is 'String?', but 'String' was expected.
```
This is a problem with the current beta.
So I am guessing I should not fix this in this PR and wait for a fix in beta to magically appear.

AverageUser125:
Too much drama.

PR #5971, opened 2026-07-01
Fix: Make DeferredItemStack lazy

lunaynx:
## What
Fixes the awful implementation of DeferredItemStack, namely:
- It no longer holds a persistent strong reference to every single ItemStack ever created by SkyHanni. The global instance holder has been removed entirely.
- ComponentsLoadedEvent and any batch building logic has been removed entirely. Instead, components are bound lazily on demand.

I've done a cursory test and the "Loading terrain" slowdown is definitely gone, including with IntelliJ's debugger attached, which would normally amplify the issue and make the freeze closer to 10 seconds than 2 seconds on my M1 Air. No errors in chat. NEU repo loading, item resolution & price info seems to work fine. As a test case, the Golden Fish Timer renderable works fine without flickering or showing missing/fallback textures. I'll likely do more battle testing of this tomorrow on a live instance to be sure.

## Changelog Fixes
+ Fixed SkyHanni causing the game to freeze for multiple seconds on the "Loading Terrain" screen on 26.1. - Luna
    * This would mainly happen the first time you joined a world (e.g. by connecting to Hypixel) after a game restart, but could also happen later.
+ Fixed memory leak and out-of-memory errors caused by the item cache never getting cleared on 26.1. - Luna


src/main/java/at/hannibal2/skyhanni/data/SackApi.kt
```diff
@@ -370,11 +370,6 @@ object SackApi {
         rebuildSackNameLists()
     }
 
-    @HandleEvent(priority = HandleEvent.LOWEST)
-    fun onComponentsLoaded() {
-        rebuildSackNameLists()
-    }
-
     private fun rebuildSackNameLists() {
```
hannibal002, line 374: i asssume this should also get delayed by one tick to be in sync with NeuItems.readAllNeuItems, right?

src/main/java/at/hannibal2/skyhanni/utils/NeuItems.kt
```diff
@@ -13,6 +13,7 @@ import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.test.command.ErrorManager
 import at.hannibal2.skyhanni.utils.ItemPriceUtils.getPriceOrNull
 import at.hannibal2.skyhanni.utils.ItemUtils.getInternalName
+import at.hannibal2.skyhanni.utils.ItemUtils.getRepoItemNameFromJson
```
hannibal002, line 15: unused import import at.hannibal2.skyhanni.utils.SafeItemStack

src/main/java/at/hannibal2/skyhanni/utils/NeuItems.kt
```diff
@@ -88,21 +85,6 @@ object NeuItems {
     fun onNeuRepoReload(event: NeuRepositoryReloadEvent) {
```
hannibal002: missing sweeping change:
Parameter "event" is never used

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/NeuItems.kt
lunaynx: Your IntelliJ is drunk, the file mentions SafeItemStack 8 times outside the import.

src/main/java/at/hannibal2/skyhanni/data/SackApi.kt
lunaynx: Probably, shouldn't hurt.

src/main/java/at/hannibal2/skyhanni/utils/NeuItems.kt
hannibal002: I trust you but I wanna investigate if I can replicate the gray format

hannibal002: my intellij is lying

hannibal002 (approved):

PR #5973, opened 2026-07-01
Backend: Slightly better contributor mention regex

AverageUser125:
## What
Rain had someone write:
"who is the skyhannu contrubiter?"
So, I made the regex slightly more permissive.

exclude_from_changelog


lunaynx:
A better idea would be something like checking the Levenshtein distance of the words to "skyhanni" and "contributor" and setting some reasonable threshold.

AverageUser125:
> A better idea would be something like checking the Levenshtein distance of the words to "skyhanni" and "contributor" and setting some reasonable threshold.

does there exist some already existing function in skyhanni for calculating the Levenshtein distance?
Also, this is already good enough IMO.

lunaynx:
I don't think there is, but it should be pretty simple to implement.

AverageUser125:
> I don't think there is, but it should be pretty simple to implement.

let me be lazy.

CalMWolfs (approved):

PR #5976, opened 2026-07-02
Fix + Improvement: Fruit Digging Solver improvements

lunaynx:
## What
Improved Fruit Digging Solver logic so that it doesn't treat Bomb & Watermelon explosions as player-dug fruits, added an error in case we go above max digs again, and made it stop early on the last round because Hypixel has a delay of several seconds before actually ending the game. Also made one of the patterns more lenient by not depending on a fixed amount of spaces.

TODO:
- [ ] Change round detection to be scoreboard-based to fix remaining reliability issues
- [ ] Fix entities rendering over the in-world solver text (may end up delaying this and fixing it in a separate PR)

Waiting for another Carnival:

waiting_on_hypixel_alpha

## Changelog Improvements
+ Fruit Digging Solver will now warn you if you're trying to dig a square with the wrong shovel mode. - Luna
+ Fruit Digging Solver will now stop making suggestions immediately once the last round has been reached. - Luna

## Changelog Fixes
+ Fixed Fruit Digging Solver treating Bomb and Watermelon explosions as fruits dug by the player. - Luna


lunaynx:
I'm adding more fixes and looking into a misdetection issue either still present or caused by this.

Vivo-Y12:
@lunaynx  foxy is arrived time to test make more better, latest beta fruit feature is  not worth it

lunaynx:
Will do.

Vivo-Y12:
> Will do.

when?

PR #5982, opened 2026-07-04
Fixe: Test

hannibal002:
## Dependencies
- #5998
- #6002

would be ideal if this would not get merged

## Changelog Fix
+ Fixed an super important bug. -hannibal2

exclude_from_changelog


AverageUser125:
a 9 hours test.

AverageUser125:
Is it a good idea to treat build warnings as errors?
Or Atleast show build warnings in the review message?

hannibal002:
> Is it a good idea to treat build warnings as errors?

yes, thats exactly what i am trying to do right now.

AverageUser125:
Surely it’s as simple as doing
```gradle 
compilerOptions {
     allWarningsAsErrors.set(true)
}
```

hannibal002:
> Surely it’s as simple as doing

if we want spam, yes. but we want no spam. we want logic

AverageUser125:
What do you mean? #5940 got merged. So there should not be any spam warnings.

what kinds of spam are you talking about?

hannibal002:
> Is it a good idea to treat build warnings as errors?

i have no idea. why would we do that?

also, this is not what i am trying to do right now.


i just try to make the workflow send a comment in the pr with the build fail stack trace

AverageUser125:
I mean, there should ideally not be any build warnings at all.
Since stuff like redundant Elvis operator, or redundant null check, don’t seem that good an idea to keep.

hannibal002:
> I mean, there should ideally not be any build warnings at all.

i dont think this converstaion is productive. lets move over to discord.

AverageUser125:
Could you test for the case of a missing space before the author name? Like doing:
`-Author` instead of `- Author`.
Since it’s quite a common mistake

hannibal002:
>  Since it’s quite a common mistake

this will show the "missing author" reply

AverageUser125:
> > Since it’s quite a common mistake
> 
> this will show the "missing author" reply

Alright, cool.

PR #5983, opened 2026-07-04
Backend: Bump language version to 2.4

AverageUser125:
## What
I figured out why shadowJar failed.

exclude_from_changelog


AverageUser125:
Should I also bump the minimum version for the fabric-kotlin mod in fabric.json?

lunaynx:
Probably, since we're actively setting language version 2.4 now I'd expect older versions to cause runtime breakage eventually.

CalMWolfs (approved):

PR #5985, opened 2026-07-04
Backend: Fix all current import orders

AverageUser125:
## What
Fixes all the wrong order in imports.

This PR does not enable the detekt CustomImportRule cause it doesn't work, and I can't seem to fix it in a way I like.

exclude_from_changelog


src/main/java/at/hannibal2/skyhanni/api/minecraftevents/RenderEvents.kt
```diff
@@ -10,15 +10,14 @@ import at.hannibal2.skyhanni.utils.compat.MinecraftCompat
 import at.hannibal2.skyhanni.utils.render.SkyHanniRoundedShapeRenderManager
 import at.hannibal2.skyhanni.utils.render.item.SkyHanniItemRenderCoordinator
 import at.hannibal2.skyhanni.utils.render.item.SkyHanniPipCoordinatorRenderer
+//~ if < 26.1 'PictureInPictureRendererRegistry' -> 'SpecialGuiElementRegistry'
```
hannibal002: ehhhh u sure?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/api/minecraftevents/RenderEvents.kt
AverageUser125: no, which is why this is a draft lol.

AverageUser125:
tommorow problem.

AverageUser125:
I can't rework the detekt rule code in a way I like it.
So I just made this PR only fix the imports.

The fix the `CustomImportRule` PR can be some other time, by someone smarter then me.

lunaynx:
I don't think `kotlinx` imports should be moved to be among the normal imports. This looks like more of an oversight from whoever originally made the rule. It would make more sense for them to go after `kotlin` imports.

AverageUser125:
> I don't think `kotlinx` imports should be moved to be among the normal imports. This looks like more of an oversight from whoever originally made the rule. It would make more sense for them to go after `kotlin` imports.

but.... sorting in lexicographic order.

lunaynx:
I'm not sure if I'm a big fan of specifically moving aliases to the end either, though that *is* explicitly called out in the current rule, so it should be discussed before changing it.

AverageUser125:
> I'm not sure if I'm a big fan of specifically moving aliases to the end either, though that _is_ explicitly called out in the current rule, so it should be discussed before changing it.

you speak magic words, I got no clue what you are talking about.
So just... what do you want me to change?

lunaynx:
> > I don't think `kotlinx` imports should be moved to be among the normal imports. This looks like more of an oversight from whoever originally made the rule. It would make more sense for them to go after `kotlin` imports.
> 
> but.... sorting in lexicographic order.

What even is your argument here? I'm saying that `java` and `kotlin` are already special-cased to be at the end. Why should `kotlinx` be in the middle instead of with `kotlin`?

lunaynx:
Technically the current order is correct if you strictly follow the existing rule, but I'd move `kotlinx` after `kotlin` and adjust the wording of the rule. The aliases one, just leave it as-is for now until we discuss it.

AverageUser125:
I give up trying to fix the rule itself to match what it should be.
So I will just move kotlinx to the end and keep the rest as you said.

hannibal002 (approved):

PR #5987, opened 2026-07-05
Fix: Century Party Invitation highlighter not working

CalMWolfs:
## What
Made the code work as we were pulling the wrong entity name
Made it actually highlight all missing colors and not just the first one listed
Made it work for level 400+ as well

Also added search tags to gift clean display

## Changelog Improvements
+ Improved Gift Clean Display to also appear when searching for Century Cake Slice in the config.

## Changelog Fixes
+ Fixed Century Party Invitation Highlighter not working. - CalMWolfs


legentpc:
Gift Clean Display and century cake slice both are different feature?

MiscConfig.kt#L130 I think the 'Century Cake Slice' tag should be here.

hannibal002:
> Gift Clean Display and century cake slice both are different feature?
> 
> MiscConfig.kt#L130 I think the 'Century Cake Slice' tag should be here.

good catch! [fixed it](https://github.com/hannibal002/SkyHanni/commit/4fbd6fc8888f3b1a7a17a0eddb3e57d7d5c2eaac).

PR #5990, opened 2026-07-05
Fix: Pet storage selected pet skin parsing

Akinsoft:
## What

Fixes a Pet Storage crash when reading selected pet lore for pets without alternate skin data.

Missing nullable matcher groups are now handled safely, matching the existing regex helper behavior. Added a regression test, bazinga!

## Changelog Fixes
+ Fixed error when opening pets menu when there are pets without alternate skin data. - akinsoft


AverageUser125:
@CalMWolfs There seems to be a problem with AutoPet rule detection, so hold off from making a beta.

PR #5994, opened 2026-07-05
Fix: Revert "Fix: Sodium-Extra incompatiblity"

AverageUser125:
Reverts hannibal002/SkyHanni#5991

this could make skyhanni incompatible with other mods, so unsure about this.
Idk if this even causes any real incompatibilities this is just a suspicion.

Also if this gets merged the changelog from the other PR needs to get removed.

exclude_from_changelog


CalMWolfs:
Is this still needed, I haven't seen anyone complain yet

AverageUser125:
Honestly, no clue.
It is could hypothetically cause problems, but yeah none that I have seen.

PR #6004, opened 2026-07-06
Fix: Pest spawn chat message shown twice when using Compact chat format

RemainingDelta:
## What
Pest spawn messages in the Garden appear twice when Chat Message Format is set to Compact.

`PestSpawn.onChat` was listening on `SystemMessageEvent.Allow`. Since `SkyHanniChatEvent.Allow` is a subclass of it, the event system triggers the handler once when `SkyHanniChatEvent.Allow` fires, then again when `PlayerChatManager` posts a separate `SystemMessageEvent.Allow` for the same message. Two triggers means the spawn message gets sent twice.

This was introduced in #5839, which switched the event from `SkyHanniChatEvent.Allow` to `SystemMessageEvent.Allow` as extra protection against a false positive, on top of tightening the regex from `.*!` to `^\w+!`. I tested and confirmed the regex change alone is enough to prevent the false positive, so the event switch was unnecessary and caused this bug.

Fix: switched `PestSpawn.onChat` back to `SkyHanniChatEvent.Allow`, which fires once per message. The `^\w+!` regex stays, and still prevents the original false positive.

Bug identified in:
- https://discord.com/channels/997079228510117908/1518858917835440209

## Changelog Fixes
+ Fixed the Pest Spawn chat message appearing twice when Chat Message Format is set to Compact. - RemainingDelta

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestSpawn.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestSpawn.kt
@@ -3,8 +3,8 @@ package at.hannibal2.skyhanni.features.garden.pests
 import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.config.features.garden.pests.PestSpawnConfig
 import at.hannibal2.skyhanni.data.IslandType
-import at.hannibal2.skyhanni.data.hypixel.chat.event.SystemMessageEvent
 import at.hannibal2.skyhanni.data.title.TitleManager
+import at.hannibal2.skyhanni.events.chat.SkyHanniChatEvent
 import at.hannibal2.skyhanni.events.garden.pests.PestSpawnEvent
 import at.hannibal2.skyhanni.features.garden.pests.PestApi.lastPestSpawnTime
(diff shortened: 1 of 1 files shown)


AverageUser125:
1. Are you certain player messages won’t be detected? Like guild messages
2. Is It a good idea to add like a 1 second period to ensure that it will ignore any other chat messages, or am I over engineering ?

RemainingDelta:
> 1. Are you certain player messages won’t be detected? Like guild messages

The regex is checking for some word followed by "!" (e.g. "GROSS!" or "YUCK!"). 
Cases that I can think of:
- Guild messages start with "GUILD >" (Regex will not pick it up)
- Party messages start with "PARTY >" (Regex will not pick it up)
- All chat messages start with level like "[371]" (Regex will not pick up)

There might be more cases that I can't think of but I don't think there is any case that start with a word followed up by "!".

> 2. Is It a good idea to add like a 1 second period to ensure that it will ignore any other chat messages, or am I over engineering ?

I think you're over engineering.

AverageUser125:
Frick, the future changelog is gonna be slightly broken

hannibal002:
> Frick, the future changelog is gonna be slightly broken

luckily we can still edit it post merge

PR #6008, opened 2026-07-07
Improvement: Add reminders hud display

RemainingDelta:
## What
Adds an on-screen HUD display for `/shremind` reminders. Instead of only being notified via chat when a timer fires, active reminders are now visible on screen with their time remaining, updating every second.
 
The HUD is toggleable under **Misc → Reminders → Show Reminders HUD** and is repositionable via `/shmovegui`. 

<details>
<summary>Images</summary>

New config option:   

image

New HUD:   

image

</details>

Requested in:
- https://discord.com/channels/997079228510117908/1524016187007176704

## Changelog Improvements
+ Added an on-screen display for active reminders. - RemainingDelta

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/misc/RemindersConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/misc/RemindersConfig.kt
@@ -1,6 +1,5 @@
 package at.hannibal2.skyhanni.config.features.misc
 
-import at.hannibal2.skyhanni.config.FeatureToggle
 import at.hannibal2.skyhanni.config.core.config.Position
 import com.google.gson.annotations.Expose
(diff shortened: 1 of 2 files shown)


AverageUser125:
I am surprised that “formatToHud” is not applicable/already created somewhere else.

AverageUser125:
Also, wouldn’t the “formatToHud” make more sense to be in `SimpleTimeMark` or I guess just in `TimeUtils`?

RemainingDelta:
> I am surprised that “formatToHud” is not applicable/already created somewhere else.

Hmm, that's good point let me check. Surely there is one. 

Edit: There is `.format()` from `TimeUtils.kt`, which I can use.

CalMWolfs (approved):

PR #6011, opened 2026-07-08
Fix: MaxwellApi detecting Air as a tuning

AverageUser125:
## What
The stack name could have been air, and it would have failed.
Also made it use clean name for the item stack and moved the pattern to repo pattern while I was here.

## Changelog Fixes
+ Fixed Maxwell tuning stats sometimes not being detected correctly. - AverageUser125

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/MaxwellApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/MaxwellApi.kt
@@ -231,9 +231,8 @@ object MaxwellApi {
     private fun loadThaumaturgyTuningsFromTuning(inventoryItems: Map<Int, SafeItemStack>) {
         val map = mutableListOf<ThaumaturgyPowerTuning>()
         for (stack in inventoryItems.values) {
-            stack.takeUnlessEmpty() ?: continue
+            val stackName = stack.takeUnlessEmpty()?.cleanName() ?: continue
             for (line in stack.getLore()) {
-                val stackName = stack.cleanName()
                 statsTuningDataPattern.readTuningFromLine(line)?.let {
                     it.name = tuningNamePattern.matchMatcher(stackName) {
                         group("name")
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
```diff
@@ -1118,5 +1118,5 @@ object ItemUtils {
         }
     }
 
-    fun SafeItemStack.takeUnlessEmpty(): SafeItemStack? = takeUnless { it.isEmpty() }
+    fun SafeItemStack.takeUnlessEmpty(): SafeItemStack? = takeUnless { it.isEmpty || it.item == Items.AIR }
```
lunaynx: Are you sure this is necessary or actually fixes the issue? `ItemStack.isEmpty` already checks for essentially the same thing:
- 1.21.11: https://mcsrc.dev/1/1.21.11/net/minecraft/world/item/ItemStack#L307-309
- 26.1.2: https://mcsrc.dev/1/26.1.2/net/minecraft/world/item/ItemStack#L323-325

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
AverageUser125: then I am really confused what could have possible caused it to be "Air"...
Well, the pattern thing is still needed so imma keep that atleast.

But imma revert the change you mentioned

Rainymei: it's shift clicking the tuning slot. sends the item to inventory and removes it from thaumaturgy gui

AverageUser125: then I hope the change I did with the `takeUnlessEmpty` fixes that.

src/main/java/at/hannibal2/skyhanni/data/MaxwellApi.kt
```diff
@@ -214,13 +231,15 @@ object MaxwellApi {
     private fun loadThaumaturgyTuningsFromTuning(inventoryItems: Map<Int, SafeItemStack>) {
         val map = mutableListOf<ThaumaturgyPowerTuning>()
         for (stack in inventoryItems.values) {
+            stack.takeUnlessEmpty() ?: continue
             for (line in stack.getLore()) {
+                val stackName = stack.cleanName()
```
hannibal002: stack name can go outside of the get lore for loop

hannibal002 (changes requested):

AverageUser125:
Well, at least I managed to test this before the server went into maintenance, so I know this PR works.

hannibal002 (approved):

PR #6012, opened 2026-07-08
Fix: colorless HOTX

AverageUser125:
## What
Can't currently test it since hypixel is down.
But please double check I didn't miss any usage.

IS NOT TESTED, LIKE AT ALL.

## Changelog Fixes
+ Fixed Skymall / Lottery perks giving an error. - AverageUser125

+ Made Hotx features colorless. - AverageUser125


AverageUser125:
well, I need to go urgently, it seems so I won't be able to finish this.
So someone can just copy and continue this ig.

AverageUser125:
I am not changing the symbol, it should be replaced in the entire skyhanni project all at once.

PR #6018, opened 2026-07-08
Fix: Fix dungeon location being broken

AverageUser125:
## What
Would it be a better solution to follow the TODO:
```
        // TODO: move this under inDungeon check when we use Hypixel's ModAPI for island detection
```
probably, but I decided that Imma follow:
```
    // TODO repo patterns
```

## Changelog Fixes
+ Fixed dungeon floor detection being broken due to new update. - AverageUser125

+ Made dungeon floor detection use repo patterns. - AverageUser125

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonApi.kt
@@ -42,15 +42,55 @@ import net.minecraft.world.level.block.Blocks
 @Suppress("MemberVisibilityCanBePrivate")
 @SkyHanniModule
 object DungeonApi {
+    private val patternGroup = RepoPattern.group("dungeon")
 
-    // TODO repo patterns
-    private val floorPattern = " §7⏣ §cThe Catacombs §7\\((?<floor>.*)\\)".toPattern()
-    private val uniqueClassBonus = "^Your ([A-Za-z]+) stats are doubled because you are the only player using this class!$".toRegex()
+    /**
+     * WRAPPED-REGEX-TEST: " ⏣ The Catacombs (F7)"
+     */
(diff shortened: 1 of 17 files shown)


Fazfoxy:
> Would it be a better solution to follow the TODO:

Floor detection cannot be done under mod API, the detection of any dungeon could be though

AverageUser125:
> > Would it be a better solution to follow the TODO:
> 
> Floor detection cannot be done under mod API, the detection of any dungeon could be though

so... do I remove that TODO, or not?

src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonApi.kt
```diff
@@ -275,7 +315,7 @@ object DungeonApi {
             started = true
             DungeonStartEvent(floor).post()
         }
-        if (event.cleanMessage.matches(uniqueClassBonus)) {
+        uniqueClassBonus.matches(event.cleanMessage).let {
```
hannibal002: let checks nullability, not a boolean. you need an if for checking a boolean. or use .takeIf {it }. but a if is enough here

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/dungeon/DungeonApi.kt
AverageUser125: my bad.

hannibal002 (approved):

Fazfoxy:
> so... do I remove that TODO, or not?

actually maybe I misunderstood the TODO when I last looked at this as well, It means that an any dungeon detection should be fast enough with modAPI that it can have a dungeon check? But the TODO can stay I think.

PR #6032, opened 2026-07-09
Backend: Bump Fabric loader to 0.19.3

AverageUser125:
## What
This is only for the builtin intelij running, so it's not a user facing changelog.

exclude_from_changelog


CalMWolfs (changes requested):
I dont think its a good idea to change the fabric loader / fabric api version we compile with and not change the version the mod needs to run as this is just waiting for some sort of crash to happen without warning the user they need to update dependencies

lunaynx:
Can't wait for all the support messages asking how to update it... at least we have a tag for it now!

AverageUser125:
Good point, I bumped the fabric loader requirement.
And the fabric API is bumped “automatically”.

PR #6033, opened 2026-07-09
Backend: Fix log files not being grouped properly due to Lazy fullFormat

Fazfoxy:
## What
This is maybe the stupidest fix but it works somehow to make the log files *actually* stick together

+ Changed fullFormat into the companion object of SkyHanniLogger to fix log folder splitting. - Fazfoxy


AverageUser125:
Psss, this PR looks simple.
Also I should probably stop doing this but it’s hella funny.
So this is the last time lol.

CalMWolfs (approved):

PR #6036, opened 2026-07-09
Fix: wardrobe not working after pressing button

AverageUser125:
## What
It currently sets currentPage to null always, even if you cycle inside the wardrobe menu.
Also fixed the equipment thingy while I was here.

Should be quite resilient.

## Changelog Improvements
+ Added equipment wardrobe value estimation. - AverageUser125

## Changelog Fixes
+ Fixed custom wardrobe freezing after the first click. - AverageUser125


AverageUser125:
Psss, me next

CalMWolfs (approved):
worked for me

PR #6037, opened 2026-07-09
Backend: particle unsafe player distance usage

AverageUser125:
## Dependencies
- #5887

## What
Fixes ParticleEvent giving error when the world is loading.
Which happens very rarely, but still. just incase.

+ Fixed particle events being fired while the player hadn't finished loading. - Avrg


hannibal002 (changes requested):
1. i dont think we need to check if the player is null in the listeners, we can just not fire the event in the first place when the player is null.

2. the phrase "Fixed not ignoring particles while the world is still loading" describes not what the user facing bug was. if there was none, this is a technical details bug fix.

3. this is confusing:
  - What section: "gives error when leaving the world"
  - Changelog: "while the world is still loading"

hannibal002 (approved):

PR #6038, opened 2026-07-09
Fix: Incorrect grand feast detection when Finnegan is minister

RemainingDelta:
## What
Fixes Harvest Feast incorrectly assuming a Grand Feast is active when Finnegan is minister with Blooming Business.
The check was using `it.perks` (all possible perks for the candidate) instead of `it.activePerks` (only the currently active perk), causing the Grand Feast perk to always be detected whenever Finnegan holds any position.
  
This resulted in SkyHanni spamming "Harvest feast data is not yet available" every 10 minutes.

Hotfix in REPO to silence the error:
- https://github.com/hannibal002/SkyHanni-REPO/pull/782

## Changelog Fixes
+ Fixed Harvest Feast incorrectly detecting a Grand Feast when Finnegan is minister. - RemainingDelta


CalMWolfs (approved):

f4bii:
nice review should have also change mayorGrandFeast to check for active :pleading_face:

PR #6040, opened 2026-07-09
Fix: Grand Feast detection v2 + rename perks to allPerks

lunaynx:
## What
Follow-up fix to #6038. While I'm at it, I just rewrote the code to be perk-centered rather than mayor/minister-centered, as that is the preferred way to check (we may get Grand Feast through e.g. Aura v2 for all we know, though Jerry is now ruled out). I've also renamed `ElectionCandidate.perks` to `allPerks` to reduce future mistakes.

## Changelog Fixes
+ Fixed Harvest Feast incorrectly detecting a Grand Feast when Finnegan is mayor. - Luna

+ Renamed ElectionCandidate.perks to allPerks. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/ElectionApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/ElectionApi.kt
@@ -250,7 +250,7 @@ object ElectionApi {
             if (lastMayor?.name != currentMayorName) {
                 Perk.resetPerks()
                 val oldMayor = currentMayor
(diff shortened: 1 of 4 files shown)


CalMWolfs (approved):

lunaynx:
I added a quick follow-up guard, gotta enable auto-merge again.

lunaynx:
Surely third time's the charm...

lunaynx:
Close enough. It built, but Detekt went "nuh uh".

PR #6047, opened 2026-07-10
Backend: WardrobeApi be abstract

f4bii:
## What
Refactored Wardrobe to be abstract for Armor/Equipment

+ Made WardrobeApi be abstract so it has different storage. - FabiHBBBT


src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/EstimatedWardrobePrice.kt
```diff
@@ -20,32 +20,39 @@ object EstimatedWardrobePrice {
...
+            tooltip.add(index++, "")
         } catch (e: IndexOutOfBoundsException) {
             ErrorManager.logErrorStateWithData(
                 "Can not show Estimated Wardrobe Price",
                 "IndexOutOfBoundsException while trying to add the estimated wardrobe price line to the tooltip",
+                "index" to index,
                 "lore" to lore,
             )
         }
+        tooltip.addAll(index, lore)
```
AverageUser125: For some reason, in equipment wardrobe. this `addAll` randomly creates an `IndexOutOfBoundException`.
With it being size 2 and the index being 4.
Which is why I changed it to just add to the end of the tooltip.

If your plan was to override "Click to equip!" text then do some regex and find the line with a fallback to put at the end if it's not found.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/WardrobeApi.kt
```diff
@@ -23,62 +19,52 @@ import at.hannibal2.skyhanni.utils.compat.DyeCompat.Companion.isDye
...
+        const val MAX_PAGES = 3
 
-    /**
-     * REGEX-TEST: (1/2) Equipment Sets
-     */
-    private val equipmentInventoryPattern by patternGroup.pattern(
-        "inventory.name.equipment",
-        "\\((?<currentPage>\\d+)/\\d+\\) Equipment Sets"
-    )
+        internal fun emptyArmor(): List<SafeItemStack?> = listOf(null, null, null, null)
```
AverageUser125: rename to "emptyItems" at this point I think.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/WardrobeApi.kt
```diff
@@ -196,25 +161,18 @@ object WardrobeApi {
...
-            }
+            inThisWardrobe = inventoryPattern.matches(inventoryName)
+            if (!inThisWardrobe) currentPage = null
         }
     }
 
-    @HandleEvent
-    fun onDebugDataCollect(event: DebugDataCollectEvent) {
-        event.title("Wardrobe")
+    protected fun handleDebugDataCollect(event: DebugDataCollectEvent) {
```
AverageUser125, line 175:
```
setOf("Helmet", "Chestplate", "Leggings", "Boots")
```
it should not be those names for the equipment wardrobe.

src/main/java/at/hannibal2/skyhanni/features/inventory/EquipmentApi.kt
```diff
@@ -51,7 +51,7 @@ object EquipmentApi {
 
     fun getAll(): List<SafeItemStack> = equipment?.filterNotNull() ?: emptyList()
 
-    private fun setEquipment(slot: EquipmentSlot, itemStack: SafeItemStack?) {
```
AverageUser125, line 54: make an event instead of just making the private setter public.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/WardrobeApi.kt
```diff
@@ -23,62 +19,52 @@ import at.hannibal2.skyhanni.utils.compat.DyeCompat.Companion.isDye
...
+import java.util.regex.Pattern
 import kotlin.time.Duration.Companion.milliseconds
 
-@SkyHanniModule
-object WardrobeApi {
+abstract class WardrobeApi {
 
-    val storage get() = ProfileStorageData.profileSpecific?.wardrobe
+    companion object {
+        private val patternGroup = RepoPattern.group("inventory.wardrobe")
```
AverageUser125: I think it would be a good idea to make this `patternGroup` be public/protected and then let the class children access it instead of duplicating it.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/WardrobeApi.kt
```diff
@@ -160,20 +134,11 @@ object WardrobeApi {
...
-            } != null
-        ) {
-            return true
-        }
-
-        return equipmentInventoryPattern.matchMatcher(inventoryName) {
-            inEquipmentWardrobe = true
+    private fun checkInventory(inventoryName: String): Boolean =
+        inventoryPattern.matchMatcher(inventoryName) {
+            inThisWardrobe = true
```
AverageUser125, line 140: shouldn't this set it to false if it's not in the inventory?
(unsure)

src/main/java/at/hannibal2/skyhanni/features/inventory/EquipmentApi.kt
AverageUser125: Maybe make an enum for the wardrobe types and then:
`WardrobeUpdateEvent(val type: WardrobeType, val items: List<SafeItemStack?>`

or just make seperate event for each wardrobe type, your choice.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/WardrobeApi.kt
f4bii: Nah there are other places like inventorycloseevent that will make it false

src/main/java/at/hannibal2/skyhanni/features/inventory/EquipmentApi.kt
AverageUser125: I am mostly suggesting this since the loadout system will later also utilize this event.

f4bii: There is already that event. Should i replace all calls to fire this event instead and make this a eventlistener?

AverageUser125: Uh, good question.
I didn’t see that this fires an event already lol.

AverageUser125: I guess keep this as is? Honestly now you got me conflicted.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/WardrobeApi.kt
```diff
@@ -202,7 +203,7 @@ abstract class WardrobeApi {
 
     class WardrobeData(
         @Expose val id: Int,
-        @Expose var armor: List<SafeItemStack?>,
+        @Expose var items: List<SafeItemStack?>,
```
AverageUser125: This will break backwards compatibility and may cause problems.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/EquipmentWardrobeApi.kt
```diff
@@ -36,7 +38,19 @@ object EquipmentWardrobeApi : WardrobeApi() {
...
-    fun onInventoryClose(event: InventoryCloseEvent) = handleInventoryClose()
+    fun onInventoryClose(event: InventoryCloseEvent) {
+        handleInventoryClose()
+
+        val currentEquipped = currentSlot?.let {
+            slots[it]
+        }?.getData()?.items ?: return
+        EquipmentSlot.entries.forEach {
+            val itemStack = currentEquipped[it.slot]
+            if (itemStack != null && !itemStack.isStainedGlassPane()) {
```
AverageUser125, line 48: I could have sworn this check had a dedicated function in WardrobeApi for it already

AverageUser125 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/EquipmentWardrobeApi.kt
```diff
@@ -0,0 +1,57 @@
...
+import kotlin.collections.forEach
+
+@SkyHanniModule
+object EquipmentWardrobeApi : WardrobeApi() {
+
+    /**
+     * REGEX-TEST: (1/2) Equipment Sets
+     */
+    override val inventoryPattern by patternGroup.pattern(
+        "inventory.name.equipment",
```
AverageUser125: `inventory.wardrobe.inventory.name.equipment` seems a bit too long.

Maybe just
`inventory.wardrobe.equipment` is enough?

AverageUser125:
This is technically an important fix, but like. Fabi is probably busy.

f4bii:
got you

AverageUser125:
I mean, I think changing loadouts still needs to notify EquipmentApi. But otherwise probably good.

CalMWolfs (approved):

PR #6048, opened 2026-07-10
Fix: skymall & lottery

AverageUser125:
## What
due to blue goblin omelate, color is needed to find out the level of a HOTM perk.
Which is what caused it to not detect the perk in the UI.

Also I renamed "perkDescription" to "displayDescription", and readded colors to it.

## Changelog Fixes
+ Fixed Skymall/Lottery display being colorless. - AverageUser125
+ Fixed Skymall/Lottery not being detected in the hotm/hotf menu. - AverageUser125


CalMWolfs (approved):
Code looks fine

PR #6050, opened 2026-07-10
Fix: Rift time features not working

CalMWolfs:
## What
Another outdated symbol

## Changelog Fixes
+ Fixed rift time features not working with the hypixel resource pack. - CalMWolfs

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/rift/everywhere/RiftTimer.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/rift/everywhere/RiftTimer.kt
@@ -36,7 +36,7 @@ object RiftTimer {
      */
     private val nametagPattern by patternGroup.pattern(
         "nametag.timer",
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/rift/everywhere/RiftTimer.kt
```diff
@@ -31,11 +32,11 @@ object RiftTimer {
     private val patternGroup = RepoPattern.group("rift.everywhere")
 
     /**
-     * REGEX-TEST: 3150 ф
+     * REGEX-TEST: 3150 
      */
     private val nametagPattern by patternGroup.pattern(
         "nametag.timer",
-        "(?<time>\\d+) ф",
+        "(?<time>\\d+) \uE020",
```
AverageUser125: Why does it below use the SkyblockStat and here it does not?

PR #6051, opened 2026-07-10
Fix: CME in HideSlayerSpawnParticles

lunaynx:
## What
Fixes a [ConcurrentModificationException](https://discord.com/channels/997079228510117908/1000669238035497022/1525143182583791817) in HideSlayerSpawnParticles by making it use a ~~ConcurrentHashMap~~ TimeLimitedCache. Also removes a dumbass suppression that uses more code than just fixing it does.

## Changelog Fixes
+ Fixed ConcurrentModificationException error in Hide Slayer Spawn Particles. - Luna


AverageUser125:
Would a TimeLimitedCache make more sense?

lunaynx:
Oh yeah, definitely, I was too focused on fixing the code to think of that.

AverageUser125:
I hope TimeLimitedCache is thread safe then, it probably is.

lunaynx:
It seems to be backed by Guava and ConcurrentMap, so it's probably fine.

CalMWolfs (approved):

PR #6052, opened 2026-07-10
Fix: Same Chat Color not working

TomekoYT:
## What
Same Chat Color now actually colors private and non-rank players' messages instead of doing nothing

## Changelog Fixes
+ Fixed Same Chat Color not working. - Tomeko


CalMWolfs (approved):
worked for me

PR #6058, opened 2026-07-10
Backend: Lenient max line length detekt rule

AverageUser125:
## What
It kept annoying me that the MaxLineLength rule kept flagging strings.
So I made it NOT count any string length inside the calculation for the total line length.

Except if you are inside a function, which then it doesn't ignore string length.
or lambda, property getter/setter, or init block.

+ Relaxed MaxLineLength detekt rule requirement. - Avrg


lunaynx:
To be fair, there's a lot of strings (regular text) that *should* be wrapped for readability. It might make more sense to only exclude repo patterns, because while those could technically be wrapped, we currently don't bother because it can be often awkward to try.

detekt/detekt.yml
```diff
@@ -63,9 +65,7 @@ style:
             - 'HandleEvent'
             - 'Mod.EventHandler'
     MaxLineLength:
```
lunaynx:
```suggestion
    MaxLineLength: # handled by custom rule
```

hannibal002:
Most of the times when the line is too long it is because of the string, you are right. Thought, ignoring the string length entirely will make this flag skip normal lines that have medium sized string as well. And those should surely be separated. So the approach is not ideal.

Also in general I think if the string is too long for a coding text, either split it up at a end of sentence or maybe the text doesn't event belong hard coded? Think of skin textures being stored in repo.

I'm against implementing the change as is.

AverageUser125:
So, should I make it ignore a line if it’s like, more than half string length?
Or some other ratio?

AverageUser125:
I could make this only be for RepoPattern, I just wanted a more nucleus option.
Aka, the first idea that came to mind lol.

I wonder if I can make it ignore a string if it’s just not inside a function…

hannibal002:
No Ratio at all. I Like the idea if the line is regex text or the repo pattern regex itself, that we ignore the flag then. But e.g. normal chat messages we sent to the user should not be skipped from the flag. Are there other cases?

AverageUser125:
Regex detection sounds cool, but is probably impossible since that would have a lot of false positives.

AverageUser125:
Okay, it doesn’t do exactly what you asked.
Instead I just made it not skip strings if it’s inside anything that could be counted as a function.
(Lambda, named function, property accessor, anonymous init constructor)

AverageUser125:
Should I just make this rule completely ignore anything outside a function?
Would make it really simple.

that way I don’t have to specifically ignore string length and stuff.

lunaynx:
I don't know, that sounds potentially too broad. If we can't find a good way to have a tightly scoped rule, it's probably better to just continue using annotations for the few cases where we don't want to wrap.

AverageUser125:
I can see class inheritance being a problem, so fair.

So I guess just ignoring string length outside of functions is good enough.

AverageUser125:
Too neglected

PR #6064, opened 2026-07-11
Improvement: Async Repo Reload Event Handlers

lunaynx:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/6482 (can be removed if absolutely necessary, but it was agreed upon on Discord, since this lets us do a little more cleanup and kill two birds with one stone)

## What
Adds `AsyncSkyHanniEvent`, which `RepositoryReloadEvent` and `NeuRepositoryReloadEvent` are now subclasses of. These require `suspend fun`s as handlers, meaning we don't need to set up and launch a coroutine in each feature class. All other events continue to require normal `fun`s as handlers for the time being, but it's extensible and any event can be turned into an async one in the future.

Also migrates all existing synchronous repo reload handlers to be async and completely replaces the old synchronous `getConstant` & async `getConstantAsync` with a new async `getConstant`.

Note that the current behavior of this implementation is to sequentially `await` each repository reload event handler on a shared coroutine. This matches current synchronous behavior on `beta` (112 uses of `event.getConstant` across 103 files at the time of writing) without blocking the render thread, but differs from current async behavior on `beta` (12 uses of `event.getConstantAsync` across 10 files) which is fire-and-forget coroutines executing in parallel.

There are a few good reasons to not run all of them in parallel:
- Priority would only guarantee submission order.
- Cancellation would not be reliably able to prevent further handlers of the same event from executing.
- Error handling may be less robust.

In the future, we could possibly look into parallelizing execution of event handlers where that is likely safe to do (e.g. handlers of the same priority for non-cancellable events).

This is a refactor that touches a lot of files, but that's largely unavoidable here (at most, I could leave `getConstantAsync` as deprecated alias and get rid of 12 line changes across 10 files), and any merge conflicts with other PRs should be trivially resolvable.

The new event system also required changing some `crashInDevEnv`s into hard `require`s for safety, but it's incredibly unlikely that a broken event handler would slip through all other guardrails (KSP in local build, KSP in GitHub Actions, (hopefully) manual testing) and make its way into a beta/release (and if it somehow does, we can just do an emergency update to fix it).

## Changelog Improvements
+ Reduced lag during SkyHanni & NEU repo reloads. - Luna

+ Added AsyncSkyHanniEvent, which takes "suspend fun" event handlers. - Luna
+ Made RepositoryReloadEvent and NeuRepositoryReloadEvent subclasses of AsyncSkyHanniEvent. - Luna
+ Removed getConstantAsync from RepositoryReloadEvent and NeuRepositoryReloadEvent and made getConstant asynchronous. - Luna


AverageUser125:
This is gonna be a fun conflict to resolve, just a hunch.

AverageUser125:
Future PR: Make all listeners of the same priority run in parallel

src/main/java/at/hannibal2/skyhanni/features/misc/ContributorManager.kt
```diff
@@ -27,7 +26,7 @@ import at.hannibal2.skyhanni.utils.SimpleTimeMark
 import at.hannibal2.skyhanni.utils.collection.CollectionUtils.mapKeysNotNull
 import at.hannibal2.skyhanni.utils.compat.append
 import at.hannibal2.skyhanni.utils.compat.appendWithColor
-import at.hannibal2.skyhanni.utils.coroutines.CoroutineSettings
+import at.hannibal2.skyhanni.utils.compat.hover
```
hannibal002: unused

src/main/java/at/hannibal2/skyhanni/api/event/SkyHanniEvent.kt
```diff
@@ -23,8 +29,7 @@ abstract class SkyHanniEvent protected constructor() {
 
     interface Cancellable {
```
hannibal002: why is this still here and not also in AbstractSkyHanniEvent?

Cast still goes to SkyHanniEvent while cancelEvent() sits on AbstractSkyHanniEvent, so a cancellable AsyncSkyHanniEvent throws CCE. Moving the interface alone doesn't fix that. Rendering should stay where it is.

src/main/java/at/hannibal2/skyhanni/api/event/AsyncEventHandler.kt
```diff
@@ -0,0 +1,62 @@
...
+
+class AsyncEventHandler<T : AsyncSkyHanniEvent> private constructor(
+    val name: String,
+    private val listeners: List<EventListeners.Listener>,
+) {
+    val invokeLog = SkyHanniEvents.EventInvokeLog()
+
+    constructor(event: Class<T>, listeners: List<EventListeners.Listener>) : this(
+        (event.name.split(".").lastOrNull() ?: event.name).replace("$", "."),
+        listeners.sortedBy { it.priority }.toList(),
```
hannibal002: `onlyOnIsland`, `onlyOnIslands`, `onlyOnIslandTypeTag` and `onlyOnSkyblock` are silently ignored for async events.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/api/event/AsyncEventHandler.kt
AverageUser125: Solution: ignore them loudly 🥺

PR #6068, opened 2026-07-11
Backend: quite a few random old stat usage

AverageUser125:
## What
I ain't making this a fix, since then long changelog.

+ Moved some old stat icons regex usage to new ones. - AverageUser125


CalMWolfs (changes requested):
Quite a few regex tests need to be updated with this

I know the detekt rule doesn't run on them but that should still be accurate. Also in future idk how possible it is but we should see if we can make detekt run on patterns that use string templates

CalMWolfs (approved):

PR #6073, opened 2026-07-12
Backend: Event thread annotation

AverageUser125:
## What
Currently this annotation marks an event for which thread it is from, later it should be used for static analysis like how android linter has it to ensure no cross-thread access.

But currently, it means the KDoc doesn't need to mention which thread the event is from.
And for developments environments it gives a warning about it.

The changes have been auto-generated, by me just putting a logger inside `EventHandler` which logs the event and it's thread name. meaning this thread annotations are accurate.
Also, I love context senstive enums since they also work in annotations and not just in `when`, so that's nice.

Weird exceptions:
1. PacketSentEvet also has connector, ping, render, and alot of threads. so it's any
2. TooltipTextEvent is suppose to be only from render thread, but mods like JEI need tooltip to do advance searching. so it's any.

+ Added thread annotation to events. - Avrg


AverageUser125:
I know I technically modified an event and should add a KDoc to it,
but I am _NOT_ gonna do that.

hannibal002:
> I know I technically modified an event and should add a KDoc to it, but I am _NOT_ gonna do that.

Super sad, that would be supe helpful if happening some time.
But I agree this or doesnt have to add documentarions to all events.
Also have you removed the comments in events that mention what Thread an event gets fired from already?

Also have you pushed the logic by what you detected what is the thread type? And can we reuse this to throw an error to contributors if an event gets fired from a threat that it is not supposed to according to the annotation? Optional, if complicated.

AverageUser125:
1. About comments, no. But probably should.
2. I didn’t push it, I made it only log for unique event posting with thread name… so I guess I can instead make it write in chat instead?
I remember their being some way to log all skyhanni events, could probably add thread name there. (Once I find it)
3. Hmm, shouldn’t be too hard to check if correct thread. It will probably only do the check for development builds.

AverageUser125:
1. okay, I can't find any comment about thread that don't make sense. 
(They mostly same, render thread from mixin into X so it's fine).
2. I added a check for if an event is posted for unexpected thread. it's only enabled in development builds.
3. the current thread detection is just using the thread name, so I sure hope it isn't different between 1.21.11, 26.1.2, and 26.2. Surely they are the same thread names.

AverageUser125:
Doesn’t seem to be wanted

PR #6075, opened 2026-07-13
Improvement: Add hide known visitors toggle for visitor shopping list

legentpc:
## What
Added toggle Hide Known Visitors (default false). When toggle OFF, shows current display with shopping list, new visitors, known visitors. When toggle ON, shows old display like in screenshots: only shopping list and new visitors, hides known visitor names.

<details>
<summary>Images</summary>

Before (toggle OFF, default):
Shopping list + new visitors + Visitor section with clickable names
image

After (toggle ON):
Shopping list + new visitors only, no Visitor section
image

</details>

## Changelog Improvements
+ Added Hide Known Visitors toggle for the visitor shopping list. - legentpc
    * Added toggle Hide Known Visitors (default false), shows only shopping list and new visitors, hides known visitor names.


lunaynx:
This seems like a weirdly named toggle, though I don't know what I'd name it that isn't also confusing. I also don't know why the GUI even looks like this (the "new" style) now, feels like a weird choice by whoever made it. I'd rather have a single list of visitors and have (NEW) suffix, and then an option to hide ones that aren't new.

legentpc:
> This seems like a weirdly named toggle, though I don't know what I'd name it that isn't also confusing. I also don't know why the GUI even looks like this (the "new" style) now, feels like a weird choice by whoever made it. I'd rather have a single list of visitors and have (NEW) suffix, and then an option to hide ones that aren't new.


Yeah the name is confusing, should I rename it to Hide Known Visitors? Please suggest a better name if you have one.

PR #6080, opened 2026-07-13
Backend: add fallback api name to IslandType

AverageUser125:
## What
Added fallback API name to island type incase repo is missingi t.
Also to ensure we never forgot adding it.

+ Added fallback island api name. - AverageUser125


AverageUser125:
should I sort the entries in some sensibale order while I am here? Cause right now they are ordered by when they were added.

CalMWolfs (approved):

PR #6083, opened 2026-07-13
Backend+ Fix: Move Repo directory outside config

AverageUser125:
## What

New file structure:
```
.minecraft/
├── skyhanni/
│   ├── shrepo/
│   │   └── stuff
│   ├── neurepo/
│   │   └── stuff
│   ├── shrepo.tar.gz
│   ├── neurepo.tar.gz
│   ├── shrepo.meta.json
│   └── neurepo.meta.json
└── logs/
    └── skyhanni/
```

Also centerlize config and data dir path references, data incase we move tracker and other stuff into the data dir too. and config since it felt off keeping it the same. 

Deleted old SBE migration, since that felt off to keep.
and  `blocked-features.txt` logic. which has been replaced with disabled events handlers in repo for a while.

Also moved the logs to the logs directory.

+ Moved the SkyHanni and NotEnoughUpdates repos out of the config folder. - Avrg
+ Moved log files into the Minecraft logs folder. - Avrg

## Changelog Fixes
+ Fixed expired config backups not being deleted. - Avrg
+ Fixed old log files not always being deleted after expiring. - Avrg


Fazfoxy:
.minecraft/repo/neu/
Isn't the intended (current) folder layout
.minecraft/repo/skyhanni-neu/{Files}
based on https://discord.com/channels/997079228510117908/1513973394855432244/1526326501136400424

src/main/java/at/hannibal2/skyhanni/data/repo/SkyHanniRepoManager.kt
```diff
@@ -2,20 +2,20 @@ package at.hannibal2.skyhanni.data.repo
...
+import java.io.File
 
 @SkyHanniModule
 object SkyHanniRepoManager : AbstractRepoManager<RepositoryReloadEvent>() {
     override val commonName = "SkyHanni"
     override val commonShortNameCased = "SH"
-    override val configDirectory = ConfigManager.configDirectory
     override val config get() = SkyHanniMod.feature.dev.repo
     override val backupRepoResourcePath: String = "assets/skyhanni/repo.tar.gz"
+    override val legacyConfigDirectory = File("config/skyhanni")
```
hannibal002, line 20: ConfigManager.configDirectory already exists

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
```diff
@@ -561,7 +571,23 @@ abstract class AbstractRepoManager<E : AbstractRepoReloadEvent> {
 
     private fun deleteArchiveFiles() {
```
hannibal002, line 592: Legacy cleanup runs on every reload and update. Doing it once in initRepo, plus a TODO with a date for removal, would fit better.

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
```diff
@@ -561,7 +571,23 @@ abstract class AbstractRepoManager<E : AbstractRepoReloadEvent> {
...
         repoTgzFile.delete()
-        legacyRepoZipFile.delete()
+        deleteLegacyFiles()
+    }
+
+    private fun deleteLegacyFiles() {
+        legacyRepoDirectory?.let {
+            if (it.exists()) {
+                logger.warn("Deleting legacy repo directory: ${it.absolutePath}")
+                it.deleteRecursively()
```
hannibal002: Moving the old directory instead of deleting it would keep users with repo auto update off working, right now they silently drop to the bundled backup repo, and in dev env to no repo at all.

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
```diff
@@ -52,15 +52,28 @@ abstract class AbstractRepoManager<E : AbstractRepoReloadEvent> {
...
     open val backupRepoResourcePath: String? = null
 
+    open val legacyConfigDirectory: File? = null
+
     abstract val config: AbstractRepoConfig
-    abstract val configDirectory: File
+
+    /**
+     * The root directory for this specific repo.
+     * Inheriting classes should provide the path, e.g.: `File(mcDataDir, "repo/skyhanni")`
```
hannibal002: The property is final now, and mcDataDir does not exist.

src/main/java/at/hannibal2/skyhanni/api/enoughupdates/EnoughUpdatesManager.kt
```diff
@@ -325,20 +319,6 @@ object EnoughUpdatesManager {
         }
     }
 
-    fun reportItemStatus() {
```
hannibal002, line 328: repoFileSystem.list("items").size gives the same count without a File handle, why remove it instead of porting it?

hannibal002 (changes requested):
the Repo part in the contributing md file needs to get updated as well

lunaynx:
I know this is "late" but I'd honestly rather have something like `.minecraft/data/skyhanni` that can hold more non-config stuff rather than try to use a "standardized" folder but then still have `skyhanni-` prefixes because it isn't actually standardized yet. If we actually manage to standardize on a single copy of the repo in the future, we can move it again, not a big deal. And it would have to be moved regardless because of the current `skyhanni-` prefix.

src/main/java/at/hannibal2/skyhanni/data/repo/SkyHanniRepoManager.kt
AverageUser125: Yeah, but if the config directory is changed. This should remain the same.

so feels off doing that.

lunaynx:
To clarify: I think the `.minecraft/repo` direction implies scope creep here. It's based on the entirely theoretical idea that other mods may want to use SkyHanni's repo and/or have their own repo handled by a central repo parser library. To me, both seem pointless and unlikely, but regardless of their merits, we should follow the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle and discuss that separately from the actual immediate issue this is trying to solve (Modrinth pack exports).

src/main/java/at/hannibal2/skyhanni/data/repo/SkyHanniRepoManager.kt
hannibal002: Fair, then please add a short comment saying it is hardcoded because the legacy path must not follow the config directory.

hannibal002:
Fair point, let's go with data/skyhanni and revisit a shared folder when something actually shares it.

AverageUser125:
Wooops forgot to update contributing md

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -97,7 +96,8 @@ class ConfigManager {
     }
 
     private fun deleteOldBackups() {
-        OSUtils.deleteExpiredFiles(File("skyhanni/config/backup"), SkyHanniMod.feature.dev.configBackupExpiryTime.days)
+        val file = SkyHanniMod.configDir.resolve("backup")
```
hannibal002, line 99: this fixes the backup config not deleting. needs a explicit bug fix changelog entry.

src/main/java/at/hannibal2/skyhanni/utils/SkyHanniLogger.kt
```diff
@@ -3,34 +3,37 @@ package at.hannibal2.skyhanni.utils
...
+    }
 
     companion object {
         private var deletedExpired = false
-        private val fullFormat by lazy {
-            SimpleDateFormat("yyyy_MM_dd/HH_mm_ss").formatCurrentTime()
-        }
+
+        private val startTimeFormatted = SimpleDateFormat("yyyy_MM_dd/HH_mm_ss").formatCurrentTime()
+        private val format = SimpleDateFormat("HH:mm:ss")
```
hannibal002, line 28: why is this now shared for all objects? thats wrong, please keep using one instance per instance of logger

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
```diff
@@ -141,6 +142,18 @@ object SkyHanniMod : CompatCoroutineManager by SkyHanniCoroutineManager(
...
+    val dataDir: File by lazy {
+        PlatformUtils.dataDir.resolve(MODID).toFile()
+    }
+
+    val configDir: File by lazy {
+        PlatformUtils.configDir.resolve(MODID).toFile()
+    }
+
+    val logsDir: File by lazy {
+        PlatformUtils.logsDir.resolve(MODID).toFile()
```
hannibal002, line 154: doesnt this hit `.minecraft/data/logs/skyhanni`, and is outside everything else skyhanni stores: in `.minecraft/data/skyhanni`.

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
```diff
@@ -52,15 +51,55 @@ abstract class AbstractRepoManager<E : AbstractRepoReloadEvent> {
...
+     */
+    private val repoTgzFile: File by lazy {
+        SkyHanniMod.dataDir.resolve("$repoFolderName.tar.gz")
+    }
+
+    /**
+     * Stores commit metadata for this repo.
+     *
+     * For example:
+     * `.minecraft/data/skyhanni/repo/sh-currentCommit.json`
```
hannibal002: invalid path

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
```diff
@@ -561,7 +591,50 @@ abstract class AbstractRepoManager<E : AbstractRepoReloadEvent> {
...
+                legacyFile.delete()
+                return@let
+            }
+            logger.warn("Moving legacy commit file to: ${commitFile.absolutePath}")
+            commitFile.parentFile?.mkdirs()
+            runCatching {
+                Files.move(legacyFile.toPath(), commitFile.toPath())
+            }.onFailure {
+                runCatching { legacyFile.copyTo(commitFile, overwrite = false) }
+                legacyFile.delete()
```
hannibal002: The copyTo result is discarded, so this deletes the file even when the fallback failed.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
lunaynx: I think it's supposed to be `.minecraft/logs/skyhanni`, we discussed this on Discord and came to the conclusion that it makes sense to write to an existing established directory. Not a dealbreaker either way though.

src/main/java/at/hannibal2/skyhanni/utils/SkyHanniLogger.kt
AverageUser125: If I understood correctly from @Fazfoxy it’s cause there is 1 folder per session. And this ensures all log files for this session are in the same folder.

which I think was a bug previously that they weren’t. So I should probably add a fix changelog tbh.

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
AverageUser125: I didn’t even notice 😭

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
AverageUser125: This is `.minecraft/logs/skyhanni` not inside the data folder but the dedicated logs folder.

unless I made a stupid mistake

hannibal002: logsDir resolves against dataDir, not gameDir, so this lands in .minecraft/data/logs/skyhanni. Please make it data/skyhanni/logs so all our stuff stays in one folder.

AverageUser125: So I did something stupid, let me fix that real quick

AverageUser125: Data folder is for persistent stuff between runs, while logs does not have to persistent between runs.

so I perfer the separation.

src/main/java/at/hannibal2/skyhanni/utils/SkyHanniLogger.kt
hannibal002: This is about format, not startTimeFormatted. startTimeFormatted belongs in the companion, format does not, it was per instance before and SimpleDateFormat is not thread safe.

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
```diff
@@ -52,15 +51,52 @@ abstract class AbstractRepoManager<E : AbstractRepoReloadEvent> {
...
+    open val legacyConfigDirectory: File? = null
+
     abstract val config: AbstractRepoConfig
-    abstract val configDirectory: File
+
+    /**
+     * The root directory for this specific repo.
+     *
+     * For example:
+     * `.minecraft/data/skyhanni/repo` or `.minecraft/data/skyhanni/neurepo`
```
hannibal002, line 64: All three example paths still say data/skyhanni/repo, it is skyhanni/shrepo now.

CONTRIBUTING.md
```diff
@@ -458,7 +458,8 @@ SkyHanni uses KSP via the `annotation-processors` module to generate code at com
 
 SkyHanni uses a repo system to easily change static variables without the need for a mod update.
 The repo is located at https://github.com/hannibal002/SkyHanni-REPO.
-A copy of all JSON files is stored on the computer under `.minecraft\config\skyhanni\repo`.
+A copy of all JSON files is stored on the computer under `.minecraft\data\skyhanni\repo`.
```
hannibal002, line 461: Should be .minecraft\skyhanni\shrepo.

hannibal002 (changes requested):
Ive already fixed the changelog for you, just those small changes in the code are remaining. then the pr is ready imo.

hannibal002 (changes requested):
lunas three points on discord

hannibal002 (changes requested):
bump

Vivo-Y12:
but prismlaucher not make .minecraft

lunaynx:
@Vivo-Y12 Please stop commenting on technical PRs as a uninformed user when you clearly have no idea what you're talking about. We do not hardcode `.minecraft` anywhere, we get the actual path from Fabric Loader. It will work just fine on Prism Launcher.

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
```diff
@@ -561,7 +589,62 @@ abstract class AbstractRepoManager<E : AbstractRepoReloadEvent> {
...
+                            OnErrorAction.TERMINATE
+                        }
+                    }
+                )
+            }.onFailure { e ->
+                logger.error("Uncaught exception while migrating legacy repo: ${e.message}")
+            }.getOrDefault(false)
+
+            if (copied) {
+                legacyRepoDirectory.deleteRecursivelySafe()
```
hannibal002, line 619: Path.deleteRecursively throws instead of returning a boolean, and this line sits outside the runCatching, so a locked file here aborts initRepo again. Guarding the whole function body would be safer.

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
```diff
@@ -493,7 +521,7 @@ abstract class AbstractRepoManager<E : AbstractRepoReloadEvent> {
 
     private fun prepCleanRepoFileSystem(progress: ChatProgressUpdates) {
         progress.update("deleteRecursively")
-        repoDirectory.listFiles()?.forEach { if (it != logger.logsDir) it.deleteRecursively() }
+        repoDirectory.listFiles()?.forEach { it.deleteRecursively() }
```
hannibal002, line 524: Both this and DiskRepoFileSystem.deleteRecursively still use the unsafe variant, worth switching them while we are here. Note they would need the same exception handling as the migration.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/repo/AbstractRepoManager.kt
AverageUser125: I just made `deleteRecursivelySafe` return a boolean so it has the same behavior there was previously.

PR #6085, opened 2026-07-13
Backend: Mark as incompatible with more spoofer mods

lunaynx:
## What
Like ClientSpoofer, these mods break our ability to detect that the player is on Hypixel and receive location information via the Mod API. I am not going to bother writing a compatibility layer for them, anyone who really wants to use them can use local dependency overrides and figure it out themselves.

+ Marked SkyHanni as incompatible with the mods "OpSec" and "Spoofifed". - Luna
    * These mods break our ability to communicate with the Hypixel Mod API.

### DIFF
--- a/src/main/resources/fabric.mod.json
+++ b/src/main/resources/fabric.mod.json
@@ -55,7 +55,9 @@
     },
     "breaks": {
         "clientspoofer": "*",
-        "exordium": "*"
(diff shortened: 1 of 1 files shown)


lunaynx:
My cleanup script thought this was merged and deleted the branch 🎉

PR #6091, opened 2026-07-14
Improvement: Make "Increase Pest Warning Volume" bypass volume restrictions

AverageUser125:
## What
Okay, I made this bypass the clamping of the volume, and it now directly lets you sets the volume above the "maximum".
Is this a good solution? heck no.
Is sound actually logarithmic so you barely notice the difference after a while? yeah.

## Changelog Improvements
+ Improved pest warning by letting you increase it's volume. - Avrg


AverageUser125:
This code feels so wrong, it works. but like. even the slider feels wrong.
I know sound is logarithmic with like the equation `20*log(amplitude)` for some reason. but still.

CalMWolfs (changes requested):
we currently have `/shplaysound` can we either have that allow for testing this new type of sound or make a new command for it

AverageUser125:
Replaced by #6418

PR #6096, opened 2026-07-15
Fix: Rare tile crash

AverageUser125:
## What
https://discord.com/channels/997079228510117908/1526894409579495495
Basically, somehow this block tile entity can be null somehow. maybe mod incompatibility.

## Changelog Fixes
+ Fixed rare error/crash with Crimson Isle Miniboss Timer (likely a mod conflict). - Avrg


lunaynx:
Since we've had at least two people get this error so far on the exact same feature, I'd add a changelog, something like:
- Fixed rare error/crash with Crimson Isle Miniboss Timer (likely a mod conflict).

PR #6098, opened 2026-07-16
Backend: Rename Self

AverageUser125:
## What
This doesn't really matter, and since the mod uses modrinth to get the changelog and this doesn't even update it.
But eh.
I would rather start being consistent.
I just changed "AverageUser125" -> "Avrg".
This way definitely no more changelogs over 2000 characters.

exclude_from_changelog


CalMWolfs (approved):

hannibal002:
Don't forget to change the author names in your open and future prs.

AverageUser125:
I did already, don't worry.
But thanks anyway.

PR #6103, opened 2026-07-16
Fix: Widget Highlight

AverageUser125:
## What
they added widget numbers to the gui. and while I was here I made it colorless.

## Changelog Fixes
+ Fixed tab widget highlight not working in multi-page settings pages. - Avrg

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/TabWidgetSettings.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/TabWidgetSettings.kt
@@ -67,7 +67,7 @@ object TabWidgetSettings {
      * REGEX-TEST: Currently: ENABLED
      */
     private val enabledPattern by patternGroup.pattern(
-        "is.enabled",
+        "is.enabled.colorless",
         ".*ENABLED",
     )
 
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/TabWidgetSettings.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/TabWidgetSettings.kt
@@ -6,7 +6,7 @@ import at.hannibal2.skyhanni.events.GuiContainerEvent
 import at.hannibal2.skyhanni.events.InventoryCloseEvent
 import at.hannibal2.skyhanni.events.InventoryFullyOpenedEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
-import at.hannibal2.skyhanni.utils.ItemUtils.getLore
+import at.hannibal2.skyhanni.utils.ItemUtils.getCleanLore
 import at.hannibal2.skyhanni.utils.LorenzColor
 import at.hannibal2.skyhanni.utils.RegexUtils.anyMatches
 import at.hannibal2.skyhanni.utils.RegexUtils.matches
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/misc/TabWidgetSettings.kt
```diff
@@ -21,28 +21,30 @@ object TabWidgetSettings {
     /**
      * REGEX-TEST: Widgets on Private Island
      * REGEX-TEST: Widgets in Crystal Hollows
+     * REGEX-TEST: (1/2) Widgets on Galatea
+     * REGEX-TEST: (1/2) Widgets in Crystal Hollows
      */
     private val mainPageSettingPattern by patternGroup.pattern(
         "gui",
```
lunaynx, line 28: Change the pattern key so that we don't break it for older versions on islands it's still working on.

```suggestion
        "gui.colorless",
```

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/TabWidgetSettings.kt
AverageUser125: It wasn't colorless already?
Also I am guessing you want "is.enabled" to also be added the .colorless at the end?

AverageUser125: Wait, this specific REGEX is for inventory name. it never ever had color to begin with.

lunaynx: Never mind, I misread.

AverageUser125: well, should I do "it.enabled.colorless" then I guess?

lunaynx (approved):

src/main/java/at/hannibal2/skyhanni/features/misc/TabWidgetSettings.kt
lunaynx: Shouldn't be needed, because the existing pattern is theoretically already permissive enough to support both.

AverageUser125: Yeah, I thought the same. just wanted to be sure.

src/main/java/at/hannibal2/skyhanni/features/misc/TabWidgetSettings.kt
```diff
@@ -54,15 +56,15 @@ object TabWidgetSettings {
...
+        "click.disable.colorless",
         ".*disable!",
     )
 
     /**
-     * REGEX-TEST: §7Currently: §aENABLED
+     * REGEX-TEST: Currently: ENABLED
      */
     private val enabledPattern by patternGroup.pattern(
         "is.enabled",
```
CalMWolfs:
```suggestion
        "is.enabled.colorless",
```

CalMWolfs (changes requested):

PR #6104, opened 2026-07-16
Fix: PestFinder wrong thread

AverageUser125:
## What
I tried to figure out why the pest timer randomly didn't work, and had a wrong thread happend in pest finder. so fixed that.
I couldn't figure out why, so I for now just added an error in the case it couldn't fix the time remaining time part.
and the widget wasn't completely empty. (which happened when joining the island)

I just hope this error won't have alot of false positive for when they first join the garden.

## Changelog Fixes
+ Fixed Pest Finder giving an error when joining the Garden. - Avrg

+ Made IslandLeaveEvent, IslandJoinEvent and IslandChangeEvent run on the render thread. - Avrg

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/hypixelapi/HypixelLocationApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/hypixelapi/HypixelLocationApi.kt
@@ -160,8 +160,9 @@ object HypixelLocationApi {
             }
         }
         if (island != IslandType.NONE) {
+            val captured = previousIsland
             DelayedRun.runOrNextTick {
-                IslandJoinEvent(island = island, previousIsland = previousIsland).post()
+                IslandJoinEvent(island = island, previousIsland = captured).post()
             }
             previousIsland = island
         }
--- a/src/main/java/at/hannibal2/skyhanni/api/hypixelapi/HypixelLocationApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/hypixelapi/HypixelLocationApi.kt
@@ -15,6 +15,7 @@ import at.hannibal2.skyhanni.events.hypixel.modapi.HypixelApiServerChangeEvent
 import at.hannibal2.skyhanni.events.minecraft.ScoreboardTitleUpdateEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.ChatUtils
+import at.hannibal2.skyhanni.utils.DelayedRun
 import at.hannibal2.skyhanni.utils.RegexUtils.matchMatcher
 import at.hannibal2.skyhanni.utils.SkyHanniLogger
 import at.hannibal2.skyhanni.utils.StringUtils.removeColor
@@ -154,14 +155,20 @@ object HypixelLocationApi {
         EventListeners.markEventCacheDirty()
 
         if (oldIsland != IslandType.NONE) {
-            IslandLeaveEvent(oldIsland).post()
+            DelayedRun.runOrNextTick {
+                IslandLeaveEvent(oldIsland).post()
+            }
         }
         if (island != IslandType.NONE) {
-            IslandJoinEvent(island = island, previousIsland = previousIsland).post()
+            DelayedRun.runOrNextTick {
+                IslandJoinEvent(island = island, previousIsland = previousIsland).post()
+            }
             previousIsland = island
         }
 
-        IslandChangeEvent(island, oldIsland).post()
+        DelayedRun.runOrNextTick {
+            IslandChangeEvent(island, oldIsland).post()
+        }
     }
 
     @HandleEvent
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestFinder.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestFinder.kt
@@ -26,6 +26,7 @@ import at.hannibal2.skyhanni.features.garden.GardenPlotApi.sendTeleportTo
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.ConfigUtils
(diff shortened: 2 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestFinder.kt
```diff
@@ -123,10 +123,12 @@ object PestFinder {
         }
     }
 
-    @HandleEvent
-    fun onIslandChange(event: IslandChangeEvent) {
-        display = listOf()
-        update()
+    @HandleEvent(onlyOnIsland = IslandType.GARDEN)
+    fun onIslandJoin() {
+        DelayedRun.runOrNextTick {
```
hannibal002: when is island join event getting sent from an async thread?

if you intend to delay the logic by one tick, then use runNextTick instead.

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestSpawnTimer.kt
```diff
@@ -111,6 +112,14 @@ object PestSpawnTimer {
                 hasReminderShown = false
                 pestSpawned = false
             }
+        } ?: run {
+            if (event.widget.lines.all { it.string.isBlank() }) return
```
hannibal002: use widgetLines here instead

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestSpawnTimer.kt
```diff
@@ -111,6 +112,14 @@ object PestSpawnTimer {
...
+
+            ErrorManager.logErrorStateWithData(
+                "Could not find pest cooldown time in widget update event.",
+                internalMessage = "Could not find pest cooldown time in widget update event.",
+                "widgetLines" to event.widget.lines.map { it.string },
```
hannibal002: extract event.widget.lines.map { it.string } into a value, reuse from above in this function.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestFinder.kt
AverageUser125: I do not intend to delay it by one tick, I am merely trying to move it from network to render thread.
And I assumed `runOrNextTick` was the convention function for doing that.

hannibal002: > And I assumed runOrNextTick was the convention function for doing that.

it is!

though, it would be best to just make island join event never be async to begin with.

i assumed you know why the event can be async, and therefore can fix the actual issue instead of just the follow up bug in this class.

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestSpawnTimer.kt
```diff
@@ -111,6 +112,15 @@ object PestSpawnTimer {
                 hasReminderShown = false
                 pestSpawned = false
             }
+        } ?: run {
+            val widgetLines = event.widget.lines.map { it.string }
```
hannibal002: please also merge the usage in line 95

hannibal002 (changes requested):
and the one open review above

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestFinder.kt
AverageUser125: oh, it's cause it uses HypixelLocationApi, and that uses the packets to detect the island. so it's from the network thread.

Do you wish to move the DelayedRun.runOrNextTick there?

hannibal002: > Do you wish to move the DelayedRun.runOrNextTick there?

Yes! a event that gets sometimes fired from sync and sometimes from async is super bad design, plus if it were fully async we would need to add a comment to the event desc, and also do siimilar workarounds in all usages of this event.

src/main/java/at/hannibal2/skyhanni/api/hypixelapi/HypixelLocationApi.kt
```diff
@@ -154,14 +155,20 @@ object HypixelLocationApi {
...
-            IslandJoinEvent(island = island, previousIsland = previousIsland).post()
+            DelayedRun.runOrNextTick {
+                IslandJoinEvent(island = island, previousIsland = previousIsland).post()
+            }
             previousIsland = island
         }
 
-        IslandChangeEvent(island, oldIsland).post()
+        DelayedRun.runOrNextTick {
+            IslandChangeEvent(island, oldIsland).post()
```
hannibal002, line 171: IslandChangeEvent is depreacated

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestSpawnTimer.kt
```diff
@@ -111,6 +113,14 @@ object PestSpawnTimer {
                 hasReminderShown = false
                 pestSpawned = false
             }
+        } ?: run {
+            if (widgetLines.isEmpty() || widgetLines.all { it.isBlank() }) return
```
hannibal002: ´widgetLines.isEmpty()´ is redundant. ´all()´ already returns ´true´ for empty lists

hannibal002 (changes requested):
PestFinder.onIslandChange should use the new onIslandJoin event instead

src/main/java/at/hannibal2/skyhanni/api/hypixelapi/HypixelLocationApi.kt
AverageUser125: It should still be posted. Even if deprecated.

hannibal002: yeah my bad.

hannibal002 (approved):

PR #6107, opened 2026-07-16
Backend: Shwords causing crash if text is somehow null

AverageUser125:
## What
Apparently the text can be null. so just made it compatible.
This ain't a fix since too techincal.

+ Fixed shwords Kotlin hook causing a crash on null strings. - Avrg

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/mixins/hooks/VisualWordsHook.kt
+++ b/src/main/java/at/hannibal2/skyhanni/mixins/hooks/VisualWordsHook.kt
@@ -12,17 +12,18 @@ object VisualWordsHook {
     val caxtonLoaded by lazy { PlatformUtils.isModInstalled("caxton") }
 
     @JvmStatic
-    fun modifyOrderedText(value: FormattedCharSequence): FormattedCharSequence =
+    fun modifyOrderedText(value: FormattedCharSequence?): FormattedCharSequence? =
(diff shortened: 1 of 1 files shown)


lunaynx:
If it's a user-facing error/crash, it should still be a "Fix". Just make a best-effort attempt at making the changelog entry at least somewhat understandable by the average user. There's (probably) no absolute ban on mentioning a technical term like "null" if you can't express it better, but it's up to you, could probably also just generalize it and say something like "Fixed a rare crash with /shwords likely caused by buggy mods".

AverageUser125:
as an average user, I got no clue.
Also this bug happens even with /shwords disabled. so I can't even mention it being enabled as the cause.

CalMWolfs (approved):

hannibal002:
> There's (probably) no absolute ban on mentioning a technical term

Yes. We want the changelog to be meaningful when a user reads them. That's the whole point.

If the non technical savvy reader has an idea what's going on / the context is overal clear, then it's alright ish. If we can expect the user to ask questions what this means then this is not ok

hannibal002:
> This ain't a fix since too techincal.

For the question wether to add a user facing changelog, the technicality of the code change doesn't matter. The user impact matters. 

Not sure if I understand the context enough but a phrase like "fixed rare crash with shwords" would be fine.

PR #6109, opened 2026-07-17
Backend: Repofiy DungeonChatFilter

AverageUser125:
## What
It was hardcded, and had startsWith and equals checks. now it's all regex.
I still need to make it colorless, but I hope I didn't break anything yet.

+ Made DungeonChatFilter use the repo for patterns. - Avrg


CalMWolfs (changes requested):
I like the idea of this pr as it is both removing color matching and making stuff use the repo but there are a few issues currently.

Patterns like `(.*) (.*) (.*) -> (.*)` become nonsese
There area also a lot of unnamed capture groups that just arent needed. As if most of these were normal repo patterns rather than pattern lists then detekt would be warning you about these things.

These things need to be addressed before merging

PR #6110, opened 2026-07-17
Fix: Minion upgrade Helper

AverageUser125:
## What
What is with them and adding spaces at the front?
Anyways, fixed that.
Also fixed did some TODOs in MinionFeatures while I was here.

## Changelog Fixes
+ Fixed Minion Upgrade Helper not appearing. - Avrg

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/minion/MinionFeatures.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/minion/MinionFeatures.kt
@@ -122,7 +122,7 @@ object MinionFeatures {
         "Minion [^➜]",
     )
     private val minionCollectItemPattern by patternGroup.pattern(
-        "item.collect",
+        "item.collect.colorless",
         "^Collect All$",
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/features/minion/MinionFeatures.kt
```diff
@@ -107,7 +123,7 @@ object MinionFeatures {
     )
     private val minionCollectItemPattern by patternGroup.pattern(
         "item.collect",
```
CalMWolfs:
```suggestion
        "item.collect.colorless",
```

CalMWolfs (approved):

src/main/java/at/hannibal2/skyhanni/features/minion/MinionFeatures.kt
AverageUser125: Oh, good catch.

PR #6111, opened 2026-07-17
Fix: Equipped equipment being wrong

f4bii:
## What
Fixed EquipmentApi having the wrong equipped Equipment when changing loadouts etc.

## Changelog Fixes
+ Fixed Zorro's Cape protection not working due to Equipment change. - FabiHBBBT
+ Fixed Estimated Armor Wardrobe Price showing while using Custom Wardrobe. - FabiHBBBT

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/WardrobeApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/WardrobeApi.kt
@@ -53,6 +53,8 @@ abstract class WardrobeApi {
 
     abstract val storage: ProfileSpecificStorage.WardrobeStorage?
 
+    protected open fun onEquippedSlotUpdated(items: List<SafeItemStack?>) {}
+
     var slots: List<WardrobeSlot> = emptyList()
         private set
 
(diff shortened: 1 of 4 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/WardrobeApi.kt
```diff
@@ -53,6 +53,8 @@ abstract class WardrobeApi {
 
     abstract val storage: ProfileSpecificStorage.WardrobeStorage?
 
+    protected open fun onEquippedSlotUpdated(items: List<SafeItemStack?>) {}
```
AverageUser125, line 56: this function is for both equipment and armor wardrobe.
So maybe call it:
"onWardrobeSlotUpdated" instead?

AverageUser125 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/WardrobeApi.kt
f4bii: Idk its the equipped slot in both ui's

AverageUser125: hm, fair enough.

hannibal002 (approved):

PR #6113, opened 2026-07-17
Fix: Not detecting 4th harvest feast crop

AverageUser125:
## What
In the alpha, they currently added a 4th one.
This are confirmed to be the correct slot using the Ted contact.

## Changelog Fixes
+ Fixed skyhanni not detecting the 4th harvest feast in-season crop. - Avrg


lunaynx:
It might be worth adding a repo-based override just in case, though it's probably safe to just merge a change to 4 once we know there's no more Harvest Feasts before Tuesday.

Also, maybe look into detecting based on slot text rather than hardcoded IDs.

AverageUser125:
This is such a one-off change, that's it's probably fine to just hardcode it.

Also this way we prune the none updaters, so it's a win win.

AverageUser125:
Now just go pray I chose the correct slots.

AverageUser125:
Better solution, I just support both "T" and "-- --" patterns.
And it shouldn't cause any problems.

And can be removed in 3 days.

CalMWolfs (approved):

PR #6116, opened 2026-07-17
Fix: Giant's Sword item ability cooldown detection

Maratons4:
## What
The item ability cooldown display does not currently work for the Giant's Sword. This PR fixes it by changing the volume.

## Changelog Fixes
+ Fixed the item ability cooldown detection for the Giant's Sword. - Maratons4


AverageUser125 (approved):
Just tested it, and it works.

PR #6122, opened 2026-07-18
Backend: Fix Check PR Dependencies failing after merge (hopefully)

lunaynx:
## What
This should hopefully stop Check PR Dependencies from failing on every PR after it's merged. We love GitHub jank.

exclude_from_changelog


CalMWolfs (approved):
surely this works

PR #6112, opened 2026-07-17
Backend: ComposterUpgrades + HideMobNames regex repofiable

AverageUser125:
## What
I saw some easy "TODO: use SH-REPO". and either did that, or made it colorless.
Not much to say.

exclude_from_changelog

PR #6123, opened 2026-07-18
Fix: HideMobNames + Composter were wrong, me sorry

AverageUser125:
## What
Fixes my stupidity from #6112 

It's broken, and would wipe config.
So now it's just a repo pattern list.

Also a companion object of an enum is apparently lazy initialized.

exclude_from_changelog

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/garden/ComposterUpgradesData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/garden/ComposterUpgradesData.kt
@@ -3,24 +3,39 @@ package at.hannibal2.skyhanni.data.garden
 import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.data.IslandType
 import at.hannibal2.skyhanni.data.model.ComposterUpgrade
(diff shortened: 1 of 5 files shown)


src/main/java/at/hannibal2/skyhanni/data/garden/ComposterUpgradesData.kt
```diff
@@ -9,18 +9,32 @@ import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
...
+     * REGEX-TEST: Fuel Cap I
+     * REGEX-TEST: Organic Matter Cap IV
+     * REGEX-TEST: Cost Reduction V
+     */
+    val composterUpgradePattern by RepoPattern.pattern(
+        "composter.upgrade.name",
+        "(?<name>${regexValues()})(?: (?<level>.*))?",
+    )
+    private fun regexValues() = ComposterUpgrade.entries.joinToString("|") { it.displayName }
+
```
CalMWolfs, line 29: we have enumJoinToPattern for this i think

CalMWolfs (changes requested):

CalMWolfs (approved):

PR #5894, opened 2026-06-19
Feature: Add 26.2 support

lunaynx:
## Dependencies
<details>
<summary>Merged</summary>

- #5877 (I don't want to work around the janky old build workflow breaking)
- #5970 (pulled in to avoid intermittent build failures)
- #5971 (pulled in to prevent freezes/memory leaks for testers)
- #6145
- #6150
- #6160
- #6194
- #6242
- #6268
- #6271
- #6272
- #6276
- #6359
</details>

## What
Adds support for Minecraft 26.2 (both OpenGL and Vulkan).

## Changelog New Features
+ Added support for Minecraft 26.2. - Luna

PR #6124, opened 2026-07-18
Backend: Random Cleanup

lunaynx:
## Dependencies
- #6145

## What
Includes a bunch of miscellaneous cleanup split out from #5894. Removed any 26.2-exclusive code, I think.

One notable change is that I made specifically imports & classtweaker use only basic Stonecutter annotations, with no local replacements – this is for increased clarity and to make it easier for a future custom Detekt import order rule to make sense of preprocessed imports. More advanced Stonecutter annotations are still fair game for the rest of the code.

Mixin classes were also renamed to match latest target names, and subfolders for mixins were removed because the categorization was somewhat arbitrary – there isn't an objective way to determine which folder to place a mixin class in every single time, which often leads to accidental duplication.

+ Added an opt-in development environment-only toggle allowing developers to test release candidate versions of Minecraft on Hypixel. - Luna


AverageUser125:
Would it make sense to change `Minecraft.getInstance().font` into `MinecraftCompat.font`?
Or is it not worth it/needed.

lunaynx:
Maybe, but not in this PR. The main purpose of a compat class is for things that have or are expected to have differences between versions, though honestly some of them are just wrappers for convenience/safety at this point. We could expand it into a more general utils class like SkyblockAPI's `McClient` at some point.

hannibal002:
this is a huge pr and it gets merge conflicts all the time. ill try to review it very soon. in the meantime, do you think it is worth the effort to extract "Fixed some text not rendering over entities when it should." into a standalone pr?

lunaynx:
Yeah, I'll look into it. Originally this was a 26.2-specific fix, but it turned out 26.1 was also bugged in some areas.

lunaynx:
Closing, this has been/will be split off into smaller PRs.

PR #6125, opened 2026-07-18
Backend: Scatter Debug options

AverageUser125:
## What
Moves the debug options to the places it makes sense using the "OnlyDebug" annotation.
Also removed a config move on a debug option.

Whoever is reading this: should I also make it not have the "Expose" annotation?

+ Moved debug config options to relevant places. - Avrg


CalMWolfs (changes requested):
im against a lot of this pr as sometimes if users are having issues it is useful for them to enable the debug options and making this change makes it harder. I know they can add themself as a contributor locally somehow but not easily

hannibal002:
Im also against the idea of scattering debug options around the whole config. ideally, the whole dev/debug/test category would be even hidden by default from the config, so users dont accidentally break something. (think NEU times iq test) or similar approaches.

PR #6126, opened 2026-07-19
Improvement: Superpairs Slot Highlighting

RemainingDelta:
### What 
Adds slot highlighting to the Superpairs experiment in the Experimentation Table, similar to NEU's enchanting solver. Cards are highlighted based on what is known about them:
  - **Green** — collected pairs
  - **Orange** — known matches (both cards seen, not collected yet)
  - **Yellow** — cards revealed once whose partner is still unknown
  - **Dark Purple** — uncovered powerups  

All four colors are configurable (Inventory → Experimentation Table → Superpairs), with a single "Highlight Pairs" toggle. Tracking also correctly ignores the round-end reveal of leftover cards, so only pairs you actually collected show as collected.
  
<details>
<summary>Images</summary>

Superpairs overlay similar to NEU:
Superpairs (Metaphysical

</details>

## Changelog Improvements
+ Added NEU-style Superpairs slot highlighting in the Experimentation Table. - RemainingDelta
    * Showing collected pairs, known matches, seen cards and powerups.


CalMWolfs (changes requested):
Works great I think all of my suggestion is only about superpairs data

At the end of the experiment it says you have collected everything which is not true

Image

RemainingDelta:
> Works great I think all of my suggestion is only about superpairs data
> 
> At the end of the experiment it says you have collected everything which is not true

Fixed it by watching the "Remaining Clicks" counter: the first time it hits 0 is your genuine last click (still counted normally), but any update after that while it's still at 0 gets ignored, since that's the round-end reveal, not something you actually collected. If any powerup that gives any additional clicks is collected and happens to bounce the counter back up from 0, instead of the round actually ending, tracking just resumes normally the moment it sees clicks above 0 again.

CalMWolfs (approved):

PR #6130, opened 2026-07-19
Backend: add Log Only flag to errors

AverageUser125:
## What
I am 80% I misunderstood the request. but still.
And this _MAY_ cause problems using a sealed interface in the case of a stack overflow.
(ErrorStateT exists for a reason)

+ Added logOnly flag to ErrorManager.logError. - Avrg


AverageUser125:
I think Hannibal made a PR about stack trace and stuff, so eh

PR #6136, opened 2026-07-19
Backend: Detekt Custom Import Rule

AverageUser125:
## What
Fixes the CustomImportRule.
And that's it for now.
Maybe I should also make the rule show the exact line with the wrong import, but this is fine.

~~The rule can only actually be enabled after #6124 is merged probably. and the things are fixed.~~
Screw that, it somehow works right now.

+ Fixed the CustomImportOrdering detekt rule. - Avrg


hannibal002:
how about as first step just fix the import ordering kt bug, but keep the detetk rule disabled?

AverageUser125:
I mean, then that's done.
But I just want to ensure we all agree on the actual import ordering itself too.

AverageUser125:
Why does this still have the "Waiting on Dependency PR" tag?

hannibal002:
> Why does this still have the "Waiting on Dependency PR" tag?

a bug, ill investigate

AverageUser125:
well, atleast remove the label manually for now.

detekt/src/main/kotlin/imports/CustomImportOrdering.kt
```diff
@@ -1,65 +1,150 @@
...
-        if (rawText.isBlank()) {
+        val rawText = importList.text.lines()
+        if (rawText.all(String::isBlank)) {
             return
         }
 
-        val importsCorrect = isImportsCorrectlyOrdered(importList.imports, rawText.lines())
+        val blocks = createImportBlocks(rawText)
+
+        require(blocks.isNotEmpty()) { "No import blocks found in the import list." }
```
hannibal002: do we really want to error out here? isnt an early return enouth?

detekt/src/main/kotlin/imports/CustomImportOrdering.kt
```diff
@@ -1,65 +1,150 @@
...
+        if (!isPreprocessingBlocksLast(blocks)) {
+            importList.reportIssue(
+                "Preprocessed import blocks must be at the end of the import list.",
+            )
+        }
+
+        if (!isValidImportSpacing(rawText, blocks)) {
+            importList.reportIssue(
+                "Preprocessed import blocks must be separated by empty lines. " +
+                    "There should not be any empty lines between non-preprocessed imports.",
```
hannibal002: this check does not happen in fact. we only check aginst empty lines between the preprocessed and the non preprocessed blocks, not  between the non preporcessed lines.
e.g.:

import a.A

import a.B

this would not get flagged.

src/main/java/at/hannibal2/skyhanni/api/minecraftevents/RenderEvents.kt
```diff
@@ -16,8 +16,11 @@ import net.minecraft.client.DeltaTracker
 import net.minecraft.client.gui.GuiGraphicsExtractor
 import net.minecraft.resources.Identifier
 
-//~ if < 26.1 'PictureInPictureRendererRegistry' -> 'SpecialGuiElementRegistry'
+//? if >= 26.1 {
```
hannibal002, line 19: was the old code here just a comment or actual valid stonecutter? im not sure here

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/api/minecraftevents/RenderEvents.kt
AverageUser125: It was valid stonecutter.
I just made it basic stoncutter so the rule can detect it.

detekt/src/main/kotlin/imports/CustomImportOrdering.kt
AverageUser125: good catch. and fixed.

detekt/src/main/kotlin/imports/CustomImportOrdering.kt
AverageUser125: since there is the:
```
        if (rawText.all(String::isBlank)) {
            return
        }
```
This should never happen unless the import processing had a mistake.

AverageUser125:
hm, this rule currently doesn't check the import order if it's commented out.
but I am not gonna rewrite ImportOrder to work with strings rather then KtImportDirective right now.

detekt/src/main/kotlin/SkyHanniRule.kt
```diff
@@ -1,12 +1,70 @@
...
+            source = SourceLocation(lineAndColumnRange.start.line, lineAndColumnRange.start.column),
+            endSource = SourceLocation(lineAndColumnRange.end.line, lineAndColumnRange.end.column),
+            text = TextLocation(offset, offset + line.length),
+            path = file.absolutePath(),
+        )
+
+        report(Finding(Entity.from(ktElement, location), message))
+    }
+
+
```
hannibal002, line 49: duplicate empty lines

hannibal002 (approved):

PR #6139, opened 2026-07-19
Backend: Use commit status for Check PR Dependencies

lunaynx:
## What
Makes "Check PR Dependencies" report as a failing status check rather than a workflow failure, which stops GitHub from sending unnecessary "workflow failed" emails.

Before, you would get a workflow failure email from GitHub every time you pushed to a PR that is waiting on one or more dependencies.

exclude_from_changelog


lunaynx:
The "Verify build results" one wasn't working correctly, so I reverted that one for now.

PR #6145, opened 2026-07-20
Backend: Update mixin naming + misc cleanup

lunaynx:
d mixin classes to match lates## Dependencies
- #6148
- #6158

## What
Split off from #6124.

Changes are mainly:
- Standardized mixin classes so they're all named `Mixin<TargetClass>` with the latest 26.1 class name, unless they really need an additional suffix, e.g. `MixinClientPacketListenerEarly`
- Added `fabric` subfolder for mixins targeting Fabric rather than Minecraft.
- Removed `gui` and `render` subfolders and moved mixins from them into the root folder, because this is a subjective categorization, which is bad for mixins with a rigid naming structure, and leads to accidental creation of duplicate mixin files.
- Merged multiple mixin classes targeting the same class into a single mixin class (except when there was a good reason to do so, e.g. different priorities for `ClientPacketListener` mixins).
- Made all mixin classes `abstract` (no change in functionality, just good practice as it prevents accidental attempts to instantiate them directly).
- Cleaned up imports in mixin classes.
- Cleaned up classtweaker file to be more readable instead of having version-conditional stuff all over the place.

Should not change any functionality.

+ Updated Mixin classes to match latest target names, merged duplicate Mixin classes into one file where appropriate, removed GUI and render subfolders, and added Fabric subfolder for Mixins. - Luna


AverageUser125:
> Updated mixin classes to match latest target names, and removed subfolders for mixins. - Luna

Is probably the place for this changelog to go.

lunaynx:
Yeah, I guess I could put a technical details changelog, it's kinda arbitrary how I decide when to put one for changes like that or not.

lunaynx:
I added it with slightly adjusted wording. I think that's one of the type of changes that potentially reaches the threshold of "notable enough to call out as a sort of developer changelog", the rest maybe not so much.

AverageUser125:
Would it be worth it to make some detekt rule or something to enforce that mixins must be abstract?

lunaynx:
Detekt is for Kotlin only, so that wouldn't work. But we could do it in the future if we add some Java static analysis tool. Or update the IntelliJ plugin, but that wouldn't actually enforce it, unless it can be integrated with Qodana or something.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinChatComponent.java
```diff
@@ -1,19 +1,44 @@
-//? if < 26.1 {
-/*package at.hannibal2.skyhanni.mixins.transformers;
+package at.hannibal2.skyhanni.mixins.transformers;
 
+import at.hannibal2.skyhanni.features.chat.ChatPeek;
+import at.hannibal2.skyhanni.features.chroma.ChromaFontManagerKt;
+import at.hannibal2.skyhanni.features.misc.visualwords.ModifyVisualWords;
+import at.hannibal2.skyhanni.mixins.hooks.GuiChatHook;
```
hannibal002: unused import

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinChatComponent.java
```diff
@@ -1,19 +1,44 @@
...
+import com.llamalad7.mixinextras.injector.wrapmethod.WrapMethod;
 import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
-import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
-import net.minecraft.client.multiplayer.chat.GuiMessage;
-import net.minecraft.client.multiplayer.chat.GuiMessageTag;
+import net.minecraft.client.Minecraft;
+import net.minecraft.client.gui.Gui;
```
hannibal002: unused import

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinChatComponent.java
```diff
@@ -1,19 +1,44 @@
...
 import net.minecraft.client.gui.components.ChatComponent;
-import net.minecraft.util.FormattedCharSequence;
+import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
```
hannibal002: unused import

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinChatComponent.java
```diff
@@ -1,19 +1,44 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import java.util.ListIterator;
```
hannibal002: unused import

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinItemFeatureRenderer.java
```diff
@@ -1,21 +1,35 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.ModifyArg;
+
+//? if >= 26.1 {
 import at.hannibal2.skyhanni.data.entity.EntityTransparencyManager;
 import at.hannibal2.skyhanni.mixins.hooks.EntityRenderDispatcherHookKt;
 import com.llamalad7.mixinextras.injector.ModifyExpressionValue;
+import com.mojang.blaze3d.vertex.VertexConsumer;
 import net.minecraft.client.renderer.feature.ItemFeatureRenderer;
```
hannibal002: duplicate import

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinChatComponent.java
```diff
@@ -32,5 +57,39 @@ private GuiMessage.Line addMessageId(
...
+    *///?}
+    private void wrapRender(
+        ChatComponent.ChatGraphicsAccess chatGraphicsAccess,
+        int screenHeight,
+        int ticks,
+        //~ if < 26.1 'ChatComponent.DisplayMode' -> 'boolean'
+        ChatComponent.DisplayMode displayMode,
+        Operation<Void> original
+    ) {
+        ChromaFontManagerKt.setRenderingChat(true);
```
hannibal002, line 78: too much code in java, extract to a hook please

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinLevelRenderer.java
```diff
@@ -105,6 +111,33 @@ private PoseStack onCreateMatrixStack(PoseStack matrixStack) {
         return matrixStack;
     }
 
+    @Inject(method = "extractVisibleEntities", at = @At(value = "HEAD"))
+    public void resetRealGlowing(CallbackInfo ci) {
+        RenderLivingEntityHelper.check();
```
hannibal002: too much code in java, extract to a hook please

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinLevelRenderer.java
```diff
@@ -105,6 +111,33 @@ private PoseStack onCreateMatrixStack(PoseStack matrixStack) {
...
+        RenderEntityOutlineEvent noXrayOutlineEvent = new RenderEntityOutlineEvent(RenderEntityOutlineEvent.Type.NO_XRAY, null);
+        RenderLivingEntityHelper.setCurrentGlowEvent(noXrayOutlineEvent);
+        noXrayOutlineEvent.post();
+    }
+
+    @WrapOperation(method = "extractVisibleEntities", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/renderer/entity/state/EntityRenderState;appearsGlowing()Z"))
+    public boolean shouldAlsoGlow(EntityRenderState instance, Operation<Boolean> original, @Local Entity entity) {
+        Integer glowColor = RenderLivingEntityHelper.getEntityGlowColor(entity);
+        if (glowColor == null) return original.call(instance);
```
hannibal002: can be compacted to return glowColor != null || original.call(instance);

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinChatComponent.java
lunaynx: How do you expect me to extract this more? It's two separate features' guards around the original function call. I condensed the `original.call` into one line since it's not long, but that's about it.

The other option is `@Inject` at `HEAD`/`TAIL`, but `@WrapMethod` is probably better because it's harder to accidentally delete just one or the other.

hannibal002: ill fix it in a follow up pr

hannibal002 (approved):

PR #6146, opened 2026-07-20
Fix: LoadOutApi does not update MaxwellApi about powerstone & tunings

AverageUser125:
## What
Made it tell MaxwellApi the new tunings and powerstone.
Don't know if making this public is the best solution, but it was the easiest.

Has been tested and it works.

## Changelog Fixes
+ Fixed changing loadouts not updating Power Stone and tunings in Custom Scoreboard. - Avrg

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/MaxwellApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/MaxwellApi.kt
@@ -106,7 +106,7 @@ object MaxwellApi {
         "gui.thaumaturgy.start",
         "§7Your tuning:",
     )
-    private val thaumaturgyDataPattern by patternGroup.pattern(
+    val thaumaturgyDataPattern by patternGroup.pattern(
         "gui.thaumaturgy.data",
         "§(?<color>.)\\+(?<amount>[^ ]+)(?<icon>.) (?<name>.+)",
     )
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/data/MaxwellApi.kt
```diff
@@ -118,7 +118,7 @@ object MaxwellApi {
         "gui.thaumaturgy.magicalpower",
         "§7Total: §6(?<mp>[\\d.,]+) Magical Power",
     )
-    private val statsTuningGuiPattern by patternGroup.pattern(
+    val statsTuningGuiPattern by patternGroup.pattern(
```
CalMWolfs: i dont see why this is being made public

src/main/java/at/hannibal2/skyhanni/data/MaxwellApi.kt
CalMWolfs, line 1: Then I really dont like an extension function on Pattern that reads the tuning from it. As now anywhere in the mods code you can run this extension function on a pattern and 99.9% of the time that is not something you want.
Instead id just make it a normal function that either takes in a string and a pattern or just make it take in a string and then loop through both patterns that this extension function currently uses

CalMWolfs (changes requested):

CalMWolfs (approved):

PR #6147, opened 2026-07-20
Backend: Component AdvancedPlayerList

AverageUser125:
## What
tbh, I gave an AI the task to this task as a joke. (a free tier).
And it seemed to work?

And then I added basic tests, and everything still seems working ingame.
I wish I had the original example with the custom color codes to check that this actually works.

+ Make AdvancedPlayerList allow custom hex value names. - Avrg


AverageUser125:
Needed for when RGB hypixel level drops.
But currently it needs to support both component and legacy color codes cause hypixel.
So when RGB then only component and must simpler

PR #6149, opened 2026-07-21
Fix: Render see-through world text above entities on 26.1

lunaynx:
## What
Workaround for entities rendering above text marked as `SEE_THROUGH` on 26.1 due to https://mojira.dev/MC-265743. 

## Changelog Fixes
+ Fixed some text (e.g. Damage Indicator) rendering behind entities on 26.1. - Luna


src/main/java/at/hannibal2/skyhanni/utils/render/WorldRenderUtils.kt
```diff
@@ -271,6 +288,26 @@ object WorldRenderUtils {
...
+                    text,
+                    x,
+                    0f,
+                    color?.rgb ?: LorenzColor.WHITE.toColor().rgb,
+                    shadow,
+                    matrix,
+                    bufferSource,
+                    Font.DisplayMode.SEE_THROUGH,
+                    backGroundColor,
+                    15728880,
```
AverageUser125: I don’t know much about rendering. 
But I am curious. What is this magic number?

CalMWolfs: > I don’t know much about rendering. But I am curious. What is this magic number?

light texture coordinates afaik, likely copied from where mc uses it themself

hannibal002: > light texture coordinates afaik, likely copied from where mc uses it themself

then the magic number should be extracted to a variable with a name that explains that please @lunaynx

hannibal002: I just realized we use the same magic n umber 15728880 four times in the file. please replace all of them

hannibal002 (changes requested):
see comment above

hannibal002 (approved):

PR #6158, opened 2026-07-21
Backend: Ban @Overwrite and @Redirect

lunaynx:
## What
We should never use `@Overwrite` or `@Redirect` in mixins unless we are intentionally trying to be incompatible with other mods. This PR bans importing those annotations and changes existing mixins that used `@Redirect` to safer alternatives (no existing uses of `@Overwrite`).

The `deleteMessageOrDelay` mixins have been removed entirely per discussion on Discord as they appear to be unnecessary now and are invasive blanket changes not gated to our own message deletion.

+ Banned `@Overwrite` and `@Redirect` and changed all existing uses of `@Redirect` to safer alternatives. - Luna


AverageUser125:
add Overwrite annotation too.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/gui/MixinGui.java
```diff
@@ -1,6 +1,7 @@
 package at.hannibal2.skyhanni.mixins.transformers.gui;
 
 import at.hannibal2.skyhanni.api.minecraftevents.RenderEvents;
+import at.hannibal2.skyhanni.data.ScoreboardData;
 import at.hannibal2.skyhanni.events.TitleReceivedEvent;
 import at.hannibal2.skyhanni.features.chat.ChatPeek;
```
hannibal002, line 6: two unsed improts below

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinEntityBlaze.java
```diff
@@ -1,19 +1,26 @@
...
-        }
+    @Inject(
+        method = "aiStep",
+        at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/world/level/Level;addParticle(Lnet/minecraft/core/particles/ParticleOptions;DDDDDD)V"
+        ),
+        cancellable = true
+    )
+    private void onLivingUpdate(CallbackInfo ci) {
```
hannibal002: does this cancel the inside of the for loop in the aiStep fun, or the whole aiStep fun? becuase if this cancels the whole aiStep fun, then `super.aiStep();` never gets called at the end of the fun, meaning that in the Monster class those functions arent called either:

        this.updateSwingTime();
        this.updateNoActionTime();

this doesn't feel right to me.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinBaseFireBlock.java
```diff
@@ -1,19 +1,26 @@
 package at.hannibal2.skyhanni.mixins.transformers;
 
 import at.hannibal2.skyhanni.features.misc.ParticleHider;
+import net.minecraft.core.particles.ParticleOptions;
```
hannibal002, line 6: two unused imports here

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinBlaze.java
```diff
@@ -1,19 +1,26 @@
 package at.hannibal2.skyhanni.mixins.transformers;
 
 import at.hannibal2.skyhanni.features.misc.ParticleHider;
+import net.minecraft.core.particles.ParticleOptions;
```
hannibal002, line 6: two unused imports

hannibal002 (changes requested):
we do need detekt in java files :(

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinEntityBlaze.java
lunaynx: Good point, this is probably not safe as-is. I'll change it to a WrapOperation.

hannibal002 (approved):

PR #6160, opened 2026-07-21
Backend: Update dependencies

lunaynx:
## What
Updates a bunch of dependencies to likely non-breaking versions and simplifies some mapping-related logic (we never use anything other than Mojmap or unobfuscated anymore).

REI is intentionally left out of this, #6150 takes care of that. 

Detekt baseline had to be updated because of a bunch of new errors in existing code that are equivalent to already baselined errors.

+ Updated several dependencies to newer versions. - Luna


detekt/detekt.yml
```diff
@@ -136,6 +137,10 @@ ktlint:
         active: false
     TrailingCommaOnCallSite:
         active: false
+    # causes a lot of false positives with Stonecutter annotations forcing ugly indentation
+    # TODO implement a custom rule to handle this
+    Indentation:
+        active: false
```
CalMWolfs: This doesnt fit into this pr

CalMWolfs (changes requested):
all looks fine apart from this one thing

detekt/detekt.yml
lunaynx: Oops, you're right. I blindly copied over the new Detekt rules from the other PR.

gradle/wrapper/gradle-wrapper.jar
CalMWolfs, line 1: is this meant to be gone

CalMWolfs (changes requested):

gradle/wrapper/gradle-wrapper.jar
lunaynx: It's not gone, the new one is just smaller in size.

CalMWolfs (approved):

PR #6161, opened 2026-07-21
Backend: Fsync after atomic config save

AverageUser125:
## What
This is mostly just an idea that popped up. idk if it's a really good idea though.

Bassically,  the OS can be lazy and just, not write the config to disk after writing it,
Even after the atomic rename.
So I just force it to sync to disk.
This does reduce performance by ALOT.

I also removed the backup file, since it should never happen.
as the entire point of fsync is that the entire file must be written to the real disk.

+ Ensured config saving actually writes to the disk. - Avrg


AverageUser125:
Probably too much performance cost without the saving being queued.

PR #5402, opened 2026-03-12
Backend: Don't pass CallbackInfo outside mixins

lunaynx:
## What
Eliminated instances where mixins pass `CallbackInfo` or `CallbackInfoReturnable` to outside code. It is best to not expose raw control flow like this.

## Changelog Technical Details
+ Refactored mixins to not pass CallbackInfo or CallbackInfoReturnable to outside code. - Luna

PR #6162, opened 2026-07-22
test workflows

hannibal002:
## What
this is a copy of #5402 to test the workflows

ignore_from_changelog


hannibal002:
way better, but still duplicates that should collapse, and the two versions should also collpase? not sure, will keep investigating

PR #6166, opened 2026-07-22
Fix: Random CME in RiftWiltedBerberisHelper

AverageUser125:
## What
I could make the "list" itself not mutable and be atomic. but it's easier to to do this.
Could also do `CopyOnWriteArrayList` or do `synchronized(list)` I guess.

## Changelog Fixes
+ Fixed rare error while using the rift berberis helper. - Avrg


lunaynx:
Yeah, I'd probably just go for a `CopyOnWriteArrayList`, should be fine at this scale and doesn't add clutter at usage time. We can't do the usual `ConcurrentHashMap.newKeySet` shenanigans here because the `WiltedBerberis` objects do not have immutable identities, unless we figure out a way to give them one.

CalMWolfs (approved):

PR #6167, opened 2026-07-22
Improvement: Made CorpseTracker account resourceful perk

AverageUser125:
## What
Made it account for the fossil essence resourceful perk not consuming the corpse key.
Couldn't really test it since well, kinda hard to replicate.

the "resourceful" message appears after the loot message.
So I just did DelayedRun.runNextTickOld, which should work even the message appears in the next tick.
(since runNextTickOld is actually runNextTickEnd)

## Changelog Improvements
+ Made Corpse Tracker account for the key being saved by the Resourceful perk. - Avrg


CalMWolfs (approved):
couldnt test the perk triggering but the command worked and the code seemed fine

PR #6174, opened 2026-07-23
Backend: Cache detekt RepoPatternElement creation

AverageUser125:
## What
Most of the repo detekt rule all do `delegate.asRepoPatternElement()`. meaning it does that work 4x times.
This makes a centeral cache for all the rules so this work is not duplicated.
The actual number of unique delegates is about 1500, so I rounded it up to 2048.

I also made it cache null values/failure to parse.

+ Made repo pattern Detekt rule cache the parsing result between the rules. - Avrg


detekt/src/main/kotlin/repo/RepoPatternElement.kt
```diff
@@ -27,6 +29,8 @@ class RepoPatternElement private constructor(
     }
 
     companion object {
+        val SENTINEL_VALUE = RepoPatternElement("SENTINEL", null, "", emptyList(), emptyList())
```
hannibal002: why does this exist? please a comment above. also, does this need to be public?

detekt/src/main/kotlin/repo/RepoPatternContext.kt
```diff
@@ -0,0 +1,21 @@
+package repo
+
+import org.jetbrains.kotlin.psi.KtPropertyDelegate
+import repo.RepoPatternElement.Companion.asRepoPatternElement
+import java.util.IdentityHashMap
+
+class RepoPatternContext {
+    private val cache = IdentityHashMap<KtPropertyDelegate, RepoPatternElement>(2048)
```
hannibal002: does this handle async detekt threads?

hannibal002 (changes requested):

detekt/src/main/kotlin/repo/RepoPatternElement.kt
AverageUser125: Cause I didn’t want to do an “Any”. And I needed some way to mark “null” in the cache.
Imma make it internal though.

detekt/src/main/kotlin/repo/RepoPatternContext.kt
```diff
@@ -0,0 +1,31 @@
+package repo
+
+import org.jetbrains.kotlin.psi.KtPropertyDelegate
+import repo.RepoPatternElement.Companion.asRepoPatternElement
+import java.util.IdentityHashMap
```
hannibal002: unused import. detekt why are you slacking?

detekt/src/main/kotlin/repo/RepoPatternElement.kt
```diff
@@ -27,10 +29,14 @@ class RepoPatternElement private constructor(
     }
 
     companion object {
+        // A sentinel value to indicate that a KtPropertyDelegate has been processed but does not correspond to a valid RepoPatternElement.
+        // Must be public so that it can be accessed from RepoPatternContext.
```
hannibal002: desc said public, kotlin is internal

hannibal002 (changes requested):

hannibal002 (approved):

PR #6176, opened 2026-07-23
Improvement: Use Hypixel Items API for Minion XP

lunaynx:
## What
Hypixel recently added how much XP each item grants when collected from a minion to their API. This PR changes our feature to use the Hypixel API directly rather than the `MinionXP` SkyHanni-REPO constant.

## Changelog Improvements
+ Improved Minion XP Display to show the XP for newly added items right away. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/HypixelItemApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/HypixelItemApi.kt
@@ -1,80 +1,144 @@
 package at.hannibal2.skyhanni.features.inventory.bazaar
 
-import at.hannibal2.skyhanni.SkyHanniMod
+import at.hannibal2.skyhanni.SkyHanniMod.launch
 import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.config.ConfigManager
 import at.hannibal2.skyhanni.data.jsonobjects.other.SkyblockItemsDataJson
 import at.hannibal2.skyhanni.data.jsonobjects.repo.neu.NeuGeorgeJson
+import at.hannibal2.skyhanni.events.DebugDataCollectEvent
 import at.hannibal2.skyhanni.events.NeuRepositoryReloadEvent
 import at.hannibal2.skyhanni.features.rift.RiftApi
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
+import at.hannibal2.skyhanni.test.command.ErrorManager
+import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.ItemUtils
 import at.hannibal2.skyhanni.utils.NeuInternalName
 import at.hannibal2.skyhanni.utils.NeuItems
+import at.hannibal2.skyhanni.utils.SimpleTimeMark
+import at.hannibal2.skyhanni.utils.SkyBlockUtils
 import at.hannibal2.skyhanni.utils.api.ApiStaticGetPath
 import at.hannibal2.skyhanni.utils.api.ApiUtils
+import at.hannibal2.skyhanni.utils.coroutines.CoroutineSettings
 import at.hannibal2.skyhanni.utils.json.fromJson
+import kotlin.concurrent.atomics.AtomicBoolean
+import kotlin.time.Duration.Companion.hours
 import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
 
-class HypixelItemApi {
-    @SkyHanniModule
-    companion object {
-        // prices = george prices + npc prices
-        private var prices = mapOf<NeuInternalName, Double>()
-        private var npcPrices = mapOf<NeuInternalName, Double>()
-        private var georgePrices = mapOf<NeuInternalName, Double>()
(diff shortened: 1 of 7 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/HypixelItemApi.kt
```diff
@@ -44,28 +67,14 @@ class HypixelItemApi {
     private suspend fun loadItemData() {
         val (_, apiResponseData) = ApiUtils.getJsonResponse(hypixelItemStatic).assertSuccessWithData() ?: return
```
hannibal002: if assertSuccessWithData fails for networking issues, there is no recover. should retry periodically, like the bazaar data.

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/HypixelItemApi.kt
```diff
@@ -17,23 +17,46 @@ import at.hannibal2.skyhanni.utils.json.fromJson
...
-
         // prices = george prices + npc prices
         private var prices = mapOf<NeuInternalName, Double>()
         private var npcPrices = mapOf<NeuInternalName, Double>()
         private var georgePrices = mapOf<NeuInternalName, Double>()
+        private var minionStorageXp = mapOf<NeuInternalName, Map<String, Double>>()
 
         fun getNpcPrice(internalName: NeuInternalName) = prices[internalName]
 
+        fun getMinionStorageXp(internalName: NeuInternalName): Map<String, Double>? =
```
hannibal002: xp needs to be written as XP, as well as in all other cases in this pr

hannibal002 (changes requested):
> Minion XP values are now fetched directly from Hypixel's API.

that is not a improvement. that is a implementation detail. is there a user facing benefit of doing so? if so, name it. if not -> technical details

lunaynx:
> > Minion XP values are now fetched directly from Hypixel's API.
> 
> that is not a improvement. that is a implementation detail. is there a user facing benefit of doing so? if so, name it. if not -> technical details

Reworded it to a better user-facing explanation.

hannibal002 (changes requested):
bump

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/HypixelItemApi.kt
```diff
@@ -1,71 +1,144 @@
...
+        this.npcPrices = npcPrices
+        this.minionStorageXP = minionStorageXP
+        prices = georgePrices + npcPrices
     }
 
-    fun start() {
-        SkyHanniMod.launchIOCoroutine("hypixel item api fetch", timeout = 1.minutes) {
-            loadItemData()
-            prices = georgePrices + npcPrices
+    private fun onError(e: Exception) {
```
hannibal002, line 124: this should mimic the behaviour of HypixelItemApi.onError. no error message if `rawResponse == null || rawResponse == "{}"`, only chat message. nothing to report for the user to us here.

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/HypixelItemApi.kt
```diff
@@ -1,71 +1,144 @@
...
+        val userMessage = "Failed fetching item data from hypixel"
+        failedAttempts++
+        if (failedAttempts <= HIDDEN_FAILED_ATTEMPTS) {
+            nextFetchTime = SimpleTimeMark.now() + 15.seconds
+            ChatUtils.debug("$userMessage. (errorMessage=${e.message}, failedAttempts=$failedAttempts)")
+            e.printStackTrace()
```
hannibal002, line 130: dont print stack traces this way. always use ErrorManager.logErrorWithData for this. also, you can just wrap the whole function call behind a "is the user having the debug toggle enabled" check

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/HypixelItemApi.kt
hannibal002: when we now make the player wait 15 mins after couple misses, the user is left with missing/outdated data and no way to force a new trial without restarting/waiting. i suggest we add a new command similar to /shupdatebazaarprices

hannibal002 (changes requested):
also ive changed the changelog to be less "technical details", only the change users notice.

src/main/java/at/hannibal2/skyhanni/features/inventory/bazaar/HypixelItemApi.kt
```diff
@@ -1,71 +1,144 @@
...
+    private var lastSuccessfulFetch = SimpleTimeMark.farPast()
+    private var nextFetchTime = SimpleTimeMark.farPast()
+    private var failedAttempts = 0
 
-        @HandleEvent
-        fun onNeuRepoReload(event: NeuRepositoryReloadEvent) {
-            val constant = event.getConstant<NeuGeorgeJson>("george")
-            georgePrices = constant.prices ?: return
-            prices = georgePrices + npcPrices
+    fun getNpcPrice(internalName: NeuInternalName) = prices[internalName]
```
hannibal002, line 50: those two public function should explicitly mention their return type

hannibal002 (changes requested):
also the comments above are still open

PR #6178, opened 2026-07-23
Fix: Visitor supercraft resolving pumpkin recipe to werewolf sea creature

AverageUser125:
## What
The way it worked previously was quite funny.
But now I made it filter for just CRAFTING recipes.
And aslong as hypixel does not make some alternative recipe it should work.

## Changelog Fixes
+ Fixed Visitor Supercraft assuming killing a Werewolf Sea Creature is the only way to craft Enchanted Pumpkin. - Avrg


AverageUser125:
This is a very funny changelog atleast.

PR #6187, opened 2026-07-24
Fix + Backend: InventoryDetector Cleanup

AverageUser125:
## What
They changed the inventory name for /stats

## Changelog Fixes
+ Fixed not detecting /stats menu due to Hypixel changing the menu name. - Avrg
+ Fixed Enigma Soul Helper not detecting the menu. - Avrg

+ Replaced InventoryDetector direct string and lambda checks with repo-backed pattern groups where possible. - Avrg

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/garden/ComposterUpgradesData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/garden/ComposterUpgradesData.kt
@@ -47,7 +47,7 @@ object ComposterUpgradesData {
 
     @HandleEvent(onlyOnIsland = IslandType.GARDEN)
     fun onInventoryFullyOpened(event: InventoryFullyOpenedEvent) {
-        if (!composterInventoryPattern.matches(event.inventoryName)) return
+        if (!composterUpgradesInventoryPattern.matches(event.inventoryName)) return
         for (item in event.inventoryItems.values) {
             composterUpgradePattern.matchMatcher(item.cleanName) {
                 val name = group("name")
--- a/src/main/java/at/hannibal2/skyhanni/data/garden/ComposterUpgradesData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/garden/ComposterUpgradesData.kt
@@ -10,10 +10,12 @@ import at.hannibal2.skyhanni.utils.EnumUtils.enumJoinToPattern
 import at.hannibal2.skyhanni.utils.ItemUtils.cleanName
 import at.hannibal2.skyhanni.utils.NumberUtil.romanToDecimalIfNecessary
 import at.hannibal2.skyhanni.utils.RegexUtils.matchMatcher
+import at.hannibal2.skyhanni.utils.RegexUtils.matches
 import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
 
 @SkyHanniModule
 object ComposterUpgradesData {
+    private val patternGroup = RepoPattern.group("composter")
 
     /**
      * REGEX-TEST: Composter Speed II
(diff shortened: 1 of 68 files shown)


src/main/java/at/hannibal2/skyhanni/features/inventory/CurrentEquipmentApi.kt
```diff
@@ -38,7 +38,17 @@ enum class EquipmentSlot(val slot: Int, vararg val categories: ItemCategory) {
...
+
+    /**
+     * REGEX-TEST: Stats & Equipment
+     */
+    val equipmentAndStatsMenuPattern by patternGroup.pattern(
+        "equipment-and-stats-menu",
+        "Stats & Equipment",
+    )
+
+    val inventory = InventoryDetector(equipmentAndStatsMenuPattern)
```
CalMWolfs: I could be wrong but i think the last time i looked at the inventory detector code, constucting it with a repo pattern like this doesnt allow the pattern to actually be updataed through the repo so whatever the pattern was when it was compiled is what it will match (so loses the benefit of a repo pattern)

CalMWolfs (commented):
Not exactly related to your pr and doesnt have to be fixed in it but it would be good if this can be confirmed

src/main/java/at/hannibal2/skyhanni/features/inventory/CurrentEquipmentApi.kt
AverageUser125: Looking at the code, you are semi-correct.
It will only use the updated pattern when the game launches again.

Not if someone does /shupdaterepo.

AverageUser125: You want me to change all usage of the constructor with Pattern with just a function that is like:
```{ name -> equipmentAndStatsMenuPattern.matches(name) }```
?

File: src/main/java/at/hannibal2/skyhanni/data/garden/ComposterUpgradesData.kt [hunk 2]
```diff
@@ -27,9 +29,25 @@ object ComposterUpgradesData {
...
+     */
+    val composterUpgradesInventoryPattern by patternGroup.pattern(
+        "upgrades.inventory",
+        "Composter Upgrades",
+    )
+
     @HandleEvent(onlyOnIsland = IslandType.GARDEN)
     fun onInventoryFullyOpened(event: InventoryFullyOpenedEvent) {
-        if (event.inventoryName != "Composter Upgrades") return
+        if (!composterInventoryPattern.matches(event.inventoryName)) return
```
hannibal002: why did this here change from "composter upgrades" to "composter"? this looks unintended

src/main/java/at/hannibal2/skyhanni/features/foraging/ForagingBeaconSolver.kt
```diff
@@ -238,11 +249,14 @@ object ForagingBeaconSolver {
...
         },
-    ) { name ->
-        upgradingStrength = (name == "Upgrade Signal Strength")
-        val inInv = (name == "Tune Frequency" || upgradingStrength)
-        inInv
-    }
+        checkInventoryName = { name ->
+            beaconInventoryNamePattern.matchMatcher(name) {
+                upgradingStrength = group("upgrade") != null
+                true
```
hannibal002: this is dead code. you want to return here

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/foraging/ForagingBeaconSolver.kt
AverageUser125: it works as it, but I get why it's probably unclear.

src/main/java/at/hannibal2/skyhanni/features/foraging/ForagingBeaconSolver.kt
```diff
@@ -238,11 +249,14 @@ object ForagingBeaconSolver {
...
         },
-    ) { name ->
-        upgradingStrength = (name == "Upgrade Signal Strength")
-        val inInv = (name == "Tune Frequency" || upgradingStrength)
-        inInv
-    }
+        checkInventoryName = { name ->
+            beaconInventoryNamePattern.matchMatcher(name) {
+                upgradingStrength = group("upgrade") != null
+                return@InventoryDetector true
```
hannibal002: instead of the label, you might want to use true } ?: false

hannibal002 (changes requested):

hannibal002 (approved):

PR #6194, opened 2026-07-25
Backend: Fix mixin audit masking some failures on 1.21.11

lunaynx:
## What
This will probably become irrelevant in ~6 days, but just in case.

exclude_from_changelog


lunaynx:
Bump, 1.21.11 is still alive.

lunaynx:
Looks like there's an invalid mixin in the 26.2 PR that's not being caught even with this 😭

PR #6195, opened 2026-07-25
Backend: ChatFilter V2

AverageUser125:
## What
I can't for the life of me figure out a good API for it.
But this API works, it defers config stuff till ConfigLoadEvent happens and not earlier.
And ensuring no circular including and that all the objects are initalized at pre-init time for RepoPattern to work.
Also no "unused" problems.

Also made everything using RepoPattern and be colorless.

Also if some filter is disabled it's removed from the list of activeFilters and never even considered.
Same for a filter being island specific. theyt are not even processed.

If someone got a better API idea, please share.
As This is overengineered mess.

+ Split up ChatFilter logic into seperate files, and make it all repo patterns and colorless. - Avrg


AverageUser125:
There is not good architecture for this crap

PR #6196, opened 2026-07-25
Backend: Rate Limit Config saving

AverageUser125:
## What
The config should not be saved any faster then then some threshold.

Someone had QuiverApi be updated each frame, meaning AchievmentManager trying saving the updated achievment value each tick, meaning it trying saving the file every single tick.
And they got a huge lag cause of that.

Maybe making the file saving async is possible, but eh.

+ Added Rate Limiting for config saving. - Avrg


AverageUser125:
Luna got a better solution.

PR #6197, opened 2026-07-25
Fix: Queue config saves from AchievementManager

lunaynx:
## What
Fixes a user-reported issue saying their game lags every time they shoot an arrow, which turned out to be because AchievementManager synchronously saves the config to disk each time the fake quiver menu item updates.

Now, low-priority config saves (currently only AchievementManager) will be queued and dealt with by a background thread that runs once every 5 seconds and saves the config only if there is a pending config save, plus whenever the game is closed cleanly.

So, I guess, if you get a super rare once-in-a-lifetime achievement, do /shconfigsave real quick or pray your game doesn't crash within the next 5 seconds...

## Changelog Fixes
+ Fixed lag spikes when shooting arrows with achievements enabled. - Luna


AverageUser125:
tbh, I more thought of the direction that the ConfigType contains some boolean on whether it should get queued.
That way a storage can just be marked as "not crucial".

But this gives more control, so probably better.

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -79,9 +87,19 @@ class ConfigManager {
...
         // TODO use SecondPassedEvent
-        fixedRateTimer(name = "skyhanni-config-auto-save", daemon = true, period = 60_000L, initialDelay = 60_000L) {
-            saveConfig(ConfigFileType.FEATURES, "auto-save-60s")
+        fixedRateTimer(
+            name = "skyhanni-config-auto-save",
+            daemon = true,
+            period = QUEUED_SAVE_INTERVAL.inWholeMilliseconds,
+            initialDelay = QUEUED_SAVE_INTERVAL.inWholeMilliseconds,
+        ) {
+            if (lastFeatureAutoSave.passedSince() >= FEATURE_AUTO_SAVE_INTERVAL) {
```
hannibal002: errors in hte fixed rate timer will silently get ignored currently. please either migrate to the second passed event as the todo suggests or add a try catch with error manager calls here.

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
```diff
@@ -107,7 +107,8 @@ object SkyHanniMod : CompatCoroutineManager by SkyHanniCoroutineManager(
 
     @HandleEvent
     fun onClientShutdown() {
-        configManager.saveConfig(ConfigFileType.FEATURES, "shutdown-hook")
+        configManager.queueSave(ConfigFileType.FEATURES, "shutdown-hook")
```
hannibal002, line 122: scenario that might corrupt the config: the timer thread gets fired exactly in between queueSave and flushQueuedSaves. meaning , the "write file" process already started, and the queue below skips features. results in an corrupt or empty config file.

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -208,11 +226,30 @@ class ConfigManager {
     }
 
     fun saveConfig(fileType: ConfigFileType, reason: String) {
+        queuedSaves.remove(fileType)
```
hannibal002, line 212: why do we remove first? shoudnt we only remove after successful save file? at the end of the fun

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -208,11 +226,30 @@ class ConfigManager {
...
+    }
+
+    fun flushQueuedSaves() {
+        for (fileType in ConfigFileType.entries) {
+            val reason = queuedSaves[fileType] ?: continue
+            saveConfig(fileType, reason)
+        }
+    }
+
     private fun saveFile(file: File, fileName: String, data: Any, reason: String) {
```
hannibal002, line 236: the move from main thread to async thread for the save file logic means the `gson.toJson(data)` can produce a concurrent modification exception. thats not ideal, especially since we dont catch for CME in the try block.

hannibal002 (changes requested):
this pr touches sacred code here, we dont want a wave of config wipes, so I try to find every edge case that might happen that might cause race conditions eventually probably unlikely

hannibal002 (changes requested):
Do we need the background thread at all? The reported bug is that we save on every arrow, not that we save. Moving `flushQueuedSaves()` into `SecondPassedEvent` on the main thread already fixes the lag (one write per 5s instead of one per arrow) and removes the shutdown race, the CME risk and the swallowed exceptions in one go, plus it resolves the existing `// TODO use SecondPassedEvent`.

If we do want the write off the main thread, then `gson.toJson(data)` still has to run on the main thread and only the file write should be async, otherwise the CME stays.

lunaynx:
Sure, let's try that. If people still complain, we can explore a hardened background thread approach.

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
lunaynx: I think it's better to leave it at the beginning for now. Putting it at the end would be a behavioral change outside the scope of this PR: failed config saves would now be retried every 5s instead of 60s, and some of those are classes of errors that are not recoverable by just saving the same config again:
- If `jsonHolder` fails, that's basically a failure that's unrecoverable without a game restart, because we init it in `firstLoad` and never try to repair it if it somehow fails.
- If we get a `ClassCastException`, just trying to save again is almost certainly going to produce the same exception.
- The last case, `IOException`, could be plausibly retried and possibly work a second time.

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
```diff
@@ -104,9 +105,22 @@ object SkyHanniMod : CompatCoroutineManager by SkyHanniCoroutineManager(
         this.screenToOpen = null
     }
 
+    @HandleEvent
+    private fun onSecondPassed(event: SecondPassedEvent) {
+        if (!::configManager.isInitialized) return
+
+        if (!event.repeatSeconds(ConfigManager.QUEUED_SAVE_INTERVAL)) return
```
hannibal002, line 112: this if condition here should only box flushQueuedSaves, not the additional if below. otherwise re rely on one of hte timer value to be a multiplier of the other. unnecessary complexity

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
```diff
@@ -104,9 +105,22 @@ object SkyHanniMod : CompatCoroutineManager by SkyHanniCoroutineManager(
         this.screenToOpen = null
     }
 
+    @HandleEvent
+    private fun onSecondPassed(event: SecondPassedEvent) {
```
hannibal002, line 109: the event now fires on the first moment the user joins hypixel. thats not a good timing to start saving the file. earlier code was delaying by one minute with the first auto save

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
```diff
@@ -104,9 +105,22 @@ object SkyHanniMod : CompatCoroutineManager by SkyHanniCoroutineManager(
...
+        if (!::configManager.isInitialized) return
```
hannibal002, line 110: dead check. second passed event only fires on hypixel, the config gets initialized during setup phase, before the main screen event shows up.

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -208,11 +209,30 @@ class ConfigManager {
...
     fun saveConfig(fileType: ConfigFileType, reason: String) {
+        queuedSaves.remove(fileType)
         val json = jsonHolder[fileType] ?: error("Could not find json object for $fileType")
         saveFile(fileType.file, fileType.fileName, json, reason)
         saveFile(fileType.backupFile, fileType.fileName, json, reason)
     }
 
+    /**
+     * Marks [fileType] as dirty instead of writing it to disk right away. Repeated calls collapse into a single
+     * write done by the [onSecondPassed] event handler, at most once every [QUEUED_SAVE_INTERVAL].
```
hannibal002, line 220: onSecondPassed is private, broken link in kdocs. either qualify fully or not at all. currently this gets flagged by intellij as unresolves symbol

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -208,11 +209,30 @@ class ConfigManager {
     }
 
     fun saveConfig(fileType: ConfigFileType, reason: String) {
+        queuedSaves.remove(fileType)
         val json = jsonHolder[fileType] ?: error("Could not find json object for $fileType")
         saveFile(fileType.file, fileType.fileName, json, reason)
```
hannibal002, line 214: Now that this runs on the main thread, `saveConfig` serializes the whole config twice per save, and for `FEATURES` that happens inside the tick every 60 seconds.

Two ideas to solve that a bitt:

1. Serialize once in `saveConfig` and pass the resulting string to both `saveFile` calls. That halves the cost immediately and is worth doing on its own, independent of everything else.
2. Keep `gson.toJson` on the main thread, which is the only place the config objects are ever mutated, and move only the file write off it. That keeps the ConcurrentModificationException risk gone and takes the IO back out of the tick.

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
hannibal002: SecondPassedEvent only fires while on hypixel. since we have features that per settings work outside hypixel, it would be bad if those would not save if changed.

mabye use SkyyHanniTickEvent, it fires outside of skyblock and hypixel, e.g. on singleplayer.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
lunaynx: > SecondPassedEvent only fires while on hypixel

This sounds like a flaw in SecondPassedEvent itself to be honest, maybe we need a `@HandleEvent(onlyOnHypixel = true)` or something. But fixing that is probably outside the scope of this PR.

hannibal002: > But fixing that is probably outside the scope of this PR.

I agree.

hannibal002 (changes requested):
bump

lunaynx:
Reverted to the preexisting `fixedRateTimer` method for the time being, added a try-catch (preexisting issue, but trivial to solve) and a `@Synchronized` guard.

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
lunaynx: > **Finding 2 (CME in `gson.toJson`): not a regression, no code change.** The timer already serialized FEATURES off the client thread before this branch. The achievements map is a `LinkedHashMap` whose keys are all inserted at init, and later updates only replace existing keys or mutate fields, so no structural modification occurs. A CME in `toJson` throws before any file write, so it cannot corrupt output. The branch's `catch (Throwable)` in the timer is what keeps such an exception from silently killing the `java.util.Timer` thread, which was a pre-existing hazard. Moving serialization onto the client thread would reintroduce the lag spike the queue exists to avoid.

(The `catch` wasn't there at the time of this comment, to be fair, but it is now.)

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -73,15 +81,28 @@ class ConfigManager {
...
         configDirectory.mkdirs()
 
-
         for (fileType in ConfigFileType.entries) {
             val clazzInstance = fileType.clazz.getDeclaredConstructor().newInstance()
             setConfigHolder(fileType, firstLoadFile(fileType.file, fileType, clazzInstance))
         }
 
+        lastFeatureAutoSave = SimpleTimeMark.now()
         // TODO use SecondPassedEvent
```
hannibal002, line 90: the todo is now misguided. remove

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -73,15 +81,28 @@ class ConfigManager {
...
-        fixedRateTimer(name = "skyhanni-config-auto-save", daemon = true, period = 60_000L, initialDelay = 60_000L) {
-            saveConfig(ConfigFileType.FEATURES, "auto-save-60s")
+        fixedRateTimer(
```
hannibal002, line 91: `gson.toJson` still runs on the timer thread, so the CME risk from my earlier review is unchanged. Either serialize on the main thread and only write async, or say that you are fine with the risk.

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -204,15 +225,37 @@ class ConfigManager {
         return output
     }
 
+    @Synchronized
     fun saveConfig(fileType: ConfigFileType, reason: String) {
```
hannibal002, line 229: Both `saveFile` calls serialize the same object again, could be one `toJson` and two writes.

src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
hannibal002, line 174: > So, I guess, if you get a super rare once-in-a-lifetime achievement, do /shconfigsave real quick or pray your game doesn't crash within the next 5 seconds...

`/shconfigsave` only saves FEATURES, so it needs `flushQueuedSaves()` too.

hannibal002 (changes requested):
> + Fixed frequent config saves causing lag in some cases, e.g. when shooting an arrow. - Luna

Too technical, from the user side this is "fixed lag spikes when shooting arrows", regardless of the config.


`resetConfig()` in SkyHanniDebugsAndTests drops the queue as well, easiest is flushing inside `disableSaving()`.

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
lunaynx: The reason I'm confused here is that you're complaining about behavior that is already identical in `beta` today. I did not introduce it. It may very well be a valid concern worth fixing, but I don't understand why it's blocking this PR.

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
lunaynx: Valid, but once again, why is this blocking this pull request? I did not change that part whatsoever, it's the same as on `beta`.

hannibal002 (changes requested):
`resetConfig()` in SkyHanniDebugsAndTests calls `disableSaving()` without flushing, so queued achievement progress is dropped. Flushing inside `disableSaving()` covers every caller.

PR #6203, opened 2026-07-26
Fix + Backend: Pest Repellent Not Being Detected + EffectsApi Cleanup

AverageUser125:
## What

Changes Done:
1. The code had color codes for matching, so removed all that.
2. Hypixel changed it from "REGULAR" -> "Regular" for pest repellant, so did "uppercase"
3. It only checks for seconds when "55m" was possible and the such for pest repellant. so It now uses this partial data.
4. onInventoryFullyOpened didn't use the repo pattern for the inventory name

(If wanted any of this changes can either be split, or just removed)

## Changelog Fixes
+ Fixed Pest Repellent time not being detected. - Avrg

+ Made EffectApi matching colorless. - Avrg

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/effect/EffectApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/effect/EffectApi.kt
@@ -29,6 +29,7 @@ import at.hannibal2.skyhanni.utils.chat.TextHelper
 import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
 import net.minecraft.network.chat.Component
 import kotlin.time.Duration
+import kotlin.time.Duration.Companion.days
 import kotlin.time.Duration.Companion.hours
 import kotlin.time.Duration.Companion.minutes
 import kotlin.time.Duration.Companion.seconds
@@ -300,7 +301,8 @@ object EffectApi {
      * to the range implied by the displayed precision.
      *
      * Hypixel truncates omitted units:
-     * - 1h      -> [1h, 2h)
+     * - 1w      -> [1w, 2w)
+     * - 1d 5h   -> [1d5h, 1d6h)
      * - 1h 20m  -> [1h20m, 1h21m)
      * - 20m     -> [20m, 21m)
      * - 20s     -> [20s, 21s)
@@ -313,25 +315,17 @@ object EffectApi {
             return duration
         }
 
-        val hours = duration.inWholeHours
-        val minutes = duration.inWholeMinutes % 60
-        val seconds = duration.inWholeSeconds % 60
-
-        val hasHours = hours > 0
-        val hasMinutes = minutes > 0
-        val hasSeconds = seconds > 0
-
-        val lowerBound = duration
-
         val upperBound = when {
-            hasSeconds -> duration + 1.seconds
-            hasMinutes -> duration + 1.minutes
-            hasHours -> duration + 1.hours
+            duration.inWholeSeconds % 60 != 0L -> duration + 1.seconds
+            duration.inWholeMinutes % 60 != 0L -> duration + 1.minutes
+            duration.inWholeHours % 24 != 0L -> duration + 1.hours
+            duration.inWholeDays % 7 != 0L -> duration + 1.days
+            duration.inWholeDays > 0L -> duration + 7.days
             else -> duration + 1.seconds
         }
 
         return when {
-            existing < lowerBound -> lowerBound
+            existing < duration -> duration
             existing >= upperBound -> upperBound - 1.seconds
             else -> existing
         }
--- a/src/main/java/at/hannibal2/skyhanni/data/effect/EffectApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/effect/EffectApi.kt
@@ -16,6 +16,7 @@ import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.ItemUtils.cleanName
 import at.hannibal2.skyhanni.utils.ItemUtils.getCleanLore
 import at.hannibal2.skyhanni.utils.RegexUtils.firstMatcher
+import at.hannibal2.skyhanni.utils.RegexUtils.groupOrNull
 import at.hannibal2.skyhanni.utils.RegexUtils.matchAll
 import at.hannibal2.skyhanni.utils.RegexUtils.matchAllComponents
 import at.hannibal2.skyhanni.utils.RegexUtils.matchMatcher
@@ -70,7 +71,7 @@ object EffectApi {
      * REGEX-TEST: Filter
      */
     private val filterPattern by RepoPattern.pattern(
-        "inventory.effects.filter.colorlesss",
+        "inventory.effects.filter.colorless",
         "Filter",
     )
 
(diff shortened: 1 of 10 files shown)


src/main/java/at/hannibal2/skyhanni/features/misc/effects/NonGodPotEffectDisplay.kt
```diff
@@ -85,6 +87,44 @@ object NonGodPotEffectDisplay {
             EffectDurationChangeType.REMOVE -> {
                 effectDuration.remove(event.effect)
             }
+
+            EffectDurationChangeType.PARTIAL_SET -> {
+                val existing = effectDuration[event.effect]?.duration ?: Duration.ZERO
+                if (existing == Duration.ZERO) {
+                    effectDuration[event.effect] = Timer(duration)
+                    return
```
hannibal002: you sure we want to skip the update() call at the bottom of the function?

we maybe should extract the whole PARTIAL_SET case into a funciton, then the return can stay there unchanged

src/main/java/at/hannibal2/skyhanni/data/effect/EffectApi.kt
```diff
@@ -199,16 +218,14 @@ object EffectApi {
         if (!event.isWidget(TabWidget.PESTS)) return
 
         repellentPattern.firstMatcher(event.lines.map { it.string }) {
-            // Update repellent timer when near expiration to sync with the in-game countdown delay (which is slow)
-            val time = group("time")?.toIntOrNull() ?: return@firstMatcher
-            val tier = group("tier")
-            val duration = time.toDuration(DurationUnit.SECONDS)
+            val duration = TimeUtils.getDurationOrNull(group("time")) ?: return@firstMatcher
```
hannibal002: the regex group "time" is optional

src/main/java/at/hannibal2/skyhanni/data/effect/EffectApi.kt
```diff
@@ -25,11 +26,8 @@ import at.hannibal2.skyhanni.utils.SafeItemStack
...
 import at.hannibal2.skyhanni.utils.TimeUtils
 import at.hannibal2.skyhanni.utils.chat.TextHelper
-import at.hannibal2.skyhanni.utils.compat.formattedTextCompatLeadingWhiteLessResets
 import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
 import net.minecraft.network.chat.Component
-import kotlin.time.DurationUnit
-import kotlin.time.toDuration
 
 @SkyHanniModule
 object EffectApi {
```
hannibal002, line 40: you touch so much code in that file, please also do sweeping changes fixing the non matching primary function names for three listener functions in here, and the unused e in Exception

src/main/java/at/hannibal2/skyhanni/data/effect/EffectApi.kt
```diff
@@ -44,62 +42,64 @@ object EffectApi {
...
     )
 
     /**
-     * REGEX-TEST: §aFilter
+     * REGEX-TEST: Filter
      */
     private val filterPattern by RepoPattern.pattern(
-        "inventory.effects.filter",
-        "§aFilter",
+        "inventory.effects.filter.colorlesss",
```
hannibal002: one too many s

src/main/java/at/hannibal2/skyhanni/data/effect/EffectApi.kt
```diff
@@ -122,6 +122,26 @@ object EffectApi {
         "tab.salts-no-color",
         " (?<effect>[\\w\\-' ]+)*: *(?<time>[dhms0-9 ]+)",
     )
+
+    /**
+     * WRAPPED-REGEX-TEST: " Press TAB or type /effects to view your active effects!"
+     */
+    private val tabListFooterPattern by RepoPattern.pattern(
```
hannibal002: im pretty sure this pattern can never match. as far as i remember, this line here is a suffix to the normal message, not a standalone message. so not only is this pattern unused, all other patterns that previously were checking for this are now broken too! the old code was intentionally first stripping this suffix before parsing the message further. did hypixel change their formatting?

src/main/java/at/hannibal2/skyhanni/events/effects/EffectDurationChangeEvent.kt
```diff
@@ -13,5 +13,6 @@ class EffectDurationChangeEvent(
 enum class EffectDurationChangeType {
     ADD,
     REMOVE,
-    SET
+    SET,
```
hannibal002, line 31: missing explanation of what this is

src/main/java/at/hannibal2/skyhanni/data/effect/EffectApi.kt
```diff
@@ -133,32 +153,31 @@ object EffectApi {
...
 
         for (effect in NonGodPotEffect.entries) {
-            if (effect.effectRemovedPattern?.pattern() == modifiedMessage) {
+            if (effect.effectRemovedPattern?.pattern() == msg) {
                 EffectDurationChangeEvent(effect, EffectDurationChangeType.REMOVE, null).post()
                 return
             }
 
-            if (effect.effectGainedPattern?.pattern() != modifiedMessage) continue
+            if (effect.effectGainedPattern?.pattern() != msg) continue
```
hannibal002: commenting this here since you have not touched the file i need totalk about and intellij doesnt let me add comments there:

the line further down in the file `EffectDurationChangeEvent(nonGodPotEffect, EffectDurationChangeType.SET, duration).post()`:

should probably be partial set, not set. the relevant pattern tabEffectPattern has no preize time examples. if that is what i assume that set and partial set means.

hannibal002 (changes requested):
is the name repellent or repellant? code and pr desc dont align.

src/main/java/at/hannibal2/skyhanni/data/effect/EffectApi.kt
AverageUser125: Since there are multiple event listeners for the same event type, just different island annotation/requirement.
I can’t really do the PrimaryFunction thing.

AverageUser125: I mean, I could do JvmName thingy, but that seems like bad practice.

hannibal002: thats not a pattern we use so far. how about only use one event but call the other funcitons from there?

src/main/java/at/hannibal2/skyhanni/data/effect/EffectApi.kt
```diff
@@ -168,68 +208,68 @@ object EffectApi {
     }
 
     @HandleEvent(onlyOnSkyblock = true)
-    fun onTabUpdate(event: TablistFooterUpdateEvent) {
-        val footerLines = TextHelper.split(event.footer, "\n") ?: listOf(event.footer)
-        footerLines.readNonGodPotEffects()
+    private fun onTabUpdate(event: TablistFooterUpdateEvent) {
+        if (!activeEffectsFooterPattern.anyMatchesComponent(event.footer)) return
```
hannibal002, line 212: i cant find a way to test this. /shcopytablist does not show any footer info containing active effects. are you sure this is valid? have you tested this?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/effect/EffectApi.kt
AverageUser125: /shcopytablist DOES infact show the footer.
but the footer only exists if the "Active Effects" Widget is disabled and there is space.
But yeah, I run this with a debugger and tested that it works.

hannibal002 (approved):

PR #6204, opened 2026-07-26
Fix: Unbreak DialogUtils

lunaynx:
## What
Fixes `DialogUtils` not working ever since the original 1.21.5 port. On modern Minecraft, GLFW effectively owns the graphics stack, so it intentionally sets `java.awt.headless=true` to prevent compatibility issues. Forcing headless mode to false is not safe – Windows and Linux X11 could theoretically work, but macOS and Linux Wayland would definitely break.

The solution is to use LWJGL's bundled TinyFileDialogs library instead. This is more bare-bones and doesn't support formatting, but it does the job – boosted Jacob Contest crops are now indicated with a "(boosted)" suffix instead of bold, which is arguably more clear anyway.

The implementation uses a single shared coroutine with a mutex – both for ease and use, and because TinyFD uses global state and therefore it is not safe to open multiple dialogs at once. There's also a couple other guards in place to prevent unexpected behavior.

It can be tested using a new `/shtestdialog` command.

## Changelog Fixes
+ Fixed Jacob Contest Warning Popup always failing to open. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
@@ -436,16 +436,14 @@ object GardenNextJacobContest {
         TitleManager.sendTitle("§eFarming Contest!")
         SoundUtils.playBeepSound()
 
-        val cropTextNoColor = crops.joinToString(", ") {
-            if (it == boostedCrop) "<b>${it.cropName}</b>" else it.cropName
-        }
         if (config.warnPopup && !Minecraft.getInstance().isWindowActive) {
-            CoroutineSettings("garden jacob contest openPopupWindow").launchCoroutine {
-                DialogUtils.openPopupWindow(
-                    title = "SkyHanni Jacob Contest Notification",
-                    message = "<html>Farming Contest soon!<br />Crops: $cropTextNoColor</html>",
-                )
+            val cropTextNoColor = crops.joinToString(", ") {
+                if (it == boostedCrop) "${it.cropName} (boosted)" else it.cropName
             }
+            DialogUtils.openPopupWindow(
+                title = "SkyHanni Jacob Contest Notification",
+                message = "Farming Contest soon!\nCrops: $cropTextNoColor",
+            )
         }
     }
 
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
@@ -57,8 +57,8 @@ import at.hannibal2.skyhanni.utils.renderables.primitives.ItemStackRenderable.Co
 import at.hannibal2.skyhanni.utils.renderables.primitives.text
 import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
 import com.google.gson.JsonPrimitive
-import kotlinx.coroutines.sync.Mutex
 import net.minecraft.client.Minecraft
+import kotlinx.coroutines.sync.Mutex
 import kotlin.time.Duration.Companion.hours
 import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.Duration.Companion.minutes
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
@@ -57,12 +57,12 @@ import at.hannibal2.skyhanni.utils.renderables.primitives.ItemStackRenderable.Co
 import at.hannibal2.skyhanni.utils.renderables.primitives.text
 import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
 import com.google.gson.JsonPrimitive
+import kotlinx.coroutines.sync.Mutex
 import net.minecraft.client.Minecraft
 import kotlin.time.Duration.Companion.hours
 import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.Duration.Companion.minutes
 import kotlin.time.Duration.Companion.seconds
-import kotlinx.coroutines.sync.Mutex
 
 @SkyHanniModule
 object GardenNextJacobContest {
@@ -145,7 +145,7 @@ object GardenNextJacobContest {
     )
 
     @HandleEvent
-    fun onDebugDataCollect(event: DebugDataCollectEvent) {
+    private fun onDebugDataCollect(event: DebugDataCollectEvent) {
         event.title("Garden Next Jacob Contest")
 
         if (!GardenApi.inGarden()) {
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/utils/DialogUtils.kt
```diff
@@ -1,46 +1,94 @@
...
+    private fun messageBox(title: String, message: String): Boolean {
+        //? if >= 26.1 {
+        return TinyFileDialogs.tinyfd_messageBox(title, message, "ok", "info", 1) != 0
+        //?} else {
+        /*return TinyFileDialogs.tinyfd_messageBox(title, message, "ok", "info", true)
+        *///?}
+    }
+
+    @HandleEvent
+    private fun onCommandRegistration(event: CommandRegistrationEvent) {
```
hannibal002, line 93: dont use private (we want to migrate, but the migration is not done yet so this is inconsistent and may even break)

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/DialogUtils.kt
lunaynx: What? Why would I *not* use a private event handler in a utility class? This is arguably one of the most important places where we *should* use private event handlers to not leak it to callers. It's not going to stop us from moving it elsewhere in the future.

lunaynx: Per Discord, this should not break anything, other than the SkyHanni IntelliJ plugin giving you a now false warning. Any logic or checks within SkyHanni itself preventing private event handlers from working correctly are already corrected or removed. Detekt didn't flag it either, so there's no remaining rule.

hannibal002: working on the intellij plugin issue:

image

hannibal002: the intellij plugin is now fixed, and also got the first release: https://github.com/hannibal002/SkyHanniDevelopment/releases

src/main/java/at/hannibal2/skyhanni/utils/DialogUtils.kt
```diff
@@ -1,46 +1,94 @@
...
-        )
-    }.onFailure { e ->
-        ErrorManager.logErrorWithData(
-            e, "Failed to open a popup window",
-            "message" to message,
-        )
+    private fun currentBackend(): String? = TinyFileDialogs.tinyfd_getGlobalChar("tinyfd_response")
+
+    /**
+     * Opens a modal message box outside of the game window.
```
hannibal002: the word "of" can go

src/main/java/at/hannibal2/skyhanni/utils/DialogUtils.kt
```diff
@@ -1,46 +1,94 @@
...
+                ErrorManager.logErrorWithData(
+                    e, "Failed to open a popup window",
+                    "title" to title,
+                    "message" to message,
+                )
+            }
+        }
+    }
+
+    private fun String.stripQuotes(): String = filterNot { it in forbiddenCharacters }
```
hannibal002: the term "quotes" is a bit confusing or misleading. how about rename to stripForbiddenChars?

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
```diff
@@ -436,16 +436,14 @@ object GardenNextJacobContest {
...
         if (config.warnPopup && !Minecraft.getInstance().isWindowActive) {
-            CoroutineSettings("garden jacob contest openPopupWindow").launchCoroutine {
-                DialogUtils.openPopupWindow(
-                    title = "SkyHanni Jacob Contest Notification",
-                    message = "<html>Farming Contest soon!<br />Crops: $cropTextNoColor</html>",
-                )
+            val cropTextNoColor = crops.joinToString(", ") {
+                if (it == boostedCrop) "${it.cropName} (boosted)" else it.cropName
             }
+            DialogUtils.openPopupWindow(
```
hannibal002: the logic runs async, maybe a bit delayed. so the condition above !Minecraft.getInstance().isWindowActive might not be valid anymore. intended?

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
```diff
@@ -436,16 +436,14 @@ object GardenNextJacobContest {
         TitleManager.sendTitle("§eFarming Contest!")
         SoundUtils.playBeepSound()
 
-        val cropTextNoColor = crops.joinToString(", ") {
-            if (it == boostedCrop) "<b>${it.cropName}</b>" else it.cropName
-        }
         if (config.warnPopup && !Minecraft.getInstance().isWindowActive) {
```
hannibal002: use early return

hannibal002 (changes requested):
> + Fixed Jacob Contest Warning Popup feature not working. - Luna

"not working" is a bit vague. how about using "not appearing" or "failing to open" or something?

src/main/java/at/hannibal2/skyhanni/utils/DialogUtils.kt
lunaynx: I mean, what kind of quotes would it be stripping, Shakespeare quotes? But sure, renamed.

src/main/java/at/hannibal2/skyhanni/features/garden/GardenNextJacobContest.kt
lunaynx: Yeah fair, especially because of the mutex. I added a secondary condition check inside the coroutine.

src/main/java/at/hannibal2/skyhanni/utils/DialogUtils.kt
lunaynx: Oh wait, it includes backticks too, yeah, good call then.

src/main/java/at/hannibal2/skyhanni/utils/DialogUtils.kt
```diff
@@ -1,46 +1,95 @@
...
-            baseFrame, message, title,
-            JOptionPane.DEFAULT_OPTION, JOptionPane.INFORMATION_MESSAGE, null,
-            listOf(okButton).toTypedArray(), okButton,
-        )
-    }.onFailure { e ->
-        ErrorManager.logErrorWithData(
-            e, "Failed to open a popup window",
-            "message" to message,
-        )
+    fun openPopupWindow(title: String, message: String, condition: () -> Boolean = { true }) = popupCoroutine.launch {
```
hannibal002: please specify the return type for public functions

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/DialogUtils.kt
lunaynx: In this case we don't actually *need* the return value and I'm only really using the expression body to avoid double nested braces, but sure.

hannibal002 (approved):

hannibal002:
> GardenNextJacobContest.kt:60: Imports must be ordered in lexicographic order with "java", "javax", "kotlin", "kotlinx" and aliases in the end.

my intellij cant auto reforamt this. is detekt dumb or is my intellij missing something?

lunaynx:
IntelliJ's defaults are a little silly, will look if we can safely commit custom import rules to the repo later.

hannibal002:
id say we either find a way to automatically (via gradle?) or the intlellij plugin tweak the default intellij rules, or we tweak detekt until it follows whatever intellij uses as default.

PR #6206, opened 2026-07-26
Fix: Anita Extra Farming Fortune Helper

AverageUser125:
## What
Literally nothing was writing what the value was.
So I just used the lore + tooltip to figure it out.
And also pulled out the lore parsing to a seprate function.
Also made everything use the REPO.

## Changelog Fixes
+ Fixed Anita's Extra Farming Fortune Helper not showing. - Avrg

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/garden/inventory/AnitaExtraFarmingFortune.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/garden/inventory/AnitaExtraFarmingFortune.kt
@@ -63,6 +63,14 @@ object AnitaExtraFarmingFortune {
         "You have: \\+(?<farmingFortune>\\d+)${SkyblockStat.FARMING_FORTUNE.hypixelIcon} Farming Fortune",
     )
 
+    /**
+     * REGEX-TEST: Click to trade!
+     */
+    private val clickToTradePattern by patternGroup.pattern(
+        "clicktotrade",
+        "Click to trade!",
+    )
+
     private var levelPrice = mapOf<Int, AnitaUpgradePrice>()
 
     @HandleEvent
@@ -87,7 +95,7 @@ object AnitaExtraFarmingFortune {
         }
         jacobTickets = (contributionFactor * jacobTickets).toInt()
 
-        val index = event.toolTip.lastIndex - 1
+        val index = event.toolTip.indexOfFirst { clickToTradePattern.matches(it) }.let { it - 1 }
         if (index < 0) return
 
         val price = jacobTickets * "JACOBS_TICKET".toInternalName().getPrice()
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/inventory/AnitaExtraFarmingFortune.kt
```diff
@@ -65,31 +95,66 @@ object AnitaExtraFarmingFortune {
...
+            )
         }
     }
 
+    private fun parseExtraFarmingFortuneLore(
+        lore: List<String>,
+    ): ExtraFarmingFortuneLore? {
+        val farmingFortune = farmingFortunePattern.firstMatcher(lore) {
+            group("farmingFortune").toInt()
+        } ?: 0
```
hannibal002, line 129: what does the zero mean here?  if there is no lore, do we want to override the value in the storage by 0? is this intended?

src/main/java/at/hannibal2/skyhanni/features/garden/inventory/AnitaExtraFarmingFortune.kt
```diff
@@ -65,31 +95,66 @@ object AnitaExtraFarmingFortune {
...
+                "§7Current Tier: §e$anitaUpgrade/${levelPrice.size}",
+            )
         }
     }
 
+    private fun parseExtraFarmingFortuneLore(
+        lore: List<String>,
+    ): ExtraFarmingFortuneLore? {
+        val farmingFortune = farmingFortunePattern.firstMatcher(lore) {
+            group("farmingFortune").toInt()
```
hannibal002: do not use toInt. use formatInt() from NumberUtil

src/main/java/at/hannibal2/skyhanni/features/garden/inventory/AnitaExtraFarmingFortune.kt
```diff
@@ -65,31 +95,66 @@ object AnitaExtraFarmingFortune {
...
+        } ?: 0
+
+        var contributionFactor = realAmountPattern.firstMatcher(lore) {
+            group("realAmount").formatDouble()
+        } ?: return null
```
hannibal002: do we need to return? dont we want to just keep the default factor to be 1?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/inventory/AnitaExtraFarmingFortune.kt
AverageUser125: if there is no lore, that means you never upgraded the perk. which means it's level 0.
So yes. intentional.

hannibal002 (approved):

PR #6211, opened 2026-07-27
Backend: Private Event Listener Detekt

AverageUser125:
## What
As discussed on discord.
This rule says all event listeners shall be private.
(Should I call them listeners or handlers?)
I baselined the rule for now.

+ Added Detekt rule to enforce private event listeners. - Avrg

### DIFF
--- a/detekt/src/main/kotlin/style/PrivateEventListener.kt
+++ b/detekt/src/main/kotlin/style/PrivateEventListener.kt
@@ -15,7 +15,7 @@ class PrivateEventListener(config: Config) : SkyHanniRule(
     override fun visitNamedFunction(function: KtNamedFunction) {
         if (function.hasAnnotation("HandleEvent") && !isEffectivelyPrivate(function)) {
             function.reportIssue(
-                "Event listener functions should be private. Please change the visibility of this function to private.",
+                "Event listener functions should be private.",
             )
         }
         super.visitNamedFunction(function)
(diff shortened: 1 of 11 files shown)


hannibal002:
> Should I call them listeners or handlers?

Listeners

> I baselined the rule for now.

That's fine, ideally we never touch the baseline file ever manually, only to remove already fixed stuff. The sweeping changes will happen mostly via the IntelliJ plugin, this Detekr rule is for new code mostly imo.

detekt/src/main/kotlin/style/PrivateEventListener.kt
```diff
@@ -0,0 +1,31 @@
...
+import utils.DetektUtils.hasAnnotation
+
+class PrivateEventListener(config: Config) : SkyHanniRule(
+    config,
+    "Event listener should be private.",
+) {
+    override fun visitNamedFunction(function: KtNamedFunction) {
+        if (function.hasAnnotation("HandleEvent") && !isEffectivelyPrivate(function)) {
+            function.reportIssue(
+                "Event listener functions should be private. Please change the visibility of this function to private.",
```
hannibal002: remove second sentence.

hannibal002 (changes requested):
this is the only issue i see

hannibal002:
ups wrong pr

hannibal002 (approved):

PR #6216, opened 2026-07-27
Backend: Make @HandleEvent outside a @SkyHanniModule a hard error

lunaynx:
## What
Currently, if you annotate a function with `@HandleEvent` and the declaring class is *not* annotated with `@SkyHanniModule`, it silently fails – the event handler is never registered. This PR makes that a hard error by adding two guards: KSP validation and `crashInDevEnv`. If somehow such code makes it into production runtime, the event handler is still silently skipped, consistent with the existing behavior around invalid event handlers.

Note that this required moving `@SkyHanniModule` into the main source set for proper verification.

This change also surfaced three config migrations that never ran due to this previously being a silent failure. They have been corrected, so anyone still updating from older versions now will get the intended migrations, but it will be a no-op for players already on newer versions.

+ Made \@HandleEvent in non-\@SkyHanniModule classes fail loudly instead of silently. - Luna


annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/ModuleProcessor.kt
```diff
@@ -89,6 +96,33 @@ class ModuleProcessor(
...
+     */
+    private fun validateModuleMembership(resolver: Resolver) {
+        val functions = processBuildPaths(
+            resolver.getSymbolsWithAnnotation(HANDLE_EVENT)
+                .filterIsInstance<KSFunctionDeclaration>()
+                .toList(),
+        )
+        for (function in functions) {
+            val parent = function.parentDeclaration as? KSClassDeclaration
+            if (parent?.annotations?.any { it.shortName.asString() == "SkyHanniModule" } == true) continue
```
hannibal002: do we want this? `if (parent?.annotations?.any { it.annotationType.resolve().declaration.qualifiedName?.asString() == SKYHANNI_MODULE } == true)
continue`

would be more precise. but your call, if this KSP perf hit is worth it.

hannibal002 (changes requested):

annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/ModuleProcessor.kt
lunaynx: I benchmarked 10 builds of each, doesn't seem too bad.
* Before: avg. 19.642s
* After: avg. 20.609s

lunaynx: Actually, there's more things using `shortName`, let me benchmark with all of them changed.

lunaynx: Okay, now it's averaging at 19.245s so I guess there's basically no performance hit at all, just a little contention, since I'm running on a quite beefy droplet but it *is* shared CPU.

annotation-processors/src/main/kotlin/at/hannibal2/skyhanni/skyhannimodule/ModuleProcessor.kt
```diff
@@ -116,46 +151,52 @@ class ModuleProcessor(
...
-                                "with @HandleEvent",
+                val handleEvent = function.annotations.find { it.getQualifiedName() == HANDLE_EVENT }
+                    ?: continue
+
+                val event = skyHanniEvent ?: return symbol
+                val eventParameterType = function.extensionReceiver?.resolve()
+                    ?: function.parameters.firstOrNull()?.type?.resolve()
+                val parameterCount = function.parameters.size + if (function.extensionReceiver != null) 1 else 0
+                val hasPrimaryFunction = function.simpleName.asString() in primaryFunctionNames
+                val hasExplicitEventSpec = handleEvent?.hasExplicitEventSpec() == true
```
hannibal002: handleEvent is non null, remove unnecessary nullability checks

hannibal002 (changes requested):

hannibal002 (approved):

PR #6217, opened 2026-07-27
Backend: BarnFishingTimer INFINITE timer fix

AverageUser125:
## What
If it's somehow unable to find the oldest, it will just reset the display.
I got no proof this actually fixes the issue, so no changelog.
Also I think it's better to just reset the state then have a broken state.

exclude_from_changelog


AverageUser125:
Issue couldn’t be replicated

PR #6218, opened 2026-07-27
Backend: fix MinecraftCompat Detekt Rule

AverageUser125:
## What
I copied the ForbiddenMethodCall from the offiicial detekt, and just changed a few lines to suit our needs.

+ Switched MinecraftCompat detekt rule to ForbiddenMethodCall. - Avrg


detekt/src/main/kotlin/compat/MinecraftCompat.kt
```diff
@@ -4,49 +4,72 @@ import SkyHanniRule
...
+        "gui" to "MinecraftCompat.hud",
+        "options.hideGui" to "MinecraftCompat.hideGui",
+        "levelRenderer.allChanged()" to "MinecraftCompat.reloadChunks()",
+    )
+
+    private val replacements = buildList {
+        listOf(
+            "Minecraft.getInstance()",
+            "mc",
+            "client",
```
hannibal002: this is too broad. stuff like "val client = HttpClient()" would accidentally also get flagged that way. should contain a type check as well

detekt/src/main/kotlin/compat/MinecraftCompat.kt
```diff
@@ -4,49 +4,72 @@ import SkyHanniRule
...
+                add(
+                    Regex(
+                        """\b${Regex.escape("$prefix.$access")}(?=\.|\(|$)"""
+                    ) to replacement
+                )
+            }
+        }
+    }
+
+    private val allowedPaths = listOf(
```
hannibal002: why? cant we just use @Suppress("MinecraftCompat")

detekt/src/main/kotlin/compat/MinecraftCompat.kt
```diff
@@ -4,49 +4,72 @@ import SkyHanniRule
...
+    private val replacements = buildList {
+        listOf(
+            "Minecraft.getInstance()",
+            "mc",
+            "client",
+        ).forEach { prefix ->
+            minecraftReplacements.forEach { (access, replacement) ->
+                add(
+                    Regex(
+                        """\b${Regex.escape("$prefix.$access")}(?=\.|\(|$)"""
```
hannibal002: does not include null safety support (mc.level?.let). am i missing something or would nullability break the detection?

hannibal002 (changes requested):

AverageUser125:
I should probably look at what these are called in the 26.2 PR.
But it’s fine.

detekt/src/main/kotlin/compat/MinecraftCompat.kt
AverageUser125: I forgot those exist

AverageUser125:
Ignore that.

AverageUser125:
Should probably just be part of the intelij plugin.

PR #6220, opened 2026-07-27
Backend: Add IntelliJ required plugins + custom import order

lunaynx:
## What
Adds "Required Plugins" for IntelliJ (as far as I can tell, this is not "brick the editor if the plugins are not installed" – it is "yell at the user if they're missing one of these plugins"), and added our custom import ordering rules to `.editorconfig`, which is a nice place to put them because it can be put in VCS and won't get overwritten by the user messing with their IntelliJ settings.

exclude_from_changelog

### DIFF
--- a/.idea/externalDependencies.xml
+++ b/.idea/externalDependencies.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ExternalDependencies">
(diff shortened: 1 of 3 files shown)


.idea/externalDependencies.xml
hannibal002, line 1: lf detekt

hannibal002 (changes requested):

hannibal002 (approved):

PR #6224, opened 2026-07-28
Fix: Deployable Reminder depends on Deployable Display

AverageUser125:
## What
It wouldn't detect the deployables if the display was turned off.
The check itself is expensive so idk if a good idea to make it always enabled.

## Changelog Fixes
+ Fixed deployable reminder not detecting deployables when deployable display was turned off. - Avrg


hannibal002 (changes requested):
the changes done in the pr are good, just not enough to actually fix the issues here.

getActiveDeployables() does not filter if the entry is actulally active

AverageUser125:
The variable is called "activeDeployables" so I assumed it's only for entires that are actually active.

hannibal002:
sadly no, this is the list of all deployables that are actively known about. there is an extra .isActive() funciton. if every entry in there would be active, the fun would just return true

hannibal002 (approved):

PR #6225, opened 2026-07-28
Backend + Fix: Make CustomWardrobe an actual Screen

AverageUser125:
## What
makes the CustomWardrobe an actual screen, and not just redirecting the Container rendering.
Thanks Aaron for the code examples!

I did have to make the edit menu an actual screen too to avoid sending a packet when leaving edit mode back into the custom menu.
(Also means no sound)

I made a seperate PR for moving the rendering logic to the screen and a utility class.

## Changelog Fixes
+ Fixed other mods' inventory buttons showing in the Custom Wardrobe. - Avrg

+ Made CustomWardrobe an actual Screen instead of just overwriting the container rendering. - Avrg
+ Made it easy to add item list support to custom SkyHanni UIs. - Avrg


src/main/java/at/hannibal2/skyhanni/mixins/transformers/MenuScreensConstructorMixin.java
```diff
@@ -0,0 +1,38 @@
...
+        T containerMenu = type.create(id, player.getInventory());
+        String inventoryName = name.getString().trim();
+
+        if (containerMenu instanceof ChestMenu chestMenu) {
+            if (CustomWardrobeScreen.shouldReplace(inventoryName)) {
+                client.player.containerMenu = chestMenu;
+                MinecraftCompat.setScreen(new CustomWardrobeScreen(chestMenu, name));
+                ci.cancel();
+            }
+        }
```
lunaynx: This should probably be replaced with a hook written in Kotlin to avoid putting too much logic in mixins, something like this:
```kt
        if (CustomWardrobeScreen.tryReplace(name, type, client, id)) {
            ci.cancel();
        }
```

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
```diff
@@ -0,0 +1,93 @@
...
+            return true
+        }
+        return false
+    }
+
+    // Called every client tick to ensure player safety
+    override fun tick() {
+        super.tick()
+        val player = MinecraftCompat.localPlayerOrNull ?: return
+        if (!player.isAlive || player.isRemoved) {
```
lunaynx: - The `isRemoved` check is redundant because `!isAlive` already checks for that: https://mcsrc.dev/1/26.1.2/net/minecraft/world/entity/LivingEntity#L1717
- This check is largely useless because you can never "die" in the traditional sense on SkyBlock except via a glitch. Dying just means getting teleported to spawn.
- I question if this whole function is actually needed at all. The game and/or server should already be handling automatically closing the container as needed, and we should respect that and automatically close our custom GUI when the underlying container is closed.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
```diff
@@ -0,0 +1,93 @@
...
+    // Capture inventory key press (e.g. 'E') to trigger safe close sequence
+    override fun keyPressed(input: KeyEvent): Boolean {
+        if (super.keyPressed(input)) {
+            return true
+        } else if (Minecraft.getInstance().options.keyInventory.matches(input)) {
+            this.onClose()
+            return true
+        }
+        return false
+    }
```
lunaynx: Checking for this specifically feels hacky. Surely there's a better way to pass through all inputs other than slot actions (we send those ourselves, since we obviously don't want a player clicking or pressing their drop key over where the original invisible container's item happens to be to trigger an action).

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
AverageUser125: I copied Skyblocker code, so no clue.
And aaron said closing is very particular.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
AverageUser125: ditto

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MenuScreensConstructorMixin.java
AverageUser125: I mean, the entire "editMode" feature is not working with this concept currently. but fair enough.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
```diff
@@ -0,0 +1,94 @@
...
+
+    // Called every client tick to ensure player safety
+    override fun tick() {
+        super.tick()
+        val player = MinecraftCompat.localPlayerOrNull ?: return
+        if (!player.isAlive || player.isRemoved) {
```
lunaynx: At least change this 🥺 I know Skyblocker is doing it too, but the second check can never succeed if the first one failed, so it's just dead code. (Confirmed on 1.21.11 & 26.2 too.)

```suggestion
        if (!player.isAlive) {
```

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
AverageUser125: okay, time to send a patch to Minecraft and tell them to fix `AbstractContainerScreen`.
(Which is where all this weird logic comes from)

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/AbstractCustomMenuScreen.kt
```diff
@@ -0,0 +1,65 @@
...
+        handler.removeSlotListener(this)
+        handler = newHandler
+        handler.addSlotListener(this)
+        ToolTipData.lastSlot = null
+    }
+
+    override fun keyPressed(input: KeyEvent): Boolean {
+        if (super.keyPressed(input)) {
+            return true
+        } else if (Minecraft.getInstance().options.keyInventory.matches(input)) {
```
lunaynx: Can be written as early returns.
```suggestion
        }
        if (Minecraft.getInstance().options.keyInventory.matches(input)) {
```

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/AbstractCustomMenuScreen.kt
```diff
@@ -0,0 +1,65 @@
...
+    override fun onClose() {
+        MinecraftCompat.localPlayerOrNull?.closeContainer()
+        super.onClose()
+    }
+
+    override fun removed() {
+        MinecraftCompat.localPlayerOrNull?.let {
+            handler.removed(it)
+            handler.removeSlotListener(this)
+        }
```
lunaynx: Can avoid nesting here.
```suggestion
        val player = MinecraftCompat.localPlayerOrNull ?: return
        handler.removed(player)
        handler.removeSlotListener(this)
```

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeEditScreen.kt
```diff
@@ -0,0 +1,61 @@
...
+import at.hannibal2.skyhanni.utils.renderables.Renderable
+import net.minecraft.client.gui.GuiGraphicsExtractor
+import net.minecraft.client.gui.screens.inventory.ContainerScreen
+import net.minecraft.network.chat.Component
+import net.minecraft.world.inventory.ChestMenu
+import java.awt.Color
+
+class CustomWardrobeEditScreen(
+    handler: ChestMenu,
+    inventory: net.minecraft.world.entity.player.Inventory,
```
lunaynx: Import instead of writing out the fully qualified name.
```suggestion
    inventory: Inventory,
```

lunaynx (changes requested):

AverageUser125:
Luna, wait a second. I am trying to move the rendering logic to CustomWardrobeScreen.

hannibal002:
are those fixes hard bound to the big change in this pr? otherwise, would be better to separate them in a small pr

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobe.kt
```diff
@@ -46,45 +38,73 @@ import at.hannibal2.skyhanni.utils.renderables.primitives.text
...
     val config: CustomWardrobeConfig get() = SkyHanniMod.feature.inventory.customWardrobe
 
+    internal var switchingScreens = false
+    val inCustomWardrobe get() = MinecraftCompat.screen is CustomWardrobeScreen
+    val editMode get() = MinecraftCompat.screen is CustomWardrobeEditScreen
+
     private var displayRenderable: Renderable? = null
     private var inventoryButton: Renderable? = null
-    var editMode = false
     private var waitingForInventoryUpdate = false
```
hannibal002, line 54: this value never gets set to false again after getting set to true once

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
```diff
@@ -0,0 +1,62 @@
...
+import net.minecraft.client.gui.navigation.ScreenRectangle
+import net.minecraft.network.chat.Component
+import net.minecraft.world.inventory.AbstractContainerMenu
+import net.minecraft.world.inventory.ChestMenu
+
+class CustomWardrobeScreen(
+    menu: ChestMenu,
+    title: Component,
+) : AbstractCustomMenuScreen(menu, title) {
+    var waitingForInventoryUpdate = false
```
hannibal002: this is dead

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
```diff
@@ -0,0 +1,62 @@
...
+            CustomWardrobe.update()
+        }
+    }
+
+    override fun removed() {
+        val player = MinecraftCompat.localPlayerOrNull ?: return
+        if (!CustomWardrobe.switchingScreens) {
+            menu.removed(player)
+        }
+        menu.removeSlotListener(this)
```
hannibal002: do we want to omit the removeSlotListener call if the player is null? why? why not?

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobe.kt
```diff
@@ -46,45 +38,73 @@ import at.hannibal2.skyhanni.utils.renderables.primitives.text
...
 
-@SkyHanniModule
+@Suppress("TooManyFunctions")
 object CustomWardrobe {
-
     val config: CustomWardrobeConfig get() = SkyHanniMod.feature.inventory.customWardrobe
 
+    internal var switchingScreens = false
+    val inCustomWardrobe get() = MinecraftCompat.screen is CustomWardrobeScreen
+    val editMode get() = MinecraftCompat.screen is CustomWardrobeEditScreen
```
hannibal002, line 51: unused

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobe.kt
```diff
@@ -46,45 +38,73 @@ import at.hannibal2.skyhanni.utils.renderables.primitives.text
...
+
     private var displayRenderable: Renderable? = null
     private var inventoryButton: Renderable? = null
```
hannibal002: unused

src/main/java/at/hannibal2/skyhanni/utils/AbstractCustomMenuScreen.kt
```diff
@@ -0,0 +1,68 @@
...
+        if (!player.isAlive) {
+            player.closeContainer()
+        }
+    }
+
+    override fun guiClosed() {
+        MinecraftCompat.localPlayerOrNull?.closeContainer()
+    }
+
+    override fun removed() {
```
hannibal002, line 62: when does this get fired?

src/main/java/at/hannibal2/skyhanni/mixins/hooks/MenuScreensHook.kt
```diff
@@ -0,0 +1,43 @@
...
+        type: MenuType<T>,
+        client: Minecraft,
+        id: Int,
+    ): Boolean {
+        val player = client.player ?: return false
+        if (!SkyBlockUtils.inSkyBlock) return false
+
+        val inventory = player.inventory
+        val menu = type.create(id, inventory)
+        val inventoryName = name.string.trim()
```
hannibal002: why do we use `name.string.trim()` here? in `OtherInventoryData.onInventoryDataReceiveEvent` we use `packet.title.unformattedTextCompat()` for the same thing

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/ArmorWardrobeApi.kt
```diff
@@ -24,20 +24,23 @@ object ArmorWardrobeApi : AbstractWardrobeApi() {
...
 
     @HandleEvent
-    fun onInventoryClose(event: InventoryCloseEvent) = handleInventoryClose()
+    private fun onInventoryClose(event: InventoryCloseEvent) = handleInventoryClose()
 
     @HandleEvent
-    fun onDebugDataCollect(event: DebugDataCollectEvent) = handleDebugDataCollect(event)
+    private fun onDebugDataCollect(event: DebugDataCollectEvent) = handleDebugDataCollect(event)
+
+    // Sets inThisWardrobe to true if it matches
```
hannibal002: this comment is wrong. it sets `inThisWardrobe` not only to true, but also to false, depending on the group in the pattern.

src/main/java/at/hannibal2/skyhanni/compat/SkyHanniReiPlugin.kt
```diff
@@ -1,41 +1,37 @@
...
-                if (CustomWardrobe.renderableTopCorner == Pair(0, 0)) {
-                    listOf(screen.fullRectangle())
-                } else {
-                    listOf(screen.customWardrobeExclusionRect())
-                }
-            } else {
-                listOf()
-            }
+private class SkyHanniDisplayBoundsProvider : DisplayBoundsProvider<SkyHanniBaseScreen> {
+    override fun <R : Screen?> shouldScreenBeOverlaid(screen: R?): InteractionResult? {
```
hannibal002, line 18: this line changes not only custom wardrobe, but sh screen that extends SkyHanniBaseScreen. is this intended? if so, then we should mention that in the changelog.

src/main/java/at/hannibal2/skyhanni/compat/SkyHanniReiPlugin.kt
```diff
@@ -1,41 +1,37 @@
...
+        return if (customScreen.shouldShowItemList()) {
+            InteractionResult.SUCCESS
+        } else {
+            InteractionResult.FAIL
         }
     }
 
-    private fun ContainerScreen.fullRectangle(): Rectangle {
-        return Rectangle(0, 0, this.width, this.height)
+    override fun getScreenBounds(screen: SkyHanniBaseScreen): Rectangle? {
```
hannibal002: have you tested that returning null here is acceptable/expected by REI? if not, this might cause issues

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeEditScreen.kt
```diff
@@ -0,0 +1,66 @@
...
+    override fun removed() {
+        if (!CustomWardrobe.switchingScreens) {
+            super.removed()
+        }
+    }
+
+    //~ if < 26.1 'extractRenderState' -> 'render' {
+    override fun extractRenderState(graphics: GuiGraphicsExtractor, mouseX: Int, mouseY: Int, partialTick: Float) {
+        super.extractRenderState(graphics, mouseX, mouseY, partialTick)
+        //~}
```
hannibal002: why do we close stonecutter comment here in the middle of the function?

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    override fun onKeyTyped(typedChar: Char?, keyCode: Int?) {
+        CustomWardrobeKeybinds.handlePress()
```
hannibal002, line 49: why do we ignore the result of the key press action? in the old code, we canceled the event, how is the new structure?

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobe.kt
```diff
@@ -46,45 +38,73 @@ import at.hannibal2.skyhanni.utils.renderables.primitives.text
...
+        switchingScreens = true
+        MinecraftCompat.screen = CustomWardrobeEditScreen(
+            screen.menu,
+            player.inventory,
+            screen.title,
+        )
+    }
+
+    fun exitEditMode() {
+        val screen = MinecraftCompat.screen ?: return
```
hannibal002: we do not check the screen type. should this work for everything or just for custom wardrobe?

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobe.kt
```diff
@@ -1,24 +1,17 @@
 package at.hannibal2.skyhanni.features.inventory.wardrobe
 
+
```
hannibal002: empty line

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobe.kt
```diff
@@ -46,45 +38,73 @@ import at.hannibal2.skyhanni.utils.renderables.primitives.text
...
+import net.minecraft.world.inventory.ChestMenu
 import java.awt.Color
 import kotlin.math.min
-import kotlin.time.Duration.Companion.milliseconds
 
-@SkyHanniModule
+@Suppress("TooManyFunctions")
 object CustomWardrobe {
-
     val config: CustomWardrobeConfig get() = SkyHanniMod.feature.inventory.customWardrobe
```
hannibal002, line 47: re-add the empty line after class name

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobe.kt
```diff
@@ -46,45 +38,73 @@ import at.hannibal2.skyhanni.utils.renderables.primitives.text
 import net.minecraft.client.Minecraft
 import net.minecraft.network.chat.Component
 import net.minecraft.world.entity.player.Inventory
+import net.minecraft.world.inventory.ChestMenu
 import java.awt.Color
 import kotlin.math.min
-import kotlin.time.Duration.Companion.milliseconds
 
-@SkyHanniModule
+@Suppress("TooManyFunctions")
```
hannibal002, line 44: eventually we want to fix the error by extracting stuff out of this monster of 600 lines. don't know if we want ot suppress the error for now.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MenuScreensConstructorMixin.java
```diff
@@ -0,0 +1,30 @@
...
+import net.minecraft.client.gui.screens.MenuScreens;
+import net.minecraft.network.chat.Component;
+import net.minecraft.world.inventory.AbstractContainerMenu;
+import net.minecraft.world.inventory.MenuType;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+// Refrence: https://github.com/SkyblockerMod/Skyblocker/blob/main/src/main/java/de/hysky/skyblocker/mixins/MenuScreensConstructorMixin.java
```
hannibal002, line 14: tyopo. should be "Reference"

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
```diff
@@ -0,0 +1,62 @@
...
+) : AbstractCustomMenuScreen(menu, title) {
+    var waitingForInventoryUpdate = false
+        private set
+    private var updateScheduled = false
+
+    override fun getRectangle(): ScreenRectangle = ScreenRectangle(
+        CustomWardrobe.renderableTopCorner.first,
+        CustomWardrobe.renderableTopCorner.second,
+        CustomWardrobe.renderableDimensions.first,
+        CustomWardrobe.renderableDimensions.second
```
hannibal002: missing trailing comma

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
```diff
@@ -0,0 +1,62 @@
...
+    )
+    override fun shouldShowItemList(): Boolean = CustomWardrobe.config.showReiItems
```
hannibal002, line 24: missing empty line above this function

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeEditScreen.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    private fun addReEnableButton(): Renderable {
+        val color = Color(116, 150, 255, 200)
+        return CustomWardrobe.createLabeledButton(
+            "§bEdit",
+            hoveredColor = color,
+            unhoveredColor = color.darker(0.8),
+            onClick = {
+                CustomWardrobe.exitEditMode()
+            }
```
hannibal002: missing trailing comma

src/main/java/at/hannibal2/skyhanni/utils/AbstractCustomMenuScreen.kt
```diff
@@ -0,0 +1,68 @@
...
+import net.minecraft.client.Minecraft
+import net.minecraft.client.gui.screens.inventory.MenuAccess
+import net.minecraft.client.input.KeyEvent
+import net.minecraft.network.chat.Component
+import net.minecraft.world.inventory.AbstractContainerMenu
+import net.minecraft.world.inventory.ChestMenu
+import net.minecraft.world.inventory.ContainerListener
+
+// Reference: https://github.com/SkyblockerMod/Skyblocker/blob/main/src/main/java/de/hysky/skyblocker/skyblock/dungeon/LeapOverlay.java
+abstract class AbstractCustomMenuScreen(
```
hannibal002, line 19: why is this file in wardrobe package?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/compat/SkyHanniReiPlugin.kt
AverageUser125: I have, looked at REI source code.
Also the API has Nullable annotation.

src/main/java/at/hannibal2/skyhanni/compat/SkyHanniReiPlugin.kt
AverageUser125: I could make it just for custom wardrobe, but skyhanni got a couple custom which could also benefit from item list support.
But yeah, probably changelog worthy.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeEditScreen.kt
AverageUser125: Cause it’s just those 2 lines that need it.
I can make it be the whole function I suppose.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
AverageUser125: Since this is a custom screen, there is no one else listening to the click event.
There is nothing to cancel.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobe.kt
AverageUser125: I made a follow up PR for that.

src/main/java/at/hannibal2/skyhanni/utils/AbstractCustomMenuScreen.kt
AverageUser125: I uh, great question.
Should it be utils, or compat?

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobe.kt
AverageUser125: I still want this to exist as it’s useful, even if unused.

src/main/java/at/hannibal2/skyhanni/utils/AbstractCustomMenuScreen.kt
AverageUser125: When setScreen is called, Minecraft calls "removed" on the overwritten screen.

src/main/java/at/hannibal2/skyhanni/utils/AbstractCustomMenuScreen.kt
hannibal002: up to you. either way, please add a short docs explaining the point of the file

src/main/java/at/hannibal2/skyhanni/compat/SkyHanniReiPlugin.kt
hannibal002: good to know. i dont see the nuolable annotation in intellij, maybe add a comment in sh code explaining what null means to the api?

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
AverageUser125: cause that's what AbstractContainerScreen inside minecraft did.
Also what skyblocker does.
Other then that, no clue.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
hannibal002: look at keyPressed in AbstractCustomMenuScreen.
if onKeyTyped in CustomWardrobeScreen gets handelled, and returns true, this means we already did an action. but now if we dont pass that boolean further along to AbstractCustomMenuScreen, the key pressed function returns false, the early return gets skipped, and the onClose logic might run.

this is not ideal, we would have two actions in one key press.

src/main/java/at/hannibal2/skyhanni/compat/SkyHanniReiPlugin.kt
hannibal002: the comment is still missing. what does null mean to REI?

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobe.kt
```diff
@@ -119,57 +138,10 @@ object CustomWardrobe {
...
-        val posX = gui.leftPos + (1.05 * gui.imageWidth).toInt()
-        val posY = gui.topPos + (gui.imageHeight - renderable.height) / 2
-        inventoryButtonPosition.moveTo(posX, posY)
-            .renderRenderable(renderable, posLabel = GUI_NAME, addToGuiManager = false)
-    }
-
-    @HandleEvent
-    fun onInventoryClose(event: InventoryCloseEvent) {
+    internal fun update() {
         waitingForInventoryUpdate = false
```
hannibal002, line 139: in `fun WardrobeSlot.clickSlot() {` we set this to true, but afterwards this function gets called, resetting the state.

hannibal002 (changes requested):
also see two repoened review further up

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
AverageUser125: This is `onKeyTyped` NOT `keyPressed`. meaning it's the version that `SkyHanniBaseScreen` creates that wraps `keyPressed`.
So it doesn't need to return true.

Also, if the user presses "E" (the inventory button) even if the key was handled by our custom screen, hypixel expects us to close the menu no matter what, since it assumes we are still in the container screen.
Also the `Screen.keyTyped` is literaly empty.

src/main/java/at/hannibal2/skyhanni/compat/SkyHanniReiPlugin.kt
AverageUser125: I gave up trying to decipher the REI source code. so I just deleted that check.
It shouldn't run anyways.

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
AverageUser125: I am way to tired to think,
but I assume you just want this function to return true and do an early return in such case when the click is handled.
Imma do that tomorrow.

src/main/java/at/hannibal2/skyhanni/utils/AbstractCustomMenuScreen.kt
```diff
@@ -0,0 +1,69 @@
...
+        if (!player.isAlive) {
+            player.closeContainer()
+        }
+    }
+
+    override fun guiClosed() {
+        MinecraftCompat.localPlayerOrNull?.closeContainer()
+    }
+
+    override fun removed() {
```
hannibal002, line 62: the code inside this function will never get called. confusing design. intentional?

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeEditScreen.kt
```diff
@@ -0,0 +1,64 @@
...
+            super.removed()
+        }
+    }
+
+    override fun extractRenderState(graphics: GuiGraphicsExtractor, mouseX: Int, mouseY: Int, partialTick: Float) {
+        super.extractRenderState(graphics, mouseX, mouseY, partialTick)
+        DrawContextUtils.setContext(graphics)
+        try {
+            onDrawScreen()
+        } finally {
```
hannibal002, line 46: why no catch

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MenuScreensConstructorMixin.java
hannibal002, line 14: typo reference

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/ArmorWardrobeApi.kt
```diff
@@ -24,20 +24,22 @@ object ArmorWardrobeApi : AbstractWardrobeApi() {
...
 
     @HandleEvent
-    fun onInventoryClose(event: InventoryCloseEvent) = handleInventoryClose()
+    private fun onInventoryClose(event: InventoryCloseEvent) = handleInventoryClose()
 
     @HandleEvent
-    fun onDebugDataCollect(event: DebugDataCollectEvent) = handleDebugDataCollect(event)
+    private fun onDebugDataCollect(event: DebugDataCollectEvent) = handleDebugDataCollect(event)
+
+    // This modifies the inWardrobe property aswell
```
hannibal002: typo as well

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeScreen.kt
```diff
@@ -0,0 +1,61 @@
...
+import net.minecraft.world.inventory.AbstractContainerMenu
+import net.minecraft.world.inventory.ChestMenu
+
+class CustomWardrobeScreen(
+    menu: ChestMenu,
+    title: Component,
+) : AbstractCustomMenuScreen(menu, title) {
+    private var updateScheduled = false
+
+    override fun getRectangle(): ScreenRectangle = ScreenRectangle(
```
hannibal002, line 17: this will return 0,0 and 0,0 until renderWardrobeOverlay gets called. thats different to what it was before. we were checking if the render top corner is 0,0, and if this is the case we would return screen.fullRectangle(). this means in the first tick (or however long init long it takes), REI could not have drawn over the screen. now, this protection is gone.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/MenuScreensHook.kt
```diff
@@ -0,0 +1,44 @@
...
+        name: Component,
+        type: MenuType<T>,
+        client: Minecraft,
+        id: Int,
+    ): Boolean {
+        val player = client.player ?: return false
+        if (!SkyBlockUtils.inSkyBlock) return false
+
+        val inventory = player.inventory
+        val menu = type.create(id, inventory)
```
hannibal002: only run create() after `CustomWardrobe.shouldReplace(inventoryName)` returns true. heavy code after light checks

src/main/java/at/hannibal2/skyhanni/features/inventory/wardrobe/CustomWardrobeKeybinds.kt
```diff
@@ -52,15 +37,12 @@ object CustomWardrobeKeybinds {
 
             slot.clickSlot()
             lastClick = SimpleTimeMark.now()
-            return true
         }
-
-        return false
     }
 
     fun allowMouseClick() = isEnabled() && keybinds.filter { it < 0 }.any { it.isKeyHeld() }
```
hannibal002: those two functions can go. they don't serve a purpose anymore. also remove their calls in gui data. do not remove the funcitons with the same name in gui data, they need to stay.

hannibal002 (changes requested):
only minor stuff now:

1.  I am not sure if this is because of this PR, but update() gets called twice per wardrobe opening.

2. The global scale setting does not change the size of the wardrobe anymore.

3. typo in SkyHanni:

> Made it easy to add item list support to custom skyhanni UIs. - Avrg

4. the "what" part is outdated:

> I also moved MOST of the rendering code from CustomWardrobe to CustomRenderUtils, but it's untouched otherwise. (that is why the lines changed a big number)

src/main/java/at/hannibal2/skyhanni/utils/AbstractCustomMenuScreen.kt
AverageUser125: Intentional, since for wardrobes specifically with the “edit screen” I need to skip the remove player.
But for any other custom screen without edit menu this is correct.

hannibal002: Any custom screen that swaps screens needs exactly this. Can we put the hook in the base class instead, so CustomWardrobeScreen.removed() disappears entirely?

AverageUser125: Not really, since “switchingScreens” needs to exist outside the screen lifetime (meaning part of some seperate object).
So can’t really make it be in the base class.

AverageUser125: I could put an abstract getter for “isSwitchingScreens” (default false) and then it should work.

AverageUser125:
image
The global scale looks working to me.
It's just coerced to not be bigger then the screen size.

hannibal002:
> image The global scale looks working to me. It's just coerced to not be bigger then the screen size.

What have you done to replate?

I set the value to 100, opened the wardrobe, saw the normal size, then changed it to 50, opened the wardrobe and nothing happened.

AverageUser125:
Changed it to a value like 20.

AverageUser125:
I guess the global scale was more tuned for a 16k monitor?

hannibal002:
when you change the value from 100 to 50, the scale should  be half, doesnt matter the resolution

hannibal002 (changes requested):
bump

AverageUser125:
I would rather not touch the actual rendering code in this PR.
I got a follow up PR to seperate the rendering code though.

hannibal002:
this pr breaks the global scale in wardrdrobe, so you already touched rendering code. i cant merge this pr while there are open bugs caused by this pr.

hannibal002 (approved):

PR #6226, opened 2026-07-28
Fix: Hunting Box Value Detection

hannibal002:
## What

Previously the display was built once on `InventoryFullyOpenedEvent`. Slot updates that arrived afterward fired only `InventoryUpdatedEvent`, so those shards were missed.

The fix replaces this with a handler on `InventoryUpdatedEvent` that reads `event.inventoryItems` directly. The event always carries the full accumulated inventory state, so the display is updated correctly each time new slots arrive.

## Changelog Fixes
+ Fixed Hunting Box Value not showing all shards when switching pages, especially with high server ping. - hannibal2


CalMWolfs (approved):
worked for me

PR #6231, opened 2026-07-29
Fix: Skill Progress Display adding cap XP twice when a goal is set

lunaynx:
## What
https://discord.com/channels/997079228510117908/1531798408111591584

## Changelog Fixes
+ Fixed Skill Progress Display adding the level 60 XP twice when custom goal is enabled and set. - Luna
+ Fixed All Skill Display showing wrong overflow level and current XP for skills with a cap below level 60. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/skillprogress/SkillProgress.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/skillprogress/SkillProgress.kt
@@ -269,6 +269,17 @@ object SkillProgress {
         return newList
     }
 
+    /**
+     * Progress towards [SkillApi.SkillInfo.customGoalLevel], expressed as cumulative XP out of the
+     * cumulative XP that goal level requires.
+     *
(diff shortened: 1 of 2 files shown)


hannibal002 (changes requested):
Does the same bug exist in "All Skills Display" in the function drawAllDisplay() ?

the fix here looks good, just asking since if we already fix it in once spot for one feature,we should also fix it couple lines away in the same file for another feature..

lunaynx:
That path uses `SkillLevel(skillInfo.overflowLevel, have, need, xp)` rather than our new `SkillLevel(skill.overflowLevel, xp, need, xp)`, but the XP displayed is correct and identical between the two, so it's probably fine.

lunaynx:
This is weird though, because looking more closely at the code and understanding it better, it sounds like it *should* have the same bug...

lunaynx:
Okay, that part is not bugged, but I caught another bug with it.

CalMWolfs (approved):

PR #6234, opened 2026-07-29
Improvement: Use materials in sacks for MinionCraftHelper

hugocornago:
# What
With the new beginner sacks available right as the start of bingo, players using them will deceive the current logic of MinionCraftHelper.

This commit makes uses of the current SackApi to get the required items from it.

Unfortunately, if the player is using a sack, the currently amount of materials the player has will be delayed by a significant amount of time, instead of being instantly without the sack.

## Changelog Improvements
+ MinionCraftHelper now also uses materials in sacks. - cornago


hannibal002:
Most likely fix:
merge the beta branch into this PR.

PR #6235, opened 2026-07-29
Improvement: Make /shlastopened respect PreventEarlyCommands settings

AverageUser125:
## What
Made the "/shlastopened" respect the "prevent early command settings" so that running it does not give:
"You need to wait 4 more seconds to run that".
And it thinking it works.

The implementation is a bit dumb, but it's the simplest and easiest solution.
And the first idea I had.

## Changelog Improvements
+ Made /shlastopened compatible with prevent early commands feature. - Avrg


AverageUser125:
Too hacky a solution

PR #6240, opened 2026-07-30
Fix: shmarkplayer not working on players with capital players

AverageUser125:
## What
This feature is really old.
Also had the problem where it would show in suggestions both lowercase and regular names.
So I just decided to store the regular names in the list, and then make all matching with it ignore casing.

## Changelog Fixes
+ Fixed Marked Players not working on usernames with capital letters. - Avrg


CalMWolfs (changes requested):
Erm

Image

AverageUser125:
That was like that previously.
But I should probably make it use ComponentSpan and stuff.
No clue if possible lol.

CalMWolfs:
> That was like that previously. But I should probably make it use ComponentSpan and stuff. No clue if possible lol.

I checked out beta and it worked fine

AverageUser125:
I can’t figure out chat components with hypixel combing them

PR #6251, opened 2026-08-01
Feature + Fix + Improvement + Backend: shnavall command

CalMWolfs:
## What
Added /shnavall which navigates to all nodes with a certain tag.
In adding this I fixed our travelling salesman algorithm being super super slow. Now hub fairy souls on my device takes ~0.15 seconds compared to the 8.4 seconds it was taking before (50x speed up).
We were running and saving dijkstra calculations way too frequently along with not considering that we have a directional graph leading wrong shortest path calculations.

The new NavigateAllHelper logic is currently just used for the new command but can later allow for existing features to be abstracted and for new features to be made easily. 

## Changelog New Features
+ Added /shnavall. - CalMWolfs
    * Navigate to all occurrences of a target on an island.
    * May be useful for new points of interest on Honey Hives (Torrhus Canyon), and the Bells and Hideyho (Safari).

## Changelog Improvements
+ Sped up pathfinding and optimized routes for the Fast Fairy Souls, Spider's Den Relics, and Shulker Finder. - CalMWolfs

## Changelog Fixes
+ Fixed /shnav <coordinates> making the end location offset by a block diagonally. - CalMWolfs

+ Allowed EnumArgumentType to take in a filtered subset of the whole enum. - CalMWolfs

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/arguments/EnumArgumentType.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/arguments/EnumArgumentType.kt
@@ -78,7 +78,7 @@ class EnumArgumentType<E : Enum<E>> private constructor(
         }
 
         /**
-         * To filtering of an enum should not change during runtime.
+         * The filtering of an enum should not change during runtime.
          */
         inline fun <reified E : Enum<E>> filtered(
             noinline toString: (E) -> String,
--- a/src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/arguments/EnumArgumentType.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/arguments/EnumArgumentType.kt
@@ -18,9 +18,10 @@ class EnumArgumentType<E : Enum<E>> private constructor(
     clazz: Class<E>,
     val isGreedy: Boolean = false,
     toString: (E) -> String,
+    predicate: (E) -> Boolean,
 ) : ArgumentType<E> {
 
-    private val mapping: Map<String, E> = clazz.enumConstants.associateBy { constant ->
+    private val mapping: Map<String, E> = clazz.enumConstants.filter { predicate(it) }.associateBy { constant ->
         val string = toString(constant)
         require(string.none { it.isWhitespace() }) {
             "String representation of constant ${constant.name} of enum ${clazz.simpleName} contains whitespace: '$string'"
@@ -53,25 +54,43 @@ class EnumArgumentType<E : Enum<E>> private constructor(
     override fun getExamples(): Collection<String> = mapping.keys
 
     companion object {
-        fun <E : Enum<E>> create(clazz: Class<E>, isGreedy: Boolean = false, toString: (E) -> String): EnumArgumentType<E> {
-            return EnumArgumentType(clazz, isGreedy, toString)
+        fun <E : Enum<E>> create(
+            clazz: Class<E>,
+            isGreedy: Boolean = false,
+            toString: (E) -> String,
+            predicate: (E) -> Boolean = { true },
+        ): EnumArgumentType<E> {
+            return EnumArgumentType(clazz, isGreedy, toString) { predicate(it) }
         }
 
         /**
          * To use enum name arguments do `EnumArgumentType.name<Enum>()`
          */
         inline fun <reified E : Enum<E>> name(isGreedy: Boolean = false): EnumArgumentType<E> {
-            return create(E::class.java, isGreedy) { it.name }
+            return create(E::class.java, isGreedy, toString = { it.name })
         }
 
         /**
          * To use enum lowercase name arguments do `EnumArgumentType.lowercase<Enum>()`
          */
         inline fun <reified E : Enum<E>> lowercase(isGreedy: Boolean = false): EnumArgumentType<E> {
-            return create(E::class.java, isGreedy) { it.name.lowercase() }
+            return create(E::class.java, isGreedy, toString = { it.name.lowercase() })
         }
 
-        /** The string representation of the enum should not change during runtime. */
+        /**
+         * To filtering of an enum should not change during runtime.
+         */
+        inline fun <reified E : Enum<E>> filtered(
+            noinline toString: (E) -> String,
+            isGreedy: Boolean = false,
+            crossinline predicate: (E) -> Boolean,
+        ): EnumArgumentType<E> {
+            return create(E::class.java, isGreedy, toString) { predicate(it) }
+        }
+
+        /**
+         * The string representation of the enum should not change during runtime.
+         */
         inline fun <reified E : Enum<E>> custom(noinline toString: (E) -> String, isGreedy: Boolean = false): EnumArgumentType<E> {
             return create(E::class.java, isGreedy, toString)
         }
--- a/src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/arguments/LorenzVecArgumentType.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/arguments/LorenzVecArgumentType.kt
@@ -59,7 +59,7 @@ sealed class LorenzVecArgumentType : ArgumentType<LorenzVec> {
         override fun toVec(x: kotlin.Double, y: kotlin.Double, z: kotlin.Double) = LorenzVec(x, y, z)
 
         override fun getExamples(): Collection<String> =
-            listOf("1.0 2.5 -3", "0.0 0.0 0.0", "-1.7 ~ ~", "-78.8:68.0:-28.7", "LorenzVec(-91.7, 70.0, 29.3), x=-262.0, y=58.0, z=117.0")
+            listOf("1.0 2.5 -3", "0.0 0.0 0.0", "-1.7 ~ ~", "-78.8:68.0:-28.7", "LorenzVec(-91.7, 70.0, 29.3)", "x=-262.0, y=58.0, z=117.0")
     }
 
     @SkyHanniModule
@@ -107,7 +107,7 @@ sealed class LorenzVecArgumentType : ArgumentType<LorenzVec> {
          */
         private val namedParameterPattern by patternGroup.pattern(
             "named-parameter",
-            "x=(?<x>-?\\d+(?:\\.\\d+)?),\\s*y=(?<y>-?\\d+(?:\\.\\d+)?),\\s*z=(?<z>-?\\d+(?:\\.\\d+)?)"
+            "x=(?<x>-?\\d+(?:\\.\\d+)?),\\s*y=(?<y>-?\\d+(?:\\.\\d+)?),\\s*z=(?<z>-?\\d+(?:\\.\\d+)?)",
         )
 
         private val patterns = listOf(lorenzVecPattern, colonPattern, spacePattern, namedParameterPattern)
--- a/src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/arguments/LorenzVecArgumentType.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/arguments/LorenzVecArgumentType.kt
@@ -51,14 +51,15 @@ sealed class LorenzVecArgumentType : ArgumentType<LorenzVec> {
         override fun toVec(x: kotlin.Double, y: kotlin.Double, z: kotlin.Double) =
             LorenzVec(x.toInt(), y.toInt(), z.toInt())
 
-        override fun getExamples(): Collection<String> = listOf("1 2 3", "-4 0 5", "~ 64 ~", "1:2:3", "LorenzVec(1, 2, 3)")
+        override fun getExamples(): Collection<String> =
+            listOf("1 2 3", "-4 0 5", "~ 64 ~", "1:2:3", "LorenzVec(1, 2, 3), x=-262, y=58, z=117")
     }
 
     data object Double : LorenzVecArgumentType() {
         override fun toVec(x: kotlin.Double, y: kotlin.Double, z: kotlin.Double) = LorenzVec(x, y, z)
 
         override fun getExamples(): Collection<String> =
-            listOf("1.0 2.5 -3", "0.0 0.0 0.0", "-1.7 ~ ~", "-78.8:68.0:-28.7", "LorenzVec(-91.7, 70.0, 29.3)")
+            listOf("1.0 2.5 -3", "0.0 0.0 0.0", "-1.7 ~ ~", "-78.8:68.0:-28.7", "LorenzVec(-91.7, 70.0, 29.3), x=-262.0, y=58.0, z=117.0")
     }
 
     @SkyHanniModule
@@ -74,7 +75,7 @@ sealed class LorenzVecArgumentType : ArgumentType<LorenzVec> {
          */
         private val lorenzVecPattern by patternGroup.pattern(
             "lorenz",
-            """LorenzVec\((?<x>-?\d+(?:\.\d+)?),\s*(?<y>-?\d+(?:\.\d+)?),\s*(?<z>-?\d+(?:\.\d+)?)\)""",
+            "LorenzVec\\((?<x>-?\\d+(?:\\.\\d+)?),\\s*(?<y>-?\\d+(?:\\.\\d+)?),\\s*(?<z>-?\\d+(?:\\.\\d+)?)\\)",
         )
 
         /**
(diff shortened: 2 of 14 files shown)


src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigateAllHelper.kt
```diff
@@ -0,0 +1,131 @@
...
+
+            recursiveNavigate()
+        }
+    }
+
+    private fun recursiveNavigate() {
+        if (route.isEmpty()) return
+
+        val target = route.firstOrNull() ?: error("No more nodes found")
+        route = route.drop(1)
```
hannibal002, line 77: why do we drop the last entry? this makes it so that the condition is empty below, meaning we never navigate the user to the last point

src/main/java/at/hannibal2/skyhanni/utils/navigation/NavigationUtils.kt
```diff
@@ -1,286 +1,161 @@
...
-        val output = calculateTravelingSalesman(input, maxIterations, neighborhoodSize)
+    fun getRoute(
+        nodes: List<GraphNode>,
+        maxIterations: Int = 300,
+        neighborhoodSize: Int = 50,
+    ): List<LorenzVec> {
+        if (nodes.isEmpty()) return emptyList()
+        val graph = IslandGraphs.currentIslandGraph ?: error("no active island graph")
+
+        val closestNode = graph.minByOrNull { it.position.distanceSqToPlayer() } ?: error("no closest node")
```
hannibal002: minByOrNull ignores disabled nodes. for this case, we already have  `GraphUtils.nearestNodeOnCurrentIsland()`

src/main/java/at/hannibal2/skyhanni/utils/navigation/NavigationUtils.kt
```diff
@@ -1,286 +1,161 @@
...
-        clusters.forEach { cluster ->
-            cluster.parallelStream().forEach { node ->
-                val dijkstraTree = GraphUtils.findAllShortestDistances(node)
-                result[node] = cluster.associateWith { target ->
-                    dijkstraTree.distances[target] ?: Double.POSITIVE_INFINITY
-                }.toMutableMap()
-            }
-        }
+        while (visited.size < distanceMap.size) {
+            val next = distanceMap[current]?.filterKeys { it !in visited }?.minByOrNull { it.value }?.key ?: break
```
hannibal002, line 82: please explain the reasoning here. the logic here can contain unreachable nodes now, the old code dealt with POSITIVE_INFINITY to avoid this.

src/main/java/at/hannibal2/skyhanni/data/model/graph/Graph.kt
```diff
@@ -30,6 +29,8 @@ value class Graph(
         nodes.filter { node -> tag.all { node.hasTag(it) } }
     fun getNodesWithName(name: String): List<GraphNode> =
         nodes.filter { it.name == name }
+    fun getActiveNodeTags(): List<GraphNodeTag> =
```
hannibal002: doesnt this return duplicate entries?

src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigateAllHelper.kt
```diff
@@ -0,0 +1,131 @@
...
+            ChatUtils.userError("No current navigation to skip. §eUse /shnavigateall to start navigation")
+            return
+        }
+
+        // TODO handle recalculation of remaining route without the current node
+        recursiveNavigate()
+    }
+
+    @HandleEvent
+    private fun onIslandChange() {
```
hannibal002: do not use deprecated events

src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/arguments/LorenzVecArgumentType.kt
```diff
@@ -51,14 +51,15 @@ sealed class LorenzVecArgumentType : ArgumentType<LorenzVec> {
         override fun toVec(x: kotlin.Double, y: kotlin.Double, z: kotlin.Double) =
             LorenzVec(x.toInt(), y.toInt(), z.toInt())
 
-        override fun getExamples(): Collection<String> = listOf("1 2 3", "-4 0 5", "~ 64 ~", "1:2:3", "LorenzVec(1, 2, 3)")
+        override fun getExamples(): Collection<String> =
+            listOf("1 2 3", "-4 0 5", "~ 64 ~", "1:2:3", "LorenzVec(1, 2, 3), x=-262, y=58, z=117")
```
hannibal002, line 55: looks wrong. is there a missing string end/start in between `LorenzVec(1, 2, 3), x=-262` ?

src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/arguments/EnumArgumentType.kt
```diff
@@ -53,25 +54,43 @@ class EnumArgumentType<E : Enum<E>> private constructor(
...
          * To use enum lowercase name arguments do `EnumArgumentType.lowercase<Enum>()`
          */
         inline fun <reified E : Enum<E>> lowercase(isGreedy: Boolean = false): EnumArgumentType<E> {
-            return create(E::class.java, isGreedy) { it.name.lowercase() }
+            return create(E::class.java, isGreedy, toString = { it.name.lowercase() })
         }
 
-        /** The string representation of the enum should not change during runtime. */
+        /**
+         * To filtering of an enum should not change during runtime.
```
hannibal002: typo. to -> the

src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigateAllHelper.kt
```diff
@@ -0,0 +1,131 @@
...
+            ) { nodeType ->
+                navigateAll(nodeType)
+            }
+            literalCallback("skip") {
+                handleSkip()
+            }
+        }
+    }
+
+    private fun getValidNodeNames(): List<GraphNodeTag> {
```
hannibal002: wrong function name. returns tags, not names

src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigateAllHelper.kt
```diff
@@ -0,0 +1,131 @@
...
+            ChatUtils.chat("Navigating to $total ${nodeType.displayName}")
+
+            recursiveNavigate()
+        }
+    }
+
+    private fun recursiveNavigate() {
+        if (route.isEmpty()) return
+
+        val target = route.firstOrNull() ?: error("No more nodes found")
```
hannibal002: dead code, we alr check 2 lines above if the list is empty. can be replaced by first()

src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigateAllHelper.kt
```diff
@@ -0,0 +1,131 @@
...
+
+        IslandGraphs.pathFind(
+            target,
+            "${currentNodeType?.displayName} ${total - route.size}/$total",
+            onFound = { recursiveNavigate() },
+            condition = { route.isNotEmpty() },
+        )
+    }
+
+    private fun calculateRoute(targetNodes: List<GraphNode>): List<LorenzVec> =
```
hannibal002: can be inlined

src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigateAllHelper.kt
```diff
@@ -0,0 +1,131 @@
...
+                handleSkip()
+            }
+        }
+    }
+
+    private fun getValidNodeNames(): List<GraphNodeTag> {
+        val activeTags = IslandGraphs.currentIslandGraph?.getActiveNodeTags() ?: return emptyList()
+        return activeTags.filter { it in allowedMultiNavigationTags }
+    }
+
```
hannibal002: empty line

src/main/java/at/hannibal2/skyhanni/utils/navigation/NavigationUtils.kt
```diff
@@ -1,286 +1,161 @@
...
     }
 
-    private fun adjustRouteForCurrentLocation(
-        route: List<GraphNode>,
-        currentLocation: LorenzVec,
-    ): List<GraphNode> {
-        val closestNode = route.minByOrNull { it.position.distanceSq(currentLocation) } ?: route.first()
-        val idx = route.indexOf(closestNode)
-        return route.drop(idx) + route.take(idx)
+    private fun edge(
```
hannibal002, line 129: unnecessary nullability. can be dropped in the function without replacement.

src/main/java/at/hannibal2/skyhanni/utils/navigation/NavigationUtils.kt
```diff
@@ -1,286 +1,161 @@
...
-            cluster.add(node)
-            for (neighbor in node.neighbors) {
-                // Ensure neighbor.key equals one of the target nodes.
-                if (targetNodes.contains(neighbor.key) && !visited.contains(neighbor.key)) {
-                    dfs(neighbor.key, cluster)
-                }
-            }
-        }
+    private fun directedTwoOpt(
+        routeInput: MutableList<GraphNode>,
```
hannibal002: this para could be list instead of mutable list list

src/main/java/at/hannibal2/skyhanni/features/nether/MatriarchHelper.kt
```diff
@@ -84,7 +84,6 @@ object MatriarchHelper {
         if (config.useShortestDistance) {
             val path = tspCache ?: NavigationUtils.getRoute(
                 pearlList.map { it.second },
```
hannibal002, line 86: why do we drop the maxIterations 5 here?

src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigateAllHelper.kt
```diff
@@ -0,0 +1,132 @@
...
+        val targetNodes = graph.getNodesWithTags(nodeType)
+        currentNodeType = nodeType
+
+        pathfindCoroutine.launch {
+            route = emptyList()
+            route = calculateRoute(targetNodes)
+            total = route.size
+
+            ChatUtils.chat("Navigating to $total ${nodeType.displayName}")
+
```
hannibal002, line 68: is this intended? we first run this function async, then all future calls are on main thread. dont we want to only use one consistent thread for the function logic?

hannibal002, line 68: this looks unintended, but why do we run this function here in async , but in all other cases in a sync context?

src/main/java/at/hannibal2/skyhanni/utils/navigation/NavigationUtils.kt
```diff
@@ -1,286 +1,161 @@
...
-            visited.add(node)
-            cluster.add(node)
-            for (neighbor in node.neighbors) {
-                // Ensure neighbor.key equals one of the target nodes.
-                if (targetNodes.contains(neighbor.key) && !visited.contains(neighbor.key)) {
-                    dfs(neighbor.key, cluster)
-                }
-            }
-        }
+    private fun directedTwoOpt(
```
hannibal002, line 92: what does "directed" mean here?

src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/arguments/LorenzVecArgumentType.kt
```diff
@@ -98,10 +99,18 @@ sealed class LorenzVecArgumentType : ArgumentType<LorenzVec> {
...
-            """(?<x>~|-?\d+(?:\.\d+)?)\s+(?<y>~|-?\d+(?:\.\d+)?)\s+(?<z>~|-?\d+(?:\.\d+)?)""",
+            "(?<x>~|-?\\d+(?:\\.\\d+)?)\\s+(?<y>~|-?\\d+(?:\\.\\d+)?)\\s+(?<z>~|-?\\d+(?:\\.\\d+)?)",
+        )
+
+        /**
+         * REGEX-TEST: x=-262.0, y=58.0, z=117.0
+         */
+        private val namedParameterPattern by patternGroup.pattern(
+            "named-parameter",
+            "x=(?<x>-?\\d+(?:\\.\\d+)?),\\s*y=(?<y>-?\\d+(?:\\.\\d+)?),\\s*z=(?<z>-?\\d+(?:\\.\\d+)?)"
```
hannibal002: missing trailing comma

src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/arguments/EnumArgumentType.kt
```diff
@@ -53,25 +54,43 @@ class EnumArgumentType<E : Enum<E>> private constructor(
...
     companion object {
-        fun <E : Enum<E>> create(clazz: Class<E>, isGreedy: Boolean = false, toString: (E) -> String): EnumArgumentType<E> {
-            return EnumArgumentType(clazz, isGreedy, toString)
+        fun <E : Enum<E>> create(
+            clazz: Class<E>,
+            isGreedy: Boolean = false,
+            toString: (E) -> String,
+            predicate: (E) -> Boolean = { true },
+        ): EnumArgumentType<E> {
+            return EnumArgumentType(clazz, isGreedy, toString) { predicate(it) }
```
hannibal002, line 63: why do we return EnumArgumentType? is this necessary here?

hannibal002 (changes requested):
changelog:

> Added /shnavall which allows you to navigate to all occurrences of a target on an island. - CalMWolfs 

this new command does not "improve" the old exisitng /shnav command, this is a new feature in my eyes.

> Sped up pathfinding and optimised routes for the Fast Fairy Souls and Shulker Finder.

what about spider den relic and matriach helper?

>  Improved the solve time for the traveling salesman pathfinding

in technical details, this is a duplicate line.


the /shnavigate improvement/fix is not mentioned, that we no longer off set by -1 -1 -1

src/main/java/at/hannibal2/skyhanni/utils/navigation/NavigationUtils.kt
CalMWolfs: Breaking early will lead to this if statement triggering which sends an error message already `if (route.size != nodes.size + amountOffset)`

src/main/java/at/hannibal2/skyhanni/utils/navigation/NavigationUtils.kt
CalMWolfs: It is the 2-opt algorithm for traveling salesman but it is taking into account that our graph is directed (you can go A -> B but that doesnt mean you can go B -> A along the same edge) therefore directed 2 opt algorithm.

src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigateAllHelper.kt
CalMWolfs: We are running calculateRoute in a coroutine as even after my optimisations there is a .2 second time it takes for routing out all fairy soul locations on the hub and on weaker computers this could be up to a second or two, we dont want the game to freeze for a second or two while the route is being calculated. On the other hand, recursiveNavigate is instantaneous but it needs to be called after the route is calculated so thats why here it is called off thread whereas later it can be on the main thread

src/main/java/at/hannibal2/skyhanni/config/commands/brigadier/arguments/EnumArgumentType.kt
CalMWolfs: This is how we always did it and yes its necessary because otherwise `Public-API inline function cannot access non-public-API constructor.`

src/main/java/at/hannibal2/skyhanni/features/nether/MatriarchHelper.kt
CalMWolfs: The navigation is now instant so we dont need to worry about limiting iterations. Tested it and  it only took 1 iteration each time but i guess it doesnt hurt to limit how many iterations it takes, (From looking at the code matriarch helper recalculates the path every frame 💀. Ill fix this when i abstract every use of multi navigation

hannibal002 (approved):

PR #6256, opened 2026-08-02
Backend: Graph editor flags one way paths

hannibal002:
+ Updated the Graph Editor, especially around errors. - hannibal2
    * Added an error message when a path is one-way and doesn't allow you to go back.
    * Neighbors are now sorted by id to avoid funny-looking diffs.
    * Now only showing the closest 10 errors to improve readability.
    * Now showing error count and category number in chat if there are any.


hannibal002:
maybe changelog verification should not run on draft prs

CalMWolfs (approved):

PR #6257, opened 2026-08-02
Feature: Bestiary Mob Glowing

AverageUser125:
## Dependencies
- #6254

## What
Lets you click any bestiary entry and make said mob glow.
Note that you need to click inside the bestiary entry to see all the mob variants and click each individual variant you want.
it seems mob name + level is unique so I use that to match the mob.

## Changelog New Features
+ Added Bestiary Mob Glowing. - Avrg
    * You can middle click any mob variant bestiary entry to make said mob glow in the world.


src/main/java/at/hannibal2/skyhanni/data/jsonobjects/repo/BestiaryJson.kt
CalMWolfs, line 1: What data do you need that the neu bestiary file does not provide?

AverageUser125: Didn't even know that exist.
But literaly just the info on which bestiary categories are valid in which islands.
Though it's just for optimization reason and not functionality reasons.

AverageUser125: Eh, I can hardcode that probably.

PR #6238, opened 2026-07-30
Fix: SlayerApi Parsing Error

AverageUser125:
## What

added protection for if the tablist/scoreboard is invalid, then do not report error unless it happend for 3 updates in a row. (widget & scoreboard updates)
Also removed "onTick", and seperated it into seperate functions for each thing.

## Changelog Fixes
+ Fixed a false error report when the Slayer quest scoreboard/widget was parsed while incomplete. - Avrg

PR #6258, opened 2026-08-02
Fix: Slayer Cocoon fix

AverageUser125:
## Dependencies
- #6238

## What
Adds a new COCOONED state to the current slayer.
The "cocoonedTimestamp" is ONLY as a failsafe, I can remove it and everything is fine.

## Changelog Improvements
+ Added a Discord RPC message for when your Slayer boss is cocooned. - Avrg

## Changelog Fixes
+ Fixed the "You have no Auto-Slayer active!" warning showing when your Slayer boss gets cocooned. - Avrg


src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
```diff
@@ -361,6 +390,16 @@ object SlayerApi {
         }
     }
 
+    @HandleEvent
+    private fun onWorldChange() {
+        val data = outsideRiftData
```
hannibal002, line 402: why do we use here outside rift data instead of getCurrentData()? please explain in code

src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
```diff
@@ -193,6 +207,14 @@ object SlayerApi {
                 data.currentStateRaw = "no slayer"
                 SlayerStateChangeEvent(FAILED).post()
             }
+            cocoonPattern.matches(message) -> {
+                val data = getCurrentData()
+                ChatUtils.debug("SlayerApi: Slayer boss cocooned, posting SlayerStateChangeEvent")
+                data.currentState = COCOONED
+                data.currentStateRaw = "cocooned"
+                SlayerStateChangeEvent(COCOONED).post()
```
hannibal002, line 221: run the event after the timestamp got updated. otherwise listeners on that event still read the old data

src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
```diff
@@ -68,6 +69,14 @@ object SlayerApi {
         "quest.failed",
         "\\s*SLAYER QUEST FAILED!",
     )
+
+    /**
+     * WRAPPED-REGEX-TEST: "  YOU COCOONED YOUR SLAYER BOSS"
+     */
+    private val cocoonPattern by RepoPattern.pattern(
```
hannibal002: why no patternGroup?

src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
```diff
@@ -296,7 +318,13 @@ object SlayerApi {
 
         data.currentStateRaw = progress
 
-        val newState = detectState(progress)
+        var newState = detectState(progress)
+
+        val cocooned = data.currentState == COCOONED && lastCocoonTimestamp.passedSince() <= 6.seconds
```
hannibal002: make magic number 6.seconds a variable with name that explains what this is

src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
```diff
@@ -105,6 +114,11 @@ object SlayerApi {
      */
     private var invalidUpdates = 0
 
+    /**
+     * The last time we saw a cocoon message, used to ensure it doesn't get stuck in a state where we think we are cocooned when we are not
+     */
+    private var lastCocoonTimestamp = ServerTimeMark.farPast()
```
hannibal002: we have in this file now latestWrongAreaWarning and questStartTime, and now this. please use a consistent suffix

hannibal002: every other case here uses SimpleTimeMark. why ServerTimeMark only here?

hannibal002 (changes requested):
1. in `DiscordStatus.kt`, `COCOONE` is still missing in `when (SlayerApi.state) {`

2. please use the same phrase as in game: 

> "Fixed The slayer boss being cocooned causing "Auto Slayer Inactive" warning. - Avrg"

src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
AverageUser125: The uniform suffix is fair. Will just use “Time” suffix

the ServerTimeMark is since cocoon time is effected by server TPS (I think?) so makes more sense to use that.

src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
```diff
@@ -296,7 +324,13 @@ object SlayerApi {
...
+        var newState = detectState(progress)
+
+        val cocooned = data.currentState == COCOONED && lastCocoonTime.passedSince() <= GRACE_COCOON_TIME
+        if (cocooned && (newState == NO_ACTIVE_QUEST || newState == SLAIN)) {
+            ChatUtils.debug("SlayerApi: Cocooned state detected, overriding $newState to COCOONED")
+            newState = COCOONED
+        }
 
         // If the player kills the boss immediately after the boss spawns
         if (data.currentState == BOSS_FIGHT && newState == GRINDING) {
```
hannibal002: does this support COCOONED? can it jump from COCOONED directly to GRINDING? in that case, we would not send the slain event

src/main/java/at/hannibal2/skyhanni/features/misc/discordrpc/DiscordStatus.kt
```diff
@@ -202,6 +202,7 @@ enum class DiscordStatus(private val displayMessageSupplier: DiscordStatus.() ->
                 BOSS_FIGHT -> "Slaying $article ${SlayerApi.activeType?.displayName}"
                 SLAIN -> "Finished slaying $article ${SlayerApi.activeType?.displayName}"
                 FAILED -> "Lost to $article ${SlayerApi.activeType?.displayName}"
+                COCOONED -> "Waiting for $article ${SlayerApi.activeType?.displayName} Cocoon to hatch"
```
hannibal002: lowercase Cocoon

src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
```diff
@@ -44,6 +45,12 @@ object SlayerApi {
 
     private const val GRACE_UPDATE_COUNT = 3
 
+    /**
+     * Cocoons are actually 5 seconds, but +1 second just in case.
+     * This is for when the chat message appears, but even after this time has passed, the boss hasn't spawned yet.
```
hannibal002: this reads confusing. the extra second is for the time between chat message and respawn of the boss, right?

src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
```diff
@@ -105,6 +120,11 @@ object SlayerApi {
      */
     private var invalidUpdates = 0
 
+    /**
+     * The last time we saw a cocoon message, used to ensure it doesn't get stuck in a state where we think we are cocooned when we are not
+     */
+    private var lastCocoonTime = ServerTimeMark.farPast()
```
hannibal002: can we rename this to latestCocoonTime, so that it is consistent to latestCategory, latestWrongAreaWarningTime and latestProgress? or rename all 3 to last please.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
AverageUser125: It’s more just to accommodate lag.
The chat message is when the boss is killed and then cocoon takes 5 seconds to burst.

After said 5 seconds the boss SHOULD respawn immediately.
If it somehow doesn’t (idk how?) then the timer will expire and everything will be fine.

AverageUser125: Tbh, should I just remove the timer? Since it shouldn’t be possible anyways?

hannibal002: > Tbh, should I just remove the timer? Since it shouldn’t be possible anyways?

no the timer is necessary as a fallback if the change from cocooned to another state cant be properly tracked

src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
```diff
@@ -296,11 +323,17 @@ object SlayerApi {
...
+        if (cocooned && (newState == NO_ACTIVE_QUEST || newState == SLAIN)) {
+            ChatUtils.debug("SlayerApi: Cocooned state detected, overriding $newState to COCOONED")
+            newState = COCOONED
+        }
 
         // If the player kills the boss immediately after the boss spawns
-        if (data.currentState == BOSS_FIGHT && newState == GRINDING) {
-            ChatUtils.debug("SlayerApi: Intermediate state change detected: BOSS_FIGHT -> SLAIN -> GRINDING")
+        if ((data.currentState == BOSS_FIGHT || data.currentState == COCOONED) && newState == GRINDING) {
+            ChatUtils.debug("SlayerApi: Intermediate state change detected: ${data.currentState.name} -> SLAIN -> GRINDING")
```
hannibal002: .name is unnecessary

src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
```diff
@@ -361,6 +395,18 @@ object SlayerApi {
...
     }
 
+    @HandleEvent
+    private fun onWorldChange() {
+        // Using outsideRiftData since rift does not have slayer cocoon
+        // and using getCurrentData is ambiguous while changing worlds (inside/outside rift)
+        val data = outsideRiftData
+        if (data.currentState != COCOONED) return
+        ChatUtils.debug("SlayerApi: World change detected, resetting cocooned state")
+        data.currentState = NO_ACTIVE_QUEST
```
hannibal002, line 405: why do we change the state but not send the SlayerStateChangeEvent here?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/SlayerApi.kt
AverageUser125:
```
                    "source" to source.name,
```
This line exists so I assume doing `.name` is okay.
But done.

hannibal002: the mapping string -> string that happens in this line here requires a string. an enum is no string. but the line i flagged was inside a string, here kotlin automatically applies a .tostring() to any non string object. same reason why you can use `"i am " + 20 + " years old`,  kotlin calls the .toString function of Int. and the toString function of an enum, if not specified, returns the field "name".

thats the reason why you dont need .toString() or .name for enums inside a string, but you do need them when they are the only statement.

AverageUser125: cool, thanks for explaining it!

AverageUser125: though ErrorManager specificity takes `Any?`, so it probably works anyways.
But I get your point.

hannibal002: > though ErrorManager specificity takes Any?, so it probably works anyways.

yes, thats exactly why it is `any` and not `string`, you can omit the .name in that one line as well :D

AverageUser125: done.

hannibal002 (approved):

PR #6260, opened 2026-08-02
Fix: More Perfomant Better Tree Progress feature - “This should not use onTick” said Hannibal

AverageUser125:
## What
Made the TreeProgressDisplay take custom entity name change events instead of looking at every entity.

## Changelog Fixes
+ Fixed excessive memory usage when the Tree Progress Display is enabled. - Avrg
    
+ Made TreeProgressDisplay use events instead of checking all entities every tick. - Avrg


src/main/java/at/hannibal2/skyhanni/features/foraging/TreeProgressDisplay.kt
```diff
@@ -36,32 +46,59 @@ object TreeProgressDisplay {
...
+    }
+
+    @HandleEvent(onlyOnIslandTypeTag = [IslandTypeTag.FORAGING_CUSTOM_TREES])
+    private fun onIslandJoin() {
+        progressDisplay = null
+    }
+
+    @HandleEvent(onlyOnIslandTypeTag = [IslandTypeTag.FORAGING_CUSTOM_TREES])
+    private fun onEntityRemoved(event: EntityRemovedEvent<ArmorStand>) {
+        if (!isEnabled()) return
```
hannibal002: the check here is wrong. we always want to clear the display, even if not enabled.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/foraging/TreeProgressDisplay.kt
AverageUser125: done

hannibal002 (approved):

AverageUser125:
I noticed that switching to AOTV makes it clear the display, so I moved the held item axe detection to the rendering part.

AverageUser125:
My apologies for noticing this only now.

hannibal002:
still LGTM

PR #6261, opened 2026-08-03
Feature + Improvement: Safari Stuff

hannibal002:
## What
Since we now have graph area data on safari island, lets use it!

## Changelog New Features
+ Added Safari Names in Center. - hannibal2
    * Shows the names of the 4 areas while in the center of the Critter Safari.
    * This works identically to the Names in Core feature in Crystal Hollows.

## Changelog Improvements
+ Added Hide Not Clickable support to the Birdfeeder inventory in Safari. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/hunting/safari/SafariNamesInCenter.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/hunting/safari/SafariNamesInCenter.kt
@@ -1,4 +1,4 @@
-package at.hannibal2.skyhanni.features.foraging.safari
+package at.hannibal2.skyhanni.features.hunting.safari
 
 import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.api.event.HandleEvent
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/HideNotClickableItems.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/HideNotClickableItems.kt
@@ -96,7 +96,7 @@ object HideNotClickableItems {
     )
 
     // TODO add more special named fossils (hypixel why)
-    private val fossilNames = setOf(
+    private val specialFossils = setOf(
         "HELIX".toInternalName(),
     )
 
@@ -262,7 +262,7 @@ object HideNotClickableItems {
             return false
         }
 
-        hideReason = "§cNot a chisel or scrap!"
+        hideReason = "Not a chisel or scrap!"
         return true
     }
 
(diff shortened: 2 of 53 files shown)


CalMWolfs (commented):
Have not tested yet or looked at code, just before I do did this also fix the custom scoreboard showing your cold value when you weren't cold (weren't in icy biome)

hannibal002:
no this does not touch custom scoreboard code at all

CalMWolfs:
> no this does not touch custom scoreboard code at all

I think the fix would be just checking every second or so if the current stored cold is > 0, cold is not visible in the real scoreboard, and the user is not in the icy biome and if all 3 conditions are met post a cold update event with the new cold level being 0

CalMWolfs (changes requested):
As I said in my comment above, I think the proper solution for both the cold overlay and cold on scoreboard is sending the cold event as even with your current code the mod still internally thinks the user has the wrong cold value and we shouldn’t be fixing every feature that uses cold with a workaround but instead should be fixing the problem in just one spot and letting the events handle the rest

hannibal002:
> As I said in my comment above, I think the proper solution for both the cold overlay and cold on scoreboard is sending the cold event as even with your current code the mod still internally thinks the user has the wrong cold value and we shouldn’t be fixing every feature that uses cold with a workaround but instead should be fixing the problem in just one spot and letting the events handle the rest


Agreed, fixing it at the source is the right call. I've reverted the cold overlay and scoreboard changes from this PR, so it is now feature plus improvement only.

The follow up PR will do it in MiningApi.

src/main/java/at/hannibal2/skyhanni/features/hunting/safari/SafariNamesInCenter.kt
CalMWolfs, line 1: The safari check could be in the event annotation with `onlyOnIsland = IslandType.SAFARI` and then you can just put the !config.namesInCenter in the if statement
but this is up to you, I dont have too much preference on which early return pattern is used here

src/main/java/at/hannibal2/skyhanni/features/inventory/HideNotClickableItems.kt
CalMWolfs: Instead of using strings we should be using `stack.getItemCategoryOrNull() `

CalMWolfs (changes requested):

src/main/java/at/hannibal2/skyhanni/features/inventory/HideNotClickableItems.kt
hannibal002: we want to check for armor or equipment, we dont have that check yet afaik. adding a new helper variable.

hannibal002: ok not even the mask check is there yet. i dont like the solution i come up with, but its better than nothing

PR #6262, opened 2026-08-03
Fix: Cold not resetting outside cold areas

hannibal002:
## What
Reproduced on the alpha Critter Safari without this PR: when leaving the Icy Biome, Hypixel removes the Cold line from the scoreboard, but the Cold Overlay stays on screen indefinitely. Once Hypixel stopped showing it, the last known value stayed forever, and every feature built on `ColdUpdateEvent` kept believing the player was freezing.

Cold now follows the same shape the Heat handling a few lines below already used: read the value from the full current scoreboard, and reset it to 0 when it is no longer there.

On the Critter Safari, Cold only applies inside the Icy Biome, and Hypixel does not reliably remove the line when leaving it. `MiningApi.inColdArea()` therefore also checks the graph area, and a `GraphAreaChangeEvent` handler resets the value the moment the player leaves the Icy Biome instead of waiting for the scoreboard to catch up.

Both the Cold Overlay and the Custom Scoreboard Cold line are fixed by this without a single change of their own. `ColdOverlay` fades out once its Cold value reaches 0, and `ScoreboardElementCold` returns no line at 0. That was the whole point of fixing it in `MiningApi` instead of adding an area check to every feature that reads Cold.

Split out of #6261 after review feedback from CalMWolfs, since this affects every cold island and not just the Safari.

## Changelog Fixes
+ Fixed the Cold Overlay and the Custom Scoreboard Cold line staying visible after leaving a cold area. - hannibal2
    * Hypixel removes Cold from the scoreboard when leaving the Glacite Tunnels in the Dwarven Mines or the Icy Biome on the Critter Safari, but SkyHanni kept the last known value.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/MiningApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/MiningApi.kt
@@ -217,7 +217,8 @@ object MiningApi {
 
     /**
      * Whether cold can currently apply to the player.
-     * On the Critter Safari cold only exists inside the Icy Biome, on the other cold islands it applies everywhere.
+     * On the Critter Safari cold is limited to the Icy Biome, which is why the area is checked here.
+     * The other cold islands are not narrowed down further, there we rely on Hypixel removing the scoreboard line.
      */
     fun inColdArea(): Boolean {
         if (!IslandTypeTag.IS_COLD.isInIsland()) return false
--- a/src/main/java/at/hannibal2/skyhanni/data/MiningApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/MiningApi.kt
@@ -13,6 +13,7 @@ import at.hannibal2.skyhanni.events.ServerBlockChangeEvent
 import at.hannibal2.skyhanni.events.chat.SkyHanniChatEvent
 import at.hannibal2.skyhanni.events.mining.OreMinedEvent
 import at.hannibal2.skyhanni.events.player.PlayerDeathEvent
+import at.hannibal2.skyhanni.events.skyblock.GraphAreaChangeEvent
 import at.hannibal2.skyhanni.events.skyblock.ScoreboardAreaChangeEvent
 import at.hannibal2.skyhanni.features.dungeon.DungeonApi.dungeonRoomPattern
 import at.hannibal2.skyhanni.features.mining.OreBlock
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/data/MiningApi.kt
```diff
@@ -212,20 +215,37 @@ object MiningApi {
...
+        if (!IslandTypeTag.IS_COLD.isInIsland()) return false
+        if (IslandType.SAFARI.isInIsland()) return icyBiomePattern.matches(SkyBlockUtils.graphArea)
+        return true
+    }
+
     @HandleEvent
-    fun onScoreboardChange(event: ScoreboardUpdateEvent) {
+    private fun onScoreboardChange(event: ScoreboardUpdateEvent) {
         if (IslandTypeTag.IS_COLD.isInIsland()) {
             dungeonRoomPattern.firstMatcher(event.new) {
```
CalMWolfs: While we are in this file can we add DungeonApi to the front of this

CalMWolfs (changes requested):
code looks fine just one thing i noticed

src/main/java/at/hannibal2/skyhanni/data/MiningApi.kt
hannibal002: what do you mean with "to the front of this"? check dung room pattern independently of is cold island tag?

CalMWolfs: > what do you mean with "to the front of this"? check dung room pattern independently of is cold island tag?

I mean it should say DungeonApi.dungeonRoomPattern instead of statically importing the pattern variable

hannibal002: ahh, good catch!

CalMWolfs (approved):

AverageUser125:
Resetting the cold to 0 when leaving a cold area feels a bit strange.
Since if you go to glacite tunnels and then /warp forge.
You will still have that cold in the scoreboard.

hannibal002:
> Resetting the cold to 0 when leaving a cold area feels a bit strange. Since if you go to glacite tunnels and then /warp forge. You will still have that cold in the scoreboard.

lucky us there is a merge conflict, good catch!

hannibal002:
> Resetting the cold to 0 when leaving a cold area feels a bit strange. Since if you go to glacite tunnels and then /warp forge. You will still have that cold in the scoreboard.

Ive checked the code again:
The line is gone after /warp forge though, cold only applies in the Glacite Tunnels, not at the base camp. This PR only makes the mod notice the line disappearing, which it never did before.

AverageUser125:
I remember having like 80 cold and standing in the forge lava to see if that heats you faster.
Which is why I asked.

Though it seems they made the scoreboard line for cold not appear outside the area anymore, so nvm.

PR #6263, opened 2026-08-03
Backend: Area Change Events cleanup

hannibal002:
## What
Added a docs and primary function to the two area change events, fixed the listener names of their usages, and made all listeners in all related files private.

+ Cleaned up Area Change Events.  - hannibal2


CalMWolfs (approved):
We should prob remove the notation of `Event::class' in the future as if you are doing that you should be using an existing primary function or making a new one imo

PR #6265, opened 2026-08-03
Improvement: No Contributor Impersonation

AverageUser125:
## What
Made the EEK achievement also send a "A wild SkyHanni contributor appears!"

## Changelog Improvements
+ Improved the contributor discovery message to show who the contributor is. - Avrg
    * Now includes a player head sprite.


src/main/java/at/hannibal2/skyhanni/features/misc/ContributorManager.kt
```diff
@@ -319,10 +317,11 @@ object ContributorManager {
     }
 
     private fun isContributorMentionMessage(message: String): Boolean {
-        if (message.startsWith(CONTRIBUTOR_ACHIEVEMENT_GOT)) return true
-        if (message.startsWith(FOUND_WILD_CONTRIBUTOR)) return true
+        val cleanMessage = message.removePrefix("[SkyHanni] ")
+        if (cleanMessage.startsWith(CONTRIBUTOR_ACHIEVEMENT_GOT)) return true
+        if (cleanMessage.startsWith(FOUND_WILD_CONTRIBUTOR)) return true
```
hannibal002: FOUND_WILD_CONTRIBUTOR can go. this is already checked in the pattern

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
```diff
@@ -120,8 +126,19 @@ object ContributorAchievement {
         }
     }
 
-    fun onUniqueContributorSeen() {
-        AchievementManager.completeAchievement(CONTRIBUTOR_ACHIEVEMENT)
+    fun onUniqueContributorSeen(username: String) {
+        val completed = AchievementManager.completeAchievement(CONTRIBUTOR_ACHIEVEMENT)
+        if (completed || config.discoverContributorMessage) {
```
hannibal002: why do we show this here even if the config option is disabled?

src/main/java/at/hannibal2/skyhanni/features/achievements/AchievementManager.kt
```diff
@@ -92,15 +92,16 @@ object AchievementManager {
         SkyHanniMod.configManager.saveConfig(ConfigFileType.ACHIEVEMENTS, "achievement progress update")
     }
 
-    fun completeAchievement(id: String) {
-        if (HypixelData.hypixelAlpha) return
+    fun completeAchievement(id: String): Boolean {
```
hannibal002, line 97: what does the return value mean here? please add a docs

src/main/java/at/hannibal2/skyhanni/features/misc/ContributorManager.kt
```diff
@@ -63,8 +61,8 @@ object ContributorManager {
     private val contributorMentions get() = SkyHanniMod.seenContributorStorage.contributorMentions
     private val contributorMentionersThisSession = mutableSetOf<String>()
 
-    private const val CONTRIBUTOR_ACHIEVEMENT_GOT = "[SkyHanni] Achievement Get! EEEEKK!!"
-    private const val FOUND_WILD_CONTRIBUTOR = "A wild SkyHanni contributor appears!"
+    private const val CONTRIBUTOR_ACHIEVEMENT_GOT = "Achievement Get! EEEEKK!!"
+    const val FOUND_WILD_CONTRIBUTOR = "A wild SkyHanni contributor appears!"
```
hannibal002: move the value to ContributorAchievement

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
AverageUser125: Since it has "hover" people can actually see which person is the skyhanni contributor.
And not larp about being it since the chat message shows it.

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
```diff
@@ -92,36 +104,51 @@ object ContributorAchievement {
...
         if (event.playerName in ContributorManager.contributorNames) {
             AchievementManager.completeAchievement(CONTRIBUTOR_REJECTED_ACHIEVEMENT)
         }
     }
 
-    fun onUniqueContributorSeen() {
-        AchievementManager.completeAchievement(CONTRIBUTOR_ACHIEVEMENT)
+    fun onUniqueContributorSeen(profile: GameProfile, username: String) {
+        val completed = AchievementManager.completeAchievement(CONTRIBUTOR_ACHIEVEMENT)
+        if (completed || config.discoverContributorMessage) {
```
hannibal002: early return

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
```diff
@@ -92,36 +104,51 @@ object ContributorAchievement {
...
+            val resolvedProfile = ResolvableProfile.createResolved(profile)
+            val sprite = PlayerSprite(resolvedProfile, false)
```
hannibal002: boolean parameter needs a label

src/main/java/at/hannibal2/skyhanni/features/misc/ContributorManager.kt
```diff
@@ -319,43 +316,31 @@ object ContributorManager {
...
+        getSuffix(uuid)?.let {
+            recordSeenContributor(gameProfile, uuid, gameProfile.name)
             if (!config.contributorNametags) return
             if (contribNametagAppendSpacePattern.find(event.chatComponent)) event.chatComponent.append(" ")
             event.chatComponent.append(it)
         }
     }
 
-    private fun recordSeenContributor(uuid: UUID, username: String) {
+    private fun recordSeenContributor(gameProfile: GameProfile, uuid: UUID, username: String) {
```
hannibal002: no need to pass uuid and username, both are already in game profile

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
hannibal002: If the option is turned off, nothing should ever appear in the chat. That's the whole point of the option.

hannibal002 (changes requested):
the term "larping" is confusing to me. i needed to check it to make sure i understand the first time i saw the pr. and now. so i  guess you just replace the changelog phrasing to "impersonate" or similar?

also, that is a improvement, not a fix.
also, the player sprite in hover maybe needs an improvement entry, at least a technical changes one.

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
AverageUser125: it's impossible to do label parameter for java functions.

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
hannibal002: this is still not correct. we do not want to show anything in chat if the setting is disabled.

hannibal002 (changes requested):
see above

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
AverageUser125: I changed the description to accommodate that change.

also, the point of this PR is TO SHOW the thing in chat even if the setting is disabled.
That’s literally the entire point.

hannibal002: why would we want to show it still? if thats the goal of the pr, then i miissed that, big time. pleas explain.
whats the point of the setting if its not hiding it fully? 

even if we want the current behaviour, the toggle's desc is then wrong atm

AverageUser125: So, the goal was: Tell the person getting the achievement which player is the contributor.

the simplest solution was reusing the message from the setting, I could have made it a separate chat message but reusing it was the simplest.

The chat message already shows the ACHIEVEMENT GET message, so I assumed adding this extra line is fine.

the point of the setting being default disabled is that people don’t bother skyhanni contributors too much, and that if only shows for the first one found.
(If for every unique it would send a chat message it would get annoying fast, for both the users and the contributors)

hannibal002 (changes requested):
Thanks for explaining the goal, that helps a lot. I think the tricky part is
that one boolean is doing two jobs here: "tell me about every contributor" and
"confirm the one-time achievement". The second one belongs to the achievement,
which already has its own toggle in `misc.achievementMessages`.

If we gate it there instead, you get exactly what you want and
`discoverContributorMessage` keeps a real off state. Either put the player name
into the achievement message itself, or keep your separate line and gate it on
the achievement toggle, whichever you prefer.

src/main/java/at/hannibal2/skyhanni/config/features/dev/DevConfig.kt
```diff
@@ -106,7 +106,8 @@ class DevConfig {
     @Expose
     @ConfigOption(
         name = "Discover Contributor Message",
-        desc = "Sends a message to your chat when you see a unique contributor.",
+        desc = "Sends a chat message when seeing a contributor. " +
```
hannibal002: the desc is no longer accurate

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
```diff
@@ -92,36 +105,56 @@ object ContributorAchievement {
...
     }
 
-    fun onUniqueContributorSeen() {
-        AchievementManager.completeAchievement(CONTRIBUTOR_ACHIEVEMENT)
+    fun onUniqueContributorSeen(profile: GameProfile) {
+        val completed = AchievementManager.completeAchievement(CONTRIBUTOR_ACHIEVEMENT)
+
+        // If the achievement was completed and achievement messages are enabled,
+        // or if the discover contributor message is enabled, then show the message.
+        if ((!completed || !achievementMessages) && !config.discoverContributorMessage) {
```
hannibal002: the line is complicated to read, please extract as local boolean or similar

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
```diff
@@ -9,21 +10,33 @@ import at.hannibal2.skyhanni.events.FriendRequestExpiredEvent
...
+import com.mojang.authlib.GameProfile
+import net.minecraft.ChatFormatting
+import net.minecraft.network.chat.Component
+import net.minecraft.network.chat.contents.objects.PlayerSprite
+import net.minecraft.world.item.component.ResolvableProfile
 
 @SkyHanniModule
 object ContributorAchievement {
+    private val config get() = SkyHanniMod.feature.dev
+    private val achievementMessages get() = SkyHanniMod.feature.misc.achievementMessages
```
hannibal002: reuse AchievementManager

hannibal002 (changes requested):
The forced message is word for word the discovery message. A
user who turns "Discover Contributor Message" off and then sees "A wild SkyHanni
contributor appears!" in chat will report that as a bug, regardless of which
toggle it hangs on internally. Could you either move the name into the
achievement message itself, or reword this one so it reads as achievement
output?

"player head sprite" should not be upper case in the changelog

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
```diff
@@ -92,36 +103,65 @@ object ContributorAchievement {
...
+        ChatUtils.chat {
+            appendWithColor("A wild SkyHanni contributor appears!", ChatFormatting.GOLD)
+            appendWithColor(" (hover)", ChatFormatting.GRAY)
+            hover = message
+        }
+    }
+
+    private fun showContributorAchievement(profile: GameProfile): Boolean {
+        val completed = AchievementManager.completeAchievement(CONTRIBUTOR_ACHIEVEMENT)
+        if (!completed && !shouldShowMessages) return false
```
hannibal002: wrong operator. ues || instead of &&

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
```diff
@@ -92,36 +103,65 @@ object ContributorAchievement {
...
+        if (!config.discoverContributorMessage) return
+        val message = getDiscoverComponent(profile)
+        ChatUtils.chat {
+            appendWithColor("A wild SkyHanni contributor appears!", ChatFormatting.GOLD)
+            appendWithColor(" (hover)", ChatFormatting.GRAY)
+            hover = message
+        }
+    }
+
+    private fun showContributorAchievement(profile: GameProfile): Boolean {
```
hannibal002: The name says "show" but this also completes the achievement. Could you move the `completeAchievement` call up into `onUniqueContributorSeen` so the side effect is visible at the call site?

hannibal002 (changes requested):
The first changelog entry still needs to start with the verb

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
```diff
@@ -9,21 +10,31 @@ import at.hannibal2.skyhanni.events.FriendRequestExpiredEvent
...
 import at.hannibal2.skyhanni.utils.compat.componentBuilder
+import at.hannibal2.skyhanni.utils.compat.hover
+import com.mojang.authlib.GameProfile
+import net.minecraft.ChatFormatting
+import net.minecraft.network.chat.Component
 
 @SkyHanniModule
 object ContributorAchievement {
+    private val config get() = SkyHanniMod.feature.dev
+    private val shouldShowMessages get() = AchievementManager.shouldShowMessages
```
hannibal002: this has only one usage. inline

hannibal002 (changes requested):

hannibal002 (approved):

PR #6269, opened 2026-08-03
Backend: RGB sea creature display name

AverageUser125:
## What
Torhus canyon should be orange.
But we got gold instead.
This will let it also be orange.

Since `TitleManager` does not support `Component` I had to make that a UI element.

+ Added RGB support for sea creature display name. - Avrg


AverageUser125:
Semi-trivial to reimplement/reopen.
And is not needed for a while

PR #6271, opened 2026-08-03
Backend: Migrate remaining direct Hud and Screen uses

lunaynx:
## What
These were missed in #6144. Very small PR, just to keep the diff size of #5894 going down.

exclude_from_changelog


src/main/java/at/hannibal2/skyhanni/features/chat/CopyChat.kt
```diff
@@ -63,10 +64,10 @@ object CopyChat {
 
     private fun getChatLine(mouseX: Int, mouseY: Int): GuiMessage? {
         val mc = Minecraft.getInstance()
```
CalMWolfs, line 66: Maybe I'm not fully understanidn the intentions of mc compat but shouldn't we also be replacing the 2 usages of this variable with minecraft compat / other utils methods

lunaynx: It could, but that's not strictly necessary right now, because those didn't change in 26.2.

CalMWolfs (approved):

PR #6281, opened 2026-08-04
Backend: No Top Level Function/Property Detekt rule

AverageUser125:
## What
Hannibal made this a rule, so I made a small detekt rule to enforce it.
Baselined for now.

+ Added Detekt rule to forbid top level function and property declarations. - Avrg


AverageUser125:
Sometimes desired behavior, and I can’t formulaize into a rule with so unclear requirements.

PR #6286, opened 2026-08-04
Fix: Safari -> Critter Safari

lunaynx:
## What
Was a late-ish change on alpha. (Why is this data duplicated in code when it's already in the repo?)

<details>
<summary>Images</summary>

</details>

## Changelog Fixes
+ Fixed SkyHanni saying "Safari" instead of "Critter Safari". - Luna


CalMWolfs (approved):
should also rename it in the repo islands file

PR #6289, opened 2026-08-05
Backend: Update Detekt to 2.0.0-alpha.6

lunaynx:
## What
Updates Detekt to 2.0.0-alpha.6 and fixes new errors.

+ Updated Detekt from 2.0.0-alpha.5 to 2.0.0-alpha.6. - Luna

### DIFF
--- a/detekt/build.gradle.kts
+++ b/detekt/build.gradle.kts
@@ -1,7 +1,5 @@
 import dev.detekt.gradle.Detekt
 import dev.detekt.gradle.DetektCreateBaselineTask
-import org.gradle.api.artifacts.ComponentMetadataContext
-import org.gradle.api.artifacts.ComponentMetadataRule
 
 plugins {
     kotlin("jvm")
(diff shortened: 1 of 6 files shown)


detekt/build.gradle.kts
```diff
@@ -9,32 +9,7 @@ plugins {
     id("dev.detekt")
 }
```
hannibal002, line 9: cleanup unused imports in this file

src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
```diff
@@ -400,13 +398,12 @@ object ItemUtils {
     }
 
     class AutoUpdatingRepoSkullItemStack internal constructor(
-        private val displayName: String,
-        private val uuid: String,
+        displayName: String,
+        uuid: String,
         private val repoSkullId: String,
```
hannibal002: repoSkullId is used exactly like the other four, why does it stay a property?

detekt/build.gradle.kts
```diff
@@ -9,32 +9,7 @@ plugins {
...
-            }
-        }
-    }
-}
-
 dependencies {
-    // TODO remove this once Detekt 2.0.0-alpha.6 is out
-    // https://github.com/detekt/detekt/issues/9409
-    components {
-        withModule<DetektTestMetadataRule>("dev.detekt:detekt-test")
```
hannibal002, line 35: detekt/src has no test source set, so does anything actually resolve detekt-test in CI? Did you run :detekt:build locally?

hannibal002 (changes requested):
Please add a Technical Details entry for the version bump instead of exclude_from_changelog.

src/main/java/at/hannibal2/skyhanni/utils/ItemUtils.kt
lunaynx: Detekt didn't catch it, didn't check beyond that. Seems to still compile after removing the `val`, so probably safe to change.

detekt/build.gradle.kts
lunaynx: This workaround was previously in place because 2.0.0-alpha.4 and 2.0.0-alpha.5 not shipping those caused Detekt scans to fail with a dependency error whenever configuration cache was used. Not because we actually cared about tests. This was fixed in 2.0.0-alpha.6, so the workaround is no longer necessary.

```
$ gradlew :detekt:build
Configuration on demand is an incubating feature.
Calculating task graph as no cached configuration is available for tasks: :detekt:build
Running Stonecutter 0.9.7

BUILD SUCCESSFUL in 2s
11 actionable tasks: 11 up-to-date
Configuration cache entry stored.
```

lunaynx:
I'm honestly not sure what's going on with the diff here, so I'm going to attempt to redo the source changes from a clean slate.

hannibal002 (approved):

PR #6293, opened 2026-08-05
Feature: Mute Beeheemoth

AverageUser125:
## What
Mutes the sounds from the beeheemoth.
I need to check this doesn't have side effects.
Could also use the chat messages to ensure beeheemoth exists, or just if "loop_aggressive" is only from beeheemoth use the sound location and mute sounds from near there.

Note:
HAS BEEn tested.
AND DOES WORK, with no muting random stuff.

## Changelog New Features
+ Added Mute Beeheemoth. - Avrg


AverageUser125:
Simple enough to re open. And also other mods probably got it since it took too longer to NOT get merged

PR #6298, opened 2026-08-05
Fix: Error when using /shnav with a search term

hannibal002:
## Dependencies
- #6297

## What
`/shnav <search>` runs its lookup in a coroutine, so `TextHelper.displayPaginatedList` ended up measuring text width off the main thread. `Font.width` bakes a glyph on first use, and that bake uploads a texture, which throws when it does not happen on the render thread.

`displayPaginatedList` now runs through `DelayedRun.runOrNextTick`, so every caller reaches the font and the chat from the main thread. Callers already on the main thread are unaffected, the task runs immediately.

## Changelog Fixes
+ Fixed an error when using /shnav with a search term. - hannibal2


lunaynx:
I don't think this ever crashed, unless recent changes made it crash, but it would error in chat and stop working until lobby swap.

hannibal002:
ur right, i phrased that too dramatic. its just a red error msg and a temp broken feature

PR #6299, opened 2026-08-05
Improvement: Support SkyBlock Wiki coordinate format

hannibal002:
## What
The SkyBlock Wiki separates coordinate values with a comma followed by a narrow no-break space (U+202F) instead of a regular space, so pasting a coordinate straight from the wiki into a command failed to parse and fell through to the search argument instead.

## Changelog Improvements
+ Added support for coordinates copied from the SkyBlock Wiki in commands that take coordinates, for example /shnav. - hannibal2

## Dependencies habe ich weggelassen, es gibt keine blockierenden PRs.


lunaynx:
Why not just strip the character before matching?

hannibal002:
> Why not just strip the character before matching?

2 reasons

1. i didnt think of this beofre
2. the current regex explains this explicitly as expected. just using a `replace(fancy char, normal space)` would be maybe confusing.

lunaynx:
I mean yeah, but nothing stops you from explaining the `replace` line with comments instead.

hannibal002:
have tried it, doesnt seem to work, ill just leave the clean regex as is ig

PR #6300, opened 2026-08-05
Backend: Repo Pattern Variable Style Detekt Rule

AverageUser125:
## What
It's already implied, so might aswell make a detekt rule and make it official.

+ Added Detekt Rule to enforce repo pattern variable name convention. - Avrg


detekt/src/main/kotlin/repo/RepoPatternVariableNameStyle.kt
```diff
@@ -0,0 +1,24 @@
...
+        super.visitPropertyDelegate(delegate)
+        val repoPatternElement = ctx.getRepoPatternElement(delegate) ?: return
+        val variableName = repoPatternElement.variableName
+
+        if (!variableName.endsWith("Pattern")) {
+            delegate.reportIssue("Repo pattern variable `${variableName}` must end with 'Pattern'.")
+        }
+
+        if (!variableName.first().isLowerCase()) {
+            delegate.reportIssue("Repo pattern variable `${variableName}` must start with a lowercase letter.")
```
lunaynx: This feels like it should be a more generic rule. No variable names should start with uppercase letters unless they're `ALL_UPPERCASE` because they're constants (not always `const`, because many things can't be compile-time constants).

AverageUser125: so ensuring it's camelCase. fair enough.

AverageUser125: to be honest, there is probably another detekt rule that already does this camel case check.
So I can probably just remove it.

AverageUser125:
Simple enough to reimplement if actually wanted

PR #6302, opened 2026-08-05
Backend: Repo Integer/Double/String

AverageUser125:
## What
Extended the notion of RepoPattern to allow any string, double, float and integer.
Same with RepoPatternList.
Also moved the responsability of parsing the string/list of strings to the base class instead of the manager.
(Maybe should just provide them a String/List<String>?)

I probably need to restructure RepoPatternGroup, or maybe RepoNumber be a sealed interface or smth.
But this works and runs and functions, which is already good.

Please double check I didn't make mistakes. as broken REPO loading is a recipe for disaster.

+ Extended RepoPattern to support numbers and generic strings. - Avrg


AverageUser125:
Probably not backwards compatible the way I did it.

PR #6307, opened 2026-08-05
Backend: Parse NEU repo SNBT overlays

lunaynx:
## What
NEU repo wants to get rid of `ItemModel` (and possibly later `nbttag` entirely) from their JSON files, so it is about time to start parsing the SNBT overlays.

+ Added support for NEU repo SNBT overlays. - Luna
+ Stopped using deprecated ItemModel JSON field from NEU repo in favor of SNBT. - Luna


lunaynx:
Okay, realistically the user-facing changelog is not very useful here. Changing it back to "Backend" until I can name an actual user-facing benefit. It has a requested review anyway, so the category doesn't matter much. (It would turn into a "Fix" if NEU repo decided to remove `nbttag`/`ItemModel` today, but it hasn't yet, and it doesn't seem like they're planning to pull the plug early.)

src/main/java/at/hannibal2/skyhanni/api/enoughupdates/EnoughUpdatesManager.kt
```diff
@@ -133,6 +190,83 @@ object EnoughUpdatesManager {
...
+        overlayOps?.takeIf { overlayOpsRegistries === registries }?.let { return it }
+        return RegistryOps.create(NbtOps.INSTANCE, registries).also {
+            overlayOps = it
+            overlayOpsRegistries = registries
+        }
+    }
+
+    private fun NeuItemJson.overlayComponentPatch(): DataComponentPatch? {
+        val overlay = itemOverlay ?: return null
+        val ops = currentOverlayOps() ?: return null
```
hannibal002, line 240: `DeferredItemStack.bindComponentsIfReady` gates on `SafeItemStackUtils.canBindComponents`, which checks `BuiltInRegistries.ITEM` and not the world registries this patch depends on. The factory runs once, so a stack whose components are first read without a world keeps the old state for its lifetime, with nothing logged.

src/main/java/at/hannibal2/skyhanni/api/enoughupdates/EnoughUpdatesManager.kt
```diff
@@ -37,16 +39,59 @@ import at.hannibal2.skyhanni.utils.json.fromJsonOrNull
...
+    val path: String,
+)
+
+internal class NeuItemOverlay(
+    val itemId: String,
+    val components: CompoundTag,
+    val path: String,
+) {
+    private var patch: DataComponentPatch? = null
+    private var patchOps: RegistryOps<Tag>? = null
```
hannibal002, line 74: These two are written non atomically and read from whichever thread builds the stack first.

src/main/java/at/hannibal2/skyhanni/api/enoughupdates/EnoughUpdatesManager.kt
```diff
@@ -133,6 +190,83 @@ object EnoughUpdatesManager {
...
+            .maxByOrNull { it.version }
+            ?: candidates.minByOrNull { it.version }
+    }
+
+    internal fun parseNeuItemOverlay(snbt: String, path: String): NeuItemOverlay {
+        val rootTag = TagParser.parseCompoundFully(snbt)
+        val id = rootTag.getString("id").orElseThrow {
+            IllegalArgumentException("NEU item overlay is missing id")
+        }
+        val components = rootTag.getCompound("components").orElseThrow {
```
hannibal002, line 223: Does NEU guarantee components is always present? Otherwise the id change is dropped too.

src/test/java/at/hannibal2/skyhanni/test/NeuItemOverlayTest.kt
```diff
@@ -0,0 +1,189 @@
...
+
+    @Test
+    fun `malformed overlay does not prevent another overlay from being parsed`() {
+        val fileSystem = repoFileSystem(
+            "itemsOverlay/100/BAD.snbt" to """{id:"minecraft:stone",components:{""",
+            "itemsOverlay/100/GOOD.snbt" to validOverlay("minecraft:diamond"),
+        )
+        val candidates = EnoughUpdatesManager.run { fileSystem.findNeuItemOverlayCandidates() }
+
+        val parsed = candidates.mapNotNull { (internalName, itemCandidates) ->
```
hannibal002, line 145: This tests the loop rebuilt in the test, not `applySelectedNeuItemOverlay`.

src/main/java/at/hannibal2/skyhanni/data/repo/filesystem/MemoryRepoFileSystem.kt
```diff
@@ -43,9 +43,24 @@ class MemoryRepoFileSystem(
...
+            .filter { "/" !in it && it.endsWith(".$extension") }
+            .toList()
+    }
+
+    override fun listDirectories(path: String): List<String> {
+        val prefix = path.toPrefix()
+        return storage.keys.asSequence()
+            .filter { it.startsWith(prefix) }
+            .map { it.removePrefix(prefix) }
+            .mapNotNull { it.substringBefore('/').takeIf { dir -> dir != it } }
```
hannibal002, line 60: `it` comes from the outer lambda here, which reads like a self comparison.

src/main/java/at/hannibal2/skyhanni/api/enoughupdates/EnoughUpdatesManager.kt
```diff
@@ -133,6 +190,83 @@ object EnoughUpdatesManager {
         }
     }
 
+    internal fun RepoFileSystem.findNeuItemOverlayCandidates(): Map<String, List<NeuItemOverlayCandidate>> {
+        return listDirectories("itemsOverlay")
+            .mapNotNull { it.toIntOrNull() }
```
hannibal002, line 195: `toIntOrNull` instead of `NumberUtil.formatIntOrNull`, intentional for a folder name?

src/main/java/at/hannibal2/skyhanni/api/enoughupdates/EnoughUpdatesManager.kt
```diff
@@ -37,16 +39,59 @@ import at.hannibal2.skyhanni.utils.json.fromJsonOrNull
...
 import net.minecraft.world.item.ItemStackTemplate
 import java.io.File
 import java.util.TreeMap
+import kotlin.jvm.optionals.getOrNull
 import kotlin.math.floor
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 
+
```
hannibal002, line 61: Double blank line.

src/main/java/at/hannibal2/skyhanni/api/enoughupdates/EnoughUpdatesManager.kt
```diff
@@ -133,6 +190,83 @@ object EnoughUpdatesManager {
...
+        internalName: String,
+        candidates: List<NeuItemOverlayCandidate>,
+        fileSystem: RepoFileSystem,
+        worldDataVersion: Int,
+    ) {
+        val candidate = selectNeuItemOverlayCandidate(candidates, worldDataVersion) ?: return
+        runCatching {
+            val snbt = String(fileSystem.readAllBytes(candidate.path), Charsets.UTF_8)
+            val overlay = parseNeuItemOverlay(snbt, candidate.path)
+            itemId = overlay.itemId
```
hannibal002, line 254: The overlay id is already modern, but `toStack` still runs it through `convertMinecraftIdToModern`.

<details><summary>why</summary>

That function is a legacy id table and two of its entries collide with modern ids:
`minecraft:melon` becomes `melon_slice`, `minecraft:snow` becomes `snow_block`. Today the ItemModel from the nbttag masks this by drawing the melon block on top of the melon slice. Without it, six items render as the wrong item: `MELON_BLOCK`, `ENCHANTED_MELON_BLOCK`, `VAMPIRIC_MELON`, `ENCHANTED_VAMPIRIC_MELON`, `MELON_BARN_SKIN` and `SNOW`.

Taking the overlay id directly fixes both at once, `MELON_BLOCK` then is a real melon block and needs no item_model at all.

</details>

<details><summary>coverage numbers</summary>

Measured against the current repo: of 8,192 items with ItemModel in their nbttag, 7,102 point at the item's own id and are no-ops, 1,090 are real models. 1,084 of those are covered by an overlay carrying `minecraft:item_model`, only 2 items have no overlay file at all. The six above are the entire remainder, so dropping the nbttag field is safe apart from them.

</details>

src/main/java/at/hannibal2/skyhanni/api/enoughupdates/EnoughUpdatesManager.kt
```diff
@@ -37,16 +39,59 @@ import at.hannibal2.skyhanni.utils.json.fromJsonOrNull
...
+
+    /**
+     * Components such as enchantments or banner patterns are backed by registries that the client only receives from
+     * the server, so the patch can not be built while the repo is loading.
+     */
+    fun componentPatch(ops: RegistryOps<Tag>): DataComponentPatch? {
+        if (patchOps === ops) return patch
+        patch = DataComponentPatch.CODEC.parse(ops, components).resultOrPartial { error ->
+            ErrorManager.logErrorWithData(
+                IllegalStateException(error),
```
hannibal002, line 84: `logErrorStateWithData`, the throwable is only constructed here. See CONTRIBUTING.md, ErrorManager.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/api/enoughupdates/EnoughUpdatesManager.kt
Alex33856: Every item should have `components` (for the `minecraft:custom_data` component with the skyblock item ID).

It's possible to replace that with `getCompoundOrEmpty`, but it probably shouldn't be.

hannibal002: Thanks. All 8,190 overlay files across the six version folders have one, so keeping the throw.

PR #6315, opened 2026-08-06
Fix: Fishing Hook Display Stop not working sometimes

AverageUser125:
## What
Instead of doing a check each tick, it will see all custom name changes and create the display/renderable once.
If it's there is not known entity it uses the detected one,
if it's the same entity as the previous it just updates,
and if it's a different entity it uses the one closest to the bobber.

I have tested in torrhus, and it works.
(Some users report that it currently randomly stops showing the display like 1h after fishing, but I can't replicate that or prove this fixed it. but I hope the "closest to bobber" logic maybe fixes that)

## Changelog Fixes
+ Fixed the Fishing Hook Display not showing while other players are fishing nearby. - Avrg

+ Improved FishingHookDisplay to only run on entity custom name change instead of each tick. - Avrg


hannibal002:
damage splashes would show up in the display for short period of times, fixed it

PR #6318, opened 2026-08-06
Fix: Skymall display in Safari

AverageUser125:
## What
Safari is not a mining island.

## Changelog Fixes
+ Fixed Skymall Display showing in Safari. - Avrg


CalMWolfs:
No lie I didn't see this but I had also made my branch like a day ago I just forgot to open a pr as I started another safari run and then I went to bed

PR #6323, opened 2026-08-07
Improvement: Error on unsupported platform for See Through Farming

AverageUser125:
## Dependencies
- #6373

## What
Makes it give up after first attempt of it failing.
And actually reports to the user when it fails.

Needed for 26.3 due to Wayland being default enabled.

## Changelog Improvements
+ Added a chat message when your window system does not support See Through Farming. - Avrg


src/main/java/at/hannibal2/skyhanni/features/garden/SeeThroughWindow.kt
```diff
@@ -2,60 +2,82 @@ package at.hannibal2.skyhanni.features.garden
...
+    private fun isGlfwPlatformError(error: Int): Boolean =
+        when (error) {
+            GLFW.GLFW_PLATFORM_ERROR,
+            GLFW.GLFW_PLATFORM_UNAVAILABLE,
+            GLFW.GLFW_NOT_INITIALIZED,
+            GLFW.GLFW_FEATURE_UNAVAILABLE,
+            GLFW.GLFW_FEATURE_UNIMPLEMENTED,
+            -> true
+            else -> false
+        }
```
lunaynx: I'd change it up a bit, use `equalsOneOf` and probably also make it an extension function:
```suggestion
    private fun Int.isGlfwPlatformError(): Boolean = equalsOneOf(
        GLFW.GLFW_PLATFORM_ERROR,
        GLFW.GLFW_PLATFORM_UNAVAILABLE,
        GLFW.GLFW_NOT_INITIALIZED,
        GLFW.GLFW_FEATURE_UNAVAILABLE,
        GLFW.GLFW_FEATURE_UNIMPLEMENTED,
    )
```
(don't forget to add the import and change the usage)

AverageUser125: I didn't want it to be an extension function, since it's like a pure function which just takes a parameter and stuff.
But sure, that works too.

AverageUser125: In the end I just check if there was any error, since that list is exhaustive.
And there is no other way for setting opacity to fail.

src/main/java/at/hannibal2/skyhanni/features/garden/SeeThroughWindow.kt
```diff
@@ -2,60 +2,67 @@ package at.hannibal2.skyhanni.features.garden
...
-        } else if (opacityChanged) {
-            GLFW.glfwSetWindowOpacity(handle, 1f)
-            opacityChanged = false
+
+        if (currentOpacity == targetOpacity) return
+        if (setWindowOpacity(targetOpacity)) {
+            currentOpacity = targetOpacity
+        } else {
+            unsupportedPlatform = true
+            ChatUtils.userError("Your platform doesn't support see through windows!")
```
hannibal002: Name the feature here too, it is called See Through Farming in the config.

src/main/java/at/hannibal2/skyhanni/features/garden/SeeThroughWindow.kt
```diff
@@ -2,60 +2,67 @@ package at.hannibal2.skyhanni.features.garden
...
         }
     }
+
+    private fun setWindowOpacity(alpha: Float): Boolean {
```
hannibal002: `trySetWindowOpacity` reads better for a boolean return.

src/main/java/at/hannibal2/skyhanni/features/garden/SeeThroughWindow.kt
```diff
@@ -2,60 +2,67 @@ package at.hannibal2.skyhanni.features.garden
...
 
     private fun setOpacity() {
-        val handle = Minecraft.getInstance().window.handle()
-        if (!isActive) {
-            if (opacityChanged) {
-                GLFW.glfwSetWindowOpacity(handle, 1f)
-                opacityChanged = false
-            }
-            return
+        if (unsupportedPlatform) return
```
hannibal002, line 44: Should this also block resetting to opaque? A failure on the way back would leave the window transparent for the rest of the session.

hannibal002 (changes requested):
Changelog should name the feature and what the user actually sees: `Added a chat message when your window system does not support See Through Farming.`

src/main/java/at/hannibal2/skyhanni/features/garden/SeeThroughWindow.kt
AverageUser125: The ONLY possible error is unsupported platform or generic platform error,
there is no other possible failure for the GLFW function.

Also, if there was an error. It means the function doesn’t work. So not point trying to call it again just to reset.

src/main/java/at/hannibal2/skyhanni/features/garden/SeeThroughWindow.kt
```diff
@@ -2,60 +2,68 @@ package at.hannibal2.skyhanni.features.garden
...
+        if (currentOpacity == targetOpacity) return
+        if (trySetWindowOpacity(targetOpacity)) {
+            currentOpacity = targetOpacity
+        } else {
+            unsupportedPlatform = true
+            ChatUtils.userError("Your platform doesn't support See Through Farming Windows!")
         }
     }
+
+    // This function can only fail due to platform limitations
```
hannibal002: GLFW also lists `GLFW_NOT_INITIALIZED` and a generic `GLFW_PLATFORM_ERROR` for this call, so `only` is too strong.

src/main/java/at/hannibal2/skyhanni/features/garden/SeeThroughWindow.kt
```diff
@@ -2,60 +2,68 @@ package at.hannibal2.skyhanni.features.garden
...
-        } else if (opacityChanged) {
-            GLFW.glfwSetWindowOpacity(handle, 1f)
-            opacityChanged = false
+
+        if (currentOpacity == targetOpacity) return
+        if (trySetWindowOpacity(targetOpacity)) {
+            currentOpacity = targetOpacity
+        } else {
+            unsupportedPlatform = true
+            ChatUtils.userError("Your platform doesn't support See Through Farming Windows!")
```
hannibal002: The feature is called See Through Farming, and the changelog says window system: `Your window system doesn't support See Through Farming!`

hannibal002 (changes requested):

hannibal002 (approved):

PR #6326, opened 2026-08-07
Backend: Improved Enforced Config Values

lunaynx:
## What
This PR includes several improvements to the Enforced Config Values system, listed in the changelog.

+ Enforced config values now support enforcing enum values without crashing, and have been hardened against other invalid values. - Luna
+ Enforced config values are now applied early on game start and whenever the repo is reloaded, without requiring the player to open a GUI or be on SkyBlock. It also no longer tries to apply every tick. - Luna
+ Enforced config values are no longer persisted to the user's config unless `"persist": true` is specified, so the user's original config value is automatically restored once the value is no longer enforced instead of being lost. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
@@ -85,6 +85,8 @@ class ConfigManager {
         }
 
         val features = SkyHanniMod.feature
+        // Has to run before the config is built, so that enforced options are already blocked off
+        EnforcedConfigValues.loadFromLocalRepo()
         recreateConfig()
 
         try {
--- a/src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
@@ -73,7 +73,6 @@ class ConfigManager {
         }
         configDirectory.mkdirs()
 
-
         for (fileType in ConfigFileType.entries) {
             val clazzInstance = fileType.clazz.getDeclaredConstructor().newInstance()
             setConfigHolder(fileType, firstLoadFile(fileType.file, fileType, clazzInstance))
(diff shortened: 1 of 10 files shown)


src/main/java/at/hannibal2/skyhanni/config/EnforcedConfigValues.kt
```diff
@@ -68,25 +108,91 @@ object EnforcedConfigValues {
...
+            // The backup is re-read so that it compares equal to what the field serializes to later
+            // (e.g. a float from the repo JSON does not equal the same float serialized by Gson)
+            if (enforcedValue.persist) userValues.remove(enforcedValue.path)
+            else userValues[enforcedValue.path] = UserValue(userValue, shimmy.getJson())
+        }
+    }
+
+    private fun restoreNoLongerEnforced(config: Any, enforcedPaths: Set<String>) {
+        for (path in userValues.keys.filter { it !in enforcedPaths }) {
+            val backup = userValues.remove(path) ?: continue
```
hannibal002, line 157: Remove from userValues only after setJson succeeded. A throwing observer currently loses the user's value for good, while enforceValue guards the same case with finally.

src/main/java/at/hannibal2/skyhanni/config/EnforcedConfigValues.kt
```diff
@@ -5,50 +5,90 @@ import at.hannibal2.skyhanni.api.event.HandleEvent
...
+    // Guarded by its own monitor: the config auto-save thread reads it while it serializes the config.
+    internal val userValues = mutableMapOf<String, UserValue>()
+
+    internal class UserValue(val userValue: JsonElement, val enforcedValue: JsonElement)
+
+    /**
+     * Applies the values cached in the local repo, so that enforcement does not have to wait for the
+     * initial repo fetch. Called during config loading, before the config gets built for the first time.
+     */
+    fun loadFromLocalRepo() {
```
hannibal002, line 45: initRepo has paths that return without posting a reload event. The values applied here then stay enforced and locked in the GUI for the whole session, with nothing to clear them.

src/test/java/at/hannibal2/skyhanni/test/config/EnforcedConfigValuesTest.kt
```diff
@@ -0,0 +1,78 @@
...
+import com.google.gson.JsonPrimitive
+import io.github.notenoughupdates.moulconfig.observer.Property
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertFalse
+import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.assertThrows
+import org.junit.jupiter.api.Test
+
+class EnforcedConfigValuesTest {
```
hannibal002, line 14: Please add tests for toUserJsonTree, the restore path and persist. Only the in-memory bookkeeping is covered, and toUserJsonTree is the part that can write wrong data into config.json.

src/main/java/at/hannibal2/skyhanni/config/EnforcedConfigValues.kt
```diff
@@ -5,50 +5,90 @@ import at.hannibal2.skyhanni.api.event.HandleEvent
...
+            updateData(json)
+        } catch (e: Exception) {
+            // Gson does not enforce Kotlin nullability, so a malformed cache can still fail here.
+            // Nothing is enforced until the repo reload delivers proper data, rather than failing the config load.
+            enforcedConfigValuesData = listOf()
+            ErrorManager.logErrorWithData(e, "Failed to apply cached enforced config values")
+        }
+    }
+
+    @HandleEvent(priority = HandleEvent.HIGHEST)
```
hannibal002, line 57: The enforcement is deferred to the next tick anyway, so HIGHEST only moves the constant read. Can the HIGHEST to HIGH changes in IslandType, PowderMiningChatFilter and RemainingSlayerKills be reverted?

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -206,18 +207,25 @@ class ConfigManager {
...
         val json = jsonHolder[fileType] ?: error("Could not find json object for $fileType")
-        saveFile(fileType.file, fileType.fileName, json, reason)
-        saveFile(fileType.backupFile, fileType.fileName, json, reason)
+        saveFile(fileType.file, fileType, json, reason)
+        saveFile(fileType.backupFile, fileType, json, reason)
+    }
+
+    private fun serialize(fileType: ConfigFileType, data: Any): String {
+        if (fileType != ConfigFileType.FEATURES) return gson.toJson(data)
+        // Enforced values only ever exist in memory, the file keeps the user's own values
```
hannibal002, line 216: Wrong for persist: true, where reaching the file is the whole point. Same omission in the toUserJsonTree doc.

hannibal002 (changes requested):

PR #6327, opened 2026-08-07
Fix: Default update stream not matching installed version type

lunaynx:
## What
Fixes fresh installs of SkyHanni betas defaulting to the Full update stream instead of Beta. Also performs a one-time config migration that flips any existing beta users to the Beta update stream. (If someone first encounters this config migration on a stable release, say, 9.0.0, they will *not* be migrated to Beta.)

## Changelog Fixes
+ Fixed SkyHanni betas defaulting to the Full update stream instead of Beta. - Luna


src/main/java/at/hannibal2/skyhanni/config/features/About.kt
```diff
@@ -40,6 +42,17 @@ class About {
...
+
+        companion object {
+            /**
+             * The stream matching the currently installed version, used as the default on a clean install.
+             *
+             * This reads [VersionConstants] instead of `SkyHanniMod.isBetaVersion` because the config is
+             * constructed while `SkyHanniMod` is still initialising.
+             */
+            fun forInstalledVersion(): UpdateStream =
+                if (ModVersion.fromString(VersionConstants.MOD_VERSION).isBeta) BETA else RELEASES
```
hannibal002: can we use ` SkyHanniMod.VERSION` here?
also, this is now a bit duplciate to `SkyHanniMod.isBetaVersion`

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/config/features/About.kt
lunaynx: The reason this isn't done is explained by the comment right above. Your suggestion happens to work right now, but is risky and adds unnecessary overhead in some places (e.g. unit tests) because it introduces a circular dependency, and would blow up the moment someone moves the initialization of `SkyHanniMod.feature` above `SkyHanniMod.VERSION`.

hannibal002: Fair. I still dont like the code duplication, how about we do something like
`val installed: ModVersion by lazy { fromString(VersionConstants.MOD_VERSION) }`

in `ModVersion`?

lunaynx: That doesn't work, `ModVersion` is in a different package and can't import `VersionConstants`.

hannibal002: not sure what is going on here. i can clearly import classes from different packages and compile it and run mc with the change. see that patch:


```
Subject: [PATCH] test
---
Index: src/main/java/at/hannibal2/skyhanni/utils/system/ModVersion.kt
IDEA additional info:
<+>UTF-8
diff --git a/src/main/java/at/hannibal2/skyhanni/utils/system/ModVersion.kt b/src/main/java/at/hannibal2/skyhanni/utils/system/ModVersion.kt
--- a/src/main/java/at/hannibal2/skyhanni/utils/system/ModVersion.kt	(revision c5a5ebe4e0bffbd4aadbd8c2945ea259c7b03b5a)
+++ b/src/main/java/at/hannibal2/skyhanni/utils/system/ModVersion.kt	(date 1786783863188)
@@ -1,5 +1,7 @@
 package at.hannibal2.skyhanni.utils.system
 
+import at.hannibal2.skyhanni.utils.VersionConstants
+
 data class ModVersion(val stable: Int, val beta: Int, val bugfix: Int) : Comparable<ModVersion> {
 
     val isBeta get() = beta != 0
@@ -25,6 +27,8 @@
     }
 
     companion object {
+        val installed: ModVersion by lazy { fromString(VersionConstants.MOD_VERSION) }
+
         fun fromString(version: String): ModVersion {
             val parts = version.split('.')
             return ModVersion(
Index: src/main/java/at/hannibal2/skyhanni/config/features/About.kt
IDEA additional info:
<+>UTF-8
diff --git a/src/main/java/at/hannibal2/skyhanni/config/features/About.kt b/src/main/java/at/hannibal2/skyhanni/config/features/About.kt
--- a/src/main/java/at/hannibal2/skyhanni/config/features/About.kt	(revision c5a5ebe4e0bffbd4aadbd8c2945ea259c7b03b5a)
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/About.kt	(date 1786783830392)
@@ -3,7 +3,6 @@
 import at.hannibal2.skyhanni.features.misc.update.ConfigVersionDeprecatedDisplay
 import at.hannibal2.skyhanni.features.misc.update.ConfigVersionDisplay
 import at.hannibal2.skyhanni.utils.OSUtils.openBrowser
-import at.hannibal2.skyhanni.utils.VersionConstants
 import at.hannibal2.skyhanni.utils.system.ModVersion
 import com.google.gson.annotations.Expose
 import io.github.notenoughupdates.moulconfig.annotations.Accordion
@@ -45,13 +44,12 @@
 
         companion object {
             /**
-             * The stream matching the currently installed version, used as the default on a clean install.
+             * The stream matching the installed version, used as the default on a clean install.
              *
-             * This reads [VersionConstants] instead of `SkyHanniMod.isBetaVersion` because the config is
-             * constructed while `SkyHanniMod` is still initialising.
+             * Reads [ModVersion.installed] and not `SkyHanniMod.isBetaVersion`, because the config is also
+             * constructed while `SkyHanniMod` is still initialising, where that property may be unset.
              */
-            fun forInstalledVersion(): UpdateStream =
-                if (ModVersion.fromString(VersionConstants.MOD_VERSION).isBeta) BETA else RELEASES
+            fun forInstalledVersion(): UpdateStream = if (ModVersion.installed.isBeta) BETA else RELEASES
         }
     }
 
Index: src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
IDEA additional info:
<+>UTF-8
diff --git a/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt b/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
--- a/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt	(revision c5a5ebe4e0bffbd4aadbd8c2945ea259c7b03b5a)
+++ b/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt	(date 1786783773113)
@@ -119,7 +119,7 @@
 
     fun id(path: String): Identifier = Identifier.fromNamespaceAndPath(MODID, path)
 
-    val modVersion: ModVersion = ModVersion.fromString(VERSION)
+    val modVersion: ModVersion = ModVersion.installed
 
     val isBetaVersion: Boolean
         get() = modVersion.isBeta

```

hannibal002 (changes requested):
see above

src/main/java/at/hannibal2/skyhanni/config/features/About.kt
lunaynx: Ah, I was looking at `buildSrc/src/main/kotlin/skyhannibuildsystem/utils/ModVersion.kt`. Didn't realize there was another nearly identical copy of that file.

hannibal002 (approved):

PR #6329, opened 2026-08-08
Fix: Hideyho finder working through walls

PanTruskawka045:
## What

HideyhoFinder could previously detect Hideyho through walls when it was in a nearby room. It now only detects Hideyho when it is actually visible.

## Changelog Fixes
+ Fixed Hideyho Finder sometimes marking Hideyho as found when he wasn't. - pan_truskawka045

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/hunting/safari/HideyhoFinder.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/hunting/safari/HideyhoFinder.kt
@@ -11,6 +11,7 @@ import at.hannibal2.skyhanni.features.misc.pathfind.NavigationCondition
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.DelayedRun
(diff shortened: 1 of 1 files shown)


CalMWolfs (changes requested):
I dont see how the code was ever running when he was through a wall from you as 2 of the checks only run when u get a chat message from him (which u have to click on him for) and the other check requires you to be at the node (within 2 blocks of him and there are no location where he is within 5 blocks of a node but through a wall)
Maybe I am mistaken but do you have video evidence / coords where this is possible. As I feel making these changes will just lead to the solver intermittently failing due to him being partially obscured by blocks

PanTruskawka045:
Maybe my explanation is misleading, but

here's the video: https://youtu.be/dhsOfn9DnRo

0:00 - 0:15 - "quest" start
0:45 - end - I reached one navigation node, and as far as I understand, the mod detected Hideyho and stopped navigating. Hideyho was in the house, I was outside, and I only knew where the NPC was because there was glass in the wall. 

Also, I have increased the detection radius to 10 blocks to reproduce the issue more consistently (HideyhoFinder.kt:111).

CalMWolfs (approved):

PR #6331, opened 2026-08-08
Fix: Not registering Karma books

CalMWolfs:
## What
It has no color codes.

## Changelog Fixes
+ Fixed not properly tracking Karma books from tree gifts. - CalMWolfs


jani270:
thank you calm

PR #6332, opened 2026-08-08
Improvement: Tab completion for /sh config categories (#209)

Zelmari:
# Improvement: Tab completion for /sh config categories and settings (#209)

## What
Adds tab completion for the `/sh` command:

- `/sh <TAB>` suggests all top-level config categories (About, GUI, Garden,
  Crimson Isle, The Rift, ...).
- `/sh <TAB>` also suggests config options, matched with the same logic the
  config GUI search uses.
- Entering an exact option path navigates straight to that setting in the
  config GUI (selects its category and scrolls it into view) via
  `MoulConfigEditor.goToOption`. Non-matching input still falls back to the
  normal text search.

This closes the `/sh <categories>` part of #209. Note that `/shcroptime`
already received tab completion in #5883, and `/shtrackcollection` was
removed in #5764.

Technical details:

- The option index is built by hooking into MoulConfig's runtime config
  walker (`ConfigStructureReader` + `ConfigProcessorDriver`, the same
  mechanism `FeatureToggleProcessor` uses), collecting every
  `@ConfigOption` field with its fully-qualified path
  (`ConfigGuiManager.collectOptionPaths`). Paths are reconstructed from the
  config object tree rather than MoulConfig's path stack, which keeps them
  consistent (always prefixed with the top-level category) and in sync when
  config options are added or renamed.
- Suggestions reuse MoulConfig's own search instead of duplicating it: the
  editor's default search function is literally
  `GuiOptionEditor.fulfillsSearch(String)` (which matches option name,
  description and `@SearchTag`s), so each option is matched via
  `option.getEditor().fulfillsSearch(term)` in addition to the field path.
  Matching mirrors the editor's semantics: trimmed, case-insensitive,
  `+`-separated terms that all have to match.
- Also fixed `BrigadierUtils.toSuggestionProvider()` to match case-
  insensitively, so capitalized suggestions (like category names) work.

Regarding #5883's discussion of `/sh` completion: this suggests the config
categories that `/sh` actually searches (unlike the `CommandCategory` enum
values that were originally tried there), and exact path matches genuinely
navigate to the setting instead of only searching, addressing the
"completion suggests something the command doesn't do" concern.

Covered by unit tests (`ConfigCategoryNamesTest`, `ConfigOptionIndexTest`,
`ConfigSearchMatcherTest`).

## Changelog Improvements
+ Improved `/sh` to tab-complete config categories and settings. - Zelmari

+ Built the config option index by hooking into MoulConfig's ConfigStructureReader, reusing its search matching for suggestions. - Zelmari

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
+++ b/src/main/java/at/hannibal2/skyhanni/SkyHanniMod.kt
@@ -15,6 +15,7 @@ import at.hannibal2.skyhanni.config.StorageData
 import at.hannibal2.skyhanni.config.commands.CommandCategory
 import at.hannibal2.skyhanni.config.commands.CommandRegistrationEvent
 import at.hannibal2.skyhanni.config.commands.brigadier.BrigadierArguments
+import at.hannibal2.skyhanni.config.commands.brigadier.BrigadierUtils.dynamicContainsSuggestionProvider
 import at.hannibal2.skyhanni.config.storage.AchievementStorage
 import at.hannibal2.skyhanni.config.storage.CustomTodosStorage
 import at.hannibal2.skyhanni.config.storage.OrderedWaypointsRoutes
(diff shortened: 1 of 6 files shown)


AverageUser125:
Is it possible to hook into MoulConfig annotation processor, 
and make this auto complete for all settings instead of just categories?

Zelmari:
@AverageUser125 Will try now

AverageUser125:
1. What about “SearchTag” annotation?
2. Is it possible to just reuse the searching inside MoulConfig? Meaning whatever it would show would be suggested without needing to duplicate its searching logic? (Unsure if possible tbh)

Zelmari:
@AverageUser125 I think it should be possible, the moulconfig search is just GuiOptionEditor.fulfillsSearch() and that already includes @SearchTag

PR #6333, opened 2026-08-08
Feature: Tooltip Price Info

OnlYKai:
## What
Adds various price information to the bottom of an item's tooltip (like NEU).

Currently this feature shows:
- Lowest BIN
- Insta Buy
- Insta Sell
- NPC Price
- Raw Craft Cost

## Changelog New Features
+ Added Tooltip Price Info feature. - OnlYKai
    * Shows lowest bin, insta buy/sell price, npc price and raw craft cost in tooltip.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/misc/MiscConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/misc/MiscConfig.kt
@@ -172,6 +172,11 @@ class MiscConfig {
     val boopParty: BoopPartyConfig = BoopPartyConfig()
 
     @Expose
+    @ConfigOption(name = "Tooltip Price Info", desc = "")
+    @Accordion
+    val tooltipPriceInfo: TooltipPriceInfoConfig = TooltipPriceInfoConfig()
+
+    @Expose
     @ConfigOption(name = "Reset Search on Close", desc = "Reset the search in GUIs after closing the inventory.")
     @ConfigEditorBoolean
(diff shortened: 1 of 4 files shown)


lunaynx:
Other mods already exist to do this. We're not looking to duplicate features for no good reason.

OnlYKai:
I haven't seen another mod add craft cost.
And having a single mod show all prices in a uniform style and customizable order seems beneficial.

AzureAaron:
> I haven't seen another mod add craft cost. And having a single mod show all prices in a uniform style and customizable order seems beneficial.

Skyblocker has had a crafting cost tooltip for two years now. If style is a concern I don’t see why an option to customize it can’t be added to SkyHanni (such as a Skyblocker/NEU style).

OnlYKai:
> Skyblocker has had a crafting cost tooltip for two years now. If style is a concern I don’t see why an option to customize it can’t be added to SkyHanni (such as a Skyblocker/NEU style).

Oh, I didn't know that. Hated the old config menu and haven't tried it ever since.

As for style, I was mainly referring to having to mix different mods styles. Which Skyblocker mostly solves. And as for adding a NEU style, it would be easier to do so directly in Skyblocker instead of Skyhanni.

The only thing I'm missing is the ability to freely reorder all the info added to the bottom of the tooltip (DraggableList), but i think that is currently not possible with Dandelion.

Anyway since Skyblocker has this feature and potential improvements would be Skyblocker related, I'll close this PR.

PR #6177, opened 2026-07-23
Improvement: Send Chat message about CustomScoreboard being deprecated

AverageUser125:
## What
Sends a chat message that Custom Scoreboard is deprecated.
For some reason `ConditionalUtils.onEnabled` is called 3 times, so I can't use it.

This feature is being replaced by a standalone mod with 100% feature parity + more.

## Changelog Improvements
+ Marked Custom Scoreboard as deprecated. - Avrg
    * The built-in version is no longer maintained and will be removed in favor of the standalone mod.

PR #6336, opened 2026-08-08
Remove: Custom Scoreboard

AverageUser125:
## Dependencies
- #6177

## What
It can't be broken if it doesn't exist. :pleading_face: 

## Changelog Removed Features
+ Removed Custom Scoreboard. - Avrg
    * This feature has been unmaintained and missing updates and fixes for some time, we recommend using the seperate dedicated mod [Custom Scoreboard](https://modrinth.com/mod/skyblock-custom-scoreboard).


CalMWolfs (changes requested):
The dedicated custom scoreboard mod imports the users settings from skyhanni. If you delete all the config options it has nothing to import so you are effectively wiping the users config. 
Either you need to leave all the config options and hide them all or have some sort of special config migration where it exports all these configs to a seperate scoreboard config.json file that you communicate with the devs of the other mod so they know where to also look

Also I don't think I saw any sort of warning / popup that appears when the user first plays and they previously had the custom scoreboard enabled. As they will be very confused

Rainymei:
merge pr #6177 for a few versions first

AverageUser125:
Yeah, this PR is for after the warning PR is merged.

AverageUser125:
I re-added the config, and made it only show up in dev environment (thanks luna for the idea).

PR #6202, opened 2026-07-26
Backend: Text Display Update Events

AverageUser125:
## What

added a shared event for Armor Stands AND Text displays incase hypixel decides to migrate 1 evening.
Also EntityEquipmentChange + EntityDeathEvent were using Entity when they were from a mixin in LivingEntity, so I fixed that.

The source of the events probably needs to be some other file, but I got no clue what's a good spot.

## Changelog Technical Details
+ Added TextDisplayEvents which fire for both armor stand and text display entity. - Avrg

PR #6337, opened 2026-08-08
Improvement: Update Lasso Display from entity events instead of every tick

lunaynx:
## What
Improves performance of Lasso Display by no longer fetching entities every tick while the player is holding a lasso, and instead using proper entity events.

## Changelog Improvements
+ Improved performance of Lasso Display. - Luna


AverageUser125:
This is like the 3rd PR where #6202 is useful lol.

PR #6348, opened 2026-08-10
Backend: Replace Apache Commons Net with a self-contained SNTP client

lunaynx:
## What
Removes the dependency on Apache Commons Net and replaces it with a self-contained SNTP client.

## Why
Pulling in the entire Commons Net library for a single feature that can be implemented without dependencies in under 100 lines is unnecessary. Dropping it saves nearly 1 MB on the JAR size.

+ Dropped the dependency on Apache Commons Net in favor of a self-contained SntpClient implementation. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/SntpClient.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/SntpClient.kt
@@ -1,0 +1,102 @@
+package at.hannibal2.skyhanni.utils
+
+import java.io.IOException
+import java.net.DatagramPacket
+import java.net.DatagramSocket
+import java.net.InetAddress
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.seconds
+
+/**
+ * Minimal SNTP client (RFC 4330) that queries an NTP server for the offset between its clock and the local one.
+ */
+object SntpClient {
+    private const val NTP_PORT = 123
+    private const val PACKET_SIZE = 48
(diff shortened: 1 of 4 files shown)


src/main/java/at/hannibal2/skyhanni/utils/SntpClient.kt
```diff
@@ -0,0 +1,88 @@
...
+
+            val response = DatagramPacket(buffer, buffer.size)
+            socket.receive(response)
+            val destinationTime = System.currentTimeMillis()
+
+            if (response.length < PACKET_SIZE) throw IOException("NTP response is too short (${response.length} bytes)")
+            val mode = buffer[0].toInt() and 0x7
+            if (mode != MODE_SERVER) throw IOException("NTP response has unexpected mode $mode")
+            val stratum = buffer[1].toInt() and 0xFF
+            if (stratum == 0) throw IOException("NTP server rejected the request (kiss-of-death)")
```
AverageUser125: shouldn't this throw a `SocketTimeoutException` or some other timeout exception?
Since the `ComputerTimeOffset` class catches timeouts like this and only reports when 10 of those happen.

AverageUser125 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/SntpClient.kt
lunaynx: This is not a timeout, it's usually used for rate limiting. We could probably drop down to `SLOW` mode when receiving this.

PR #6350, opened 2026-08-10
Backend: Replace libautoupdate with first-party update checking

lunaynx:
## What
Removes the dependency on libautoupdate and replaces it with first-party update checking.

## Why
We stopped doing automatic updates after the security incident, but we're still carrying around an auto-update library that needs hacks like `NoOpUpdateTarget` and pointlessly sets up a no-op shutdown hook. It's also kinda misleading to depend on a library that purports to do auto-updates when we no longer make use of that functionality.

+ Dropped the dependency on libautoupdate in favor of a first-party update checking implementation. - Luna


AverageUser125:
GitHub actions update source to a specific PR? 🥺 
(Probably a bad idea, but would mean people can use the 26.2 PR and still get updates)

PR #6352, opened 2026-08-10
Fix + Backend: Items sharing a display name

hannibal002:
## Dependencies
- https://github.com/hannibal002/SkyHanni-REPO/pull/829

## What
Two NEU repo items can share the exact same display name, here `§9Cropshot Chip`:

- [TUTORIAL_GARDEN_CHIP](https://github.com/NotEnoughUpdates/NotEnoughUpdates-REPO/blob/12f7726977b58e4a57cf4166ae0293bb90799a2d/items/TUTORIAL_GARDEN_CHIP.json)
- [CROPSHOT_GARDEN_CHIP](https://github.com/NotEnoughUpdates/NotEnoughUpdates-REPO/blob/12f7726977b58e4a57cf4166ae0293bb90799a2d/items/CROPSHOT_GARDEN_CHIP.json)

Which one wins depended on the NEU repo iteration order and could change with any repo update. Here the tutorial version won over the normal one.

Two repo lists replace the hardcoded builder block check: `ignored_internal_names` drops an item so the other one wins, `ambiguous_display_names` drops a name entirely where no candidate is the one meant. Both also apply to the mangled name fallback in `ItemResolutionQuery`. Remaining collisions are printed to the console on repo load.

## Changelog Fixes
+ Fixed Cropshot Chip being detected as the wrong item. - hannibal2

+ Moved the display name lookup exclusions into the repo. - hannibal2
+ Added a console log for duplicate item display names. - hannibal2
+ Removed the unused getInternalNameFromHypixelId. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/enoughupdates/ItemResolutionQuery.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/enoughupdates/ItemResolutionQuery.kt
@@ -15,6 +15,7 @@ import at.hannibal2.skyhanni.utils.ItemUtils.getLore
 import at.hannibal2.skyhanni.utils.ItemUtils.takeUnlessEmpty
 import at.hannibal2.skyhanni.utils.NeuInternalName
 import at.hannibal2.skyhanni.utils.NeuInternalName.Companion.toInternalName
+import at.hannibal2.skyhanni.utils.NeuItems
 import at.hannibal2.skyhanni.utils.NumberUtil.romanToDecimal
 import at.hannibal2.skyhanni.utils.RegexUtils.firstMatcher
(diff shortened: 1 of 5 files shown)


lunaynx:
Why are ambiguous display names manually declared in SkyHanni-REPO? This should be resolvable with data from the NEU repo itself. I already have working code for that part, I can PR it soon.

But the ignored internal names being in repo is definitely a good idea, and not something I added in my branch.

hannibal002:
> This should be resolvable with data from the NEU repo itself. 

how would you do that?

lunaynx:
The whole reason they're ambiguous in the first place is because the NEU repo has data on multiple items with the same display name, and some algorithm breaks the tie. You'll see in a bit, I'll submit my PR as a draft at first until I double check it.

PR #6353, opened 2026-08-10
Fix: Don't resolve ambiguous display names to an arbitrary item

lunaynx:
## What
Fixes item resolution returning an arbitrary item when given an ambiguous display name – such as the infamous "every enchanted book is Paleontologist V".

We'll have to test to make sure failing loudly won't spam users' chat, but I'd say ideally we keep loud failure and just silence known cases that don't need any more reports via SkyHanni-REPO.

## Changelog Fixes
+ Fixed SkyHanni arbitrarily picking an item when given an ambiguous name (e.g. Paleontologist V for enchanted books). - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/enoughupdates/ItemResolutionQuery.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/enoughupdates/ItemResolutionQuery.kt
@@ -42,10 +42,9 @@ import net.minecraft.world.inventory.ChestMenu
 import net.minecraft.world.item.Item
 import net.minecraft.world.item.Items
 
-// Code taken from NotEnoughUpdates
+// Code originally from NotEnoughUpdates
 class ItemResolutionQuery {
-
-    private var compound: DataComponentMap = DataComponentMap.EMPTY
+    private var compound = DataComponentMap.EMPTY
 
     private var itemType: Item? = null
     private var knownInternalName: NeuInternalName? = null
@@ -53,6 +52,7 @@ class ItemResolutionQuery {
 
     @SkyHanniModule
     companion object {
+        private const val MAX_CANDIDATE_INTERNAL_NAMES = 20
 
         private val patternGroup = RepoPattern.group("misc.itemresolution")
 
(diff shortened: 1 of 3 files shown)


hannibal002:
https://github.com/hannibal002/SkyHanni/pull/6352 and this pr are complementary rather than alternatives.

`ItemNameResolver.getInternalNameOrNull` hits `NeuItems.allItemsCache` first, and that map is built in `readAllNeuItems`, where one item silently overwrites the other. Every collision I found resolves there, so `findInternalNameByDisplayName` is never reached for them. Your change covers the opposite case, names that miss the cache and end up in the mangled name fallback. #6352 covers the cache, this covers the fallback.

On deriving the ambiguous names from the NEU repo instead of listing them: I would rather keep that list manual. Filtering automatically can hide false positives, and a new duplicate would silently become unresolvable instead of showing up as something that needs a decision in `ignored_internal_names`.

One merge note: whichever goes second, the `ignored_internal_names` check has to run before the ambiguity check in `filterInternalNameCandidates`. Otherwise an item we deliberately excluded still makes the name ambiguous.

lunaynx:
I'll have another look later.

> On deriving the ambiguous names from the NEU repo instead of listing them: I would rather keep that list manual. Filtering automatically can hide false positives, and a new duplicate would silently become unresolvable instead of showing up as something that needs a decision in `ignored_internal_names`.

Isn't this just a matter of failing loudly, like my PR does? If we rule out loud failure, there are two options:
- Silently fail to resolve the item at all
- Silently resolve it to a random, likely wrong item

I'd argue never returning an internal name for an ambiguous match unless an explicit rule already exists (such as your `ignored_items`) is better, but like I said in my PR, ideally it should fail loudly if it hits that case, and any errors that are too noisy should be temporarily silenced via SkyHanni-REPO.

hannibal002:
For the cache the difference is not loud versus silent. #6352 already prints every colliding display name on repo load, so a new duplicate shows up immediately, before anything asks for that name. Visibility is covered either way.

PR #6356, opened 2026-08-11
Fix: Clearer config description

VK3DNS:
## What
Rewrote the config title and description for garden->custom keybinds->exclude barn to reflect the fact that it also excludes greenhouse plots. 

## Changelog Fixes
+ Fixed the misleading name and description of the Custom Keybinds "Exclude Barn" option, which also excludes Greenhouse plots. - VK3DNS
    * The option is now called "Exclude Unfarmable Plots".


hannibal002:
please also add a todo above @Expose that suggests to rename the `excludeBarn` variable to follow the new option name.

VK3DNS (commented):
added a todo 'Rename excludeBarn to excludeUnfarmablePlots'

PR #6357, opened 2026-08-11
Improvement: default enabled garden features

hannibal002:
## What
I just tested out the [26.2 branch](https://github.com/hannibal002/SkyHanni/pull/5894).
I have not migrated any configs over to the new run folder.
I saw a lot of stuff in the garden that i totally dont want to see.
This reminded me of this 3yo message on discord:

[image](https://tenor.com/en-GB/view/jarvis-iron-man-jarvis-commands-commands-meme-meme-gif-25323041)

So i changed some defaults.
If you disagree with me, pls let me know!

## Changelog Improvements
+ Improved default settings for Garden features when using SkyHanni for the first time. - hannibal2


src/main/java/at/hannibal2/skyhanni/config/features/garden/pests/StereoHarmonyConfig.kt
```diff
@@ -13,7 +13,7 @@ class StereoHarmonyConfig {
     @ConfigOption(name = "Enabled", desc = "Show a display of what pest is being boosted by your vinyl.")
```
Fazfoxy, line 13: Personally I don't think this is really that big of an inconvenient feature, we could make the default appearance smaller with turning off Show Head or something instead?

Fazfoxy: Or actually the lower option
    @ConfigOption(name = "Hide when None", desc = "Hide when no vinyl selected.")
could be default on instead to make it only show up when you have an active vinyl by standard?

hannibal002: good idea

PR #6358, opened 2026-08-11
Improvement: Show keybinds in minecraft controls menu

AverageUser125:
## What
Makes all of skyhannis key bind settings show up in the `Controls` panel.
It uses the vanila `KeyMapping` and the fabric API helper `KeyMappingHelper`.
And took some proven code from Dandelion.

The option names are not really great for controls menu, as they got no context.
So "Attack" doesn't mean much.
So I tried to convert the path into some sensible display name.
Maybe getting the parent accordion name will work, but some accordions likein Custom Wardrobe Keybinds are just called: "Keybinds".

<details>
<summary>Images</summary>
image
</details>

## Changelog Improvements
+ Added skyhanni keybinds to vanila minecraft keybinds menu. - Avrg


AverageUser125:
Maybe will be part of 26.3 , but no point as a seperate PR

PR #6360, opened 2026-08-11
Backend: Delete buildpaths-excluded.txt

AverageUser125:
## What
It annoyed me enough to make a PR to remove it.

exclude_from_changelog


hannibal002:
not so sure anymore if we can include that in 9.0.0

AverageUser125:
Alright. so begone PR for now.
(This PR either way is small enough so draft is eh)

PR #6362, opened 2026-08-12
Backend: Clean up Identifier usages

lunaynx:
## What
Migrates all code that creates `Identifier` (formerly `ResourceLocation`) objects to canonical vanilla methods and `SkyHanniMod.id` where applicable. MoulConfig demands its own `MyResourceLocation`, so it gets a helper function with a weak hash map used as a cache.

+ Standardized creation of Identifier objects across the codebase. - Luna


lunaynx:

lunaynx:
Reopening this too since it's fairly narrow, not a priority but shouldn't be hard to resolve conflicts either way.

PR #6317, opened 2026-08-06
Fix + Backend: Multi RotatingPerk support for Hotx

AverageUser125:
## What
The HOTF now got a second rotating perk called "Beekeeper" at hotm 8.
I did not add the perk itself since i ain't got hotm 8 and can't test it.
This PR instead just adds support for multiple RotatingPerks.
I am unsure about the way I implemented this change, but variadic template parameters don't exist in kotlin.

Also HOTF needs a rework since they moved perks and also added a new "powder type", but I don't got enough HOTF exp to attempt that.

## Changelog Technical Details
+ Added support for multiple rotating perks. - Avrg

## Changelog Fixes
+ Fixed an error when opening the HOTM/HOTF menu with a rotating perk. - Avrg
    * When switching HOTM/HOTF layouts, the current perk line is missing from the item lore.

PR #6363, opened 2026-08-12
Improvement + Fix: HotF Beekeeper Perk

hannibal002:
## Dependencies
- #6317

## What
Hypixel added a second rotating perk to the Heart of the Forest, called Beekeeper. It unlocks at tier 8 and rotates between four buffs every SkyBlock day. The mod did not know it, so it was missing from the tree and its daily chat message ended up in an error message in chat, once per SkyBlock day.

On top of the dependency above:

- `RotatingPerkSlot` is now a typed class with named instances, which removes the two unchecked casts and the list lookup in `HotmApi.skymall` and `HotfApi.lottery`.
- The slot registers the repo patterns of its perks itself. That was missing for `LotteryPerk`.
- Item lore reading now handles effects wrapping over several lines, which three of the four Beekeeper buffs do. The single line is still tried first, because `SkymallPerk.GOBLIN_CHANCE` has an item pattern deliberately cut off after the first line of a wrapped effect.

Three of the four Beekeeper patterns are reconstructed from a tooltip dump and unconfirmed in chat. If one is wrong, the error report surfaces the real string. The Critter icon is a private use character that does not survive copying into a text report, so its position is matched permissively.

## Changelog Improvements
+ Added the Beekeeper perk to the Heart of the Forest, shown in the Rotating Perk Display next to the Lottery buff. - hannibal002
+ Renamed "Lottery Display" to "Rotating Perk Display" and "Hide Lottery Messages" to "Hide Foraging Buff Messages". Both cover Beekeeper now. - hannibal002

## Changelog Fixes
+ Fixed an error message appearing once per SkyBlock day with the Beekeeper perk unlocked. - hannibal002
+ Fixed the Heart of the Forest reading and writing the Heart of the Mountain token count. - hannibal002


src/main/java/at/hannibal2/skyhanni/features/mining/HotxFeatures.kt
```diff
@@ -71,8 +68,8 @@ object HotxFeatures {
     }
 
     @HandleEvent(GuiContainerEvent.BackgroundDrawnEvent::class, onlyOnSkyblock = true)
```
AverageUser125: Remove the event type class while you're here.

AverageUser125 (approved):
otherwise looks super good.

PR #6365, opened 2026-08-12
Feature + Fix: NPC Trade Helper

hannibal002:
## What
Items are found by their cost lore instead of by inventory name, so the feature also works in
shops we have never seen.

`SkyblockCurrency` covers currencies written as "amount name", which the existing item parsing
cannot read, and is also where the owned amount of a currency is looked up.

The fix below comes out of replacing the old `loreCosts()` helper, which dropped the amounts,
and could not be split into its own PR.

<details>
<summary>Images</summary>

image

image

</details>

## Changelog New Features
+ Added NPC Trade Helper.  - hannibal2
    * Highlights items in NPC trade menus that you can buy right now.
    * Also shows the price, the number you own and the total cost in the item lore.

## Changelog Fixes
+ Fixed additional cost in the Chocolate Shop and SkyMart overlays ignoring the number of items. - hannibal2

+ Merged the duplicated cost lore parsers of six features into LoreCostUtils. - hannibal2
+ Added SkyblockCurrency for currencies that appear in cost lores. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/LoreCostUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/LoreCostUtils.kt
@@ -3,10 +3,10 @@ package at.hannibal2.skyhanni.utils
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.test.command.ErrorManager
 import at.hannibal2.skyhanni.utils.ItemUtils.getLoreComponent
-import at.hannibal2.skyhanni.utils.NumberUtil.formatLong
 import at.hannibal2.skyhanni.utils.RegexUtils.groupOrNull
 import at.hannibal2.skyhanni.utils.RegexUtils.matchMatcher
 import at.hannibal2.skyhanni.utils.RegexUtils.matches
+import at.hannibal2.skyhanni.utils.compat.formattedTextCompatLeadingWhiteLessResets
 import at.hannibal2.skyhanni.utils.compat.formattedTextCompatLessResets
 import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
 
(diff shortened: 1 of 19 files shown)


src/main/java/at/hannibal2/skyhanni/utils/SkyblockCurrency.kt
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        /**
+         * When the display name resolves to a repo item, [ItemNameResolver] answers before this
+         * enum is ever reached, so the entry has to use that item's id. An invented id fails
+         * silently instead: every comparison against it never matches.
+         */
+        @HandleEvent
+        private fun onNeuRepoReload() {
+            // the item name lookup needs NeuItems.allItemsCache, which is filled by another handler of this event
```
AverageUser125, line 160: why not just increase the priority of the event that populates `NeuItems.allItemsCache`?
Or decrease the priority of this event.

src/main/java/at/hannibal2/skyhanni/utils/SkyblockCurrency.kt
```diff
@@ -0,0 +1,131 @@
...
+         * REGEX-TEST: §240 Pests
+         * REGEX-TEST: §c250 Copper
+         * REGEX-TEST: §61,940,000 Coins
+         * REGEX-TEST: 46,559,892,200 Chocolate
+         * REGEX-TEST: §a400 Gems
+         * REGEX-TEST: §575 Pelts
+         */
+        private val amountPattern by RepoPattern.pattern(
+            "utils.currency.amount",
+            "(?:§.)*(?<amount>[\\d,]+) (?<name>[\\w' ]+)",
```
AverageUser125: why not just remove the color codes?

src/main/java/at/hannibal2/skyhanni/utils/SkyblockCurrency.kt
```diff
@@ -0,0 +1,131 @@
...
+ * regular item path and are listed here only so that features can refer to them.
+ *
+ * [internalName] has to be the id of the real repo item whenever the name resolves to one, even
+ * if that item has no price, as is the case for [NeuInternalName.SKYBLOCK_COPPER].
+ * [coinValue] is the worth of a single unit in coins, or null when unknown.
+ * [loreNames] are the names as written in the lore, lowercase and without color codes.
+ */
+enum class SkyblockCurrency(
+    val internalName: NeuInternalName,
+    val displayName: String,
```
AverageUser125, line 33: Alecio said something similiar but can you separate the color from the display name?
And have a seperate `getName`/`getLegacyColouredName`.
(Maybe returns a `Component`?)

src/main/java/at/hannibal2/skyhanni/utils/LoreCostUtils.kt
```diff
@@ -0,0 +1,82 @@
...
+    private val patternGroup = RepoPattern.group("utils.lore.cost")
+
+    /**
+     * REGEX-TEST: §7Cost
+     * REGEX-TEST: §5§o§7Cost
+     * REGEX-TEST: §7Cost: §b5,000 Bits
+     */
+    private val costHeaderPattern by patternGroup.pattern(
+        "header",
+        "(?:§5§o)?§7Cost(?:: (?<inline>.+))?",
```
AverageUser125: color codes 🥺 
Also why call the group "inline"?
(idk a better name but still)

AverageUser125 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/SkyblockCurrency.kt
lunaynx: You cannot really rely on repo reload events respecting priority until #6064 is merged. A synchronous event handler with `HIGHEST` priority *might* work for the time being, but all of the asynchronous ones are currently fire-and-forget, so priority only controls in what order they *start* executing – they don't wait for each other to finish.

AverageUser125: so, this PR should have #6064 as a dependecy?
And also hannibal should merge that :pleading_face:

AverageUser125: How does your PR fix this?
Does it do all event listeners of a certain priority in a batch.
And then waits till they are all done before moving to the next priority?

lunaynx: With my PR, all repo reload event handlers are individually `await`ed and executed sequentially, just in a coroutine instead of blocking the render thread. There are currently 112 uses of `event.getConstant` and 12 uses of `event.getConstantAsync` in `beta`, so this will actually largely preserve the current synchronous behavior, just delegating it to a shared coroutine to avoid freezes during repo reload and make it less annoying to use for contributors (no need for each feature class to declare and launch its own coroutine).

If it turns out it would be a noticeable improvement, we can look into executing all repo reload event handlers of the same priority execute in parallel in the future, but the current implementation in my PR is probably generally safer.

hannibal002: Priority cannot fix this one. NeuItems does not fill allItemsCache in its own handler, it
schedules readAllNeuItems through DelayedRun.runOrNextTick (NeuItems.kt:107). So no matter
where my handler sits in the dispatch order, the cache is not rebuilt yet when the reload
comes off the main thread. Running a tick later is what actually waits for it.

And yes, I will give the one pr mentioned higher priority.

PR #6366, opened 2026-08-12
Fix: Skymart shop page detection

Chicosa:
Improved store name detection to inlude store numbers so feature copper to coins would work again on farming essentials pages.

image


lunaynx:
Please change it to a `RepoPattern` while you're at it (and possibly use `InventoryDetector` as well but that's less important).

hannibal002 (changes requested):
What luna said. also, please follow the [pull request template](https://github.com/hannibal002/SkyHanni/blob/beta/pull_request_template.md).

PR #6368, opened 2026-08-12
Backend: Create SkyblockIcon & SkyblockMobType - “Good idea” said Hannibal

AverageUser125:
## What
Create a centeral class for custom hypixel icons. called `SkyblockIcon`
Also creates `SkyblockMobType` which uses it.

Made all usages of `SkyblockStat.STAT_NAME.hypixelIcon` now use `SkyblockIcon` directly.

+ Created SkyblockIcon enum to centralize hypixel symbols . - Avrg


lunaynx:
I think it would make more sense to use stuff like `${SkyblockIcon.FARMING_FORTUNE}` in regexes and such, as the string representation of that is unambiguous, while making the `toString` of `SkyblockStat` just the icon is ambiguous – what if you wanted the stat name?

AverageUser125:
`SkyblockStat.displayName` still exists.
Also since all the wisdom stats use the same symbol, it's better to write `SkyblockStat.FARMING_WISDOM` then just `SkyblockIcon.WISDOM`.

Though I agree this is not an ideal solution.

lunaynx:
> Also since all the wisdom stats use the same symbol, it's better to write `SkyblockStat.FARMING_WISDOM` then just `SkyblockIcon.WISDOM`.

Fair, though you could have aliases like `SkyblockIcon.FARMING_WISDOM = SkyblockIcon.WISDOM`. Or just change `hypixelIcon` to `icon` so it's shorter to write out (`SkyblockStat.FARMING_WISDOM.icon`).

AverageUser125:
Actually, if I just make `SkyblockIcon` just have `FARMING_WISDOM` etc just be aliases. it's probably fine.

AverageUser125:
oh, you already thought of that too lol.

AverageUser125:
Imma just do aliases and make it use `SkyblockIcon` everywhere.

AverageUser125:
> Or just change `hypixelIcon` to `icon` so it's shorter to write out (`SkyblockStat.FARMING_WISDOM.icon`).

`SkyblockStat` already defines `icon` so can't do that.

lunaynx:
Could just rename that to `iconWithColor` or something, but the alias method will also work fine.

AverageUser125:
Draft till it is confirmed the old pest icon is gone.

PR #6369, opened 2026-08-12
Fix: AtmosphericFilterDisplay still using old farming fortune icon

AverageUser125:
## What
Noticed this in #6368 so here.
It still used the old icon.

## Changelog Fixes
+ Fixed Atmospheric Filter Display not detecting the Farming Fortune buff. - Avrg


hannibal002 (changes requested):
you accidentally just use the enum name. we want to use the icon of the enum.

AverageUser125:
woops, I forgot that my other PR changed that.

hannibal002 (approved):

PR #6372, opened 2026-08-13
Fix: Don't fetch weight profile if leaderboard features are disabled

lunaynx:
## What
A user had noticed that SkyHanni was still fetching their profile information from Elite whenever they join the Garden with no way to opt out. This PR disables the automatic weight profile fetch if "Enable leaderboard features" is disabled, and makes it fetch upon enabling it if it wasn't recently fetched.

## Changelog Fixes
+ Fixed some Elite leaderboard requests still being sent with leaderboards disabled. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
@@ -1,6 +1,7 @@
 package at.hannibal2.skyhanni.data.garden
 
 import at.hannibal2.skyhanni.SkyHanniMod
+import at.hannibal2.skyhanni.SkyHanniMod.launch
 import at.hannibal2.skyhanni.api.EliteDevApi
 import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.config.ConfigManager
@@ -58,6 +59,8 @@ object FarmingWeightData {
     private var hasFetchedCropWeights = false
     private var shouldRecalculateWeight = false
 
+    private val cropWeightsCoroutine = CoroutineSettings("get crop weights")
+
     @HandleEvent(onlyOnIsland = GARDEN)
     private fun onConfigLoad() {
         config.enabled.onEnable {
@@ -95,7 +98,7 @@ object FarmingWeightData {
         if (!isEnabled()) return
         if (!event.isMod(5)) return
 
-        SkyHanniMod.launchIOCoroutine("get crop weights") {
+        cropWeightsCoroutine.launch {
             getCropWeights()
         }
     }
--- a/src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
@@ -65,6 +65,8 @@ object FarmingWeightData {
     @HandleEvent
     private fun onConfigLoad() {
         config.enabled.onEnable {
+            // This is intentionally checked inside onEnable, using onlyOnIsland would be wrong here.
+            @Suppress("IsInIslandEarlyReturn")
             if (!IslandType.GARDEN.isInIsland()) return@onEnable
             ChatUtils.debug("Updating EliteSkyBlock collections because leaderboard features were toggled on")
             updateCollections()
--- a/src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
@@ -8,6 +8,7 @@ import at.hannibal2.skyhanni.config.ConfigManager
 import at.hannibal2.skyhanni.config.commands.CommandRegistrationEvent
 import at.hannibal2.skyhanni.config.commands.brigadier.BrigadierArguments
 import at.hannibal2.skyhanni.data.HypixelData
+import at.hannibal2.skyhanni.data.IslandType
 import at.hannibal2.skyhanni.data.garden.CropCollectionApi.getCollection
 import at.hannibal2.skyhanni.data.garden.CropCollectionApi.lastGainedCrop
 import at.hannibal2.skyhanni.data.garden.CropCollectionApi.setCollectionCounter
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
```diff
@@ -60,20 +60,25 @@ object FarmingWeightData {
     private var shouldRecalculateWeight = false
 
     @HandleEvent
-    fun onIslandChange(event: IslandChangeEvent) {
-        if (event.newIsland != IslandType.GARDEN) return
-        updateCollections()
+    private fun onConfigLoad() {
```
hannibal002, line 63: this should only run in garden. maybe set a "dirty" flag that gets read and acted upon when joining garden?

src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
```diff
@@ -60,20 +60,25 @@ object FarmingWeightData {
...
+        ConditionalUtils.onToggle(config.enabled) {
```
hannibal002, line 64: use ConditionalUtils.onEnable

src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
```diff
@@ -83,7 +88,7 @@ object FarmingWeightData {
     }
 
     @HandleEvent(onlyOnSkyblock = true)
-    fun onTick(event: SkyHanniTickEvent) {
+    private fun onTick(event: SkyHanniTickEvent) {
         if (!event.isMod(5)) return
 
         SkyHanniMod.launchIOCoroutine("get crop weights") {
```
hannibal002, line 94: has no "is enabled" check. also needs a "in garden" check

src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
```diff
@@ -99,10 +104,10 @@ object FarmingWeightData {
     fun getWeight(leaderboardMode: EliteLeaderboardMode, override: Boolean = false, cropWeightOnly: Boolean = false): Double? {
         if (weightMap[leaderboardMode] == null || override) {
             when (leaderboardMode) {
-                EliteLeaderboardMode.ALL_TIME -> {}
+                ALL_TIME -> {}
```
hannibal002, line 107: either use EliteLeaderboardMode.ALL_TIME, or also drop the `EliteLeaderboardMode.` couple lines below in that function

src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
```diff
@@ -60,20 +60,25 @@ object FarmingWeightData {
...
+            if (!isEnabled()) return@onToggle
+            ChatUtils.debug("Updating EliteSkyBlock collections because leaderboard features were toggled on")
+            updateCollections()
+        }
     }
 
-    @HandleEvent(onlyOnIsland = IslandType.GARDEN)
-    fun onProfileJoin() {
+    @HandleEvent(onlyOnIsland = GARDEN)
+    private fun onIslandJoin(event: IslandJoinEvent) {
```
hannibal002, line 72: drop unused event parameter

src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
```diff
@@ -60,20 +60,25 @@ object FarmingWeightData {
...
     @HandleEvent
-    fun onIslandChange(event: IslandChangeEvent) {
-        if (event.newIsland != IslandType.GARDEN) return
-        updateCollections()
+    private fun onConfigLoad() {
+        ConditionalUtils.onToggle(config.enabled) {
+            if (!isEnabled()) return@onToggle
+            ChatUtils.debug("Updating EliteSkyBlock collections because leaderboard features were toggled on")
+            updateCollections()
+        }
```
hannibal002, line 68: lastFetchAttempt needs to get reset on profile switch. otherwise profile switches breaks the display. also on profile switch the data needs to get wiped entirely

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
lunaynx: Sure, shouldn't need a dirty flag because we already update collections on Garden join regardless.

src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
```diff
@@ -59,21 +59,32 @@ object FarmingWeightData {
     private var hasFetchedCropWeights = false
     private var shouldRecalculateWeight = false
 
+    private val cropWeightsCoroutine = CoroutineSettings("get crop weights")
+
+    @HandleEvent(onlyOnIsland = GARDEN)
```
hannibal002: the island check should not happen when the config gets loaded (before joining sb), but rather inside the oEnable block

src/main/java/at/hannibal2/skyhanni/data/garden/FarmingWeightData.kt
```diff
@@ -82,11 +93,12 @@ object FarmingWeightData {
...
-    @HandleEvent(onlyOnSkyblock = true)
-    fun onTick(event: SkyHanniTickEvent) {
+    @HandleEvent(onlyOnIsland = GARDEN)
+    private fun onTick(event: SkyHanniTickEvent) {
+        if (!isEnabled()) return
         if (!event.isMod(5)) return
 
-        SkyHanniMod.launchIOCoroutine("get crop weights") {
+        cropWeightsCoroutine.launch {
             getCropWeights()
```
hannibal002, line 107: please move the early returns inside this function in front of the launch call. so we dont unnecessary create new coroutines and exit them again.

hannibal002 (changes requested):

hannibal002:
The detect rule is wrong. The proper fix is to only check the first line / not inside a coroutine. Then the suppress annotation here can go.

But that's outside of the scope of this pr

hannibal002 (approved):

PR #6377, opened 2026-08-14
Fix: Error when opening shop profit overlays after game start

hannibal002:
## What

`InventoryFullyOpenedEvent` runs on a Netty thread, so building renderables in the handler errors in `Font.width` while the glyphs are not baked yet. Only reproduces on the first inventory after a game start.

Display building now runs in `DelayedRun.runOrNextTick`, like `SkyMartCopperPrice`.

`StarlynSisterCouponProfit` also swaps its `StarlynSisterType` loop for a local lookup with an early return, so a non-matching inventory no longer rebuilds the display from an earlier shop variant.

## Changelog Fixes
+ Fixed an error when opening a shop profit overlay as the first inventory after starting the game. - hannibal2
    * Affects Anita Medal Profit, Pesthunter Profit, Starlyn Sister Coupon Profit and Fishy Treat Profit.


lunaynx:
Is this an actual crash or again just an error caught by ErrorManager?

hannibal002:
red error in chat

PR #6379, opened 2026-08-15
Fix: Don't drain the stored shulker route

lunaynx:
## What
`route` was aliased to `storedRoute` once a lap finished, so the next lap consumed the stored copy instead of a fresh one. The route is now a single list that refills from `storedRoute` when exhausted.

## Changelog Fixes
+ Fixed Hideonleaf/Hideonsun Pathfinder breaking after completing two laps. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/hunting/ShulkerFinder.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/hunting/ShulkerFinder.kt
@@ -12,6 +12,7 @@ import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.EntityUtils.getEntitiesNearby
 import at.hannibal2.skyhanni.utils.LorenzColor
 import at.hannibal2.skyhanni.utils.LorenzVec
+import at.hannibal2.skyhanni.utils.collection.CircularList
 import at.hannibal2.skyhanni.utils.compat.MinecraftCompat
 import at.hannibal2.skyhanni.utils.navigation.NavigationUtils
 import net.minecraft.world.entity.monster.Shulker
@@ -20,8 +21,7 @@ import net.minecraft.world.entity.monster.Shulker
 object ShulkerFinder {
     private val config get() = SkyHanniMod.feature.hunting
 
-    private val route: MutableList<LorenzVec> = mutableListOf()
-    private val storedRoute: MutableList<LorenzVec> = mutableListOf()
+    private var storedRoute: CircularList<LorenzVec>? = null
 
     private var navigating = false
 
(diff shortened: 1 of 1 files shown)


hannibal002:
~~the fix describes a behaviour we dont intended. the original featue was designed to only run once and stop once you found all, not to run endlessly.~~

just saw that this is indeed intended, its just not very clear on first read. the cleanup suggestions stand though

src/main/java/at/hannibal2/skyhanni/features/hunting/ShulkerFinder.kt
```diff
@@ -21,7 +20,7 @@ import net.minecraft.world.entity.monster.Shulker
 object ShulkerFinder {
     private val config get() = SkyHanniMod.feature.hunting
 
-    private var route: MutableList<LorenzVec>? = null
+    private val route: MutableList<LorenzVec> = mutableListOf()
     private val storedRoute: MutableList<LorenzVec> = mutableListOf()
```
hannibal002, line 24: why do we have two identical lists now? either we WANT to drop it to zero, or we just keep am index and dont touch the list at all.

src/main/java/at/hannibal2/skyhanni/features/hunting/ShulkerFinder.kt
```diff
@@ -21,7 +20,7 @@ import net.minecraft.world.entity.monster.Shulker
 object ShulkerFinder {
     private val config get() = SkyHanniMod.feature.hunting
 
-    private var route: MutableList<LorenzVec>? = null
+    private val route: MutableList<LorenzVec> = mutableListOf()
```
hannibal002, line 23: unvalid/unclear name. represents the "remaining route in front of you. the other member then could be named "fullRoute" or similar.

src/main/java/at/hannibal2/skyhanni/features/hunting/ShulkerFinder.kt
```diff
@@ -72,7 +71,7 @@ object ShulkerFinder {
     @HandleEvent
     private fun onIslandChange() {
```
hannibal002, line 72: please also sweep change by no longer using the deprecated event. use world change instead

hannibal002 (changes requested):

lunaynx:
I addressed the review comments, but for future reference, I don't think it's very productive to hold up a bugfix PR with "hey, I don't like how the original code was written, so I'm making it your problem to refactor it now".

hannibal002:
Valid point. Bug fix PRs should be able to just add Todos to needed refactor parts, and only do the bare minimum of easily sweeping changes. Not the full "gold standard" of code quality as all other pr types require. Will try to remember for the next reviews

hannibal002 (approved):

PR #6380, opened 2026-08-15
Backend: No double mutability for collections

lunaynx:
## Dependencies
- #6379
- #6381

## What
This Detekt rule was previously disabled due to the sheer number of existing flags, but it's not just a pedantic rule. Declaring something as both a `var` and a `Mutable*` is a footgun and makes it easy to accidentally mess things up. This PR enables the rule and fixes all existing violations. #6379 and #6381 were a byproduct of this cleanup.

+ Disallowed double mutability for collections via a Detekt rule and fixed all existing violations. - Luna


lunaynx:

PR #6381, opened 2026-08-15
Fix: Drop unsafe lateinit usage in LimboPlaytime

lunaynx:
## What
People keep reporting a "lateinit property has not been initialized" error with the Limbo Playtime feature. This PR drops `lateinit` entirely, which should stop those errors.

## Changelog Fixes
+ Fixed errors in chat about the Limbo Playtime feature when using /playtimedetailed. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
@@ -5,7 +5,6 @@ import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.data.ProfileStorageData
 import at.hannibal2.skyhanni.events.InventoryOpenEvent
 import at.hannibal2.skyhanni.events.minecraft.ToolTipTextEvent
-import at.hannibal2.skyhanni.events.minecraft.add
 import at.hannibal2.skyhanni.events.render.gui.ReplaceItemEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.InventoryUtils
@@ -15,9 +14,11 @@ import at.hannibal2.skyhanni.utils.NeuItems.getItemStack
 import at.hannibal2.skyhanni.utils.NumberUtil.addSeparators
 import at.hannibal2.skyhanni.utils.NumberUtil.formatDoubleOrNull
 import at.hannibal2.skyhanni.utils.NumberUtil.roundTo
+import at.hannibal2.skyhanni.utils.RegexUtils.findMatcher
 import at.hannibal2.skyhanni.utils.RegexUtils.matches
 import at.hannibal2.skyhanni.utils.SafeItemStack
 import at.hannibal2.skyhanni.utils.SimpleTimeMark
+import at.hannibal2.skyhanni.utils.chat.TextHelper.asComponent
 import at.hannibal2.skyhanni.utils.itemType
 import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
 import net.minecraft.network.chat.Component
@@ -26,7 +27,7 @@ import kotlin.time.Duration.Companion.seconds
 
 @SkyHanniModule
 object LimboPlaytime {
-    private lateinit var modifiedList: MutableList<Component>
+    private var modifiedList = listOf<Component>()
     private var setMinutes = false
     private val patternGroup = RepoPattern.group("misc.limbo.tooltip")
 
@@ -49,8 +50,6 @@ object LimboPlaytime {
         "(?<hours>[\\d.,]+) hours.*$",
     )
 
-    private var tooltipPlaytime = mutableListOf<Component>()
-
     private var wholeMinutes = 0
     private var hoursString: String = ""
 
@@ -59,11 +58,11 @@ object LimboPlaytime {
 
     private val itemID = "ENDER_PEARL".toInternalName()
     private const val ITEM_NAME = "§aLimbo"
-    private lateinit var limboItem: SafeItemStack
+    private var limboItem: SafeItemStack? = null
     private var lastCreateCooldown = SimpleTimeMark.farPast()
 
     @HandleEvent
-    fun replaceItem(event: ReplaceItemEvent) {
+    private fun replaceItem(event: ReplaceItemEvent) {
         if (!enabled) return
         if (event.inventory !is SimpleContainer) return
         // TODO replace with InventoryDetector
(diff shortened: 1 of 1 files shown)


hannibal002 (changes requested):
`modifiedList` is always inizialised, thats quite clear when i read the code. that can not be the issue.

but look at the second member with the lateinit modifier: `limboItem`.
if `createItemStack` throws an error, we still do set `lastCreateCooldown`, so the next `ReplaceItemEvent` call will try to access the not yet inizlialized `limboItem`. if you want to fix the issue, please look into fixing that.

lunaynx:
You're right, this was a late night rushed out PR, I noticed the lateinit change in #6380 and went to separate the fix out without double checking. I should have waited or made it a draft. Will look into it later.

lunaynx:
Okay, should be properly fixed now. This code probably needs to get another cleanup pass at some point, but I didn't go too far for now.

src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
```diff
@@ -15,9 +15,11 @@ import at.hannibal2.skyhanni.utils.NeuItems.getItemStack
 import at.hannibal2.skyhanni.utils.NumberUtil.addSeparators
 import at.hannibal2.skyhanni.utils.NumberUtil.formatDoubleOrNull
 import at.hannibal2.skyhanni.utils.NumberUtil.roundTo
```
hannibal002, line 16: the import `import at.hannibal2.skyhanni.events.minecraft.add` is unused

src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
```diff
@@ -130,28 +133,28 @@ object LimboPlaytime {
...
-                val matcher = hoursPattern.matcher(it.string)
-                if (matcher.find()) {
-                    matcher.group("hours").formatDoubleOrNull() ?: 0.0
-                } else 0.0
-            }.toMutableList()
+            modifiedList = buildList {
+                addAll(hoursList)
+                add("§5§o§b$hoursString hours §7on Limbo".asComponent())
+            }.sortedByDescending {
+                hoursPattern.matchMatcher(it.string) {
```
hannibal002: `matchMatcher` is a full match (`Matcher.matches()`), the old code used `find()`. The line added two lines above starts with `§5§o§b` and `asComponent()` keeps those, so it never matches and falls back to `0.0`, sorting it to the bottom. `findMatcher` keeps the old behavior. Same in the minutes branch below.

src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
```diff
@@ -72,15 +74,16 @@ object LimboPlaytime {
         val playtime = storage?.playtime ?: 0
         if (playtime < 60) return
 
-        if (lastCreateCooldown.passedSince() > 3.seconds) {
+        val item = limboItem?.takeIf { lastCreateCooldown.passedSince() > 3.seconds } ?: run {
```
hannibal002: Condition is inverted: this keeps the cache when more than 3 seconds passed and rebuilds otherwise. The cached branch also never refreshes the timestamp, so after the first gap over 3 seconds the item is never rebuilt again and the lore stays frozen. Should be `< 3.seconds`.

src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
```diff
@@ -49,7 +51,7 @@ object LimboPlaytime {
         "(?<hours>[\\d.,]+) hours.*$",
     )
 
-    private var tooltipPlaytime = mutableListOf<Component>()
+    private var tooltipPlaytime = listOf<Component>()
```
hannibal002: Write-only, never read anywhere. Was removed in the first version of this PR, please remove it again.

src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
```diff
@@ -130,28 +133,28 @@ object LimboPlaytime {
...
-                if (matcher.find()) {
-                    matcher.group("minutes").toDoubleOrNull() ?: 0.0
-                } else 0.0
-            }.toMutableList()
+            modifiedList = buildList {
+                addAll(minutesList)
+                add("§5§o§a$minutes minutes §7on Limbo".asComponent())
+            }.sortedByDescending {
+                minutesPattern.matchMatcher(it.string) {
+                    group("minutes").toDoubleOrNull() ?: 0.0
```
hannibal002: `toDoubleOrNull()` here, `formatDoubleOrNull()` in the hours branch. Not aware of a case where this differs, just inconsistent.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
```diff
@@ -130,28 +129,28 @@ object LimboPlaytime {
...
-                if (matcher.find()) {
-                    matcher.group("hours").formatDoubleOrNull() ?: 0.0
-                } else 0.0
-            }.toMutableList()
+            modifiedList = buildList {
+                addAll(hoursList)
+                add("§5§o§b$hoursString hours §7on Limbo".asComponent())
+            }.sortedByDescending {
+                hoursPattern.findMatcher(it.string) {
+                    group("hours").toDoubleOrNull() ?: 0.0
```
hannibal002: This went the wrong way: the hours branch used `formatDoubleOrNull()` before. `hoursString` is built with `addSeparators()` in `onInventoryOpen`, so above 1000 hours the injected line reads `1,234.5 hours`, which `toDoubleOrNull()` cannot parse. Please use `formatDoubleOrNull()` in both branches instead.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
lunaynx: You're right, I was looking at `formatDoubleOrNull` but confused myself with the name of the function (I'd expect a function named "format" to return a string but it returns a double, so I convinced myself that thousands separators don't matter because it returns a double, but it matters in the input...)

Honestly, it makes perfect sense why the discrepancy existed in the first place now, `minutes` can never have thousands separators so it was fine using `toDoubleOrNull`, saves a couple unnecessary checks. But it's not significant enough either way and I don't want to hold up a bugfix PR even longer with your nitpicking about code that was already there and I only incidentally touched to simplify the matcher pattern, so I just changed both to `formatDoubleOrNull` as you said.

hannibal002 (approved):

PR #6382, opened 2026-08-15
Fix: Use monotonic clock for SecondPassedEvent

lunaynx:
## What
Currently, `SecondPassedEvent` is driven by a `fixedRateTimer`. This is a non-monotonic time source and does not handle large time changes well, e.g.:
- If the clock is set back by an hour while the game is running, no `SecondPassedEvent`s will be posted for an hour.
- If the clock is set forward by an hour while the game is running, 3,600 `SecondPassedEvent`s will be posted in quick succession.

This PR changes the event to be driven by a monotonic clock to avoid these issues.

## Changelog Fixes
+ Fixed mod features breaking if you change your clock by several minutes or hours while the game is running. - Luna


lunaynx:
This is a relatively rare edge case, will probably reopen as a more comprehensive and tested PR later including the same logic for SimpleTimeMark where possible.

lunaynx:
I'm not sure why I closed this honestly, I guess I was just in a bad mood. Reopening since someone actually had the issue recently, though it still needs verification before I undraft it.

`SimpleTimeMark` may end up being a separate PR, since some consumers *need* wall clock time.

PR #6383, opened 2026-08-15
Backend: No unnecessary fully qualified names

lunaynx:
## Dependencies
- #6289

## What
This PR enables the Detekt rule to forbid unnecessary use of fully qualified names and fixes all existing violations (including the one I found via grepping in a Java file, since we don't have Java linting yet).

+ Banned unnecessary use of fully qualified names in favor of importing them. - Luna


lunaynx:

lunaynx:
I'll reopen because this is actually quite useful with LLMs loving to do this randomly, and it's simple enough.

PR #6359, opened 2026-08-11
Backend: Remove dead/duplicated render code

lunaynx:
## What
Removes some dead rendering-related code that is not used anywhere, and deduplicates some that is used but unnecessarily duplicated.

## Changelog Technical Details
+ Removed some unused and duplicated rendering-related code. - Luna

PR #6384, opened 2026-08-15
Backend: Enable lateinit detekt rule

AverageUser125:
## Dependencies
- #6359

## What
Enables the "LateinitUsage" rule, which disallows `lateinit` usage, since it is bad practice most of the time.
I have fixed some places that used it, I have not baselined this rule and preferred explicitly suppressing it.

+ Enable LateinitUsage detekt rule to discourage usage of lateinit. - Avrg


AverageUser125:
What Luna said

PR #6385, opened 2026-08-15
Fix + Improvement: Clean Tree View hiding rune

AverageUser125:
## What
The new hypixel rune also has this display blocks.
So I just checked the block display scale, It does work for the tree stuff I tested.

HAS BEEN CONFIRMED WORKING BY TESTER.

I also moved some stuff to a seperate `DisplayEntitiyUtils` class.

## Changelog Improvements
+ Added a separate Clean Tree View toggle to hide the Barkshatter Rune. - Avrg
    
## Changelog Fixes
+ Fixed Clean Tree View incorrectly hiding the Barkshatter Rune. - Avrg


Rainymei:
i want it to hide the rune
make it a 2nd config option to only hide base cutting animation

AverageUser125:
done, I added a debug config for it.

lunaynx:
Why debug? It should be a normal user-facing option.

AverageUser125:
... I sometimes cannot tell when you're trolling me :sob:

lunaynx:
I'm serious.

lunaynx:
Obviously reword it so that it's understandable to an average user, like "Don't hide rune effects".

AverageUser125:
I am not gonna repurpose a bug into a feature :sob: 
And I am not gonna add a random "Hide this specific rune" config in the foraging tab.

Rainymei:
????

AverageUser125:
okay, me confused. can we talk on discord on what you mean?

src/main/java/at/hannibal2/skyhanni/config/features/foraging/CleanTreeViewConfig.kt
```diff
@@ -0,0 +1,23 @@
...
+    var enabled = true
+
+    @Expose
+    @ConfigOption(
+        name = "Hide Rune Effects",
+        desc = "Also hide floating blocks from the Barkshatter Rune's effect."
+    )
+    @ConfigEditorBoolean
+    @FeatureToggle
+    var hideRuneEffects = false
```
hannibal002, line 25: this feels like a sub toggle when following the phrasing. either make it actually a standalone feature, independently of the main one, or remove feature toggle and make it dependent in code.

src/main/java/at/hannibal2/skyhanni/features/foraging/ClearTreeLogs.kt
```diff
@@ -23,9 +28,39 @@ object ClearTreeLogs {
     }
 
     @HandleEvent(onlyOnIslandTypeTag = [IslandTypeTag.FORAGING_CUSTOM_TREES])
-    fun onRender(event: CheckRenderEntityEvent<Display.BlockDisplay>) {
-        if (!SkyHanniMod.feature.foraging.trees.cleanView) return
-        val block = event.entity.blockState
-        if (block in treeBlocks) event.cancel()
+    private fun onRender(event: CheckRenderEntityEvent<Display.BlockDisplay>) {
```
hannibal002, line 34: feature listens in on two toggles, not just one. thats wrong

src/main/java/at/hannibal2/skyhanni/features/foraging/ClearTreeLogs.kt
```diff
@@ -23,9 +28,39 @@ object ClearTreeLogs {
...
+        if (!shouldHide(entity)) return
+
+        if (entity.blockState in treeBlocks) {
+            event.cancel()
+        }
+    }
+
+    @HandleEvent
+    private fun onConfigFix(event: ConfigUpdaterMigrator.ConfigFixEvent) {
+        event.transform(144, "foraging.trees.cleanTreeView")
```
hannibal002: what is the transform doing here?

src/main/java/at/hannibal2/skyhanni/utils/DisplayEntityUtils.kt
```diff
@@ -0,0 +1,34 @@
...
+            val scale = scaleTransform.x()
+            return if (scale == scaleTransform.y() && scale == scaleTransform.z()) scale else null
+        }
+
+    val Transformation.isRotated: Boolean
+        get() {
+            return !leftRotation().isIdentity || !rightRotation().isIdentity
+        }
+
+    @Suppress("UnnecessarySafeCall")
```
hannibal002: what do we suppress?

src/main/java/at/hannibal2/skyhanni/features/foraging/ClearTreeLogs.kt
```diff
@@ -23,9 +28,39 @@ object ClearTreeLogs {
...
+            config.enabled && !config.hideRuneEffects -> floatingTree
+            !config.enabled && config.hideRuneEffects -> !floatingTree
+            else -> false
+        }
+    }
+
+    private fun isFloatingTreeBlock(transformation: Transformation): Boolean {
+        if (transformation.isRotated) return false
+        val scale = transformation.uniformScale ?: return false
+        return scale in 0.4f..1.0f
```
hannibal002: give the range a name. no magic numbers in functions

src/main/java/at/hannibal2/skyhanni/features/foraging/ClearTreeLogs.kt
```diff
@@ -23,9 +28,39 @@ object ClearTreeLogs {
...
+        event.transform(144, "foraging.trees.cleanTreeView")
+        event.move(144, "foraging.trees.cleanView", "foraging.trees.cleanTreeView.enabled")
     }
+
+    private fun shouldHide(entity: Display.BlockDisplay): Boolean {
+        if (config.enabled && config.hideRuneEffects) return true
+
+        val transformation = entity.transformation ?: return false
+        val floatingTree = isFloatingTreeBlock(transformation)
+        return when {
```
hannibal002: the when can be cleanuped to one line

src/main/java/at/hannibal2/skyhanni/config/features/foraging/TreesConfig.kt
```diff
@@ -10,11 +10,10 @@ import io.github.notenoughupdates.moulconfig.annotations.SearchTag
...
 
     @Expose
-    @ConfigOption(name = "Clean Tree View", desc = "Hides the floating blocks when mining trees in Galatea.")
-    @ConfigEditorBoolean
-    @FeatureToggle
+    @ConfigOption(name = "Clean Tree View", desc = "")
+    @Accordion
     @SearchTag("fig mangrove helix")
-    var cleanView = true
+    val cleanTreeView = CleanTreeViewConfig()
```
hannibal002: rendundant name "tree" in the variable name

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/DisplayEntityUtils.kt
lunaynx: This looks like it was just moved as-is from another file, but we probably shouldn't be suppressing it unless it's actually declared as non-nullable *and* we've seen errors without the safe call.

AverageUser125:
Major problem: for the first tick it exists. the `transform` is null still.
And since `CheckRenderEntityEvent` caches the value it makes it still display the logs for some time.

Idk, how to fix that.
So just. give me ideas. and will fix tommorow probably.

hannibal002 (changes requested):
Skipping the cache entirely is too broad. It makes the check run every frame, and some block displays never receive a block state at all, so for those the cache is off for their whole lifetime while the answer can never change. It also affects every listener of `CheckRenderEntityEvent`, not just this feature.

Two alternatives, one simple but not perfect, one harder but perfect:

1. Simple: ask for a shorter cache duration for this entity instead of none, for example one tick. The next tick corrects the result, so the block is visible for 50ms at most instead of up to 200ms. Still a flicker, just a short one.

2. Perfect: invalidate instead of shortening. When the packet that changes the entity data arrives, drop the cache entry for that entity, so the next frame recomputes once. No flicker at all, and it covers the TODO about making this work for all display entities. Note that `PacketReceivedEvent` fires on the netty network thread, so the invalidation has to be handed to the main thread.

Im fine with either one of them.

src/main/java/at/hannibal2/skyhanni/config/features/foraging/CleanTreeViewConfig.kt
```diff
@@ -0,0 +1,32 @@
...
+        desc = "Hides floating blocks when breaking tree blocks."
+    )
+    @ConfigEditorBoolean
+    @FeatureToggle
+    var hideTreeBlocks = true
+
+    @Expose
+    @ConfigOption(
+        name = "Hide Rune Effects",
+        desc = "Hides floating blocks from the Barkshatter Rune's effect."
```
Fazfoxy: Since this toggle isn't for *specifically* barkshatter (the checks would be too broad for us to pre-emptively believe it won't block future runes) should this probably not be a bit more vague
"Hides other Non Default Floating Tree blocks e.g. Barkshatter Rune's effect."
?

AverageUser125:
3. Mixin into the setBlockState function and invalidate the cache then
4. Just skip the first tick the display exists.

AverageUser125:
I did the mixin approach, and it works.
Instead of doing it for every entity data change like you proposed, I only did it for the first time it updates though.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinDisplay.java
```diff
@@ -0,0 +1,23 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(Display.class)
+public abstract class MixinDisplay {
+    @Inject(
+        method = "tick",
+        at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/world/entity/Display;createFreshRenderState()Lnet/minecraft/world/entity/Display$RenderState;",
```
hannibal002: ` createFreshRenderState` only fires in the else block of the if condtion `shouldInterpolate && this.renderState != null`. can we instad fire on `updateRenderSubState`? then we do catch all of them, without edge cases

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinBlockDisplay.java
```diff
@@ -0,0 +1,24 @@
...
+import at.hannibal2.skyhanni.data.EntityData;
+import net.minecraft.world.entity.Display;
+import net.minecraft.world.level.block.state.BlockState;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.ModifyArg;
+
+@Mixin(Display.BlockDisplay.class)
+public abstract class MixinBlockDisplay {
+    @ModifyArg(
```
hannibal002: Inject works as well instead of more aggressive modify arg here

src/main/java/at/hannibal2/skyhanni/config/features/foraging/CleanTreeViewConfig.kt
```diff
@@ -0,0 +1,32 @@
+package at.hannibal2.skyhanni.config.features.foraging
+
+import at.hannibal2.skyhanni.config.FeatureToggle
+import com.google.gson.annotations.Expose
+import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorBoolean
+import io.github.notenoughupdates.moulconfig.annotations.ConfigOption
+
+class CleanTreeViewConfig {
```
hannibal002, line 8: this function still has three feature toggles. the names of the features makes it sound like the first one is the main toggle and the two below that are just "sub options" that depend on the main toggle. either remove the feature toggles (and the code accordingly) or rename the enabled config/name/desc to reflect that those are three independent settings

src/main/java/at/hannibal2/skyhanni/features/foraging/ClearTreeLogs.kt
```diff
@@ -23,9 +30,39 @@ object ClearTreeLogs {
...
+        if (entity.blockState !in treeBlocks) return
+
+        if ((config.hideTreeBlocks == config.hideRuneEffects) && config.hideTreeBlocks) {
+            event.cancel()
+            return
+        }
+
+        val transformation = entity.transformation ?: return
+        val floatingTree = isFloatingTreeBlock(transformation)
+        if ((config.hideTreeBlocks && floatingTree) ||
```
hannibal002: this reads confusing. maybe use a when

hannibal002 (changes requested):
"Hide Breaking Tree Blocks" and "Hide Rune Effects" are new and not related to the fix. please mention them in the changelog.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinDisplay.java
AverageUser125: No, that would mean there would never be any caching.
Since the tree progress transformation is moved around each frame.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinBlockDisplay.java
AverageUser125: yeah, but I didn't want to do another `getBlockState` call. since that needs to deserialize and stuff.
Also this lets us later add an event to modify which block is shown

src/main/java/at/hannibal2/skyhanni/config/features/foraging/CleanTreeViewConfig.kt
AverageUser125: The "enabled" is required for the other 2 to work.
The hide tree progress and runes are indepedant.

src/main/java/at/hannibal2/skyhanni/features/foraging/ClearTreeLogs.kt
AverageUser125: > the when can be cleanuped to one line

Okay, will revert

src/main/java/at/hannibal2/skyhanni/config/features/foraging/CleanTreeViewConfig.kt
lunaynx: The enabled toggle is probably not necessary here, could just go back to 2 independent toggles.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinDisplay.java
AverageUser125: Also, this already catches the edge case of `transformation` not being initialized.
The block display part actually catches the rest of the edge cases.

src/main/java/at/hannibal2/skyhanni/config/features/foraging/CleanTreeViewConfig.kt
AverageUser125: > feature listens in on two toggles, not just one. thats wrong

Idk anymore

src/main/java/at/hannibal2/skyhanni/features/foraging/ClearTreeLogs.kt
hannibal002: > Okay, will revert

yeah my bad, the cleanup was an error, given how confusing this now reads. the when is more clear

src/main/java/at/hannibal2/skyhanni/config/features/foraging/CleanTreeViewConfig.kt
AverageUser125: Imma just remove the FeatureToggle annotation from the sub options.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinDisplay.java
hannibal002: > No, that would mean there would never be any caching.

Never caching would mean running each frame, at least 60 times a second.

The whole tick function in `Display.java` only gets fired at max 20 times per tick. and even that just if `updateRenderState` is set. this flag only gets set in `onSyncedDataUpdated`.

so this is exactly what we want to listen in on.

AverageUser125: eh, I don't feel it's worth it since this for the the display render TRANSFORM.
And for mobs we don't invalidate the cache if they change gear or smth.

But will do anyways.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinBlockDisplay.java
hannibal002: > since that needs to deserialize and stuff

`getBlockState` reads from an map. no modification/creation/serialization happens there.

> Also this lets us later add an event to modify which block is shown

this is fair, thoguh i would argue either add a comment explaining why we intentionally make it more complex as is currently needed or just use `@Inject`.

hannibal002 (changes requested):
> + Added to Clean Tree View a separate toggle to hide the Barkshatter rune. - Avrg

"Hide Breaking Tree Blocks" is still missing. is the optin that irrelevant that you think it would just bloat the changelog? 

also please see the two reopened comments above.

AverageUser125:
It was previously just "Hide Breaking Tree Blocks". it was just renamed.
So not really changelog worthy if it has been unchanged.

src/main/java/at/hannibal2/skyhanni/data/EntityData.kt
```diff
@@ -84,4 +86,22 @@ object EntityData {
...
+    fun onDisplayRenderStateUpdate(display: Display) {
+        lastVisibilityCheck.remove(display.id)
+    }
+
+    // TODO: Add event to allow the block in the block display to be modified
+    @JvmStatic
+    fun onBlockDisplayRenderState(display: Display.BlockDisplay, state: BlockState): BlockState {
+        val previous = display.blockRenderState()?.blockState
+
+        if (!state.isAir && (previous == null || previous.isAir)) {
```
hannibal002: this happens in the same tich where we also call onDisplayRenderStateUpdate, so this remove check here is dead code

src/main/java/at/hannibal2/skyhanni/data/EntityData.kt
```diff
@@ -84,4 +86,22 @@ object EntityData {
...
+            lastVisibilityCheck.remove(display.id)
+        }
+        return state
+    }
+
+    // TODO: add a hook for TextDisplay render state changes
```
hannibal002: doesnt this already exist?

src/main/java/at/hannibal2/skyhanni/features/foraging/ClearTreeLogs.kt
```diff
@@ -23,9 +30,42 @@ object ClearTreeLogs {
...
-    fun onRender(event: CheckRenderEntityEvent<Display.BlockDisplay>) {
-        if (!SkyHanniMod.feature.foraging.trees.cleanView) return
-        val block = event.entity.blockState
-        if (block in treeBlocks) event.cancel()
+    private fun onRender(event: CheckRenderEntityEvent<Display.BlockDisplay>) {
+        if (!isEnabled()) return
+        val entity = event.entity
+        if (entity.blockState !in treeBlocks) return
+
+        if ((config.hideTreeBlocks == config.hideRuneEffects) && config.hideTreeBlocks) {
```
hannibal002: if both are disabled, we should also return here

src/main/java/at/hannibal2/skyhanni/config/features/foraging/CleanTreeViewConfig.kt
```diff
@@ -0,0 +1,30 @@
...
+    @Expose
+    @ConfigOption(name = "Enabled", desc = "Hides floating blocks when foraging trees in Galatea.")
```
hannibal002: i read through the desc a bit more. it is not clear to me what enabled does when both sub toggles are disabled. how about we just drop the main toggle and make the two sub toggles both main toggles?

hannibal002 (changes requested):
pls dont forget the migration, changelog and config desc

src/main/java/at/hannibal2/skyhanni/data/EntityData.kt
AverageUser125: it is sadly not. idk what minecraft was cooking.

src/main/java/at/hannibal2/skyhanni/data/EntityData.kt
AverageUser125: sadly no, for some reason the individual render state of a display sometimes comes IN THE TICK AFTER the regular display render state comes.

src/main/java/at/hannibal2/skyhanni/config/features/foraging/CleanTreeViewConfig.kt
AverageUser125: Hannibal, you previously told me that it should have 1 main toggle.
So are you sure?

hannibal002: You previously had one toggle called enabled and two more also with feature toggle. that was before i noticed that that main toggle was not doing anyhting.

if the main toggle does not do anything, itts not a main toggle. its sub features are main toggles

AverageUser125:
> pls dont forget the migration, changelog and config desc

What is there to change about the changelog and config description?

src/main/java/at/hannibal2/skyhanni/utils/DisplayEntityUtils.kt
```diff
@@ -0,0 +1,29 @@
...
+import net.minecraft.world.phys.Vec3
+import org.joml.Vector3f
+
+object DisplayEntityUtils {
+    val Display.transformation: Transformation?
+        get() {
+            return renderState()?.transformation?.get(0f)
+        }
+
+    inline val Display.rotation: Vec3 get() = this.lookAngle
```
hannibal002: the name rotation is conusing/wrong: look angle is a normalized vector, not a rotation. i would suggest inlining this

src/main/java/at/hannibal2/skyhanni/test/command/CopyNearbyEntitiesCommand.kt
```diff
@@ -250,8 +252,8 @@ object CopyNearbyEntitiesCommand {
     @Suppress("UnnecessarySafeCall")
```
hannibal002: dead suppress

src/main/java/at/hannibal2/skyhanni/features/fishing/WormholeFinder.kt
```diff
@@ -73,13 +73,6 @@ object WormholeFinder {
         lastPlayerPos = playerPos
     }
```
hannibal002, line 75: empty line

hannibal002 (changes requested):

AverageUser125:
I need to stop working on this PR while it is like 11 PM+.

hannibal002 (approved):

PR #6386, opened 2026-08-15
Fix: Add new attribute shards to Pest Profit Tracker

lunaynx:
## What
Torrhus Canyon replaced the universal Pest Shard with an individual attribute shard for each pest. This PR adds support for those to the Pest Profit Tracker.

## Changelog Fixes
+ Fixed new Attribute Shards from Pests not being tracked by Pest Profit Tracker. - Luna


hannibal002 (approved):
Image

PR #6389, opened 2026-08-16
Fix: Sprayonator Spray Duration

hannibal002:
## What
The tier is read from the Sprayonator in hand, falling back to the last held one because the chat message can arrive after a slot change.

## Changelog Fixes
+ Fixed the Spray Display always showing 30 minutes for upgraded Sprayonators. - hannibal2
    * The Juicy Sprayonator lasts 45 minutes, the Salty Sprayonator 60 minutes.
+ Fixed the "New Spray Notice" showing up after spraying with an upgraded Sprayonator. - hannibal2


lunaynx (approved):
Code looks good, didn't test.

PR #6392, opened 2026-08-16
Fix: Attribute Shard Overlay using stale Hunting Box data

hannibal002:
## What

The stored Hunting Box amount was only written for shards visible in the open box, and a shard at 0 never shows up there to correct itself, so it stayed stale forever. Opening the box now clears the number for every shard not seen in it, unless a search is active or a page of the current opening has not been visited.

`hunting-box.page` must keep its `page` and `pages` groups if it goes into the SkyHanni-REPO, otherwise every box reads as a single page.

## Changelog Fixes
+ Fixed the Attribute Shard Overlay showing "Enough in Hunting Box" for shards that are no longer in the Hunting Box. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/inventory/attribute/AttributeShardOverlay.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/inventory/attribute/AttributeShardOverlay.kt
@@ -288,7 +288,11 @@ object AttributeShardOverlay {
         val priceUntilMaxed = individualPrice * actualAmountUntilMaxed
         val shardItemName = internalName.repoItemName
 
-        val coveredString = if (amountInBazaarOrders != 0) "§eOn order" else "§aEnough in Hunting Box"
+        val coveredString = when {
+            amountInBazaarOrders != 0 -> "§eOn order"
+            amountInHuntingBox != 0 -> "§aEnough in Hunting Box"
+            else -> "§aNothing needed"
+        }
         val priceColor = if (amountInBazaarOrders != 0) "§b" else "§6"
         val priceToNextTierString =
             if (actualAmountToNextTier == 0) coveredString else "$priceColor${priceUntilNextTier.shortFormat()}"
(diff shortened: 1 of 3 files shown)


lunaynx:
> Opening the box now clears the number for every shard that is not in it.

This sounds like a bad tradeoff, you are throwing away cached data from pages 2+ if the user opens their hunting box but never scrolls to those pages again. We should first check if the user has scrolled through all pages since they last opened the hunting box, and only *then* delete data of shards that were not seen. Should be fairly easy now that the inventory title has page numbers.

CalMWolfs (dismissed):
> Opening the box now clears the number for every shard that is not in it. 

I haven't tested this but more than just the hunting box value uses the shard amount and this sounds like it just breaks those other features. 
Also like what happened to this feature why does this need to be done it worked fine like a month ago with a one tick delay on populating the display why is this weird workaround needed now?

lunaynx:
I think it never fully "worked fine". The problem is: you can correct the data easily if you see an updated number in the hunting box, but if the number is 0, the shard doesn't show up at all, so it doesn't get corrected. That's why you sometimes had to manually reset hunting box data if it got desynced and scrolling through all pages didn't fix it.

hannibal002:
@CalMWolfs 

`amountInBox` is also read by FusionDisplay and ShardTrackerDisplay. Neither breaks: a shard not in the box has 0 of it, so they read a correct value instead of a stale one. Nothing is cleared while a search is active or while pages of the current opening are unvisited.

@lunaynx 

That check is already in the PR. `syncHuntingBoxAmounts` returns early while `huntingBoxSeenPages.size < totalPages`, and both sets reset on page 1, which is every time the box opens.
The description was worded badly, I have rewritten it.

CalMWolfs:
> @CalMWolfs
> 
> `amountInBox` is also read by FusionDisplay and ShardTrackerDisplay. Neither breaks: a shard not in the box has 0 of it, so they read a correct value instead of a stale one. Nothing is cleared while a search is active or while pages of the current opening are unvisited.

As i siad, I havent tested i guess i just misinterpreted how the pr was working as i just read the description not the code, it sounds like you are doing this properly then so its all good and wont break other features

PR #6394, opened 2026-08-16
Feature: Add /shnavigateallarea to navigate all nodes within the current area

gm-infinite:
## Dependencies
None

## What
Adds `/shnavigateallarea` (alias `/shnavallarea`), a variant of `/shnavigateall` that only
navigates to nodes within the SkyBlock area the player is currently standing in, instead of
the whole island. Reuses the existing `NavigateAllHelper` navigation and `skip`/`stop`
handling; a node's area is determined the same way `IslandAreaBackend` determines the player's
area, via graph search to the nearest `AREA`/`SMALL_AREA` tagged node. Falls back to full-island
navigation if the player isn't currently inside a tracked area.
 
## Changelog New Features
+ Added `/shnavigateallarea`, a variant of `/shnavigateall` that only navigates to nodes in your current area. - GM_infinite


shnavallarea.patch
lunaynx, line 1: ?

lunaynx: Yeah, fair, I just noticed this file is there that shouldn't be, but as long as you clean it up before marking it ready for review it's fine.

Vivo-Y12:
I don't think this is necessary, because no player goes everywhere without a reason, and there is no reason to be everywhere at once in Hypixel Skyblock.

shnavallarea.patch
gm-infinite: ok that is an artifact, thanks for pointing that out, I realized I acccidently opened a branch in my fork and to not lose anything while removing the branch I decided to create a patch file. while comminting It seems like I forgot about it

gm-infinite:
> I don't think this is necessary, because no player goes everywhere without a reason, and there is no reason to be everywhere at once in Hypixel Skyblock.

I created this after I made "Floor drops" navigation to the safari. right now I saw that many people that go in groups, each person  handle a spesific area (ie. caverns, forest etc..). seeing that, I decided to create a area spesfiic navigate all since right now we cannot seperate the navigation to tagged nodes and cant filter which ones we want or not at all. I did not think anymore apllications than this to be honest.

Vivo-Y12:
> I created this after I made "Floor drops" navigation to the safari. right now I saw that many people that go in groups, each person handle a spesific area (ie. caverns, forest etc..). seeing that, I decided to create a area spesfiic navigate all since right now we cannot seperate the navigation to tagged nodes and cant filter which ones we want or not at all. I did not think anymore apllications than this to be honest.

i think prosible with graph beacuse graph node use sperage area node in boarder u can you trow graph

PR #6395, opened 2026-08-16
Backend: CustomScoreboard Removal Prep

AverageUser125:
## What
Some files use it, this makes them no longer use it.
This makes the custom scoreboard feature isolated, and as such easier to remove.

+ Removed internal usage of custom scoreboard for other features. - Avrg


AverageUser125:
Folded into #6336

PR #6397, opened 2026-08-16
Improvement: Open individual tracker settings in a separate config screen

lunaynx:
## What
All "Individual Tracker Settings" options are now hidden behind a button that links out to a separate config screen, similarly to Pet Display. This makes players able to search the config for features like Estimated Item Value again without being confused with 10+ different results.

## Changelog Improvements
+ Individual Tracker Settings now link out to separate config screens. - Luna
    * This makes it easier to find features like Estimated Item Value.


hannibal002 (changes requested):
Three things:

**1. Timed trackers keep half their settings in the main config.**
`timedTracker` in `TimedGenericIndividualConfig` sits next to `trackerConfig`, not inside it. So after this change `trackerConfig` moves to the new screen while the "Timed Tracker" accordion stays behind, together with "Universal Settings", "Sync Settings" and, on the Garden variants, "Open Uptime Settings". For every timed tracker (Garden BPS Tracker, `TimedIndividualTrackerConfig`, `TimedIndividualItemTrackerConfig`, `TimedGardenIndividual*`) the settings are permanently split across two screens, and the seven options from `TimedTrackerConfig` ("New Session on Game Start", "Sessions to Keep", "Days to Keep", "Weeks to Keep", "Months to Keep", "Years to Keep", "Other") still show up once per timed tracker in the config search. That is the same problem this PR is solving, just at a smaller scale. Intentional?

**2. No way back to the main config.**
Pressing escape on the new tracker settings screen drops you into the game instead of back into `/sh config`. With one button that was tolerable, with one on every tracker it gets annoying. Same for the Pet Display button, so it does not have to be fixed here, but it would be nice.

**3. Changelog.**
The config entry is called "Individual Tracker Settings", so please capitalize it:

+ Individual Tracker Settings now link out to separate config screens. - Luna

src/main/java/at/hannibal2/skyhanni/config/IndividualTrackerConfigGuiManager.kt
```diff
@@ -0,0 +1,81 @@
...
+        ConfigUtils.openEditor(getEditorInstance(tracker))
+    }
+
+    private fun getEditorInstance(tracker: GenericIndividualTrackerConfig<*>) = editors.getOrPut(tracker) {
+        val config = IndividualTrackerSettingsConfig(findTitle(tracker), tracker.trackerConfig)
+        val processor = MoulConfigProcessor(config)
+        BuiltinMoulConfigGuis.addProcessors(processor)
+        val driver = ConfigProcessorDriver(processor)
+        driver.warnForPrivateFields = false
+        val settingsField = checkNotNull(IndividualTrackerSettingsConfig::settings.javaField) {
```
hannibal002, line 36: have you tested this in production (in dev env this will work)? see ConfigUtils.tryFindEditor for more infos

src/main/java/at/hannibal2/skyhanni/config/SkyHanniConfigScreen.kt
```diff
@@ -0,0 +1,23 @@
...
+// MoulConfig stores previousScreen but never navigates back to it on close, so that is handled here.
+internal class SkyHanniConfigScreen(
+    title: Component,
+    guiContext: GuiContext,
+    previousScreen: Screen?,
+) : MoulConfigScreenComponent(title, guiContext, previousScreen) {
+
+    override fun onClose() {
+        val previous = previousScreen ?: return super.onClose()
+        if (guiContext.onBeforeClose() == CloseEventListener.CloseAction.NO_OBJECTIONS_TO_CLOSE) {
```
hannibal002, line 18: onAfterClose is missing

detekt/src/main/kotlin/potentialbugs/NonStorageDoesntNeedExpose.kt
```diff
@@ -27,12 +26,15 @@ class NonStorageDoesntNeedExpose(config: Config) : SkyHanniRule(
             //  - The property is private
             //  - The property has a getter
             //  - The property is annotated with ConfigEditorInfoText
-            //  - The property is annotated with ConfigEditorButton
+            //  - The property is annotated with ConfigEditorButton and is a plain Runnable
+            //    (a button can also be bound to a stored config object, which does need @Expose)
             //  - The property is annotated with Transient
             val hasExplicitGetter = property.getter?.hasBody() ?: false
+            val isPlainButton = property.hasAnnotation("ConfigEditorButton") &&
+                property.typeReference?.text == "Runnable"
```
hannibal002, line 34: comparing texxt is ignoring cases without explicit types or explicit written down "java.lang.Runnable". both cases would still allow having a unnecessary expose annotation. how about matching on the last segment, so java.lang.Runnable works too? for the implicit type case, checking the initializer would avoid flagging a stored object written without an explicit type

src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
```diff
@@ -103,11 +103,18 @@ object ConfigUtils {
     }
 
     fun openEditor(editor: MoulConfigEditor<*>) {
-        SkyHanniMod.screenToOpen = createConfigScreen(editor)
+        SkyHanniMod.screenToOpen = createConfigScreen(editor, findPreviousScreen(editor))
+    }
+
+    // Closing an editor that was linked out from another config screen (e.g. Pet Display) should return to that screen
+    private fun findPreviousScreen(editor: MoulConfigEditor<*>): Screen? {
```
hannibal002, line 110: This runs for every `openEditor` caller, not only the link-out buttons, so `jumpToEditor` builds a back chain too. Probably fine, but two things: the chain is unbounded, so a few round trips between a tracker screen and the main config mean one escape press per hop, and a jump within the same editor drops an existing previous screen because of the `currentEditor !== editor` check. Passing the previous screen explicitly from the two `open` functions instead would keep this to the cases you actually meant.

hannibal002 (changes requested):
TimedIndividualItemTrackerConfig and TimedGardenIndividualItemTrackerConfig now have no usages.

src/main/java/at/hannibal2/skyhanni/config/IndividualTrackerConfigGuiManager.kt
lunaynx: I'll double check, but I was testing in production (easier than dev env with my Linux build server setup right now) and clicking the button worked fine. The issue you're thinking of *might* be gone with 26.1+.

hannibal002 (changes requested):
bump

src/main/java/at/hannibal2/skyhanni/config/SkyHanniConfigScreen.kt
lunaynx: You mean we should invoke `guiContext.onAfterClose()` here? I don't think that's right: https://github.com/NotEnoughUpdates/MoulConfig/blob/bd7b9aa3dbf7cef428280daf0156cb5e06338530/modern/templates/java/io/github/notenoughupdates/moulconfig/platform/MoulConfigScreenComponent.java#L60-L70

PR #6406, opened 2026-08-17
Backend: Reduce unnecessary formattedText comparisons

lunaynx:
## What
Small cleanup sweep searching for the regex `formattedTextCompat.*§` across the code. May be a slight win for memory usage.

Notable exclusions:
* DungeonHideItems (covered by #6404)
* HoppityCollectionStats (will be its own PR whenever I feel like it)
* InstanceChestProfit (covered by #6396)

+ Replaced some unnecessary uses of formattedTextCompat methods with colorless or component style equivalents. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/compacttablist/TabListReader.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/compacttablist/TabListReader.kt
@@ -5,6 +5,7 @@ import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.events.TabListUpdateEvent
 import at.hannibal2.skyhanni.events.TablistFooterUpdateEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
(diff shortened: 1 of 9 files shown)


src/main/java/at/hannibal2/skyhanni/features/misc/compacttablist/TabListReader.kt
```diff
@@ -238,7 +240,7 @@ object TabListReader {
         }
         upgradesPattern.matchMatcher(component) {
             if (!inUpgrades) return@matchMatcher
-            if (!component.formattedTextCompat().startsWith("§e")) return@matchMatcher
+            if (component.intoSpan().sampleStyleAtStart().color != TextColor.fromLegacyFormat(YELLOW)) return@matchMatcher
```
AverageUser125, line 242: I feel like this kind of thing with getting if a component is colored or bold should have dedicated helpers
instead of doing `intoSpan().sampleSttyleAtStart()`. maybe just `Component.startStyle`?

PR #6410, opened 2026-08-18
Fix: Auto complete error if never joined hypixel - SMALL

AverageUser125:
## What
Friends list was nullable, so if you never opened hypixel it would have been null, and getting your current friends would give a null exception.
Now it's just initialized as an empty map.

Also changed the file from a JAVA to KOTLIN file, which I am unsure why it wasn't like that previously.

## Changelog Fixes
+ Fixed an error when autocompleting commands if you have never joined Hypixel. - Avrg


src/main/java/at/hannibal2/skyhanni/data/FriendApi.kt
```diff
@@ -123,7 +114,7 @@ object FriendApi {
     }
 
     @HandleEvent
-    fun onChat(event: SkyHanniChatEvent.Allow) {
+    private fun onSystemMessage(event: SkyHanniChatEvent.Allow) {
```
lunaynx: Change this back to onChat, the IntelliJ plugin is bugged

AverageUser125: Thanks luna.

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/local/FriendsJson.kt
```diff
@@ -0,0 +1,23 @@
+package at.hannibal2.skyhanni.data.jsonobjects.local
+
+import com.google.gson.annotations.Expose
+import java.util.UUID
+
+
```
hannibal002: duplicate empty line

src/main/java/at/hannibal2/skyhanni/data/jsonobjects/local/FriendsJson.kt
```diff
@@ -0,0 +1,23 @@
...
+data class FriendsJson(
```
hannibal002, line 6: why are now all classes top level? too generic of a name, please move into sub classes

hannibal002 (changes requested):

hannibal002 (approved):

PR #6411, opened 2026-08-18
Fix + Improvement + Backend: Use RenderChest for 26.2 Outline Logic

AverageUser125:
## What
Uses https://github.com/AzureAaron/RenderChest for 26.2 glowing/outline logic instead of skyhannis custom one.
The library doesn't have a 26.1 so the existing working 26.1 remains.

Also some needed cleanup RenderLivingEntity. the outline event. and also aliased `EntityRemovedEvent` since it was just a duplicate of `EntityLeaveWorld`.

## Changelog Improvements
+ Improved entity glowing performance on Minecraft 26.2. - Avrg
## Changelog Fixes
+ Fixed excessive memory usage when many entities glow on Minecraft 26.2. - Avrg
+ Switched to a dedicated library for non-X-Ray entity glowing. - Avrg


build.gradle.kts
```diff
@@ -196,6 +196,11 @@ dependencies {
 
     shadowImpl(libs.httpclient)
     "minecraftTestClientRuntimeLibraries"(libs.httpclient)
+
+    if (target.renderChestVersion != null) {
+        shadowImpl("net.azureaaron:render-chest:${target.renderChestVersion}")
```
lunaynx: Should be JIJ'd, not shadow. Usually this is done via `include` + `implementation`.

Keep in mind that our current `includeImplementation` helper is misleading and is just an alias for `shadowImpl`, so don't use that.

lunaynx (changes requested):

build.gradle.kts
AverageUser125: Luna, I didn't understand a single word you said :sob:
But I think I did what you said.

hannibal002 (changes requested):
is it possible for the fix in this pr to not depend on other prs? if not, is it possible to move the fix up to one of the dependent prs instead of being in this one?

lunaynx (approved):
Clearing my previous review, looking good on my end after my changes.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
```diff
@@ -3,85 +3,90 @@ package at.hannibal2.skyhanni.mixins.hooks
...
-
-    @JvmStatic
-    fun <T : LivingEntity> internalSetColorMultiplier(entity: T, default: Int): Int {
-        if (GlobalRender.renderDisabled) return default
-        if (entityColorMap.containsKey(entity)) {
-            val condition = entityColorCondition[entity] ?: return default
-            if (condition.invoke()) {
-                return entityColorMap[entity]?.rgb ?: return default
-            }
+        val rgb = color.rgb.takeUnless { it == 0 } ?: return
```
hannibal002, line 95: every highlight reaches RenderChest with a reserved alpha, and white at 127 is exactly `REMOVE_GLOW`.

<details><summary>why</summary>

`Mob.highlight` replaces every opaque colour with the same colour at alpha 127, and `MobHighlight` passes 127 and 50 straight into `setEntityColor` without going through the wrapper. RenderChest documents any non opaque alpha as a reserved value and defines `REMOVE_GLOW` as `Integer.MAX_VALUE`, which is white at alpha 127. `queueEntitiesToOutline` already guards this with `or 0xFF000000`, this path does not, so a white highlight is stripped instead of shown.

</details>

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
```diff
@@ -3,85 +3,90 @@ package at.hannibal2.skyhanni.mixins.hooks
...
 
-    private val entityColorMap = mutableMapOf<LivingEntity, Color>()
-    private val entityColorCondition = ConcurrentHashMap<LivingEntity, () -> Boolean>()
+    private val entityGlowMap = mutableMapOf<Int, EntityGlowData>()
+    private var currentGlowEvent: RenderEntityOutlineEvent? = null
 
-    @JvmStatic
+    //? if >= 26.2 {
+    init {
+        CustomGlowCallback.EVENT.register { entity, _ ->
```
hannibal002, line 36: guarding `Mob.highlight` alone would not cover `MobHighlight`, which calls `setEntityColor` directly. the reserved values are a property of what this callback returns, so the guard belongs here.

build.gradle.kts
```diff
@@ -196,10 +194,16 @@ dependencies {
 
     shadowImpl(libs.httpclient)
     "minecraftTestClientRuntimeLibraries"(libs.httpclient)
+
+    if (target.renderChestVersion != null) {
+        includeImplementation("net.azureaaron:render-chest:${target.renderChestVersion}")
```
hannibal002: render-chest is jarred in but not declared in `fabric.mod.json`. skyblocker ships it too, and the loader picks the highest nested version with nothing warning about an incompatible one.

build.gradle.kts
```diff
@@ -196,10 +194,16 @@ dependencies {
...
+        "minecraftTestClientRuntimeLibraries"("net.azureaaron:render-chest:${target.renderChestVersion}")
+    }
 }
 
 fun DependencyHandler.includeImplementation(dep: Any, configure: ExternalModuleDependency.() -> Unit = {}) {
-    add("shadowImpl", dependencyNotation(dep)).also { (it as? ExternalModuleDependency)?.configure() }
+    include(dep).also { (it as? ExternalModuleDependency)?.configure() }
```
hannibal002: `dependencyNotation` existed because gradle returns null from `add` for provider notations, so the configure block never runs. `includeImplementation(libs.x) { exclude(...) }` would silently drop the exclude now.

stonecutter.gradle.kts
```diff
@@ -113,9 +113,18 @@ allprojects {
         }
 
         maven("https://jitpack.io") {
-            // NotEnoughUpdates (compiled against), Changelog builder, Preprocessor, Discord IPC
+            // MoulConfig, Changelog builder, Methanol (Modrinth Publisher)
```
hannibal002: moulconfig is `org.notenoughupdates.moulconfig` from the NEU maven, it never went through jitpack. `io.github.notenoughupdates.moulconfig` is the package name from the relocate rule, not the group.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinEntityRenderState.java
```diff
@@ -25,7 +24,8 @@ public abstract class MixinEntityRenderState implements EntityRenderStateStore {
         return skyhanni$savedEntity;
     }
 
-    @Override
+    //? if < 26.2 {
```
hannibal002, line 24: the `skyhanni$usingCustomOutline` field a few lines above can move in here too, nothing reads it on 26.2

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinLivingEntityRenderer.java
```diff
@@ -69,7 +69,8 @@ private int modifyRenderAlpha(int argb) {
         return argb;
     }
 
-    @WrapWithCondition(
+    //? if < 26.2 {
```
hannibal002, line 72: does an invisible highlighted armor stand still show only its equipment outline on 26.2?

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
hannibal002: RenderChest recolours an existing vanilla glow and then shows it through walls. `GlowingDroppedItems` opts out with `isCurrentlyGlowing`, `Mob.highlight` does not. wanted?

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
```diff
@@ -3,85 +3,90 @@ package at.hannibal2.skyhanni.mixins.hooks
...
+    @HandleEvent
+    private fun onEntityLeaveWorld(event: EntityLeaveWorldEvent<LivingEntity>) {
+        entityGlowMap.remove(event.entity.id)
     }
 
     fun <T : LivingEntity> removeEntityColor(entity: T) {
-        entityColorMap.remove(entity)
-        entityColorCondition.remove(entity)
+        val entityId = entity.id
+        DelayedRun.runOrNextTick {
```
hannibal002, line 89: is this replacing the `ConcurrentHashMap` that `entityColorCondition` was? worth a short comment, otherwise it reads as a pointless deferral.

hannibal002 (changes requested):
imo this pr does way too much, but if it truly cant be made shorter so be it

build.gradle.kts
lunaynx: I really just restored the old logic from before we dropped 1.21.11. Looking further back, before we added 26.1 support, it didn't have a configure block at all. Which makes sense to me – unlike a shadowed library, you can't exactly exclude something from a nested jar without writing additional code to manually repackage it.

stonecutter.gradle.kts
lunaynx: I blame @j10a1n15, copied this comment line from her SBIL PR to try to avoid conflicts.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinLivingEntityRenderer.java
lunaynx: Yes, I tested this.

stonecutter.gradle.kts
j10a1n15: 🥺

lunaynx: I'll just revert this change here, can fix it in my Gradle cleanup PR. Don't want a fix PR doing too much like you said.

build.gradle.kts
lunaynx: Ironically, Fabric's jar-in-jar is not as good as you suggest. If nothing declares an explicit dependency, and the mod isn't installed separately, it picks an arbitrary nested version, not the highest one.

Anyway, Aaron has promised no breaking changes within the same Minecraft version, so the only real risk here is another mod pulling in a _too old_ version. I'll declare the dependency explicitly, but this will probably require splitting our `fabric.mod.json` per-version again for the time being.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
hannibal002: bump

hannibal002 (changes requested):
bump

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
lunaynx: This was done by @AverageUser125, will try to find the discussion.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
lunaynx: Resolved (can't mark it as such)

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
lunaynx: First comment resolved, looking into the second one.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
AverageUser125: I didn’t see any reason to use a concurrent map for this really, since it’s mainly render thread anyways.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinEntityRenderState.java
lunaynx: Resolved (can't mark it as such)

AverageUser125: Will do, don’t worry Luna 🥺

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
lunaynx: Fixed. 26.1 glow logic now keeps vanilla glow xray when recoloring, just like Render Chest, meaning the Glowing Dropped Items check is no longer required and has been removed.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
AverageUser125: It’s the final review 🥺 .
But probably just write a comment about “ensures render thread”
Or smth.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
```diff
@@ -3,85 +3,92 @@ package at.hannibal2.skyhanni.mixins.hooks
...
-    @HandleEvent(SkyHanniTickEvent::class)
-    fun onTick() {
-        entityColorMap.removeIfKey { it.deceased }
-        entityColorCondition.removeIfKey { it.deceased }
+    @HandleEvent
+    private fun onEntityLeaveWorld(event: EntityLeaveWorldEvent<LivingEntity>) {
+        entityGlowMap.remove(event.entity.id)
     }
 
     fun <T : LivingEntity> removeEntityColor(entity: T) {
```
hannibal002, line 87: the convo above is exactly why it needs the comment: the deferral is what keeps the map single threaded, so it is load bearing and not just a delay. one line on `entityGlowMap` saying that covers both call sites at once.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
hannibal002: bump

src/main/java/at/hannibal2/skyhanni/mixins/hooks/EntityRenderStateStore.kt
```diff
@@ -0,0 +1,16 @@
+package at.hannibal2.skyhanni.mixins.hooks
+
+import net.minecraft.world.entity.Entity
+
+//~ if < 26.2 '{' -> ': GlowingStateStore {'
+interface EntityRenderStateStore {
```
hannibal002, line 9: `@file:Suppress("FunctionName")` with one comment would cover both, same in GlowingStateStore.kt

src/main/java/at/hannibal2/skyhanni/features/misc/GlowingDroppedItems.kt
```diff
@@ -12,19 +12,16 @@ import java.awt.Color
...
-    fun onRenderEntityOutlines(event: RenderEntityOutlineEvent) {
+    private fun onRenderEntityOutline(event: RenderEntityOutlineEvent) {
         if (!isEnabled()) return
-        if (event.type != RenderEntityOutlineEvent.Type.NO_XRAY) return
 
         event.queueEntitiesToOutline(::getGlowColor)
     }
 
     private fun getGlowColor(entity: Entity): Color? {
-        if (entity.isCurrentlyGlowing) return null
```
hannibal002, line 27: hypixel does make dropped items glow sometimes, so we now overwrite that instead of leaving it alone. intended? either way it is visible and needs a changelog entry.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/GlowingDroppedItems.kt
lunaynx: Good point. It probably doesn't matter much in practice because I expect Hypixel and SkyHanni to agree on dropped item glow color 99% of the time, but might as well restore the guard to preserve existing behavior. We can revisit this if we ever find a good reason to intentionally override Hypixel's glow color for dropped items.

For everything else (i.e., mobs), it's staying with no guard, because we don't want people to configure SkyHanni to highlight a mob in blue but it glows in red because Hypixel decided to add a red glow to it. This is distinct from Glowing Dropped Items, which explicitly says it glows items Hypixel doesn't.

The only potential additional user-facing change then is that we no longer turn Hypixel's existing X-ray glow into non-X-ray glow on entities we also glow, but I'm not aware of any actual cases where that happened previously in practice, so I'm not sure if it's worth a changelog entry. Let me know if you want me to add one (well, @AverageUser125 or you would have to add it, but yeah).

src/main/java/at/hannibal2/skyhanni/mixins/hooks/EntityRenderStateStore.kt
lunaynx: Done. In `EntityRenderStateStore` I used `@file:Suppress` as you said since putting it on the class would have meant fighting the Detekt annotation spacing rule (will make a separate PR to fix that soon). In `GlowingStateStore` I just put an interface-level `@Suppress`. Let me know if you want me to use `@file:Suppress` in both for consistency.

src/main/java/at/hannibal2/skyhanni/features/misc/GlowingDroppedItems.kt
lunaynx: The change has been made, so this is resolved unless you have any other comments.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
lunaynx: Added, along with an explicit render thread assert, because it seems useful to fail early if we mess up. A quick search suggests the underlying `Thread.currentThread()` call is cheap even when done every frame, so it's probably fine, but let me know if you want it removed.

hannibal002 (approved):

PR #6412, opened 2026-08-18
Backend + Improvement: Cleanup RenderLivingEntityHelper

AverageUser125:
## What

1. Removed the XRAY type of the `RenderEntityOutlineEvent` since it was unused
2. replaced the `onTick` usage in `RenderLivingEntityHelper` with just a `EntityLeaveWorldEvent`
3. Made `EntityRemovedEvent` a type alias to `EntityLeaveWorldEvent` (they were fired one after the other, so I just merged them)
4. Switched the `ConcurrentHashMap` to ~`Int2ObjectOpenHashMap`~ `MutableMap` + `DelayedRun`

For the last point, `Int2ObjectOpenHashMap` is from fastutils. which would be the first usage of it in SkyHanni.
(It is bundled into minecraft).
And since this map is called for every entity in the world and every tick, it's usage would be beneficial.

## Changelog Improvements
+ Improved entity glowing performance. - Avrg

+ Cleaned up RenderLivingEntityHelper. - Avrg

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
+++ b/src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
@@ -3,43 +3,38 @@ package at.hannibal2.skyhanni.mixins.hooks
 import at.hannibal2.skyhanni.api.event.HandleEvent
 import at.hannibal2.skyhanni.data.GlobalRender
 import at.hannibal2.skyhanni.events.RenderEntityOutlineEvent
-import at.hannibal2.skyhanni.events.minecraft.SkyHanniTickEvent
+import at.hannibal2.skyhanni.events.entity.EntityLeaveWorldEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
+import at.hannibal2.skyhanni.utils.DelayedRun
 import at.hannibal2.skyhanni.utils.EntityUtils.hasVisibleEquipment
-import at.hannibal2.skyhanni.utils.collection.CollectionUtils.removeIfKey
-import at.hannibal2.skyhanni.utils.compat.deceased
 import net.minecraft.world.entity.Entity
 import net.minecraft.world.entity.LivingEntity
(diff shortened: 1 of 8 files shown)


src/main/java/at/hannibal2/skyhanni/mixins/hooks/RenderLivingEntityHelper.kt
```diff
@@ -3,37 +3,32 @@ package at.hannibal2.skyhanni.mixins.hooks
...
 import java.awt.Color
-import java.util.concurrent.ConcurrentHashMap
 
 @SkyHanniModule
 object RenderLivingEntityHelper {
+    private data class EntityGlowData(val rgb: Int, val condition: () -> Boolean)
 
-    private val entityColorMap = mutableMapOf<LivingEntity, Color>()
-    private val entityColorCondition = ConcurrentHashMap<LivingEntity, () -> Boolean>()
+    private val entityGlowMap = Int2ObjectOpenHashMap<EntityGlowData>(128)
```
lunaynx: I know Minecraft uses this, but this is the only usage of fastutil in SkyHanni. Is there any reason we specifically need this here?

AverageUser125: It looks cool.
And is specifically created for integer as key.

I can use a generic map, but given this is for outline logic I wanted to try something more specialized.

lunaynx: I don't think we should arbitrarily start using classes like this without a good reason, especially not just in a single place. If we're gonna use it, it's better to have proper intent and reason other than "it looks cool".

AverageUser125: I do not feel like a generic map is appropriate here.
Since this is going to get called for every single glowing entity in the world, and every single tick.
and for dungeons that can be alot.

I don't mind using a generic map, but just feels odd for this use case.

PR #6416, opened 2026-08-18
Fix: PIP renderer performance

lunaynx:
## What
Realtime slots are cleared and re-rendered every frame, so keying them by `stableId` caused a full texture alloc/free cycle per item per frame whenever callers rebuilt their renderables. Size-keyed pooling reuses the textures regardless of caller identity, and stale settle tracker entries are now evicted instead of accumulating.

In addition, PIP textures are now also checked for eviction when we're not currently rendering any items, which is another small optimization, because previously some could have remained stale.

## Changelog Fixes
+ Fixed FPS drops caused by rendering items in SkyHanni GUIs. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinGuiRenderer.java
+++ b/src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinGuiRenderer.java
@@ -2,6 +2,7 @@ package at.hannibal2.skyhanni.mixins.transformers;
 
 import at.hannibal2.skyhanni.mixins.hooks.GuiRendererHook;
 import at.hannibal2.skyhanni.utils.render.item.SkyHanniGuiItemRenderState;
+import at.hannibal2.skyhanni.utils.render.item.SkyHanniItemRenderCoordinator;
 import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 import com.llamalad7.mixinextras.sugar.Local;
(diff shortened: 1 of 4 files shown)


src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinGuiRenderer.java
```diff
@@ -96,6 +96,14 @@ public MultiBufferSource.BufferSource getBufferSource() {
         );
     }
 
+    @Inject(
+        method = "preparePictureInPicture",
+        at = @At("TAIL")
+    )
+    private void skyhanni$evictUnusedRealtimeSlots(CallbackInfo ci) {
```
hannibal002, line 103: this runs every frame, it's not deferred like the pr desc says. the old code only evicted while skyhanni items were on screen, so slots stayed allocated once they were gone. the placement is fine, just fix the description.

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniItemRenderCoordinator.kt
```diff
@@ -14,74 +15,94 @@ import kotlin.math.abs
...
         pipStates: List<SkyHanniGuiItemRenderState>,
         //? if < 26.2
-        //bufferSource: BufferSource,
+        //bufferSource: MultiBufferSource.BufferSource,
         featureRenderDispatcher: FeatureRenderDispatcher,
         frameNumber: Int,
     ) {
-        if (pipStates.isEmpty()) return
-        handleEviction(frameNumber)
+        if (pipStates.isEmpty()) {
```
hannibal002: unnecessary check. the list is never empty in this function. we already return early if the list is empty in GuiRendererHook..

hannibal002 (changes requested):

hannibal002 (approved):

PR #6418, opened 2026-08-18
Fix: Don't corrupt category volume when playing own sounds

lunaynx:
## What
`SoundManager.updateCategoryVolume` sets a persistent engine gain multiplier rather than the options volume, so restoring the old level permanently quieted the category until a resource reload. SkyHanni sounds now bypass the volume settings via a mixin that only affects `SkyHanniSoundInstance`, leaving all other sounds and global state untouched.

This also broke our own volume boost (with "Maintain Volume During Warnings", we were supposed to be forcing the equivalent of 100% master+category volume for our own sounds, but in effect we didn't). It has now been flipped and renamed to "Boost Warning Volume" to better reflect the new behavior.

## Changelog Fixes
+ Fixed SkyHanni alert sounds being too quiet. - Luna
    * If you find that they are now too loud for your liking, turn off the "Boost Warning Volume" option.
+ Fixed SkyHanni making UI sounds from the game and other mods quieter. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/misc/MiscConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/misc/MiscConfig.kt
@@ -397,10 +397,11 @@ class MiscConfig {
     @Expose
     @ConfigOption(
         name = "Maintain Volume During Warnings",
-        desc = "Do not change game volume levels when warning sounds are played.",
+        desc = "Respect game volume settings for warning sounds instead of always playing them at 100%.",
     )
     @ConfigEditorBoolean
     @FeatureToggle
+    @SearchTag("levels loud quiet")
     var maintainGameVolume: Boolean = false
 
     @Expose
--- a/src/main/java/at/hannibal2/skyhanni/config/features/misc/MiscConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/misc/MiscConfig.kt
@@ -396,13 +396,14 @@ class MiscConfig {
 
     @Expose
     @ConfigOption(
-        name = "Maintain Volume During Warnings",
-        desc = "Respect game volume settings for warning sounds instead of always playing them at 100%.",
+        name = "Boost Warning Volume",
+        desc = "Play SkyHanni warning sounds at 100% volume regardless of game volume settings.\n" +
+            "If the game is muted, you still won't hear the sounds.",
     )
     @ConfigEditorBoolean
     @FeatureToggle
-    @SearchTag("levels loud quiet")
-    var maintainGameVolume: Boolean = false
+    @SearchTag("beep change ding during loud maintain pling quiet warnings")
+    var boostWarningVolume: Boolean = true
 
     @Expose
     @ConfigOption(
--- a/src/main/java/at/hannibal2/skyhanni/mixins/hooks/SoundEngineHook.kt
+++ b/src/main/java/at/hannibal2/skyhanni/mixins/hooks/SoundEngineHook.kt
@@ -11,7 +11,7 @@ object SoundEngineHook {
      * A complete mute is still respected and not bypassed.
      */
     @JvmStatic
-    fun modifySoundVolume(soundInstance: SoundInstance, original: Float): Float {
+    fun modifyVolume(soundInstance: SoundInstance, original: Float): Float {
         if (soundInstance !is SkyHanniSoundInstance) return original
         if (original == 0f) return original
         if (!SkyHanniMod.feature.misc.boostWarningVolume) return original
--- a/src/main/java/at/hannibal2/skyhanni/mixins/hooks/SoundEngineHook.kt
+++ b/src/main/java/at/hannibal2/skyhanni/mixins/hooks/SoundEngineHook.kt
@@ -12,6 +12,7 @@ object SoundEngineHook {
     @JvmStatic
     fun modifySoundVolume(soundInstance: SoundInstance, original: Float): Float {
         if (soundInstance !is SkyHanniSoundInstance) return original
+        if (original == 0f) return original
         if (SkyHanniMod.feature.misc.maintainGameVolume) return original
         return soundInstance.volume.coerceIn(0f, 1f)
     }
--- a/src/main/java/at/hannibal2/skyhanni/mixins/hooks/SoundEngineHook.kt
+++ b/src/main/java/at/hannibal2/skyhanni/mixins/hooks/SoundEngineHook.kt
@@ -8,12 +8,13 @@ object SoundEngineHook {
     /**
      * Makes SkyHanni's own sounds bypass the user's volume settings, without touching
      * the volume of any other sound or the sound engine's global volume state.
+     * A complete mute is still respected and not bypassed.
      */
     @JvmStatic
     fun modifySoundVolume(soundInstance: SoundInstance, original: Float): Float {
         if (soundInstance !is SkyHanniSoundInstance) return original
         if (original == 0f) return original
-        if (SkyHanniMod.feature.misc.maintainGameVolume) return original
+        if (!SkyHanniMod.feature.misc.boostWarningVolume) return original
         return soundInstance.volume.coerceIn(0f, 1f)
     }
 }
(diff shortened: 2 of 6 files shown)


src/main/java/at/hannibal2/skyhanni/utils/SkyHanniSoundInstance.kt
```diff
@@ -0,0 +1,30 @@
...
+    volume,
+    pitch,
+    SoundInstance.createUnseededRandom(),
+    false,
+    0,
+    SoundInstance.Attenuation.NONE,
+    0.0,
+    0.0,
+    0.0,
+    true,
```
AverageUser125, line 31: I know this doesn’t really matter since the attenuation is NONE. But still it probably makes sense for the “based on distance” Boolean to be false.

src/main/java/at/hannibal2/skyhanni/utils/SkyHanniSoundInstance.kt
```diff
@@ -0,0 +1,30 @@
...
+ * A UI sound created by SkyHanni. Unless the user opted out via the Maintain Volume During Warnings option,
+ * these sounds bypass the volume settings and play at their instance volume, without affecting any other sounds.
+ * See `MixinSoundEngine` and [at.hannibal2.skyhanni.mixins.hooks.SoundEngineHook].
+ */
+class SkyHanniSoundInstance(
+    soundEvent: SoundEvent,
+    pitch: Float,
+    volume: Float,
+) : SimpleSoundInstance(
+    soundEvent.location(),
```
AverageUser125: Why does it even have location?
Can’t this just be (0,0,0) and call it a day?

AverageUser125 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/SkyHanniSoundInstance.kt
lunaynx: That's an `Identifier` (formerly known as `ResourceLocation`). Blame Mojang for the confusing naming.

AverageUser125: So why create a `SoundEvent` if you’re just using it’s identifier?

src/main/java/at/hannibal2/skyhanni/utils/SkyHanniSoundInstance.kt
lunaynx: Funnily enough, apparently `isRelative` means "is this sound listener-relative?" – i.e., `isRelative=true x=0 y=0 z=0` means "the sound is centered on the player", while `isRelative=false x=0 y=0 z=0` means "play the sound at `(0, 0, 0)` world coordinates".

AverageUser125: I guess leave it on true then?
With a comment maybe.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/SoundEngineHook.kt
```diff
@@ -0,0 +1,18 @@
...
+object SoundEngineHook {
+    /**
+     * Makes SkyHanni's own sounds bypass the user's volume settings, without touching
+     * the volume of any other sound or the sound engine's global volume state.
+     */
+    @JvmStatic
+    fun modifySoundVolume(soundInstance: SoundInstance, original: Float): Float {
+        if (soundInstance !is SkyHanniSoundInstance) return original
+        if (SkyHanniMod.feature.misc.maintainGameVolume) return original
+        return soundInstance.volume.coerceIn(0f, 1f)
```
hannibal002, line 18: With the option off this discards the value vanilla computed, so the sliders drop
out entirely. For everything in between that is the point, but a slider at 0 is
not "quieter", it is off, and vanilla never starts a sound at volume 0
(`if (volume == 0.0F) ... return NOT_STARTED`). A muted client now hears every
warning at full volume.

Should 0 stay an exception?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/mixins/hooks/SoundEngineHook.kt
lunaynx: Good point, I'll keep 0% muted, looks like that's what the toggle originally did too.

src/main/java/at/hannibal2/skyhanni/mixins/hooks/SoundEngineHook.kt
```diff
@@ -0,0 +1,19 @@
+package at.hannibal2.skyhanni.mixins.hooks
+
+import at.hannibal2.skyhanni.SkyHanniMod
+import at.hannibal2.skyhanni.utils.SkyHanniSoundInstance
+import net.minecraft.client.resources.sounds.SoundInstance
+
+object SoundEngineHook {
+    /**
+     * Makes SkyHanni's own sounds bypass the user's volume settings, without touching
```
hannibal002, line 9: Both KDocs still claim an unconditional bypass. Worth mentioning the exception here and in SkyHanniSoundInstance, plus a short comment on the new line saying that 0 means muted rather than quiet.

src/main/java/at/hannibal2/skyhanni/config/features/misc/MiscConfig.kt
```diff
@@ -394,10 +394,11 @@ class MiscConfig {
     @Expose
     @ConfigOption(
         name = "Maintain Volume During Warnings",
```
hannibal002: The name still describes the old implementation, where SkyHanni changed the game volume levels. It never does that now, and the option means the opposite of what the name suggests. Do a display name and desc update, and add a TODO rename for the value.

src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinSoundEngine.java
```diff
@@ -33,4 +36,23 @@ public void handleSound(SoundInstance soundInstance, CallbackInfoReturnable<Soun
...
+
+    @ModifyExpressionValue(
+        method = "play",
+        at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/client/sounds/SoundEngine;calculateVolume(FLnet/minecraft/sounds/SoundSource;)F"
+        )
+    )
+    public float overrideSkyHanniSoundVolume(float original, @Local(argsOnly = true) SoundInstance soundInstance) {
+        return SoundEngineHook.modifySoundVolume(soundInstance, original);
```
hannibal002: modifySoundVolume repeats the class name, modifyVolume is enough

hannibal002 (changes requested):
update the changelog:  the second entry is broader than the bug was. The corrupted gain sat on SoundSource.UI, so only UI sounds got quieter, not sounds in general.

src/main/java/at/hannibal2/skyhanni/config/features/misc/MiscConfig.kt
lunaynx: Good point, I had changed the description already but not the name. I decided to just flip the logic of the toggle and add a migration because it seemed more awkward to explain otherwise.

src/main/java/at/hannibal2/skyhanni/utils/SkyHanniSoundInstance.kt
```diff
@@ -0,0 +1,32 @@
+package at.hannibal2.skyhanni.utils
+
+import net.minecraft.client.resources.sounds.SimpleSoundInstance
+import net.minecraft.client.resources.sounds.SoundInstance
+import net.minecraft.resources.Identifier
+import net.minecraft.sounds.SoundSource
+
+/**
+ * A UI sound created by SkyHanni. Unless the user opted out via the Maintain Volume During Warnings option,
```
hannibal002: The docs still names the old option and the old polarity. It is "Boost Warning Volume" now, and the bypass happens when the option is on, not unless it is off.

hannibal002 (changes requested):

hannibal002 (approved):

PR #6419, opened 2026-08-19
Backend: Fix IntelliJ run configurations + misc Gradle cleanup

lunaynx:
## What
- Fixes IntelliJ run configurations not being generated.
  * Simply re-adding `isIdeConfigGenerated = true` resulted in broken tasks being generated, so I added `preferGradleTask = true`, which makes it work *and* gives us full configuration cache compatibility for free, avoiding further issues.
- Fixes Gradle deprecation warnings and removes unnecessary safe calls in `build.gradle.kts`.
- Removes old Forge tweaker relocation that is doing nothing for us on Fabric (the class being relocated doesn't exist).
- Cleans up Gradle repository includes and removes unused Jitpack from `stonecutter.gradle.kts` (only used in `buildSrc`). Worth noting that `io.github.*` was never a thing on Jitpack – that was my mistake adding it earlier thinking of GitHub Pages.
- ~~Removes `includeImplementation` which is now just a 1:1 alias for `shadowImpl`, and a misleading one at that (include is supposed to be jar-in-jar, while shadow is just dumping the unzipped class files into the jar – two very different things).~~
  + Handled in #6411.

+ Fixed IntelliJ run configurations not being generated. - Luna
+ Cleaned up some deprecated and redundant code in the Gradle build scripts. - Luna


build.gradle.kts
```diff
@@ -56,45 +56,47 @@ loom.apply {
 
     runs {
```
hannibal002: dependencyNotation() is now dead code above

stonecutter.gradle.kts
```diff
@@ -100,22 +98,21 @@ allprojects {
...
             }
             filter {
-                includeGroup("me.shedaniel")
                 includeGroup("dev.architectury")
-                includeGroup("me.shedaniel.cloth")
+                includeGroupAndSubgroups("me.shedaniel")
             }
         }
 
+        // Changelog builder, Methanol (Modrinth Publisher)
```
hannibal002: Both only live in `buildSrc/build.gradle.kts`, which resolves through its own repositories block. Does anything in this build still pull from jitpack?

build.gradle.kts
```diff
@@ -126,19 +128,19 @@ dependencies {
...
-    target.fabricApiVersion?.let {
+    target.fabricApiVersion.let {
         implementation(it)
         "productionRuntimeMods"(it)
     }
     implementation(libs.fabricLanguageKotlin)
     "productionRuntimeMods"(libs.fabricLanguageKotlin)
 
-    target.modMenuVersion?.let {
+    target.modMenuVersion.let {
```
hannibal002: the .let is now useless

stonecutter.gradle.kts
```diff
@@ -100,22 +98,21 @@ allprojects {
...
         maven("https://jitpack.io") {
-            // NotEnoughUpdates (compiled against), Changelog builder, Preprocessor, Discord IPC
             content {
-                includeGroupByRegex("(com|io)\\.github\\..*")
+                includeGroupAndSubgroups("com.github")
```
hannibal002: settings.gradle.kts and buildSrc/build.gradle.kts still use includeGroupByRegex for the same repo. Align them in this PR too?

build.gradle.kts
```diff
@@ -56,45 +56,47 @@ loom.apply {
 
     runs {
         named("client") {
+            isIdeConfigGenerated = true
+            preferGradleTask = true
```
hannibal002, line 56: This makes the run config delegate to the Gradle task, so the "Select an appropriate Java 25 JDK" step in CONTRIBUTING.md no longer applies. The Gradle JVM step further up already covers it. Can you drop that step and `docs/run-configuration-settings.avif` here?

build.gradle.kts
```diff
@@ -198,13 +200,9 @@ dependencies {
     "minecraftTestClientRuntimeLibraries"(libs.httpclient)
 }
 
-fun DependencyHandler.includeImplementation(dep: Any, configure: ExternalModuleDependency.() -> Unit = {}) {
-    add("shadowImpl", dependencyNotation(dep)).also { (it as? ExternalModuleDependency)?.configure() }
```
hannibal002: `dependencyNotation` is unused now, this was its only caller.

hannibal002 (changes requested):

hannibal002 (approved):

PR #6420, opened 2026-08-19
Proof of Concept Feature: Added Config Dependencys

HacktheTime:
# Do not spend a lot of time on this Review. Its about the core Idea and a stylewise shitty implementation.

## What
Shows user dependencies, lets them jump to them, disabled while not matched and even show usages.

## Changelog New Features
+ Added a Big Viewer for SH Config Gui options on hover. - Hype_the_Time

## Changelog Improvements
+ Added Yellow Flash Highlight on jump in config - Hype_the_Time
    * Optional extra info.

## Changelog Fixes
+ Fixed Config always scrolling down even if in same Gui. - Hype_the_Time
    
+ Added Dependency References. - Hype_the_Time
   * Features can now define their Dependencies so they are shown to the user.

@hannibal002 as discussed.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/ConfigGuiManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/ConfigGuiManager.kt
@@ -28,8 +28,6 @@ object ConfigGuiManager {
     fun openConfigGui(search: String? = null) {
         val editor = getEditorInstance()
 
-        // A fresh config open always starts from the unfiltered view
-        ConfigUtils.clearSearchFilter()
         if (search != null) {
             editor.search(search)
         }
--- a/src/main/java/at/hannibal2/skyhanni/config/ConfigGuiManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/ConfigGuiManager.kt
@@ -28,6 +28,8 @@ object ConfigGuiManager {
     fun openConfigGui(search: String? = null) {
         val editor = getEditorInstance()
 
+        // A fresh config open always starts from the unfiltered view
+        ConfigUtils.clearSearchFilter()
         if (search != null) {
             editor.search(search)
         }
--- a/src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
@@ -3,6 +3,11 @@ package at.hannibal2.skyhanni.config
 import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.config.core.config.Position
 import at.hannibal2.skyhanni.config.core.config.PositionList
+import at.hannibal2.skyhanni.config.core.dependency.ConfigBannerProvider
+import at.hannibal2.skyhanni.config.core.dependency.FeatureDependencyResolver
+import at.hannibal2.skyhanni.config.core.dependency.GuiOptionEditorBigDescription
+import at.hannibal2.skyhanni.config.core.dependency.GuiOptionEditorDependencies
+import at.hannibal2.skyhanni.config.core.dependency.GuiOptionEditorUsedBy
 import at.hannibal2.skyhanni.config.storage.AchievementStorage
 import at.hannibal2.skyhanni.config.storage.CustomTodosStorage
 import at.hannibal2.skyhanni.config.storage.OrderedWaypointsRoutes
@@ -31,8 +36,9 @@ import at.hannibal2.skyhanni.utils.system.PlatformUtils
 import com.google.gson.Gson
 import io.github.notenoughupdates.moulconfig.annotations.ConfigLink
 import io.github.notenoughupdates.moulconfig.annotations.ConfigOption
+import io.github.notenoughupdates.moulconfig.common.RenderContext
+import io.github.notenoughupdates.moulconfig.gui.GuiContext
 import io.github.notenoughupdates.moulconfig.gui.GuiOptionEditor
-import io.github.notenoughupdates.moulconfig.gui.editors.GuiOptionEditorKeybind
 import io.github.notenoughupdates.moulconfig.processor.BuiltinMoulConfigGuis
 import io.github.notenoughupdates.moulconfig.processor.ConfigProcessorDriver
 import io.github.notenoughupdates.moulconfig.processor.MoulConfigProcessor
@@ -293,15 +299,21 @@ enum class ConfigFileType(val fileName: String, val clazz: Class<*>, val propert
     val backupFile get() = getBackupFile(file)
 }
 
-class BlockingMoulConfigProcessor : MoulConfigProcessor<SkyHanniConfig>(SkyHanniMod.feature) {
+open class BlockingMoulConfigProcessor : MoulConfigProcessor<SkyHanniConfig>(SkyHanniMod.feature) {
     val isDev by lazy { ContributorManager.isSelfDeveloper() }
 
+    @Suppress("ReturnCount")
     override fun createOptionGui(
         processedOption: ProcessedOption,
         field: Field,
         option: ConfigOption,
     ): GuiOptionEditor? {
-        val default = super.createOptionGui(processedOption, field, option) ?: return null
+        var default: GuiOptionEditor
+        try {
+            default = super.createOptionGui(processedOption, field, option) ?: return null
+        } catch (e: Exception) {
+            throw Exception("field: ${field.name} appearing class: ${field.declaringClass.simpleName}", e)
+        }
         if (processedOption !is ProcessedOptionImpl) return default
         var extraPath = ""
         val categoryParent = processedOption.category.parentCategoryId
@@ -309,20 +321,103 @@ class BlockingMoulConfigProcessor : MoulConfigProcessor<SkyHanniConfig>(SkyHanni
             extraPath = categoryParent.split(".").last() + "."
         }
         extraPath += processedOption.getPath()
-        if (default is GuiOptionEditorKeybind) {
-            UpdateKeybinds.keybinds.add(extraPath)
+
+        if (!isDev) {
+            if (field.isAnnotationPresent(OnlyDebug::class.java)) {
+                default = GuiOptionEditorHidden(default)
+            }
         }
 
         EnforcedConfigValues.isBlockedFromEditing(extraPath)?.let { extraMessage ->
             return GuiOptionEditorBlocked(default, extraMessage)
         }
 
-        if (!isDev) {
-            if (field.isAnnotationPresent(OnlyDebug::class.java)) {
-                return GuiOptionEditorHidden(default)
+        // Defer heavy dependency/third-party resolution so UI renders immediately.
+        return GuiOptionEditorBigDescription(GuiOptionEditorUsedBy(DeferredDependencyEditor(default, field), field), field)
+    }
+
+    private inner class DeferredDependencyEditor(
+        private val base: GuiOptionEditor,
+        private val field: Field,
+    ) : GuiOptionEditor(base.getOption()), ConfigBannerProvider {
+        @Volatile
+        private var resolved: GuiOptionEditor? = null
+        @Volatile
+        private var started = false
+
+        override fun bannerOffset(): Int =
+            ((resolved ?: base) as? ConfigBannerProvider)?.bannerOffset() ?: 0
+
+        private fun ensureStarted() {
+            if (started) return
+            synchronized(this) {
+                if (started) return
+                started = true
+                Thread({
+                    resolved = runCatching { buildDependencyAwareEditor() }.getOrElse { base }
+                }, "skyhanni-config-dep-resolver").apply { isDaemon = true }.start()
             }
         }
 
-        return default
+        private fun buildDependencyAwareEditor(): GuiOptionEditor {
+            val dependencyRequirements = FeatureDependencyResolver.resolve(field)
+            if (dependencyRequirements.isEmpty) return base
+            if (!dependencyRequirements.isEmpty) {
+                return GuiOptionEditorDependencies(base, dependencyRequirements, field)
+            }
+            return base
+        }
+
+        override fun render(context: RenderContext, x: Int, y: Int, width: Int) {
+            ensureStarted()
+            (resolved ?: base).render(context, x, y, width)
+        }
+
+        override fun renderOverlay(context: RenderContext, x: Int, y: Int, width: Int) {
+            (resolved ?: base).renderOverlay(context, x, y, width)
+        }
+
+        override fun mouseInput(
+            x: Int,
+            y: Int,
+            width: Int,
+            mouseX: Int,
+            mouseY: Int,
+            mouseEvent: io.github.notenoughupdates.moulconfig.gui.MouseEvent?,
+        ): Boolean {
+            ensureStarted()
+            return (resolved ?: base).mouseInput(x, y, width, mouseX, mouseY, mouseEvent)
+        }
+
+        override fun mouseInputOverlay(
+            x: Int,
+            y: Int,
+            width: Int,
+            mouseX: Int,
+            mouseY: Int,
+            mouseEvent: io.github.notenoughupdates.moulconfig.gui.MouseEvent?,
+        ): Boolean {
+            return (resolved ?: base).mouseInputOverlay(x, y, width, mouseX, mouseY, mouseEvent)
+        }
+
+        override fun keyboardInput(event: io.github.notenoughupdates.moulconfig.gui.KeyboardEvent?): Boolean {
+            return (resolved ?: base).keyboardInput(event)
+        }
+
+        override fun getHeight(): Int {
+            return (resolved ?: base).height
+        }
+
+        override fun setGuiContext(guiContext: GuiContext) {
+            (resolved ?: base).setGuiContext(guiContext)
+        }
     }
+
+    private data class ResolvedThirdParty(
+        val usesMainToggle: Boolean,
+        val requiresMainToggle: Boolean,
+        val message: String,
+        val overrideOwner: Class<*>?,
+        val overrideFieldName: String?,
+    )
 }
--- a/src/main/java/at/hannibal2/skyhanni/config/core/dependency/FeatureDependencyRequirement.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/core/dependency/FeatureDependencyRequirement.kt
@@ -1,0 +1,27 @@
+package at.hannibal2.skyhanni.config.core.dependency
+
+/**
+ * Declares feature dependencies that must be enabled before the annotated option can be edited.
+ *
+ * Accepted values inside [value]:
+ * - Third-party IDs or enum names (e.g. "BINGO_NET" or "bingo_brewers").
+ * - Config field references written as "QualifiedClass#field"; when the class part is omitted the
+ *   declaring class of the annotated option is assumed. The property has to be of the type boolean. Its recommended to use a property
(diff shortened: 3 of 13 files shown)


src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -293,36 +299,125 @@ enum class ConfigFileType(val fileName: String, val clazz: Class<*>, val propert
...
+            (resolved ?: base).renderOverlay(context, x, y, width)
+        }
+
+        override fun mouseInput(
+            x: Int,
+            y: Int,
+            width: Int,
+            mouseX: Int,
+            mouseY: Int,
+            mouseEvent: io.github.notenoughupdates.moulconfig.gui.MouseEvent?,
```
lunaynx, line 386: Unnecessary fully qualified name, import it instead.
```suggestion
            mouseEvent: MouseEvent?,
```

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -293,36 +299,125 @@ enum class ConfigFileType(val fileName: String, val clazz: Class<*>, val propert
...
+            return (resolved ?: base).mouseInput(x, y, width, mouseX, mouseY, mouseEvent)
+        }
+
+        override fun mouseInputOverlay(
+            x: Int,
+            y: Int,
+            width: Int,
+            mouseX: Int,
+            mouseY: Int,
+            mouseEvent: io.github.notenoughupdates.moulconfig.gui.MouseEvent?,
```
lunaynx, line 398: Unnecessary fully qualified name, import it instead.
```suggestion
            mouseEvent: MouseEvent?,
```

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -293,36 +299,125 @@ enum class ConfigFileType(val fileName: String, val clazz: Class<*>, val propert
...
+        ): Boolean {
+            return (resolved ?: base).mouseInputOverlay(x, y, width, mouseX, mouseY, mouseEvent)
+        }
+
+        override fun keyboardInput(event: io.github.notenoughupdates.moulconfig.gui.KeyboardEvent?): Boolean {
```
lunaynx, line 403: Unnecessary fully qualified name, import it instead.

```suggestion
        override fun keyboardInput(event: KeyboardEvent?): Boolean {
```

src/main/java/at/hannibal2/skyhanni/config/ConfigGuiManager.kt
```diff
@@ -28,6 +28,8 @@ object ConfigGuiManager {
     fun openConfigGui(search: String? = null) {
         val editor = getEditorInstance()
 
+        // A fresh config open always starts from the unfiltered view
+        ConfigUtils.clearSearchFilter()
```
Fazfoxy: This doesn't seem right to do arbitrarily within this PR unless I'm misunderstanding when this will activate?

src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
```diff
@@ -29,6 +32,62 @@ object ConfigUtils {
     )
     private val editorIndexCache = mutableMapOf<Field, Int>()
 
+    // moulconfig only consults the SearchFunction when the search text is non-blank, so the
+    // filtered config view keeps a sentinel in the search box; typing clears it and falls back
+    // to normal text search, clearing the box shows everything again.
+    private const val FILTER_SENTINEL = "###"
+    private var activeSearchFilter: Set<String>? = null
```
lunaynx, line 39: Pretty sure this can just be a non-nullable set that defaults to empty.
```suggestion
    private var activeSearchFilter: Set<String> = emptySet()
```

src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
```diff
@@ -29,6 +32,62 @@ object ConfigUtils {
...
+
+    /**
+     * Open the config GUI filtered to only show the given config option keys
+     * (see [UsedByResolver.transitiveUsedBy]).
+     */
+    fun openFilteredConfig(keys: Set<String>) {
+        val editor = ConfigGuiManager.getEditorInstance()
+        activeSearchFilter = keys.takeIf { it.isNotEmpty() }
+        if (activeSearchFilter == null) {
```
lunaynx, line 48:
```suggestion
        activeSearchFilter = keys
        if (activeSearchFilter.isEmpty()) {
```

src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
```diff
@@ -29,6 +32,62 @@ object ConfigUtils {
...
+                }
+            }
+            editor.search(FILTER_SENTINEL)
+        }
+        openEditor(editor, reuseOpenScreen = true)
+    }
+
+    /** Removes any active search filter so the full config is visible again. */
+    fun clearSearchFilter() {
+        activeSearchFilter = null
```
lunaynx, line 70:
```suggestion
        activeSearchFilter = emptySet()
```

src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
```diff
@@ -29,6 +32,62 @@ object ConfigUtils {
...
+        ConfigGuiManager.getEditorInstance().apply {
+            setSearchFunction { optionEditor, query -> optionEditor.fulfillsSearch(query) }
+            search("")
+        }
+    }
+
+    private fun MoulConfigEditor<*>.searchForJump(field: Field?) {
+        val filter = activeSearchFilter
+        if (filter == null) {
```
lunaynx, line 79:
```suggestion
        if (filter.isEmpty()) {
```

src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
```diff
@@ -96,21 +155,105 @@ object ConfigUtils {
...
+ * so the target option row can blink yellow briefly after the jump.
+ */
+object ConfigJumpHighlight {
+    private var target: String? = null
+    private var until = 0L
+    const val DURATION_MS = 3000L
+
+    fun highlight(field: Field) {
+        target = "${field.declaringClass.name}#${field.name}"
+        until = System.currentTimeMillis() + DURATION_MS
```
lunaynx, line 244: Use `SimpleTimeMark`.

src/main/java/at/hannibal2/skyhanni/config/ConfigManager.kt
```diff
@@ -293,36 +299,125 @@ enum class ConfigFileType(val fileName: String, val clazz: Class<*>, val propert
...
         processedOption: ProcessedOption,
         field: Field,
         option: ConfigOption,
     ): GuiOptionEditor? {
-        val default = super.createOptionGui(processedOption, field, option) ?: return null
+        var default: GuiOptionEditor
+        try {
+            default = super.createOptionGui(processedOption, field, option) ?: return null
+        } catch (e: Exception) {
+            throw Exception("field: ${field.name} appearing class: ${field.declaringClass.simpleName}", e)
```
lunaynx, line 315: Why are we blanket catching exceptions and then rethrowing with a non-descriptive message? Also, this should go through `ErrorManager`, I suspect it may just crash the game when this code path triggers as it is right now.

src/main/java/at/hannibal2/skyhanni/config/core/dependency/GuiOptionEditorUsedBy.kt
```diff
@@ -0,0 +1,339 @@
...
+     */
+    fun transitiveUsedBy(owner: Class<*>, fieldName: String): Set<String> {
+        if (!built) startBuild()
+        val keys = mutableSetOf<String>()
+        val queue = ArrayDeque<String>()
+        queue.add("${owner.name}#$fieldName")
+        while (queue.isNotEmpty()) {
+            val key = queue.removeFirst()
+            if (!keys.add(key)) continue
+            val dependents = synchronized(lock) { reverseMap[key]?.toList() ?: emptyList() }
```
lunaynx, line 334:
```suggestion
            val dependents = synchronized(lock) { reverseMap[key]?.toList().orEmpty() }
```

src/main/java/at/hannibal2/skyhanni/config/core/dependency/GuiOptionEditorUsedBy.kt
```diff
@@ -0,0 +1,339 @@
...
+    private val lock = Any()
+
+    /**
+     * Returns the dependents of the given config field, or an empty list when nothing requires
+     * it (or the scan has not finished yet).
+     */
+    fun findUsedBy(owner: Class<*>, fieldName: String): List<UsedByEntry> {
+        if (!built) startBuild()
+        if (!built) return emptyList()
+        synchronized(lock) { return reverseMap["${owner.name}#$fieldName"]?.toList() ?: emptyList() }
```
lunaynx, line 238:
```suggestion
        synchronized(lock) { return reverseMap["${owner.name}#$fieldName"]?.toList().orEmpty() }
```

src/main/java/at/hannibal2/skyhanni/config/core/dependency/GuiOptionEditorUsedBy.kt
```diff
@@ -0,0 +1,339 @@
...
+        if (depth > 10 || instance in visited) return
+        visited.add(instance)
+        for (field in instance.javaClass.declaredFields) {
+            try {
+                field.isAccessible = true
+            } catch (_: Throwable) {
+                continue
+            }
+            recordFieldDependents(field)
+            val value = try { field.get(instance) } catch (_: Throwable) { continue }
```
lunaynx, line 270: Why are we blanket suppressing any and all exceptions here? This will catch all sorts of things including `OutOfMemoryError` and `StackOverflowError`.

src/main/java/at/hannibal2/skyhanni/config/core/dependency/GuiOptionEditorUsedBy.kt
```diff
@@ -0,0 +1,339 @@
...
+            buildStarted = true
+            Thread({ build() }, "skyhanni-config-usedby-resolver").apply { isDaemon = true }.start()
+        }
+    }
+
+    private fun build() {
+        try {
+            walkInstance(SkyHanniMod.feature, mutableSetOf(), 0)
+        } catch (_: Throwable) {
+            // partial results are still usable
```
lunaynx, line 254: Why are we blanket suppressing any and all exceptions here? This will catch all sorts of things including `OutOfMemoryError` and `StackOverflowError`.

src/main/java/at/hannibal2/skyhanni/config/core/dependency/GuiOptionEditorUsedBy.kt
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    private fun walkInstance(instance: Any, visited: MutableSet<Any>, depth: Int) {
+        if (depth > 10 || instance in visited) return
+        visited.add(instance)
+        for (field in instance.javaClass.declaredFields) {
+            try {
+                field.isAccessible = true
+            } catch (_: Throwable) {
+                continue
```
lunaynx, line 267: Why are we blanket suppressing any and all exceptions here? This will catch all sorts of things including `OutOfMemoryError` and `StackOverflowError`.

src/main/java/at/hannibal2/skyhanni/config/core/dependency/FeatureDependencyResolver.kt
```diff
@@ -0,0 +1,151 @@
...
+                val javaField = owner.declaredFields.firstOrNull {
+                    @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
+                    it.name == fieldName && (it.type == java.lang.Boolean.TYPE || it.type == java.lang.Boolean::class.java)
+                }
+                    ?.apply { isAccessible = true }
+                if (javaField != null) {
+                    return javaField.getBoolean(receiver)
+                }
+            } catch (_: Throwable) {
+                // ignore and fall through
```
lunaynx, line 91: Why are we blanket suppressing any and all exceptions here? This will catch all sorts of things including `OutOfMemoryError` and `StackOverflowError`.

src/main/java/at/hannibal2/skyhanni/config/core/dependency/GuiOptionEditorUsedBy.kt
```diff
@@ -0,0 +1,339 @@
...
+            requirements.groups.forEach { group ->
+                group.dependencies.forEach { dep ->
+                    when (val source = dep.source) {
+                        is FeatureDependencyResolver.DependencySource.BooleanField ->
+                            addReverse("${source.owner.name}#${source.fieldName}", entry)
+                    }
+                }
+            }
+        } catch (_: Throwable) {
+            // ignore fields that cannot be resolved
```
lunaynx, line 303: Why are we blanket suppressing any and all exceptions here? This will catch all sorts of things including `OutOfMemoryError` and `StackOverflowError`.

src/main/java/at/hannibal2/skyhanni/config/core/dependency/GuiOptionEditorBigDescription.kt
```diff
@@ -0,0 +1,140 @@
...
+
+    override fun render(context: RenderContext, x: Int, y: Int, width: Int) {
+        // Only treat the actual option content as hoverable, not banners/panels reserved
+        // above it (used-by, dependencies, third-party warnings).
+        val offset = (base as? ConfigBannerProvider)?.bannerOffset() ?: 0
+        val contentTop = y + offset
+        val contentBottom = y + base.height
+        val hovered = IMinecraft.INSTANCE.mouseX in x..(x + width) &&
+            IMinecraft.INSTANCE.mouseY in contentTop until contentBottom
+        val now = System.currentTimeMillis()
```
lunaynx, line 35: Use `SimpleTimeMark`.

src/main/java/at/hannibal2/skyhanni/config/core/dependency/GuiOptionEditorBigDescription.kt
```diff
@@ -0,0 +1,140 @@
...
+            val blink = 0.5 + 0.5 * sin(progress * Math.PI * 4)
+            val alpha = (20 + 55 * blink).toInt().coerceIn(20, 75)
+            val color = (alpha shl 24) or 0x00FFFF00
+            context.drawColoredRect(x.toFloat(), y.toFloat(), (x + width).toFloat(), (y + base.height).toFloat(), color)
+        }
+    }
+
+    override fun renderOverlay(context: RenderContext, x: Int, y: Int, width: Int) {
+        val show = wasHovered && SkyHanniMod.feature.about.bigDescriptionEnabled &&
+            (context.isShiftDown() || System.currentTimeMillis() - hoverStart > delayMs())
```
lunaynx, line 56: Use `SimpleTimeMark`.

src/main/java/at/hannibal2/skyhanni/utils/ConfigUtils.kt
```diff
@@ -96,21 +155,105 @@ object ConfigUtils {
...
+    }
+
+    fun isActive(ownerName: String, fieldName: String): Boolean = remainingMs(ownerName, fieldName) > 0
+
+    /** Milliseconds left of the highlight, or 0 when inactive/expired. */
+    fun remainingMs(ownerName: String, fieldName: String): Long {
+        if (target != "$ownerName#$fieldName") return 0
+        val remaining = until - System.currentTimeMillis()
```
lunaynx, line 252: Use `SimpleTimeMark`.

src/main/java/at/hannibal2/skyhanni/config/core/dependency/FeatureDependencyRequirement.kt
```diff
@@ -0,0 +1,27 @@
+package at.hannibal2.skyhanni.config.core.dependency
+
+/**
+ * Declares feature dependencies that must be enabled before the annotated option can be edited.
+ *
+ * Accepted values inside [value]:
+ * - Third-party IDs or enum names (e.g. "BINGO_NET" or "bingo_brewers").
```
lunaynx, line 7: Why would a SkyHanni config option ever want to depend on an option from another mod? This doesn't sound easily generalizable, at most it might work if the other mod also uses a compatible version of MoulConfig, but it seems fragile and unnecessary.

lunaynx (changes requested):
I'm aware that this is a draft, but I've got some things to say.

This pull request is trying to do way too much at once. At a glance, it probably should be *at least* 4 separate pull requests. Please split it up.

Also, with all due respect, while the config dependency feature does sound useful to us, I don't appreciate you just throwing a bunch of unreviewed AI-generated code that fails to follow even basic coding standards, let alone SkyHanni's own standards, at us, especially when it clearly exists with the sole purpose of benefiting your own Bingo Net mod.

At this point, since you're clearly eager to generate a bunch of code, why not just ask Claude or whatever model you're using to rewrite your mod to be completely independent of SkyHanni? Instead, you're choosing to waste the time of SkyHanni maintainers and expecting us to either do free labor for you by reviewing thousands of lines of code, or be irresponsible and cause issues in our own mod by merging it blindly.

I'd much rather you go and do this in your own mod, where any mistakes can only possibly affect your own users.

At the end of the day, my opinion is not binding, but I urge you to consider it.

src/main/java/at/hannibal2/skyhanni/config/ConfigGuiManager.kt
HacktheTime: slipped through. yep you right.

hannibal002 (commented):
1. `ConfigGuiManager.openConfigGui` now calls `ConfigUtils.clearSearchFilter()` on every open, which also resets the normal search bar, not just the new filter. Previously a typed search term survived closing the config. Intentional? It is not in the changelog.

2. `@FeatureToggle bigDescriptionEnabled` sits in `About.kt`, but that category is otherwise meta only (version, update stream). Should this live in `gui` instead, next to `widenConfig`?

3. `EnchantParsingConfig.isColorParsing` is a new public `var` without `@Expose` and without `@ConfigOption`, existing purely as a reflection shim for the annotation. Is this meant to be the official way to make a `Property<Boolean>` option usable as a dependency? If so it needs to go into `CONTRIBUTING.md`, otherwise every contributor will solve it differently.

4. `FeatureDependencyRequirement` accepts `"SimpleClassName#field"` relative to the annotated field's package. That breaks silently once a config class moves to another package: `Class.forName` returns null and `parseDependency` drops the dependency without a word. Should an unresolvable entry at least fail loudly in the dev environment via `ErrorManager.crashInDevEnv`?

5. `UsedByResolver.build()` walks the entire live config object graph via reflection on a daemon thread while the game reads and writes those same objects. Have you seen any issues from that, or has this only been tested in the dev environment?

src/main/java/at/hannibal2/skyhanni/config/core/dependency/FeatureDependencyRequirement.kt
HacktheTime: modified from my offbranch. i only went though the compilation issues and adjusted things hani mentioned.

it was intended to be used as client implementation. TLDR it would be marking any calls to skyblock community servers such as soopy and farming elite too.

PR #6421, opened 2026-08-19
Improvement: Add /shnav stop argument to /shnav command

ThatOneDevil:
## What
Add `/shnav stop` argument to the `/shnav` command and replace the old `/shstopnavigation` command.

## Changelog Improvements
+ Added /shnav stop to the /shnav command. - ThatOneDevil
    * /shnav stop now also ends a running /shnavall.

## Changelog Removed Features
+ Removed the /shstopnavigation command. - ThatOneDevil
    * Replaced with /shnav stop.


src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigationHelper.kt
```diff
@@ -47,7 +48,7 @@ object NavigationHelper {
 
         val goBack = {
             doCommandAsync(searchTerm)
-            IslandGraphs.stopNavigation()
```
AverageUser125, line 50: Just do the `IslandGraphs.stopNavigation` for both.
(CONTRIBUTING.md says no direct function imports)

ThatOneDevil: alr i thought so but i havent contributed in a while so i forgot to read it again properly as u can see above 😭

src/main/java/at/hannibal2/skyhanni/data/IslandGraphs.kt
```diff
@@ -610,7 +613,7 @@ object IslandGraphs {
             category = CommandCategory.USERS_ACTIVE
             simpleCallback {
                 if (currentTarget != null) {
```
AverageUser125: How come this does the “currentTarget” check and the other one doesn’t?
(I guess move the check inside the function and make it print only if manual?)

ThatOneDevil: that code was previously there the function does include it already? and there is no "currentTarget" in NavigationHelper.kt

AverageUser125: I meant remove the condition and the chat user error from here into “stopNavigation”.
(Mostly the user error chat message when manual is true)

ThatOneDevil: im sorry i dont understand. what condition do u want me to remove? the manual one i assume? as it already tells the user it stopped in the navigation feedback?

AverageUser125: The call in NavigationHelper does not have currentTarget check, and as such does not give user feedback when nothing is selected.

so just move that check inside “stopNavigation” so both endpoints have it and have the same error chat message about not currently navigating.

ThatOneDevil: when navigation gets started or reset it calls stopNavigation therefore it calls `ChatUtils.userError("No navigation is currently active.")` would it be best to go back to the original impl or add another param so it doesnt send that when a player starts or resets navigation? instead if putting the check inside the function

AverageUser125: Can’t you just reuse the “manual” parameter to know if to show the user error?

ThatOneDevil: oh yes im stupid

src/main/java/at/hannibal2/skyhanni/data/IslandGraphs.kt
```diff
@@ -453,17 +453,24 @@ object IslandGraphs {
...
 
-    fun stopNavigation() {
-        if (currentTarget != null) {
+    fun stopNavigation(manual: Boolean = false) {
+        if (manual) {
+            if (currentTarget == null) {
+                ChatUtils.userError("No navigation is currently active.")
+                return
+            }
+            ChatUtils.userError("Manually stopped navigation")
```
hannibal002: this is a duplicate chat message

src/main/java/at/hannibal2/skyhanni/data/IslandGraphs.kt
```diff
@@ -566,7 +573,7 @@ object IslandGraphs {
     }
 
     fun cancelClick() {
```
hannibal002: the function name is now inaccurate. should be manualCancel.

src/main/java/at/hannibal2/skyhanni/data/IslandGraphs.kt
```diff
@@ -453,17 +453,24 @@ object IslandGraphs {
         }
     }
 
-    fun stopNavigation() {
-        if (currentTarget != null) {
+    fun stopNavigation(manual: Boolean = false) {
+        if (manual) {
```
hannibal002: Moving this into `if (manual)` drops the stop feedback for every other caller. Two of them are user triggered and now stop silently: `IslandAreaFeatures.buildAreaEntry` (tooltip "Click to stop navigating!") and the `goBack` click in `NavigationHelper.doCommandAsync`.

In GUI mode it is worse than silent: `NavigationFeedback.isActive()` stays true for another 3 seconds via `navLastActive`, so the stale "Navigating to X" line keeps rendering instead of being replaced.

src/main/java/at/hannibal2/skyhanni/data/IslandGraphs.kt
```diff
@@ -453,17 +453,24 @@ object IslandGraphs {
...
+            if (currentTarget == null) {
```
hannibal002: This can only trigger on the command path. `cancelClick()` also passes `manual = true` but is only reachable while a navigation runs. So `manual` means two things at once: "send the stop message" and "complain when nothing is active". Worth splitting?

src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigationHelper.kt
```diff
@@ -152,6 +152,9 @@ object NavigationHelper {
                     doCommandAsync(it.lowercase().removeColor())
                 }
             }
+            literalCallback("stop") {
```
hannibal002, line 155: `cancelClick()` is `stopNavigation(manual = true)` plus `onManualCancel()`, this is only the first half. Features passing `onManualCancel` to `pathFind` never learn about the command stop, although both paths now use `manual = true`. Calling `IslandGraphs.cancelClick()` here would make click and command the same path.

hannibal002: Should this also end a running `/shnavall`? It stops only the current leg, `NavigateAllHelper.route` and `currentTargetName` stay set, so `/shnavall skip` picks it back up.

hannibal002 (changes requested):
Per CONTRIBUTING.md, New Features is for standalone features independent of existing ones. This adds an argument to an existing command and replaces another one, so it belongs under Changelog Improvements, with the title prefix changing accordingly.

src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigationHelper.kt
ThatOneDevil: > Should this also end a running `/shnavall`? It stops only the current leg, `NavigateAllHelper.route` and `currentTargetName` stay set, so `/shnavall skip` picks it back up.

since both `shnavall` and `shnav` use the same system making `shnav stop` stop both would be ideal

ThatOneDevil: > `cancelClick()` is `stopNavigation(manual = true)` plus `onManualCancel()`, this is only the first half. Features passing `onManualCancel` to `pathFind` never learn about the command stop, although both paths now use `manual = true`. Calling `IslandGraphs.cancelClick()` here would make click and command the same path.

what do u suggest i do? im bit confused. i was thinking of making a function called `manualCancel` that cancels both navigation and use the same code as `cancelClick` and calling `NavigateAllApi.handleStop()` stops both already.

src/main/java/at/hannibal2/skyhanni/data/IslandGraphs.kt
ThatOneDevil: i reverted to the old code and im thinking of separating manual away from the `stopNavigation` function

src/main/java/at/hannibal2/skyhanni/data/IslandGraphs.kt
ThatOneDevil: reverting to the old code would fix this

src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigationHelper.kt
hannibal002: Yes, reverting `stopNavigation()` to always send the message is the right call. It is the low level stop, every caller should get feedback.

Then put the user facing part into `manualCancel()` and route all three entry points through it: the chat line click, the GUI element click, and `/shnav stop`. That function does the "nothing is active" check, then `stopNavigation()` and `onManualCancel()`.

A few things to watch once it also stops navigate all:
- the function is `NavigateAllHelper.handleStop`, not `NavigateAllApi`
- `handleStop` prints its own userError when no navigate all session is running, so calling it unconditionally would tell someone on a plain `/shnav` navigation to "use /shnavigateall to start navigation". Needs a guard.
- `handleStop(manual = true)` and `stopNavigation()` both print a stop message. Only one of them should survive, otherwise every cancel sends two.

ThatOneDevil: `NavigateAllHelper` doesn't exist only `NavigateAllApi` & `NavigateAllCommand` does

src/main/java/at/hannibal2/skyhanni/data/IslandGraphs.kt
```diff
@@ -467,6 +468,14 @@ object IslandGraphs {
         NavigationFeedback.setNavInactive()
     }
 
+    fun manualCancel() {
```
hannibal002, line 471: The active check is gone. `/shnav stop` with nothing running now prints "Manually stopped navigation" instead of the old "No navigation is currently active.", and it still calls `onManualCancel()`. That callback is never reset, so it re-runs the one from the last finished navigation, once per command call. The click path could not reach this before because it only exists while a navigation is active.

src/main/java/at/hannibal2/skyhanni/data/IslandGraphs.kt
```diff
@@ -467,6 +468,14 @@ object IslandGraphs {
...
+        currentTarget = null // guard to stop duplicate stop messages.
```
hannibal002: This makes `stopNavigation()` skip its `sendPathFindMessage`, so the manual path never sends a closing message to NavigationFeedback. In GUI mode the stale "Navigating to X" line then keeps rendering for the 3 second `navLastActive` window, and the only feedback is a red chat line, which is exactly what GUI mode exists to avoid.

instead, drop the `userError` below and let `stopNavigation()` send its normal "Navigation stopped!". Then this line is not needed either.

src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigateAllApi.kt
```diff
@@ -178,8 +178,8 @@ object NavigateAllApi {
         navigateToCurrent()
     }
 
-    fun handleStop(manual: Boolean = false) {
-        if (!currentlyNavigating) {
+    fun handleStop(manual: Boolean = false, errorMessage: Boolean = true) {
+        if (!currentlyNavigating && errorMessage) {
```
hannibal002: `errorMessage` now gates the early return, not only the message. With `errorMessage = false` and no navigate all running, the body falls through and calls `IslandGraphs.stopNavigation()` a second time. Keep the guard on `currentlyNavigating` and put only the message behind the flag.

hannibal002 (changes requested):
The changelog does not mention that `/shnav stop` now also ends a running `/shnavall`. That is user visible and deserves its own line.

src/main/java/at/hannibal2/skyhanni/data/IslandGraphs.kt
```diff
@@ -467,6 +468,17 @@ object IslandGraphs {
         NavigationFeedback.setNavInactive()
     }
 
+    fun manualCancel() {
+        if (currentTarget == null) {
```
hannibal002, line 472: Does this cover a navigate all run that sits between two targets? After `onFound` with a ChatMessage or SecondPassed condition, `waitingOnCondition` is true while IslandGraphs has no target, so this returns "No navigation is currently active." and the navigate all run keeps going afterwards. Guarding on the navigate all state as well would close that.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/IslandGraphs.kt
ThatOneDevil: i was unable to replicate this; also `currentTarget` would only be for normal navigation as in the `handleStop` function for `shnavall` it already checks for the navigation state

```
    if (!currentlyNavigating) {
        if (errorMessage) {
            ChatUtils.userError("No current navigation to stop. §eUse /shnavigateall to start navigation")
        }
        return
    }
```

hannibal002 (approved):

PR #6422, opened 2026-08-19
Backend: Fix Gradle deprecation warnings

lunaynx:
## What
Stop using deprecated Gradle and Loom syntax to get rid of Gradle warnings. Also removed unnecessary safe calls (`?.`) – they weren't being flagged on my end, but might as well.

+ Updated deprecated build script syntax to new syntax. - Luna

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -126,19 +126,19 @@ dependencies {
     ksp(libs.autoservice.ksp)
     implementation(libs.autoservice.annotations)
 
-    target.fabricLoaderVersion?.let {
+    target.fabricLoaderVersion.let {
         implementation(it)
         "productionRuntimeMods"(it)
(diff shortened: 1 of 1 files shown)


AverageUser125 (changes requested):
Could you also remove the unnecessary null checks on `target.fabricApiVersion`, `target.fabricLoaderVersion`, `target.modMenuVersion`?

AverageUser125 (approved):
🔥

build.gradle.kts
```diff
@@ -259,7 +259,7 @@ kotlin {
 tasks.processResources {
     from(includeBackupRepo)
     from(includeBackupNeuRepo)
-    val fapiVersion = target.fabricApiVersion?.split(":")?.last() ?: ""
+    val fapiVersion = target.fabricApiVersion.split(":")?.last()
```
AverageUser125: I mean, you probably missed the removal of "?" here.

PR #6424, opened 2026-08-20
Fix: Double Hook Leaking to next sea creature catch

AverageUser125:
## What
Made it store the timestamp for when a double hook was observed, and only counted it as a double hook if it's in the same or next tick, this way the timer naturally expires and no need to worry about intercepting messages.

Also made the regex be colorless and private listeners while I was here.

## Changelog Fixes
+ Fixed Fishing Double Hook sometimes being applied to the next caught Sea Creature. - Avrg

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/pet/PetStorageApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/pet/PetStorageApi.kt
@@ -714,7 +714,7 @@ object PetStorageApi {
         this.skinTag == skinTag
 
     fun isAutopetMessage(message: String): Boolean =
-        PetStoragePatterns.autoPetMessageColorlessPattern.matches(message.removeColor().removeResets())
+        PetStoragePatterns.autoPetMessageColorlessPattern.matches(message)
 
     fun markDirty() {
         jsonNeedsSave = true
--- a/src/main/java/at/hannibal2/skyhanni/api/pet/PetStorageApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/pet/PetStorageApi.kt
@@ -713,9 +713,6 @@ object PetStorageApi {
         this.level <= level &&
         this.skinTag == skinTag
 
-    fun isAutopetMessage(message: String): Boolean =
-        PetStoragePatterns.autoPetMessageColorlessPattern.matches(message)
-
     fun markDirty() {
         jsonNeedsSave = true
     }
--- a/src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
@@ -60,7 +60,6 @@ object SeaCreatureManager {
 
         getSeaCreatureFromMessage(message)?.let {
             SeaCreatureFishEvent(it, doubleHook).post()
-
             if (config.seaCreatureTracker.hideChat) {
                 event.blockedReason = "sea_creature_tracker"
                 doubleHook = false
@@ -103,7 +102,6 @@ object SeaCreatureManager {
             edited = when (config.compactDoubleHookPosition) {
                 CompactDoubleHookPosition.LEFT ->
                     "§e§lDOUBLE HOOK! ".asComponent().append(edited)
-
                 CompactDoubleHookPosition.RIGHT ->
                     edited.append(" §e§lDOUBLE HOOK!".asComponent())
             }
--- a/src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
@@ -2,8 +2,6 @@ package at.hannibal2.skyhanni.features.fishing
 
 import at.hannibal2.skyhanni.SkyHanniMod
 import at.hannibal2.skyhanni.api.event.HandleEvent
-import at.hannibal2.skyhanni.api.pet.PetStorageApi
-import at.hannibal2.skyhanni.data.WinterApi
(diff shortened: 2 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
```diff
@@ -71,16 +71,16 @@ object SeaCreatureManager {
 
     // if you can do it better make a pr
     @HandleEvent(onlyOnSkyblock = true)
-    fun onChat(event: SkyHanniChatEvent.Modify) {
-        if (doubleHookPattern.matches(event.message)) {
+    private fun onChat(event: SkyHanniChatEvent.Modify) {
```
hannibal002, line 65: If `config.seaCreatureTracker.hideChat` is enabled, the `Allow` handler blocks the catch message, so the `Modify` event never fires for it and the `doubleHook = false` you added here is skipped.

`Allow` returns early on that path too, so its own reset at the end of the function is skipped as well. When `Allow` swallows the message, it has to do the reset itself.

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
```diff
@@ -99,6 +99,7 @@ object SeaCreatureManager {
                 }
             }
 
+            doubleHook = false
```
hannibal002: logically correct place, but duplicate set still. either move into the if below, or set before `getSeaCreatureFromMessage(message)?.let {` and cache the old value for  the check in `if (config.compactDoubleHook && doubleHook) {`

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
```diff
@@ -151,14 +129,16 @@ object SeaCreatureManager {
...
-                counter++
             }
         }
         SeaCreatureManager.allFishingMobs = allFishingMobs
         allVariants = variants
         SpecificSeaCreatures.saveSeaCreatures(SpecificSeaCreatures.updateList())
     }
 
+    private fun isDoubleHookRecently(lastDoubleHookTime: SimpleTimeMark): Boolean =
+        lastDoubleHookTime.passedSince().inWholeTicks <= 1
```
hannibal002: instead of dealing with ticks, do a time comparison (<=100.milliseconds). we do not want a int overflow

src/main/java/at/hannibal2/skyhanni/api/pet/PetStorageApi.kt
```diff
@@ -714,7 +714,7 @@ object PetStorageApi {
         this.skinTag == skinTag
 
     fun isAutopetMessage(message: String): Boolean =
```
hannibal002: unused

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
```diff
@@ -151,14 +129,16 @@ object SeaCreatureManager {
...
                 variantFishes.add(name)
-                counter++
             }
         }
         SeaCreatureManager.allFishingMobs = allFishingMobs
         allVariants = variants
         SpecificSeaCreatures.saveSeaCreatures(SpecificSeaCreatures.updateList())
     }
 
+    private fun isDoubleHookRecently(lastDoubleHookTime: SimpleTimeMark): Boolean =
```
hannibal002: this function does not need a parameter. the field is accessible already

hannibal002: the function name ready weird. how about isDoubleHookRecent or wasDoubleHookRecently?

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
```diff
@@ -28,104 +27,83 @@ object SeaCreatureManager {
...
-            SeaCreatureFishEvent(it, doubleHook).post()
             if (config.seaCreatureTracker.hideChat) {
                 event.blockedReason = "sea_creature_tracker"
+                lastDoubleHookTime = SimpleTimeMark.farPast()
             }
             return
         }
 
-        doubleHook = false
+        // Does not reset lastDoubleHookTime on fail due to intercepting messages
```
hannibal002: comment without code. should be somewhere else

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
AverageUser125: Hold up, `inWholeTicks` can have an overflow?

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
AverageUser125: I wanted the function to be a pure function, since that feels more readable to me.

but fair enough, will do.

src/main/java/at/hannibal2/skyhanni/features/fishing/SeaCreatureManager.kt
hannibal002: inWholeMilliseconds is long , but we cast it to int in inWholeTicks. to be fair, we do div by 50 so this is somewhat secure, but better save than sorry

hannibal002 (changes requested):
`PetStoragePatterns.autoPetMessageColorlessPattern` is now also unused and can get deleted.

hannibal002 (approved):

PR #6426, opened 2026-08-21
Fix: Rendering correctness + performance fixes

lunaynx:
## What
Various rendering fixes. These have all been verified in game.

### General
- Fixed Holographic Entities making the game unplayably slow on 26.2 because it was instantiating a new `SubmitNodeStorage` object and calling `renderAllFeatures` every frame.
- Fixed lines drawn to the crosshair (e.g. burrow and pest waypoints) jumping around whenever the player sneaks or unsneaks on 26.1. The fix is to just use the 26.2 code for 26.1 as well, and then everything seems to work fine. 

### 26.1
- Fixed the world rendering using the wrong injection point (`AFTER_SOLID_FEATURES` instead of `AFTER_TRANSLUCENT_TERRAIN`). It also goes through the Fabric API event now instead of our own mixin.
  * The previous mixin was targeting `ldc=translucent`, which doesn't exist on 26.1+ – so it was silently ignoring the `ldc` filter fall back to the first `ProfilerFiller#push` call, which is `ldc=solidTerrain` (AKA Fabric API's `AFTER_SOLID_FEATURES`). The correct target is `ldc=translucentTerrain` (AKA Fabric API's `AFTER_TRANSLUCENT_TERRAIN`).
  * This is why the whole "deferred see-through text" hack was required in the first place to fix things like damage indicator text rendering below entities. It has now been removed.
  * As far as I can tell, this change doesn't currently have any known user-facing impact, but it may prevent subtle bugs in the future.

### 26.2
- Moved world rendering injection from our own mixin to Fabric API's `COLLECT_SUBMITS` (in practice, the injection point is effectively unchanged).
- Removed the unnecessarily incrementing `skyHanniTextSubmitOrder` – based on further reading on Minecraft rendering and verifying in-game, I've found that a simple `.order(1)` is sufficient to fix the issue it was created for (the Fruit Digging solver's see-through text rendering below its gray boxes).

## Changelog Fixes
+ Fixed extreme lag spikes in the Mirrorverse Crafting Room in the Rift on Minecraft 26.2. - Luna
+ Fixed some lines (e.g. for Burrow and Pest waypoints) jumping around when you sneak on Minecraft 26.1. - Luna

+ Fixed the SkyHanniRenderWorldEvent injection point on Minecraft 26.1. - Luna
+ Changed world text rendering to no longer unnecessarily increment the submit order on Minecraft 26.2. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/events/minecraft/SkyHanniRenderWorldEvent.kt
+++ b/src/main/java/at/hannibal2/skyhanni/events/minecraft/SkyHanniRenderWorldEvent.kt
@@ -19,8 +19,4 @@ class SkyHanniRenderWorldEvent(
     val submitNodeStorage: SubmitNodeStorage,
     val partialTicks: Float,
     var isCurrentlyDeferring: Boolean = true,
-) : SkyHanniEvent() {
-
-    //? if >= 26.2
-    internal var skyHanniTextSubmitOrder = 0
-}
+) : SkyHanniEvent()
--- a/src/main/java/at/hannibal2/skyhanni/events/minecraft/SkyHanniRenderWorldEvent.kt
+++ b/src/main/java/at/hannibal2/skyhanni/events/minecraft/SkyHanniRenderWorldEvent.kt
@@ -15,8 +15,8 @@ import net.minecraft.client.renderer.SubmitNodeCollector
 class SkyHanniRenderWorldEvent(
     val matrices: PoseStack,
     val camera: CameraRenderState,
-    //~ if < 26.2 'submitNodeStorage: SubmitNodeCollector' -> 'bufferSource: MultiBufferSource.BufferSource'
-    val submitNodeStorage: SubmitNodeCollector,
+    //~ if < 26.2 'submitNodeCollector: SubmitNodeCollector' -> 'bufferSource: MultiBufferSource.BufferSource'
+    val submitNodeCollector: SubmitNodeCollector,
     val partialTicks: Float,
 ) : SkyHanniEvent() {
     var isCurrentlyDeferring: Boolean = true
--- a/src/main/java/at/hannibal2/skyhanni/events/minecraft/SkyHanniRenderWorldEvent.kt
+++ b/src/main/java/at/hannibal2/skyhanni/events/minecraft/SkyHanniRenderWorldEvent.kt
@@ -3,11 +3,10 @@ package at.hannibal2.skyhanni.events.minecraft
 import at.hannibal2.skyhanni.api.event.SkyHanniEvent
 import at.hannibal2.skyhanni.skyhannimodule.PrimaryFunction
 import com.mojang.blaze3d.vertex.PoseStack
+import net.minecraft.client.renderer.SubmitNodeCollector
 import net.minecraft.client.renderer.state.level.CameraRenderState
 
-//? if >= 26.2 {
-import net.minecraft.client.renderer.SubmitNodeCollector
-//?} else {
+//? if < 26.2 {
 /*import net.minecraft.client.renderer.MultiBufferSource
 *///?}
 
@@ -15,8 +14,9 @@ import net.minecraft.client.renderer.SubmitNodeCollector
 class SkyHanniRenderWorldEvent(
     val matrices: PoseStack,
     val camera: CameraRenderState,
-    //~ if < 26.2 'submitNodeCollector: SubmitNodeCollector' -> 'bufferSource: MultiBufferSource.BufferSource'
     val submitNodeCollector: SubmitNodeCollector,
+    //? if < 26.2
+    //val bufferSource: MultiBufferSource.BufferSource,
     val partialTicks: Float,
 ) : SkyHanniEvent() {
     var isCurrentlyDeferring: Boolean = true
--- a/src/main/java/at/hannibal2/skyhanni/events/minecraft/SkyHanniRenderWorldEvent.kt
+++ b/src/main/java/at/hannibal2/skyhanni/events/minecraft/SkyHanniRenderWorldEvent.kt
@@ -6,7 +6,7 @@ import com.mojang.blaze3d.vertex.PoseStack
 import net.minecraft.client.renderer.state.level.CameraRenderState
 
 //? if >= 26.2 {
-import net.minecraft.client.renderer.SubmitNodeStorage
+import net.minecraft.client.renderer.SubmitNodeCollector
 //?} else {
 /*import net.minecraft.client.renderer.MultiBufferSource
 *///?}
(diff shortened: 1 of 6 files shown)


src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniItemRenderCoordinator.kt
```diff
@@ -36,28 +37,41 @@ internal object SkyHanniItemRenderCoordinator {
...
+
+    // Pooled slots idle for this many frames are closed, and stale settle entries dropped
+    private const val IDLE_EVICT_FRAMES = 100
     private val projectionBuffer by lazy { ProjectionMatrixBuffer("SkyHanni items") }
-    private val realtimeSlots = LinkedHashMap<Int, SkyHanniRealtimeItemSlot>()
-    private val realtimeSlotLastSeen = HashMap<Int, Int>() // stableId -> frameNumber
+
+    // Realtime slots are cleared and re-rendered every frame, so they carry no item identity.
+    // Pooling them by size lets frames reuse textures even when callers lose their stableId.
+    private class RealtimeSlotPool {
```
AverageUser125: data class is more appropriate.

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniItemRenderCoordinator.kt
```diff
@@ -70,7 +84,6 @@ internal object SkyHanniItemRenderCoordinator {
         frameNumber: Int,
     ) {
         if (pipStates.isEmpty()) return
-        handleEviction(frameNumber)
 
         val guiScale = Minecraft.getInstance().window.guiScale
         //~ if < 26.2 'SubmitNodeStorage()' -> 'bufferSource'
```
AverageUser125, line 76: You allocate a `SubmitNodeStorage` twice in this function, why not reuse a single one?

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniItemRenderCoordinator.kt
```diff
@@ -79,8 +92,9 @@ internal object SkyHanniItemRenderCoordinator {
 
         for (state in pipStates) {
             val settle = settleTracker.getOrPut(state.stableId) {
-                SettleEntry(state.rotationVector, state.adjustedScale, 0)
+                SettleEntry(state.rotationVector, state.adjustedScale, 0, frameNumber)
             }
+            settle.lastSeenFrame = frameNumber
             val stable = settle.rotationVec == state.rotationVector &&
```
AverageUser125, line 84: floating point stuff with -0.0 != 0.0 worries me.
Like how vector uses `Double.compare`

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniItemRenderCoordinator.kt
```diff
@@ -70,7 +84,6 @@ internal object SkyHanniItemRenderCoordinator {
         frameNumber: Int,
     ) {
         if (pipStates.isEmpty()) return
```
AverageUser125, line 72: if it's empty. it should also clear the `frameResources`, no?

AverageUser125 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniItemRenderCoordinator.kt
lunaynx: Not really, data classes holding mutable state is bad, and it gives us no real benefits.

src/main/java/at/hannibal2/skyhanni/utils/render/item/SkyHanniItemRenderCoordinator.kt
AverageUser125: Actually `submitBlit` and `isSettled` don't even check this rotation vector. only the scale.
Strange.

src/main/java/at/hannibal2/skyhanni/events/minecraft/SkyHanniRenderWorldEvent.kt
```diff
@@ -15,12 +15,9 @@ import net.minecraft.client.renderer.SubmitNodeStorage
 class SkyHanniRenderWorldEvent(
     val matrices: PoseStack,
     val camera: CameraRenderState,
-    //~ if < 26.2 'submitNodeStorage: SubmitNodeStorage' -> 'bufferSource: MultiBufferSource.BufferSource'
-    val submitNodeStorage: SubmitNodeStorage,
+    //~ if < 26.2 'submitNodeStorage: SubmitNodeCollector' -> 'bufferSource: MultiBufferSource.BufferSource'
+    val submitNodeStorage: SubmitNodeCollector,
```
hannibal002: wrong parameter name. previously it was a storage, this pr changes it to a collector. probably the same issue with the stonecutter names

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/events/minecraft/SkyHanniRenderWorldEvent.kt
lunaynx: Yeah, I left it alone because I wanted to keep the diff small, but I'll change it.

(Worth noting that `SubmitNodeStorage` is a subclass of `SubmitNodeCollector`, so every storage is also a collector, but not every collector is a storage. It's technically still passing the same object, but Fabric API uses the more generic superclass type.)

src/main/java/at/hannibal2/skyhanni/events/minecraft/SkyHanniRenderWorldEvent.kt
```diff
@@ -3,24 +3,21 @@ package at.hannibal2.skyhanni.events.minecraft
...
 
 @PrimaryFunction("onRenderWorld")
 class SkyHanniRenderWorldEvent(
     val matrices: PoseStack,
     val camera: CameraRenderState,
-    //~ if < 26.2 'submitNodeStorage: SubmitNodeStorage' -> 'bufferSource: MultiBufferSource.BufferSource'
-    val submitNodeStorage: SubmitNodeStorage,
+    val submitNodeCollector: SubmitNodeCollector,
+    //? if < 26.2
+    //val bufferSource: MultiBufferSource.BufferSource,
```
AverageUser125, line 19: Can you just remove buffer source at this point?
The only real usage is `WorldRenderUtils.drawString`.
And the change would just be removed the 26.1.2 stonecutter block and always using the 26.2 block.

It would make the render text bug also appear in 26.1. so maybe not worth it tbh.

lunaynx: That's too much of a change for this PR, I plan to migrate 26.1 to `SubmitNodeCollector` in another PR later.

AverageUser125: I mean, I literally just wrote to you that the change is really simple and easy and how to do it.
But alright.

lunaynx: There's a lot more than that which uses `MultiBufferSource` on 26.1, just changing only text would likely break things.

AverageUser125: my apologies then.

hannibal002 (approved):

PR #6428, opened 2026-08-21
Fix: Custom Pest Cooldown Timer showing always ready sometimes

AverageUser125:
## What
The problem was that if the last pest spawn time was far past, it would still add the custom cooldown amount, which would be nonesense.
Also it never reset the lastPestSpawn when entering/leaving the garden.

Also did generic cleanup (private event listeners + primary function)

## Changelog Fixes
+ Fixed the Custom Pest Cooldown option always showing the Pest Cooldown as ready. - Avrg

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/events/garden/pests/PestSpawnEvent.kt
+++ b/src/main/java/at/hannibal2/skyhanni/events/garden/pests/PestSpawnEvent.kt
@@ -3,7 +3,6 @@ package at.hannibal2.skyhanni.events.garden.pests
 import at.hannibal2.skyhanni.api.event.SkyHanniEvent
 import at.hannibal2.skyhanni.skyhannimodule.PrimaryFunction
 
-
 /**
  * When a pest spawn message gets detected while in the garden.
  */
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestSpawnTimer.kt
```diff
@@ -95,9 +91,9 @@ object PestSpawnTimer {
     }.get().seconds
 
     @HandleEvent(onlyOnIsland = IslandType.GARDEN)
-    fun onWidgetUpdate(event: WidgetUpdateEvent) {
+    private fun onWidgetUpdate(event: WidgetUpdateEvent) {
         if (!event.isWidget(TabWidget.PESTS)) return
-        val widgetLines = event.widget.lines.map { it.string }
+        val widgetLines = event.cleanLines.takeIfNotEmpty() ?: return
```
hannibal002: before the pr we were setting cooldownNotDetected to true if clean lines is empty

src/main/java/at/hannibal2/skyhanni/events/garden/pests/PestSpawnEvent.kt
```diff
@@ -1,11 +1,13 @@
 package at.hannibal2.skyhanni.events.garden.pests
 
 import at.hannibal2.skyhanni.api.event.SkyHanniEvent
+import at.hannibal2.skyhanni.skyhannimodule.PrimaryFunction
```
hannibal002, line 5: duplicate empty lines

hannibal002 (changes requested):

hannibal002 (approved):

PR #6439, opened 2026-08-23
Fix: Potential mod incompatibility related to Chroma Text

AzureAaron:
## What
Fixes a small mistake where a chroma text related mixin was using @WrapOperation but ignoring the operation causing it to behave as if it were a normal @Redirect. I don't think this caused any problems but you never know if it will become one later...

## Changelog Fixes
+ Fixed a potential mod incompatibility related to Chroma Text. - AzureAaron

+ Fixed a Chroma Text mixin @WrapOperation effectively behaving as a @Redirect. - AzureAaron


lunaynx (approved):
Looks good, though with no immediate user-facing fix listed this should be "Backend: Fix Text mixin ignoring Operation" + "Changelog Technical Details" since we prefer to avoid technical jargon in user-facing changelogs (everything besides "Technical Details").

(Or if you want, you could probably write a user-facing changelog like "Fixed a potential mod compatibility issue related to chroma text." and keep the PR a "Fix".)

PR #6440, opened 2026-08-23
Fix: Fishing Hook Bobber disappearing

Fazfoxy:
## What
This, Seemingly???? Fixes the bug, I couldn't manage to recreate it with this change so idk

## Changelog Fixes
+ Fixed Fishing Bobber Timer disappearing at random when swapping items. - Fazfoxy

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/fishing/FishingApi.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/fishing/FishingApi.kt
@@ -163,8 +163,7 @@ object FishingApi {
         private set
 
     @HandleEvent(onlyOnSkyblock = true)
-    fun onJoinWorld(event: EntityEnterWorldEvent<FishingHook>) {
-        if (!holdingRod) return
+    private fun onJoinWorld(event: EntityEnterWorldEvent<FishingHook>) {
         if (event.entity.playerOwner?.isLocalPlayer == false) return
 
         lastCastTime = SimpleTimeMark.now()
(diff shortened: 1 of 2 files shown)


AverageUser125 (changes requested):
Also add the `holdingRod` check to `FishingHookDisplay`

src/main/java/at/hannibal2/skyhanni/features/fishing/FishingApi.kt
```diff
@@ -163,8 +163,7 @@ object FishingApi {
         private set
 
     @HandleEvent(onlyOnSkyblock = true)
-    fun onJoinWorld(event: EntityEnterWorldEvent<FishingHook>) {
-        if (!holdingRod) return
+    private fun onJoinWorld(event: EntityEnterWorldEvent<FishingHook>) {
         if (event.entity.playerOwner?.isLocalPlayer == false) return
```
hannibal002: this line alone will fail todo what it should do if player owner is null. instead use  `playerOwner?.isLocalPlayer != true `

src/main/java/at/hannibal2/skyhanni/features/fishing/FishingApi.kt
```diff
@@ -293,7 +292,7 @@ object FishingApi {
     }
 
     @HandleEvent
-    fun onItemInHandChange(event: ItemInHandChangeEvent) {
+    private fun onItemInHandChange(event: ItemInHandChangeEvent) {
         val wasHoldingRod = holdingRod
         // TODO correct rod type per island water/lava
         holdingRod = event.newItem.isFishingRod()
```
hannibal002, line 298: the listener onRepoReload needs to be private

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/fishing/FishingApi.kt
AverageUser125:
```suggestion
        if (event.entity.playerOwner?.isLocalPlayer != true) return
```

hannibal002 (approved):

PR #6441, opened 2026-08-23
Fix: makePickblock/makeShiftClick compatibility with other mods

lunaynx:
## What
Fixes some SkyHanni features that change clicks being incompatible with other mods because we bypass the `AbstractContainerScreen` and click via `MultiPlayerGameMode` directly. This required the addition of a ThreadLocal depth guard to prevent a StackOverflowError. 

`makeShiftClick` has been moved inside the event rather than being an extension method, for consistency with `makePickblock` and `redirectClick`.

I also documented the two click methods (`InventoryUtils.clickSlot` and `InventoryUtils.mouseClickSlot`), but did not change other invocations to use the latter for now out of fear of accidentally breaking something.

Also changed the events in this file from data classes to regular classes because making them data classes is pointless overhead and would require a `@ConsistentCopyVisibility` annotation because of the private constructor.

## Changelog Fixes
+ Fixed Shift Click NPC sell bypassing item protection from certain other mods such as Skyblocker. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/events/GuiContainerEvent.kt
+++ b/src/main/java/at/hannibal2/skyhanni/events/GuiContainerEvent.kt
@@ -2,7 +2,7 @@ package at.hannibal2.skyhanni.events
 
 import at.hannibal2.skyhanni.api.event.SkyHanniEvent
 import at.hannibal2.skyhanni.skyhannimodule.PrimaryFunction
-import at.hannibal2.skyhanni.utils.InventoryUtils
+import at.hannibal2.skyhanni.utils.ItemUtils.getItemCategoryOrNull
 import at.hannibal2.skyhanni.utils.SafeItemStack
 import at.hannibal2.skyhanni.utils.compat.SkyHanniGuiContainer
 import net.minecraft.client.gui.GuiGraphicsExtractor
@@ -10,10 +10,13 @@ import net.minecraft.world.inventory.AbstractContainerMenu
 import net.minecraft.world.inventory.ContainerInput
 import net.minecraft.world.inventory.Slot
 
-abstract class GuiContainerEvent(open val gui: SkyHanniGuiContainer, open val container: AbstractContainerMenu) : SkyHanniEvent() {
+abstract class GuiContainerEvent(
+    open val gui: SkyHanniGuiContainer,
+    open val container: AbstractContainerMenu,
+) : SkyHanniEvent() {
 
     @PrimaryFunction("onBackgroundDrawn")
-    data class BackgroundDrawnEvent(
+    class BackgroundDrawnEvent(
         override val context: GuiGraphicsExtractor,
         override val gui: SkyHanniGuiContainer,
         override val container: AbstractContainerMenu,
@@ -22,7 +25,7 @@ abstract class GuiContainerEvent(open val gui: SkyHanniGuiContainer, open val co
         val partialTicks: Float,
     ) : GuiContainerEvent(gui, container), Rendering
 
-    data class PreDraw(
+    class PreDraw(
         override val context: GuiGraphicsExtractor,
         override val gui: SkyHanniGuiContainer,
         override val container: AbstractContainerMenu,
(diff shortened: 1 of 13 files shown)


src/main/java/at/hannibal2/skyhanni/features/dungeon/CroesusChestTracker.kt
```diff
@@ -224,18 +225,18 @@ object CroesusChestTracker {
     }
 
     @HandleEvent(onlyOnSkyblock = true)
-    fun onSlotClick(event: GuiContainerEvent.SlotClickEvent) {
+    private fun onSlotClick(event: GuiContainerEvent.SlotClickEvent) {
```
hannibal002, line 226: the function onInventoryClose can be made private

src/main/java/at/hannibal2/skyhanni/events/GuiContainerEvent.kt
```diff
@@ -70,28 +76,64 @@ abstract class GuiContainerEvent(open val gui: SkyHanniGuiContainer, open val co
...
+                container: AbstractContainerMenu,
+                slotId: Int,
+                clickedButton: Int,
+                clickType: ContainerInput,
+            ): SlotClickEvent? {
+                if (postDepth.get() > 0) return null
+
+                postDepth.set(postDepth.get() + 1)
+                try {
+                    return SlotClickEvent(gui, container, slotId, clickedButton, clickType).post() as SlotClickEvent
```
hannibal002: the `as` is unnecessary if we just do this:

```kt
val event = SlotClickEvent(gui, container, slotId, clickedButton, clickType)
event.post()
return event
```

src/main/java/at/hannibal2/skyhanni/events/GuiContainerEvent.kt
```diff
@@ -70,28 +76,64 @@ abstract class GuiContainerEvent(open val gui: SkyHanniGuiContainer, open val co
...
+                } finally {
+                    val depth = postDepth.get() - 1
+                    check(depth >= 0) {
+                        postDepth.remove()
+                        "SlotClickEvent postDepth underflow detected"
+                    }
+                    if (depth == 0) postDepth.remove() else postDepth.set(depth)
```
hannibal002: the "else" check here can never be reached

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/events/GuiContainerEvent.kt
lunaynx: I mean, in theory if the guard is functioning correctly, you're right. But it's not a condition that's impossible at compile-time, and this is a standard pattern. It's really just a sanity check.

src/main/java/at/hannibal2/skyhanni/events/GuiContainerEvent.kt
```diff
@@ -70,28 +76,66 @@ abstract class GuiContainerEvent(open val gui: SkyHanniGuiContainer, open val co
...
+            private val postDepth = ThreadLocal.withInitial { 0 }
+
+            fun postEvent(
+                gui: SkyHanniGuiContainer,
+                container: AbstractContainerMenu,
+                slotId: Int,
+                clickedButton: Int,
+                clickType: ContainerInput,
+            ): SlotClickEvent? {
+                if (postDepth.get() > 0) return null
```
hannibal002: The guard returns as soon as the counter is above 0, so it is always 0 on entry and 1 on exit. A depth counter makes sense when multiple levels are allowed, but here the second one is rejected, so a boolean does the same without the unreachable branches.

src/main/java/at/hannibal2/skyhanni/events/GuiContainerEvent.kt
```diff
@@ -70,28 +76,66 @@ abstract class GuiContainerEvent(open val gui: SkyHanniGuiContainer, open val co
...
+
+                postDepth.set(postDepth.get() + 1)
+                try {
+                    val event = SlotClickEvent(gui, container, slotId, clickedButton, clickType)
+                    event.post()
+                    return event
+                } finally {
+                    val depth = postDepth.get() - 1
+                    check(depth >= 0) {
```
hannibal002: if this check ever fails, we crash. no try catch/error manager to rescue us between the mixin hook and this line

src/main/java/at/hannibal2/skyhanni/events/GuiContainerEvent.kt
hannibal002: see below

hannibal002 (changes requested):

hannibal002 (approved):

PR #6449, opened 2026-08-24
Fix: Custom Wardrobe Menu wrong inventory name

AverageUser125:
## What
If you switched wardrobe screens it didn't update the screen title. which meant when it passed the title to the edit screen it was using the old title it started with instead of the new one.

## Changelog Fixes
+ Fixed Custom Wardrobe causing wrong inventory names in edit mode. - Avrg

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/AbstractCustomMenuScreen.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/AbstractCustomMenuScreen.kt
@@ -28,11 +28,12 @@ abstract class AbstractCustomMenuScreen(
         ToolTipData.lastSlot = null
     }
 
-    fun changeHandler(newMenu: ChestMenu) {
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/utils/AbstractCustomMenuScreen.kt
```diff
@@ -28,11 +28,12 @@ abstract class AbstractCustomMenuScreen(
         ToolTipData.lastSlot = null
     }
 
-    fun changeHandler(newMenu: ChestMenu) {
+    fun changeHandler(newMenu: ChestMenu, inventoryName: Component) {
         menu.removeSlotListener(this)
         menu = newMenu
         menu.addSlotListener(this)
         ToolTipData.lastSlot = null
+        this.title = inventoryName
```
hannibal002: title is final. intellij flags this as an error

PR #6452, opened 2026-08-24
Fix: Mythological Event not being detected

AverageUser125:
## What
Adds an override for whether mythological ritual is active based on chat event + finding mythological mob.

## Changelog Fixes
+ Fixed Mythological Ritual not being detected sometimes. - Avrg


hannibal002 (changes requested):
this impl would mark all players in hub with a spade in hand as "doing diana". this is clearly wrong. 
instead, how about we still check the mayor api AND at the same time listen in on the diana related chat messages? if we get chat messages even though the api said no diana mayor, we still flag a toggle locally until the user reads the "you can only use the spade during a diana event" chat message, or the player leaves the island.

also, this check should replace the existing "is diana active" check, not the "is the player doing diana" check.

src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaProfitTracker.kt
```diff
@@ -42,13 +42,38 @@ object DianaProfitTracker {
...
     private val patternGroup = RepoPattern.group("diana.chat")
+
+    /**
+     * REGEX-TEST: You dug out a Griffin Burrow!
+     * REGEX-TEST: You finished the Griffin burrow chain! (4/4)
+     */
     private val chatDugOutPattern by patternGroup.pattern(
         "burrow.dug",
-        "(?:§eYou dug out a Griffin Burrow!|§eYou finished the Griffin burrow chain!) .*",
+        "(?:You dug out a Griffin Burrow!|You finished the Griffin burrow chain!).*",
```
hannibal002, line 52: the pattern changed, the key NEEDS to get changed too otherwise we break old versions.

src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaProfitTracker.kt
```diff
@@ -42,13 +42,38 @@ object DianaProfitTracker {
...
     )
+
+    /**
+     * REGEX-TEST: Wow! You dug out 1,000 coins!
+     */
     private val chatDugOutCoinsPattern by patternGroup.pattern(
         "coins",
-        "§6§lWow! §r§eYou dug out §r§6(?<coins>.*) coins§r§e!",
+        "Wow! You dug out (?<coins>[\\d,.]+) coins!",
```
hannibal002, line 60: the pattern changed, the key NEEDS to get changed too otherwise we break old versions.

src/main/java/at/hannibal2/skyhanni/features/itemabilities/CrownOfAvariceCounter.kt
```diff
@@ -215,7 +214,7 @@ object CrownOfAvariceCounter {
     }
 
 
-    private fun isEnabled() = SkyBlockUtils.inSkyBlock && config.enable && Perk.MYTHOLOGICAL_RITUAL.isActive
+    private fun isEnabled() = config.enable && DianaApi.isRitualActive()
```
hannibal002: isRitualActive does not check for "is in skyblock". either add the check there or here.

src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaApi.kt
```diff
@@ -25,14 +28,23 @@ object DianaApi {
...
+    fun isRitualActive(): Boolean {
+        ritualActiveOverride?.let { return it }
+        return (Perk.MYTHOLOGICAL_RITUAL.isActive || Perk.PERKPOCALYPSE.isActive) ||
+            SkyHanniMod.feature.dev.debug.assumeMayor.get() == ElectionCandidate.DIANA
+    }
 
     fun hasGriffinPet() = CurrentPetApi.isCurrentPet("Griffin")
 
-    fun isDoingDiana() = IslandType.HUB.isInIsland() && isRitualActive() && hasSpadeInHotbar()
+    fun isDoingDiana(strict: Boolean = false): Boolean {
```
hannibal002: remove the strict parameter. every call of is doing diana should always also check iis ritual active.

src/main/java/at/hannibal2/skyhanni/events/ItemAbilityActivateEvent.kt
```diff
@@ -0,0 +1,13 @@
...
+import at.hannibal2.skyhanni.features.itemabilities.abilitycooldown.ItemAbility
+import at.hannibal2.skyhanni.skyhannimodule.PrimaryFunction
+
+/**
+ * Fired when an item ability is activated.
+ *
+ * @param ability the activated item ability
+ */
+@PrimaryFunction("onItemAbilityActivate")
+class ItemAbilityActivateEvent(val ability: ItemAbility) : SkyHanniEvent()
```
hannibal002: the event would only trigger when config.itemAbilityCooldown is enabled, at least for the ones that rely on action bar listeners, also, this would not trigger immediately when the user clicks with the item client side, but rather when the server responds to this click with packets. both cases should be mentioned in the docs, ideally the precondition removed entirely.

src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaApi.kt
```diff
@@ -63,21 +75,50 @@ object DianaApi {
...
 
+    @HandleEvent(onlyOnIsland = HUB)
+    private fun onChat(event: SkyHanniChatEvent.Allow) {
+        if (ritualNotActivePattern.matches(event.cleanMessage)) {
+            overrideActiveRitual(active = false)
+        }
+    }
+
+    @HandleEvent(onlyOnIsland = HUB)
+    private fun onItemAbilityActivate(event: ItemAbilityActivateEvent) {
```
hannibal002: we can not use this event here. echo gets listened in on action bar changes, meaning delayed from the actual item click. we need to check client side if the user clicks with he spade here. and the click listener needs higher priority than the ones in burrow guess code, otherwise we would break burrow detection.

hannibal002 (changes requested):
The What section still describes the first version of this PR ("You either have the mayor, or the spade"), which is not what the code does anymore. Please rewrite it to describe the override: what sets it, what clears it, and that it only overrules the mayor data rather than replacing the check.

hannibal002 (changes requested):
bump

hannibal002 (changes requested):
bump

src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaApi.kt
```diff
@@ -63,21 +75,59 @@ object DianaApi {
...
 
+    @HandleEvent(onlyOnIsland = HUB)
+    private fun onChat(event: SkyHanniChatEvent.Allow) {
+        if (ritualNotActivePattern.matches(event.cleanMessage)) {
+            overrideActiveRitual(active = false)
+        }
+    }
+
+    @HandleEvent(onlyOnIsland = HUB, priority = HandleEvent.HIGH)
+    private fun onItemInHandChange(event: ItemInHandChangeEvent) {
```
hannibal002: This still misses the actual point: nothing sets the override when the player right clicks with the spade. holdingSpade only caches what is in hand, it never calls overrideActiveRitual(). The first spade use still produces no guess.

src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaApi.kt
```diff
@@ -63,21 +75,59 @@ object DianaApi {
...
+        val item = event.newItem.takeUnless { it == NeuInternalName.NONE } ?: return
```
hannibal002: Switching to an empty slot returns early instead of setting holdingSpade to false, so it stays true. Same when the item changes outside the hub, since this listener is hub only and nothing resets on world change.

src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaApi.kt
```diff
@@ -23,16 +26,25 @@ import net.minecraft.client.player.RemotePlayer
...
 object DianaApi {
 
+    var holdingSpade = false
+        private set
+
     private var spades = emptySet<NeuInternalName>()
 
+    private var ritualActiveOverride: Boolean? = null
+
     fun hasSpadeInHand() = InventoryUtils.itemInHandId in spades
```
hannibal002, line 32: what is the difference between this and the field holdingSpade? looks confusing. can we drop one of them?

src/main/java/at/hannibal2/skyhanni/features/event/diana/GriffinBurrowHelper.kt
```diff
@@ -565,11 +563,12 @@ object GriffinBurrowHelper {
     @HandleEvent(onlyOnIsland = IslandType.HUB)
     fun onBlockClick(event: BlockClickEvent) {
         if (!isEnabled()) return
+        if (!DianaApi.holdingSpade) return
```
hannibal002: why was this added?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaApi.kt
AverageUser125: The item click event right below this does that

src/main/java/at/hannibal2/skyhanni/features/event/diana/GriffinBurrowHelper.kt
AverageUser125: what do you mean `added`? I just moved the existing check like 3 lines up.

hannibal002: missed it. thats a useful change. keep it.

src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaApi.kt
```diff
@@ -63,21 +71,53 @@ object DianaApi {
...
+            overrideActiveRitual(active = false)
+        }
+    }
+
+    @HandleEvent(onlyOnIsland = IslandType.HUB, priority = HandleEvent.HIGHEST)
+    private fun onItemClick(event: ItemClickEvent) {
+        if (!hasSpadeInHand()) return
+        if (event.clickType != RIGHT_CLICK) return
+
+        if (ritualActiveOverride != false) {
```
hannibal002: Once the override is false, no click can set it back. If the ritual starts later, everything stays off until a burrow is dug by hand. Just always set it here, the message corrects it a moment later.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/event/diana/DianaApi.kt
AverageUser125: I didn't want it to be incorrect. even if for a split second.
And I don't understand your reasoning against it tbh.

hannibal002: The guard does not prevent what you want it to. With the override still null and no ritual running, the click sets it to true anyway and it stays wrong until the message arrives. That split second happens either way, on the first click.

What it does add is the opposite case: once it is false it stays false even after the ritual actually starts, because isRitualActive() reads the override before the mayor data. That one lasts until restart or until a burrow is dug by hand.

Nothing observable happens in that short window either. No guesses exist to render yet, no Diana item gets picked up in 200ms, and there are no slayer mobs in the hub. So it trades a wrong value nothing reads for one that can last hours.

hannibal002 (changes requested):
bump

hannibal002 (approved):

PR #6453, opened 2026-08-24
Fix: Wiki Keybind triggering on menu items

hannibal002:
## Changelog Fixes
+ Fixed the Wiki Keybind triggering on menu items. - hannibal2
    * This includes Close buttons, Go Back arrows, the SkyBlock Menu item, filler panes and empty slots.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/commands/WikiManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/commands/WikiManager.kt
@@ -11,13 +11,19 @@ import at.hannibal2.skyhanni.events.RepositoryReloadEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
 import at.hannibal2.skyhanni.utils.ChatUtils
 import at.hannibal2.skyhanni.utils.InventoryUtils
+import at.hannibal2.skyhanni.utils.ItemUtils
+import at.hannibal2.skyhanni.utils.ItemUtils.cleanName
 import at.hannibal2.skyhanni.utils.ItemUtils.getInternalName
+import at.hannibal2.skyhanni.utils.ItemUtils.getInternalNameOrNull
+import at.hannibal2.skyhanni.utils.ItemUtils.takeUnlessEmpty
 import at.hannibal2.skyhanni.utils.KeyboardManager.isKeyHeld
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/commands/WikiManager.kt
```diff
@@ -28,19 +34,40 @@ object WikiManager {
 
     val wiki get() = data.unofficial
 
+    /**
+     * REGEX-TEST: Close
+     * REGEX-TEST: Go Back
+     */
+    private val ignoredKeybindItemPattern by RepoPattern.pattern(
```
AverageUser125, line 41: This should probably be a pattern list.

src/main/java/at/hannibal2/skyhanni/features/commands/WikiManager.kt
```diff
@@ -28,19 +34,40 @@ object WikiManager {
...
         event.move(136, "commands.betterWiki.sbGuide", "commands.betterWiki.skyblockGuide")
     }
 
     @HandleEvent(onlyOnSkyblock = true)
     private fun onGuiKeyPress() {
-        val stack = stackUnderCursor() ?: return
-
         if (!config.wikiKeybind.isKeyHeld()) return
+
+        val stack = stackUnderCursor()?.takeUnlessEmpty() ?: return
```
AverageUser125, line 55: Maybe take the `isBlank` from `BetterContainers` and create a `takeUnlessBlank` or smth and use it here?

Though given you do the trim later on the item name to check if empty this is probably not needed.

AverageUser125 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/commands/WikiManager.kt
hannibal002: Thanks, but I'd keep the single pattern here: the entries are plain literals, and a repo edit can extend the alternation just as well as a list can.

src/main/java/at/hannibal2/skyhanni/features/commands/WikiManager.kt
hannibal002: `sBlankStack` is tied to black panes specifically, while the blank name check here covers filler items of any type, so I'd rather keep it as is.

PR #6458, opened 2026-08-26
Fix: Use submit-based world rendering on 26.1 + see-through text dimming fix

lunaynx:
## What
On 26.1, SkyHanni currently uses a mix of `MultiBufferSource` and `SubmitNodeCollector`-based rendering for `SkyHanniRenderWorldEvent`, but unconditionally submits at `AFTER_TRANSLUCENT_TERRAIN` – which, while not known to cause any practical issues today, is technically incorrect.

I took the opportunity to migrate all users of `SkyHanniRenderWorldEvent` to `COLLECT_SUBMITS` / `SubmitNodeCollector`-based rendering.

Some things like the picture-in-picture renderer still use `MultiBufferSource` on 26.1 because that's the only API vanilla exposes for it there.

Also, there's a vanilla bug [MC-298659](https://mojira.dev/MC-298659) with the submit-based text path (fixed in 26.3) that makes certain things such as water render over it. SkyHanni's 26.2 rendering already suffers from this bug today, and changing 26.1 to submit-based rendering would have introduced the bug there as well – so I made sure it doesn't happen on 26.1 while also fixing the bug on 26.2 along the way, but it needed two different ways to fix it:
* On 26.2, we use Fabric's new `SubmitRenderPhases` API to submit it in the `AFTER_TERRAIN` phase.
* On 26.1, we defer rendering to the old `MultiBufferSource`-based `LevelRenderEvents.AFTER_TRANSLUCENT_TERRAIN` because `SubmitRenderPhases` does not exist. This is essentially the same as today, just with correct injection points (no known user-facing issues this fixes today, but potentially prevents further issues in the future).

This refactor also happened to fix a rendering bug on 26.1 where lines such as the pathfinding line would just cut off whenever they pass a water boundary.

## Changelog Fixes
+ Fixed pathfinding lines cutting off when touching water on Minecraft 26.1. - Luna
+ Fixed text shown by SkyHanni in the world being dimmed by water and stained glass on Minecraft 26.2. - Luna
    * Example: Fishing and Lily Pad Minion name tags, depending on placement.

+ Changed SkyHanniRenderWorldEvent to fire on COLLECT_SUBMITS instead of AFTER_TRANSLUCENT_TERRAIN on Minecraft 26.1, and changed the remaining MultiBufferSource-based rendering in this event to be SubmitNodeCollector-based. - Luna
+ Unified LineDrawer to use vertex-based rendering on all versions. - Luna
+ Made submitOrderedText defer rendering to after terrain to work around MC-298659. - Luna
+ Fixed drawFilledBoundingBox using the wrong PoseStack, making custom ordering for text submissions no longer necessary. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/api/minecraftevents/ClientEvents.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/minecraftevents/ClientEvents.kt
@@ -34,6 +34,10 @@ import net.minecraft.resources.Identifier
 import net.minecraft.server.packs.PackType
 import java.util.concurrent.CompletableFuture
 
+//? if < 26.2 {
+/*import at.hannibal2.skyhanni.utils.render.WorldRenderUtils
+*///?}
+
 @SkyHanniModule
 object ClientEvents {
     var totalTicks = 0
@@ -67,6 +71,12 @@ object ClientEvents {
                 Minecraft.getInstance().deltaTracker.getGameTimeDeltaPartialTick(true),
             ).post()
         }
+
+        //? if < 26.2 {
+        /*LevelRenderEvents.AFTER_TRANSLUCENT_TERRAIN.register { ctx ->
+            WorldRenderUtils.drawQueuedSeeThroughText(ctx.bufferSource())
+        }
+        *///?}
 
         ResourceLoader.get(PackType.CLIENT_RESOURCES).registerReloadListener(
             Identifier.fromNamespaceAndPath("skyhanni", "resources"),
--- a/src/main/java/at/hannibal2/skyhanni/api/minecraftevents/ClientEvents.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/minecraftevents/ClientEvents.kt
@@ -35,7 +35,7 @@ import net.minecraft.server.packs.PackType
 import java.util.concurrent.CompletableFuture
 
 //? if < 26.2 {
-/*import at.hannibal2.skyhanni.utils.render.WorldRenderUtils
+/*import at.hannibal2.skyhanni.events.minecraft.SkyHanniRenderWorldEventLegacy
 *///?}
 
 @SkyHanniModule
@@ -74,7 +74,7 @@ object ClientEvents {
 
         //? if < 26.2 {
         /*LevelRenderEvents.AFTER_TRANSLUCENT_TERRAIN.register { ctx ->
-            WorldRenderUtils.drawQueuedSeeThroughText(ctx.bufferSource())
+            SkyHanniRenderWorldEventLegacy(ctx.bufferSource()).post()
         }
         *///?}
 
--- a/src/main/java/at/hannibal2/skyhanni/api/minecraftevents/ClientEvents.kt
+++ b/src/main/java/at/hannibal2/skyhanni/api/minecraftevents/ClientEvents.kt
@@ -59,14 +59,11 @@ object ClientEvents {
             WorldChangeEvent.post()
         }
 
-        //~ if < 26.2 'COLLECT_SUBMITS' -> 'AFTER_TRANSLUCENT_TERRAIN'
         LevelRenderEvents.COLLECT_SUBMITS.register { ctx ->
             SkyHanniRenderWorldEvent(
                 ctx.poseStack(),
                 ctx.levelState().cameraRenderState,
                 ctx.submitNodeCollector(),
-                //? if < 26.2
-                //ctx.bufferSource(),
                 Minecraft.getInstance().deltaTracker.getGameTimeDeltaPartialTick(true),
             ).post()
         }
--- a/src/main/java/at/hannibal2/skyhanni/utils/render/LineDrawer.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/render/LineDrawer.kt
@@ -22,13 +22,16 @@ class LineDrawer @PublishedApi internal constructor(
         val layer = SkyHanniRenderLayers.getLines(!depth)
         val lines = queuedLines.toList()
         event.submitCustomGeometry(layer) { pose, buf ->
+            fun QueuedLine.addVertexForPoint(point: LorenzVec) {
+                buf.addVertex(pose.pose(), point.x.toFloat(), point.y.toFloat(), point.z.toFloat())
+                    .setNormal(pose, normal.x.toFloat(), normal.y.toFloat(), normal.z.toFloat())
+                    .setColor(color.red, color.green, color.blue, color.alpha)
+                    .setLineWidth(lineWidth.toFloat())
+            }
+
             for (line in lines) {
-                for (point in arrayOf(line.p1, line.p2)) {
-                    buf.addVertex(pose.pose(), point.x.toFloat(), point.y.toFloat(), point.z.toFloat())
-                        .setNormal(pose, line.normal.x.toFloat(), line.normal.y.toFloat(), line.normal.z.toFloat())
-                        .setColor(line.color.red, line.color.green, line.color.blue, line.color.alpha)
-                        .setLineWidth(lineWidth.toFloat())
-                }
+                line.addVertexForPoint(line.p1)
+                line.addVertexForPoint(line.p2)
             }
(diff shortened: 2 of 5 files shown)


src/main/java/at/hannibal2/skyhanni/api/minecraftevents/ClientEvents.kt
```diff
@@ -59,18 +63,21 @@ object ClientEvents {
...
                 ctx.levelState().cameraRenderState,
                 ctx.submitNodeCollector(),
-                //? if < 26.2
-                //ctx.bufferSource(),
                 Minecraft.getInstance().deltaTracker.getGameTimeDeltaPartialTick(true),
             ).post()
         }
 
+        //? if < 26.2 {
+        /*LevelRenderEvents.AFTER_TRANSLUCENT_TERRAIN.register { ctx ->
```
hannibal002, line 76: Every other registration here posts a SkyHanni event and runs through the ErrorManager, this one doesn't. Intentional?

src/main/java/at/hannibal2/skyhanni/utils/render/WorldRenderUtils.kt
```diff
@@ -78,18 +107,54 @@ object WorldRenderUtils {
...
-        *///?}
-    }
+    ) = submitCustomGeometry(layer) { _, buf -> render(buf) }
+
+    // The render callback runs deferred, after the current pose is popped again.
+    // Anything needing the submit-time pose must use the callback's pose copy.
+    inline fun SkyHanniRenderWorldEvent.submitCustomGeometry(
+        layer: RenderType,
+        crossinline render: (PoseStack.Pose, VertexConsumer) -> Unit,
+    ) = submitNodeCollector.submitCustomGeometry(matrices, layer) { pose, buf -> render(pose, buf) }
```
hannibal002, line 161: drawFilledBoundingBox still reads the live matrices inside the lambda. why?

src/main/java/at/hannibal2/skyhanni/utils/render/WorldRenderUtils.kt
```diff
@@ -78,18 +107,54 @@ object WorldRenderUtils {
...
+                entry.color,
+                entry.shadow,
+                entry.pose,
+                bufferSource,
+                DisplayMode.SEE_THROUGH,
+                entry.backgroundColor,
+                entry.light,
+            )
+        }
+        queuedSeeThroughText.clear()
```
hannibal002: Could the clear() move into a finally?

src/main/java/at/hannibal2/skyhanni/utils/render/WorldRenderUtils.kt
```diff
@@ -78,18 +107,54 @@ object WorldRenderUtils {
             outlineColor,
         )
     }
-    //?}
+
+    //? if < 26.2 {
+    /*private val queuedSeeThroughText = mutableListOf<SeeThroughText>()
+
+    private class SeeThroughText(
```
hannibal002: data class?

src/main/java/at/hannibal2/skyhanni/utils/render/WorldRenderUtils.kt
```diff
@@ -64,7 +65,35 @@ object WorldRenderUtils {
...
+                    SEE_THROUGH,
+                    light,
+                    color,
+                    backgroundColor,
+                    outlineColor,
+                ),
+            )
+            //?} else {
+            /*queuedSeeThroughText.add(
+                SeeThroughText(Matrix4f(matrices.last().pose()), x, y, text, shadow, light, color, backgroundColor),
```
hannibal002, line 91: outlineColor is dropped here, so it silently does nothing on 26.1.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/api/minecraftevents/ClientEvents.kt
lunaynx: Good point, will add a new event for it.

src/main/java/at/hannibal2/skyhanni/utils/render/WorldRenderUtils.kt
lunaynx: Only LineDrawer needed the fix because lines didn't render at all otherwise, but I'll take a look if these can also use the pose from the callback.

lunaynx: Okay, that was actually a good catch. Fixing this made the whole `order(1)` thing for text submissions no longer necessary.

src/main/java/at/hannibal2/skyhanni/utils/render/WorldRenderUtils.kt
lunaynx: Good catch. `Font.drawInBatch` (26.1) doesn't support specifying an outline color. There is a `drawInBatch8xOutline` method, but it seems to behave differently (for one, it doesn't support rendering see-through text). Since `outlineColor` is always unconditionally passed as 0 right now (it was added in the 26.2 PR), I decided to just remove it for now to avoid confusion. We can reevaluate it if we actually need text outline color functionality in the future (hopefully 26.1 will be dead by then 🙏).

src/main/java/at/hannibal2/skyhanni/utils/render/WorldRenderUtils.kt
```diff
@@ -946,7 +987,7 @@ object WorldRenderUtils {
         exactLocation(player).up(player.getEyeHeight(player.pose))
 
     private fun addChainedFilledBoxVertices(
-        matrices: PoseStack,
+        matrix4f: PoseStack.Pose,
```
hannibal002: confusing name. pls rename or add comment explaining

src/main/java/at/hannibal2/skyhanni/utils/render/WorldRenderUtils.kt
```diff
@@ -75,21 +104,61 @@ object WorldRenderUtils {
...
+                    entry.light,
+                )
+            }
+        } finally {
+            queuedSeeThroughText.clear()
+        }
+    }
+    *///?}
 
     inline fun SkyHanniRenderWorldEvent.submitCustomGeometry(
```
hannibal002, line 151: those two inline functions should have an explicit return type

src/main/java/at/hannibal2/skyhanni/utils/render/LineDrawer.kt
```diff
@@ -4,47 +4,33 @@ import at.hannibal2.skyhanni.events.minecraft.SkyHanniRenderWorldEvent
...
-
-                buf.addVertex(matrix.pose(), line.p2.x.toFloat(), line.p2.y.toFloat(), line.p2.z.toFloat())
-                    .setNormal(matrix, line.normal.x.toFloat(), line.normal.y.toFloat(), line.normal.z.toFloat())
-                    .setColor(line.color.red, line.color.green, line.color.blue, line.color.alpha)
-                    .setLineWidth(lineWidth.toFloat())
+        val layer = SkyHanniRenderLayers.getLines(!depth)
+        val lines = queuedLines.toList()
+        event.submitCustomGeometry(layer) { pose, buf ->
+            for (line in lines) {
+                for (point in arrayOf(line.p1, line.p2)) {
```
hannibal002: array of in a hot path looks non optimized. how about using a function in a function instead? or any other idea how we can avoid quadruplions of allocations per frame? not a big issue, just a small nit.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/render/WorldRenderUtils.kt
lunaynx: They return `Unit`, do you still want an explicit return type? We don't usually do that.

hannibal002 (approved):

PR #6461, opened 2026-08-27
Fix: /shupdate stream argument never matching

3d3n-pyc:
## What
Fixed `/shupdate release` and `/shupdate beta` not switching the update stream.
- `String.equals(Regex)` always returns `false`, replaced with `String.matches(Regex)`.
- Moved regex patterns to class-level properties to avoid recompilation on every call.
- `/shupdate release` now persists the stream change in the config. Previously, only switching to BETA was saved.

## Changelog Fixes
+ Fixed /shupdate release and /shupdate beta not switching the update stream. - 3d3n-pyc

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
@@ -245,7 +245,7 @@ object UpdateManager {
     private val betaStreamPattern = "(?i)(?:beta|latest)s?".toRegex()
 
     private fun updateCommand(arg: String) {
-        val currentStream = SkyHanniMod.feature.about.updateStream.get()
+        val currentStream = config.updateStream.get()
         val updateStream = when {
             arg.matches(releaseStreamPattern) -> UpdateStream.RELEASES
             arg.matches(betaStreamPattern) -> UpdateStream.BETA
@@ -257,15 +257,15 @@ object UpdateManager {
             ChatUtils.clickableChat(
                 "Are you sure you want to switch to beta? These versions may be less stable.",
                 onClick = {
-                    SkyHanniMod.feature.about.updateStream.set(UpdateStream.BETA)
-                    checkUpdate(true, UpdateStream.BETA)
+                    config.updateStream.set(updateStream)
+                    checkUpdate(true, updateStream)
                 },
                 "§eClick to confirm!",
                 oneTimeClick = true,
             )
         } else {
             if (updateStream != currentStream) {
-                SkyHanniMod.feature.about.updateStream.set(updateStream)
+                config.updateStream.set(updateStream)
             }
             checkUpdate(true, updateStream)
         }
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
@@ -124,11 +124,6 @@ object UpdateManager {
 
         logger.log("Starting update check (source: ${context.source.javaClass.simpleName}")
 
-        val currentStream = config.updateStream.get()
-        if (forcedUpdateStream == UpdateStream.BETA && currentStream != UpdateStream.BETA) {
-            config.updateStream.set(UpdateStream.BETA)
-        }
-
         activePromise = context.checkUpdate(forcedUpdateStream.stream)
             .orTimeout(15, TimeUnit.SECONDS)
             .whenCompleteAsync(
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
@@ -241,11 +241,14 @@ object UpdateManager {
 
     private var potentialUpdate: PotentialUpdate? = null
 
+    private val releaseStreamPattern = "(?i)(?:full|release)s?".toRegex()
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
```diff
@@ -244,8 +244,8 @@ object UpdateManager {
     private fun updateCommand(arg: String) {
         val currentStream = SkyHanniMod.feature.about.updateStream.get()
         val updateStream = when {
-            arg.equals("(?i)(?:full|release)s?".toRegex()) -> UpdateStream.RELEASES
-            arg.equals("(?i)(?:beta|latest)s?".toRegex()) -> UpdateStream.BETA
+            arg.matches("(?i)(?:full|release)s?".toRegex()) -> UpdateStream.RELEASES
```
hannibal002: regex stuff belongs in the repo pattern structure. or here at least outside the function. so that the regex is not build every time the fun gets called

hannibal002: `checkUpdate` only ever persists `BETA`, so `/shupdate release` still leaves the update stream on beta.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
3d3n-pyc: Fixed both! Moved the regex to class-level properties and also noticed that `/shupdate release` never actually persisted the stream change, so I added that too.

src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
```diff
@@ -254,14 +257,16 @@ object UpdateManager {
...
-                    newUpdateStream.set(UpdateStream.BETA)
-                    checkUpdate(true, newUpdateStream.get())
+                    SkyHanniMod.feature.about.updateStream.set(UpdateStream.BETA)
+                    checkUpdate(true, UpdateStream.BETA)
                 },
                 "§eClick to confirm!",
                 oneTimeClick = true,
             )
         } else {
+            if (updateStream != currentStream) {
```
hannibal002, line 264: why does this guard not exist in the `if (switchingToBeta) {` case above?

src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
```diff
@@ -254,14 +257,16 @@ object UpdateManager {
             ChatUtils.clickableChat(
                 "Are you sure you want to switch to beta? These versions may be less stable.",
                 onClick = {
-                    val newUpdateStream = SkyHanniMod.feature.about.updateStream
-                    newUpdateStream.set(UpdateStream.BETA)
-                    checkUpdate(true, newUpdateStream.get())
+                    SkyHanniMod.feature.about.updateStream.set(UpdateStream.BETA)
```
hannibal002: use config.updateStream instead

src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
```diff
@@ -254,14 +257,16 @@ object UpdateManager {
...
+                    checkUpdate(true, UpdateStream.BETA)
                 },
                 "§eClick to confirm!",
                 oneTimeClick = true,
             )
         } else {
+            if (updateStream != currentStream) {
+                SkyHanniMod.feature.about.updateStream.set(updateStream)
```
hannibal002: use config.updateStream instead

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
3d3n-pyc: The guard isn't needed there since `switchingToBeta` already implies `updateStream != currentStream`. It's only useful in the `else` branch where `/shupdate` with no argument keeps the same stream.

hannibal002: `switchingToBeta` is also true when the stream is already beta but the installed version is not, so it does not imply that.

src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
```diff
@@ -254,14 +257,16 @@ object UpdateManager {
...
-                    newUpdateStream.set(UpdateStream.BETA)
-                    checkUpdate(true, newUpdateStream.get())
+                    config.updateStream.set(updateStream)
+                    checkUpdate(true, updateStream)
                 },
                 "§eClick to confirm!",
                 oneTimeClick = true,
             )
         } else {
+            if (updateStream != currentStream) {
```
hannibal002, line 264: the `BETA` branch in `checkUpdate` can never run now that the persisting sits here, please remove it.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/update/UpdateManager.kt
3d3n-pyc: Right, my bad. I added the guard in both branches.

hannibal002 (approved):

PR #6462, opened 2026-08-27
Fix: Player Rank Hider not hiding ranks for players with an emblem

3d3n-pyc:
## What
Fixed `Player Rank Hider` not hiding ranks for other players in chat.
- Updated `globalPattern` in `PlayerChatManager` to support colored SkyBlock emblems before player names.
- Removed outdated `chatColor.length == 0` check in `isGlobalChat` that caused other players' chat messages to be rejected.
- Fixed `String.replace` calls in `PlayerChatModifier` using literal string replacement instead of `.toRegex()`.

## Changelog Fixes
+ Fixed Player Rank Hider not hiding ranks for players with a SkyBlock emblem in front of their name. - 3d3n-pyc

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/hypixel/chat/PlayerChatManager.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/hypixel/chat/PlayerChatManager.kt
@@ -35,10 +35,14 @@ object PlayerChatManager {
      * REGEX-TEST: [58] §7nea89o§7: haiiiii
      * REGEX-TEST: [266] ♫ §b[MVP§d+§b] lrg89§f: a
      * REGEX-TEST: [302] ♫ [MVP+] lrg89: problematic
+     * REGEX-TEST: [233] §6✿ §a[VIP] dawnbound§f: thats not mine too
+     * REGEX-TEST: [323] §b[MVP§2+§b] xatarna§f: can somebody get these beggars out of here
+     * REGEX-FAIL: [BOSS] Bonzo: Show time!
+     * REGEX-FAIL: Finding player...
      */
     private val globalPattern by patternGroup.pattern(
         "global",
-        "^(?:\\[(?<level>\\d+)] )?(?<author>(?:(?:§.)*[^ ]+ )?(?:(?:§.)*\\[[^\\]]+\\] )?[^ ]+?)(?<chatColor>§f|§7|): (?<message>.*)\$",
+        "^(?:\\[(?<level>\\d+)] )?(?<author>(?:(?:§.)*[^ ] )?(?:(?:§.)*\\[[^\\]]+\\] )?[^ ]+?)(?<chatColor>§f|§7|): (?<message>.*)\$",
     )
 
     /**
@@ -229,6 +233,10 @@ object PlayerChatManager {
     private fun ComponentMatcher.isGlobalChat(event: SkyHanniChatEvent.Allow): Boolean {
         var author = groupOrThrow("author")
         val chatColor = groupOrThrow("chatColor")
+        if (chatColor.length == 0 && !author.getText().removeColor().endsWith(PlayerUtils.getName())) {
+            // The last format string is always present, unless this is the players own message
+            return false
+        }
         val message = groupOrThrow("message").removePrefix("§f")
         if (author.getText().contains("[NPC]")) {
             NpcChatEvent.Allow(author, message, event.chatComponent).postChat(event)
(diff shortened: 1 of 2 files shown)


hannibal002 (changes requested):
A few things:

- The removed `chatColor` check was the only thing keeping system messages out of the player chat path. Without it every `<up to 3 tokens>: <text>` message is posted as `PlayerAllChatEvent`, so `SystemMessageEvent` never fires for it and `PlayerNameFormatter` rewrites the component. What separates the two now?
- `globalPattern`: why `[^ ]+` instead of `[^ ]`? The emblem is a single character, so `(?:(?:§.)*[^ ] )?` would be enough and a lot narrower.
- The two `replace` calls in `PlayerChatModifier` were literal, so they never ran. Switching them to `Regex` activates them for the first time. `colorFollowPattern` matches any `§[7ab6]<word> §<color>` in any system message, which is not always a player name. Which messages did you test against?
- `(?:\w+){2,16}` has no upper bound, it is effectively `\w{2,}`. Did you mean `(\w{2,16})`, like the two patterns above it?
- `toRegex()` is not allowed, please use `RepoPattern` (see CONTRIBUTING.md).
- `globalPattern` needs a `REGEX-TEST` for the colored emblem case, and a `REGEX-FAIL` for a system message would be good to pin down the boundary.

src/main/java/at/hannibal2/skyhanni/data/hypixel/chat/PlayerChatManager.kt
```diff
@@ -35,10 +35,14 @@ object PlayerChatManager {
      * REGEX-TEST: [58] §7nea89o§7: haiiiii
      * REGEX-TEST: [266] ♫ §b[MVP§d+§b] lrg89§f: a
      * REGEX-TEST: [302] ♫ [MVP+] lrg89: problematic
+     * REGEX-TEST: [233] §6✿ §a[VIP] dawnbound§f: thats not mine too
+     * REGEX-TEST: [323] §b[MVP§2+§b] xatarna§f: can somebody get these beggars out of here
+     * REGEX-FAIL: [BOSS] Bonzo: Show time!
```
hannibal002: this test does not fail and this is wrong it should not match.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/hypixel/chat/PlayerChatManager.kt
3d3n-pyc: lemme see

3d3n-pyc: The regex had a trailing `|` in `(?<chatColor>§f|§7|)`, so it was matching empty color codes like in `[BOSS] Bonzo:`.
Actual Hypixel chat messages always have a color code (`§f` or `§7`) before the colon, while that old `problematic` test case was unformatted.
Changed to `(?<chatColor>§[f7])` and removed that unformatted test case, now `[BOSS] Bonzo:` properly fails the regex and CI passes.

Bloxigus: this is not true, stop using claude for this

src/main/java/at/hannibal2/skyhanni/data/hypixel/chat/PlayerChatManager.kt
```diff
@@ -34,11 +34,14 @@ object PlayerChatManager {
     /**
      * REGEX-TEST: [58] §7nea89o§7: haiiiii
      * REGEX-TEST: [266] ♫ §b[MVP§d+§b] lrg89§f: a
-     * REGEX-TEST: [302] ♫ [MVP+] lrg89: problematic
```
hannibal002, line 37: Why did you delete this test? It was correct.

src/main/java/at/hannibal2/skyhanni/data/hypixel/chat/PlayerChatManager.kt
```diff
@@ -34,11 +34,14 @@ object PlayerChatManager {
...
+     * REGEX-TEST: [233] §6✿ §a[VIP] dawnbound§f: thats not mine too
+     * REGEX-TEST: [323] §b[MVP§2+§b] xatarna§f: can somebody get these beggars out of here
+     * REGEX-FAIL: [BOSS] Bonzo: Show time!
+     * REGEX-FAIL: Finding player...
      */
     private val globalPattern by patternGroup.pattern(
```
hannibal002, line 42: Making the color code mandatory breaks your own messages, those have no `§f` or `§7`. That is exactly what the `chatColor.length == 0` check below is for, and it is dead code now.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/data/hypixel/chat/PlayerChatManager.kt
3d3n-pyc: Restored the test, my bad.

src/main/java/at/hannibal2/skyhanni/data/hypixel/chat/PlayerChatManager.kt
3d3n-pyc: Makes sense, I misunderstood what that unformatted case was for. Restored `§f|§7|` and dropped the `[BOSS] regex-fail` so `isGlobalChat` handles it as intended.

hannibal002 (approved):

PR #6469, opened 2026-08-29
Backend: Fix off-by-one loop bound when reading item boosters in Estimated Item Value

3d3n-pyc:
## What
Fixed an off-by-one loop bound in `EstimatedItemValueCalculator.readBoosters()` where `0..list.size` was used instead of `0 until list.size`.

The out-of-range read was already handled safely at runtime by `ListTag.getNullable()` returning null and `getStringOrDefault` returning an empty string skipped via `isBlank()`, so this is an internal cleanup.

+ Fixed an off-by-one loop bound when reading item boosters in Estimated Item Value. - 3d3n-pyc

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/items/EstimatedItemValueCalculator.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/items/EstimatedItemValueCalculator.kt
@@ -932,7 +932,7 @@ object EstimatedItemValueCalculator {
         val list = NbtCompat.getStringTagList(extraAttributes, "boosters")
         if (list.isEmpty) return emptyList()
         val boosters = mutableListOf<NeuInternalName>()
-        for (i in 0..list.size) {
+        for (i in 0 until list.size) {
             var internalName = list.getStringOrDefault(i)
             if (internalName.isBlank()) continue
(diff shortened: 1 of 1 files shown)


hannibal002 (changes requested):
Notes on the description, the code change itself is good:

- This is better suited for `## Changelog Technical Details` than for `## Changelog Fixes`. The extra iteration never did any harm: `ListTag.getNullable()` checks the index and returns null when it is out of range, so the loop got an empty string and skipped it via `isBlank()`. Nothing was ever wrong for the user, so it counts asan internal change.
- For the same reason "error" is a bit strong. In SkyHanni an error means the red chat message from ErrorManager (see the Errors and Crashes section in CONTRIBUTING.md), and that never showed up here.
- "out of bounds" describes the error, so it should be hyphenated: "out-of-bounds".
- "calculator" is the name of the class, users know the feature as just "Estimated Item Value". Relevant if this were a user facing error. It is not, though.

Suggested entry:

+ Fixed an off-by-one loop bound when reading item boosters in Estimated Item Value. - 3d3n-pyc

It would also help to mention in the What section that the out-of-range read was already safe. That makes it clear right away why this is not a user facing fix.

hannibal002 (approved):

PR #6470, opened 2026-08-29
Fix: Correct distance calculation in BeachBallCatchHelper target selection

3d3n-pyc:
## What
Fixed the distance calculation in `BeachBallCatchHelper` when choosing the prediction closest to the average target coordinates.

Previously, `xTarget - last.x + zTarget - last.z` was used without absolute values or squares, causing points further in the positive direction to be incorrectly selected by `.minBy`. Changed to using `LorenzVec.distanceSqIgnoreY`.

## Changelog Fixes
+ Fixed Bouncy Ball Line pointing to wrong landing spots. - 3d3n-pyc

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/event/yearoftheseal/BeachBallCatchHelper.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/event/yearoftheseal/BeachBallCatchHelper.kt
@@ -218,9 +218,10 @@ object BeachBallCatchHelper {
             val xTarget = targets.mapKeys { it.key.x }.weightedAverage()
             val zTarget = targets.mapKeys { it.key.z }.weightedAverage()
 
+            val averageTarget = LorenzVec(xTarget, 0.0, zTarget)
             val target = predictions.minBy {
                 val last = it.key.last()
-                xTarget - last.x + zTarget - last.z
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/event/yearoftheseal/BeachBallCatchHelper.kt
```diff
@@ -220,7 +220,9 @@ object BeachBallCatchHelper {
 
             val target = predictions.minBy {
                 val last = it.key.last()
-                xTarget - last.x + zTarget - last.z
+                val dx = xTarget - last.x
+                val dz = zTarget - last.z
+                dx * dx + dz * dz
```
hannibal002: use the existing utils distanceSqIgnoreY instead

hannibal002 (changes requested):
changelog should name what the player sees (Bouncy Ball Line) and the symptom, not the class and the calculation

hannibal002 (changes requested):
way better, thanks.

small last thing in the changelog:
drop "due to inverted distance calculation", that belongs in What. it also wasn't inverted, it picked the largest x + z

hannibal002 (approved):

PR #6471, opened 2026-08-29
Fix: Inverted emissary scaling discount calculation in DiscountUtils

3d3n-pyc:
## What
Fixed an inverted calculation in `DiscountUtils.getEmissaryDiscountedPrice()`.

Previously, `itemDiscount` was initialized to `1.0` and set to `(1.0 - discount)` when qualifying for a reputation discount. Because `priceDecrease` was then computed as `coinPrice * itemDiscount` and subtracted from `lowestNPCPrice`, a player with 0 reputation (0% discount) would have 100% of the coin price subtracted (making items free), while a 5% discount would subtract 95% of the coin price.

Changed to tracking `discountFraction` (default `0.0`) and subtracting `coinPrice * discountFraction`.

## Changelog Fixes
+ Fixed inverted emissary reputation discount calculation in instance chest profit. - 3d3n-pyc

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/DiscountUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/DiscountUtils.kt
@@ -48,11 +48,11 @@ object DiscountUtils {
 
     private fun NeuInternalName.getEmissaryDiscountedPrice(lowestNPCPrice: Double): Double {
         val rep = ProfileStorageData.profileSpecific?.crimsonIsle?.reputation?.maxBy { it.value }?.value ?: 0
-        var itemDiscount = 1.0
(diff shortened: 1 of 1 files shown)


hannibal002 (changes requested):
The repo values are percentages, not fractions:

`"Emissary": { "1000": 5, "3000": 10, "7000": 15, "12000": 20 }`

So this needs `discount / 100.0`. As is, above 1000 reputation the price goes negative and the key shows up as revenue instead of cost.

The changelog entry describes the code instead of what the player sees, and feature names are capitalized. Also the cost was wrong in both directions, not just too low:

+ Fixed wrong Kuudra Key cost in Instance Chest Profit. - 3d3n-pyc

PR #6472, opened 2026-08-29
Fix: /shnavigate freezing the client

hannibal002:
## Changelog Fixes
+ Fixed the client freezing for a moment when running /shnavigate. - hannibal2
+ Fixed /shnavigate not stopping the navigation when clicking to return to the previous search. - hannibal2

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigationHelper.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigationHelper.kt
@@ -41,17 +41,24 @@ object NavigationHelper {
         GraphNodeTag.CRIMSON_MINIBOSS,
     )
 
-    private fun doCommandAsync(searchTerm: String) {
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/features/misc/pathfind/NavigationHelper.kt
```diff
@@ -41,17 +41,24 @@ object NavigationHelper {
         GraphNodeTag.CRIMSON_MINIBOSS,
     )
 
-    private fun doCommandAsync(searchTerm: String) {
+    private fun doCommandAsync(searchTerm: String, allowInstant: Boolean = true) {
+        SkyHanniMod.launchCoroutine("shnavigate command") {
```
lunaynx: While you're changing this, `SkyHanniMod.launchCoroutine` is deprecated, should be changed to use `CoroutineSettings` instead.

hannibal002: good catch!

PR #6473, opened 2026-08-29
Fix: Rounding calculation and fractional formatting in LimboPlaytime

3d3n-pyc:
## What
Fixed an edge-case rounding bug in `LimboPlaytime` and simplified hours calculation.

Previously, `wholeHours` (integer division) and `minutes` (fractional portion rounded to 1 decimal) were calculated separately. When `minutes` rounded up to `1.0` (e.g. 119 minutes -> `1h` and `59/60 ≈ 1.0`), `findFloatDecimalPlace(1.0f)` returned `0`, leaving the display at `1 hours` instead of `2 hours` (1 full hour too low).

Now computes `(wholeMinutes / 60.0).roundTo(1)` as a continuous float and formats decimals with an explicit dot so `addLimbo` can parse it across all locales, removing the `findFloatDecimalPlace` helper.

## Changelog Fixes
+ Fixed Limbo playtime in Detailed Playtime showing one hour less than the actual time when close to a full hour. - 3d3n-pyc


src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
```diff
@@ -116,18 +116,9 @@ object LimboPlaytime {
...
-        } else {
-            val minutes: Float = ((wholeMinutes - wholeHours * 60).toFloat() / 60).roundTo(1)
-            hoursString = wholeHours.addSeparators()
-            if (findFloatDecimalPlace(minutes) != 0) {
-                val minutesString = minutes.toString()
-                hoursString += minutesString.substring(minutesString.indexOf("."))
-            }
-        }
+        val hours = (wholeMinutes / 60.0).roundTo(1)
+        hoursString = if (hours % 1.0 == 0.0) hours.toLong().addSeparators() else hours.addSeparators()
```
hannibal002: can not use addSeparators here: the function uses user's time zone to define the foramtting, either 0.0 or 0,0. so the string parser in addLimbo will fail to read the values correctly. 12,3 will become 123, etc

hannibal002 (changes requested):
- The old code never appends the `.0` here, `findFloatDecimalPlace(1.0f)` returns 0. It shows `1 hours`, so the display was a full hour too low.

- The changelog line describes the code, not the user facing observation.
the rule is "Name the behavior a user can observe, not the code that produces it.".

src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
3d3n-pyc: Ah right, good catch with the locale decimal comma. Formatted it with an explicit `.` now and kept `addSeparators` on the whole hours only. Also updated the changelog line.

src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
```diff
@@ -116,18 +117,11 @@ object LimboPlaytime {
...
-            hoursString = "$wholeHours"
-        } else {
-            val minutes: Float = ((wholeMinutes - wholeHours * 60).toFloat() / 60).roundTo(1)
-            hoursString = wholeHours.addSeparators()
-            if (findFloatDecimalPlace(minutes) != 0) {
-                val minutesString = minutes.toString()
-                hoursString += minutesString.substring(minutesString.indexOf("."))
-            }
-        }
+        val hours = (wholeMinutes / 60.0).roundTo(1)
```
hannibal002: you could do it even more compact: Calculate directly in tenths of an hour. That way, you won't need to round the difference a second time.

src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
```diff
@@ -116,18 +117,11 @@ object LimboPlaytime {
...
+        val whole = hours.toLong()
+        val decimal = ((hours - whole) * 10).roundToInt()
+        hoursString = if (decimal == 0) whole.addSeparators() else "${whole.addSeparators()}.$decimal"
```
hannibal002: addSeparators here still causes wrong parsing with hoursPattern in fun addLimbo

hannibal002 (approved):

hannibal002 (changes requested):
missclick

src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
```diff
@@ -116,18 +115,11 @@ object LimboPlaytime {
...
-            hoursString = "$wholeHours"
-        } else {
-            val minutes: Float = ((wholeMinutes - wholeHours * 60).toFloat() / 60).roundTo(1)
-            hoursString = wholeHours.addSeparators()
-            if (findFloatDecimalPlace(minutes) != 0) {
-                val minutesString = minutes.toString()
-                hoursString += minutesString.substring(minutesString.indexOf("."))
-            }
-        }
+        val tenths = (wholeMinutes + 3) / 6
```
hannibal002: add a comment for this line, or extract into a functin explaining what this does

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/misc/limbo/LimboPlaytime.kt
```diff
@@ -116,10 +116,15 @@ object LimboPlaytime {
...
         wholeMinutes = playtime.inWholeMinutes.toInt()
-        val tenths = (wholeMinutes + 3) / 6
-        val whole = tenths / 10
-        val decimal = tenths % 10
-        hoursString = if (decimal == 0) "$whole" else "$whole.$decimal"
+        hoursString = formatMinutesAsHours(wholeMinutes)
+    }
+
+    private fun formatMinutesAsHours(minutes: Int): String {
+        // +3 is half the divisor, so the integer division rounds instead of truncating
```
3d3n-pyc: Is this comment redundant?

hannibal002 (approved):

PR #6474, opened 2026-08-29
Fix: Null safety and error handling on Mining Event API response

3d3n-pyc:
## What
Fixed potential `NullPointerException` and missing error handling in `MiningEventTracker.fetchData()` when the Mining Event API returns an error payload.

- Made `data` and `cause` nullable in `MiningEventDataReceive` to accurately represent failure payloads from Gson.
- Added missing `return` when `!miningEventData.success`, incrementing `apiErrorCount` and applying the 20-minute error cooldown.
- Added `logErrorStateWithData` and error cooldown if `data` is unexpectedly null despite `success = true`.

## Changelog Fixes
+ Fixed an error message when the Mining Event Tracker could not load data. - 3d3n-pyc

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/features/mining/eventtracker/MiningEventTracker.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/mining/eventtracker/MiningEventTracker.kt
@@ -211,6 +211,8 @@ object MiningEventTracker {
         val miningEventData = ConfigManager.gson.fromJson<MiningEventDataReceive>(receivedData)
 
         if (!miningEventData.success) {
+            apiErrorCount++
+            canRequestAt = SimpleTimeMark.now() + 20.minutes
             if (receivedData.toString().trim() == "{}") ChatUtils.chat(
                 "§cFailed loading Mining Event data!\n" +
                     "§cPlease wait until the server-problem fixes itself! There is nothing else to do at the moment.",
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/features/mining/eventtracker/MiningEventTracker.kt
```diff
@@ -221,11 +223,20 @@ object MiningEventTracker {
                 "cause" to miningEventData.cause,
                 "receivedData" to receivedData,
             )
+            return
         }
 
+        val data = miningEventData.data ?: run {
```
hannibal002, line 229: set apiErrorCount and canRequestAt again here as well

src/main/java/at/hannibal2/skyhanni/features/mining/eventtracker/MiningEventTracker.kt
```diff
@@ -221,11 +223,20 @@ object MiningEventTracker {
...
+            ErrorManager.logErrorWithData(
+                Exception("DataMissingInSuccessResponse"),
```
hannibal002: do not use an exception here. just use logErrorStateWithData with a explicit user facing message.

hannibal002 (changes requested):
The changelog entry describes the code change instead of what the player notices, and it overstates the severity.

`fetchData()` runs inside `launchIOCoroutine`, and `SkyHanniCoroutineManager.runWithErrorHandling` catches every `Throwable` and hands it to `ErrorManager`. The NPE therefore surfaces as a red chat message with a copy link, and the client keeps running. CONTRIBUTING.md (Errors and Crashes) asks to keep the two apart, since calling an error a crash overstates how bad it is.

"potential" undersells it in the other direction: as soon as the API answers with `success: false`, the NPE is guaranteed, and the existing `{}` branch shows that this does happen in practice. Players got the friendly "please wait until the server problem fixes itself" message and a red error right after it.

What the entry should name is that second message disappearing, roughly:
+ Fixed an error message when the Mining Event Tracker could not load data. - 3d3n-pyc

hannibal002 (approved):

PR #6475, opened 2026-08-30
Backend: Discount Utils discounting into negatives

Fazfoxy:
## What
Oops, I messed up discount utils in 2 big ways apparently.
 I'm pushing this fix cause https://github.com/hannibal002/SkyHanni/pull/6471 was missing some other misc things around it
Also updated the Todo to, actually, name the explicit talismans that should be implemented.

+ Fixed the Emissary discount calculation in `DiscountUtils`. - Fazfoxy

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/DiscountUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/DiscountUtils.kt
@@ -48,7 +48,7 @@ object DiscountUtils {
 
     private fun NeuInternalName.getEmissaryDiscountedPrice(lowestNPCPrice: Double): Double {
         val rep = ProfileStorageData.profileSpecific?.crimsonIsle?.reputation?.maxBy { it.value }?.value ?: 0
-        var itemDiscount = 1.0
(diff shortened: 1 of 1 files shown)


hannibal002 (changes requested):
the comment in discord somewhere

hannibal002 (changes requested):
It has no user facing effect though. NEU has no NPC shop recipe for the Kuudra keys, so `getDiscountedPrice` returns at the `?: return 0.0` above and this function never runs. `/shdebugprice KUUDRA_TIER_KEY` gives null for both `getNpcPriceOrNull` and `getRawCraftCostOrNull`.

That makes it Technical Details rather than a Fix:

```
+ Fixed the Emissary reputation discount calculation in DiscountUtils. - Fazfoxy
```

Still worth merging, and it should land before the key cost itself gets fixed, otherwise that PR turns the inverted math on for everyone at once.

Also worth naming #6471 in the What, they found the defect first.

The real bug is that the key shows as 0 in the overlay. Separate PR.

hannibal002 (approved):

PR #6480, opened 2026-08-30
Feature: Pickaxe Ability Ready Alert

3d3n-pyc:
## What
Adds an alert (title + sound) when a Pickaxe Ability comes off cooldown.

It reads the `Pickaxe Ability:` Tab widget, so the timing stays accurate even when the lobby TPS is low, and also picks up the chat message so the alert is instant when there's no lag.

## Changelog New Features
+ Added a notification when your Pickaxe Ability comes off cooldown. - 3d3n-pyc

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/mining/MiningNotificationsConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/mining/MiningNotificationsConfig.kt
@@ -25,7 +25,7 @@ class MiningNotificationsConfig {
         MiningNotificationList.COLD,
         MiningNotificationList.GOLDEN_GOBLIN,
         MiningNotificationList.DIAMOND_GOBLIN,
-        MiningNotificationList.PICKAXE_ABILITY
+        MiningNotificationList.PICKAXE_ABILITY,
     )
 
     @Expose
--- a/src/main/java/at/hannibal2/skyhanni/config/features/mining/MiningNotificationsConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/mining/MiningNotificationsConfig.kt
@@ -24,7 +24,8 @@ class MiningNotificationsConfig {
         MiningNotificationList.SCRAP,
         MiningNotificationList.COLD,
         MiningNotificationList.GOLDEN_GOBLIN,
-        MiningNotificationList.DIAMOND_GOBLIN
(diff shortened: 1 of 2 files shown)


lunaynx (changes requested):
Newly added code should avoid using color codes to match messages where possible. Drop the color codes in the regex and use `cleanMessage`.

You could probably also clean up the other uses in the same file while you're here, just make sure to append `.colorless` to the pattern key for anything that already existed before (the brand new pattern doesn't need it).

src/main/java/at/hannibal2/skyhanni/config/features/mining/MiningNotificationsConfig.kt
```diff
@@ -24,7 +24,8 @@ class MiningNotificationsConfig {
         MiningNotificationList.SCRAP,
         MiningNotificationList.COLD,
         MiningNotificationList.GOLDEN_GOBLIN,
-        MiningNotificationList.DIAMOND_GOBLIN
+        MiningNotificationList.DIAMOND_GOBLIN,
+        MiningNotificationList.PICKAXE_ABILITY
```
CalMWolfs: Missing trailing comma

src/main/java/at/hannibal2/skyhanni/features/mining/MiningNotifications.kt
```diff
@@ -31,7 +31,8 @@ object MiningNotifications {
         SCRAP("§9Suspicious Scrap", "§9Suspicious Scrap"),
         GOLDEN_GOBLIN("§6Golden Goblin", "§6Golden Goblin"),
         DIAMOND_GOBLIN("§bDiamond Goblin", "§bDiamond Goblin"),
-        COLD("§bCold", "§bCold");
+        COLD("§bCold", "§bCold"),
+        PICKAXE_ABILITY("§6Pickaxe Ability", "§6Pickaxe Ability Ready!");
```
CalMWolfs: Add the trailing comma to this line and move the semi colon onto its own line

CalMWolfs (changes requested):
In addition to what Luna said and my nitpicks, the message hypixel sends is not accurate and when the lobby has low tps this message can be very out of sync. The tab widget is much more accurate for timing, and using something like ServerTimeMark in addition to the pickaxe cooldown may be able to yield a notification that fires at the correct time

PR #6484, opened 2026-08-30
Improvement + Fix: Disable Boost Warning Volume by default and make it only boost actual warnings

lunaynx:
## What
I made a bad judgement call in #6418. The state of the toggle was migrated based on its intent, which logically makes sense, but ignores nuance. The volume boost has been broken since probably all the way back in the 1.21.5 port, so people are used to not having their ears blown out by SkyHanni warning sounds.

This PR sets "Boost Warning Volume" to default disabled, removes the FeatureToggle, and performs a one-time migration turning it off for all existing users.

Another issue that was discovered was that this was boosting literally *all* SkyHanni sounds. So now sounds can opt out with `isWarning = false`, which has been added to a bunch of existing sounds. This has been applied to mostly non-warning UI clicks, unimportant notifications (achievements), as well as anything that's replacing/emulating a Hypixel sound that wouldn't be boosted normally (overflow level up).

## Changelog Improvements
+ Disabled the "Boost Warning Volume" option by default based on user feedback. - Luna
    * The volume boost was always on by default, just under a different name, but it has been broken for a long time.
    * If you want it enabled, you'll have to turn it back on manually after this update.

## Changelog Fixes
+ Fixed "Boost Warning Volume" also boosting non-warning sounds from SkyHanni. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/config/features/combat/broodmother/BroodmotherSpawnAlertConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/combat/broodmother/BroodmotherSpawnAlertConfig.kt
@@ -1,7 +1,7 @@
 package at.hannibal2.skyhanni.config.features.combat.broodmother
 
+import at.hannibal2.skyhanni.features.combat.BroodmotherFeatures
 import at.hannibal2.skyhanni.utils.OSUtils
-import at.hannibal2.skyhanni.utils.SoundUtils
 import at.hannibal2.skyhanni.utils.SoundUtils.playSound
 import com.google.gson.annotations.Expose
 import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorButton
@@ -22,7 +22,7 @@ class BroodmotherSpawnAlertConfig {
 
     @ConfigOption(name = "Test Sound", desc = "Test current sound settings.")
     @ConfigEditorButton(buttonText = "Test")
-    val testSound: () -> Unit = { SoundUtils.createSound(alertSound, pitch).playSound() }
+    val testSound: () -> Unit = { BroodmotherFeatures.alertSound.playSound() }
 
     @Expose
     @ConfigOption(name = "Repeat Sound", desc = "How many times the sound should be repeated.")
--- a/src/main/java/at/hannibal2/skyhanni/config/features/combat/broodmother/BroodmotherSpawnAlertConfig.kt
+++ b/src/main/java/at/hannibal2/skyhanni/config/features/combat/broodmother/BroodmotherSpawnAlertConfig.kt
@@ -2,7 +2,6 @@ package at.hannibal2.skyhanni.config.features.combat.broodmother
 
 import at.hannibal2.skyhanni.features.combat.BroodmotherFeatures
 import at.hannibal2.skyhanni.utils.OSUtils
-import at.hannibal2.skyhanni.utils.SoundUtils.playSound
 import com.google.gson.annotations.Expose
 import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorButton
 import io.github.notenoughupdates.moulconfig.annotations.ConfigEditorSlider
@@ -22,7 +21,7 @@ class BroodmotherSpawnAlertConfig {
 
     @ConfigOption(name = "Test Sound", desc = "Test current sound settings.")
     @ConfigEditorButton(buttonText = "Test")
-    val testSound: () -> Unit = { BroodmotherFeatures.alertSound.playSound() }
+    val testSound: () -> Unit = { BroodmotherFeatures.playTestSound() }
 
     @Expose
     @ConfigOption(name = "Repeat Sound", desc = "How many times the sound should be repeated.")
--- a/src/main/java/at/hannibal2/skyhanni/features/dungeon/LowHealthAlert.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/dungeon/LowHealthAlert.kt
@@ -19,7 +19,7 @@ object LowHealthAlert {
     private val soundConfig get() = config.lowHealthAlertSound
     private var lastAlert: TitleContext? = null
 
-    val alertSound get() = SoundUtils.createSound(soundConfig.alertSound, soundConfig.pitch, isWarning = true)
+    private val alertSound get() = SoundUtils.createSound(soundConfig.alertSound, soundConfig.pitch, isWarning = true)
 
     @HandleEvent(onlyOnIsland = IslandType.CATACOMBS)
     fun onScoreboardChange(event: ScoreboardUpdateEvent) {
--- a/src/main/java/at/hannibal2/skyhanni/features/dungeon/LowHealthAlert.kt
+++ b/src/main/java/at/hannibal2/skyhanni/features/dungeon/LowHealthAlert.kt
@@ -15,10 +15,11 @@ import kotlin.time.Duration.Companion.seconds
 
 @SkyHanniModule
 object LowHealthAlert {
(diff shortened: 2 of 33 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/composter/ComposterOverlay.kt
```diff
@@ -564,7 +559,7 @@ object ComposterOverlay {
             }
         }
         if (havingInSacks == 0) {
-            SoundUtils.playErrorSound()
+            SoundUtils.playErrorSound(isWarning = false)
```
hannibal002, line 562: why is this one not a warning while the blocked super craft click still is?

src/main/java/at/hannibal2/skyhanni/utils/SoundUtils.kt
```diff
@@ -45,27 +46,31 @@ object SoundUtils {
...
+    // isWarning controls whether the sound is affected by the boostWarningVolume option.
+    // Set it to false for GUI interaction feedback and other non-warning sound effects.
+    fun createSound(name: String, pitch: Float, volume: Float = 50f, isWarning: Boolean): SoundInstance {
         val newSound = SoundCompat.getModernSoundName(name)
         val identifier = Identifier.parse(newSound.replace(Regex("[^a-z0-9:/._-]"), ""))
-        return SkyHanniSoundInstance(identifier, pitch, volume)
+        return SkyHanniSoundInstance(identifier, pitch, volume, isWarning)
     }
 
     fun playBeepSound(pitch: Float = 1f) {
```
hannibal002, line 57: should `playBeepSound` get the same parameter? the beep in `ReforgeHelper` line 158 is feedback, not a warning.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/composter/ComposterOverlay.kt
lunaynx: Probably just a botched edit on my part, sounds like something that should be treated as a warning.

src/main/java/at/hannibal2/skyhanni/utils/SoundUtils.kt
lunaynx: Yeah, a parameter is probably good to have. For `ReforgeHelper` specifically, I'm not sure what's best. It's technically feedback only when you've reached your desired reforge or reforging is blocked due to having a non-Blacksmith reforge rather than on every click, but if you think it makes more sense to treat it as a non-warning, I'll change it.

lunaynx: I suppose "has non-Blacksmith reforge" has more standing to be considered a warning, the other is more of a notice but doesn't strictly have to be boosted since we prevent over-clicking anyway (we also do in the former case, but that one is actually more of an edge case).

src/main/java/at/hannibal2/skyhanni/utils/SoundUtils.kt
```diff
@@ -45,27 +46,31 @@ object SoundUtils {
...
-        val beepSound = beepSoundCache.getOrPut(pitch) { createSound("entity.experience_orb.pickup", pitch) }
+        val beepSound = beepSoundCache.getOrPut(pitch) {
+            createSound("entity.experience_orb.pickup", pitch, isWarning = true)
+        }
         beepSound.playSound()
     }
 
-    fun playClickSound() {
-        clickSound.playSound()
+    fun playClickSound(isWarning: Boolean = false) {
```
hannibal002: `createSound` has no default on purpose, these two have opposite ones. same question for the new beep parameter.

src/main/java/at/hannibal2/skyhanni/utils/SoundUtils.kt
```diff
@@ -45,27 +46,31 @@ object SoundUtils {
         }
     }
 
-    fun createSound(name: String, pitch: Float, volume: Float = 50f): SoundInstance {
+    // isWarning controls whether the sound is affected by the boostWarningVolume option.
+    // Set it to false for GUI interaction feedback and other non-warning sound effects.
+    fun createSound(name: String, pitch: Float, volume: Float = 50f, isWarning: Boolean): SoundInstance {
```
hannibal002, line 51: move the required `isWarning` in front of the optional `volume`

src/main/java/at/hannibal2/skyhanni/utils/SoundUtils.kt
hannibal002, line 64: createSound has no default on purpose, these two have opposite ones. same question for the new beep parameter.

src/main/java/at/hannibal2/skyhanni/config/features/combat/broodmother/BroodmotherSpawnAlertConfig.kt
```diff
@@ -22,7 +22,7 @@ class BroodmotherSpawnAlertConfig {
 
     @ConfigOption(name = "Test Sound", desc = "Test current sound settings.")
     @ConfigEditorButton(buttonText = "Test")
-    val testSound: () -> Unit = { SoundUtils.createSound(alertSound, pitch).playSound() }
+    val testSound: () -> Unit = { BroodmotherFeatures.alertSound.playSound() }
```
hannibal002: broodmother tests via the public sound, low health via `playTestSound`, pick one.

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestSpawnSound.kt
```diff
@@ -58,10 +55,10 @@ object PestSpawnSound {
 
     private fun playPlumberTheme(soundName: String) {
         SkyHanniMod.launchCoroutine("pest spawn sound") {
-            val noteE = createSound(soundName, 0.890899f)
-            val noteC = createSound(soundName, 0.707107f)
-            val noteG = createSound(soundName, 1.059463f)
-            val noteLowG = createSound(soundName, 0.529732f)
+            val noteE = createSound(soundName, 0.890899f, isWarning = true)
```
hannibal002, line 58: these replace the hypixel sound as well, why warning here but not in `replaceWithOther`?

src/main/java/at/hannibal2/skyhanni/features/dungeon/LowHealthAlert.kt
```diff
@@ -15,11 +15,12 @@ import kotlin.time.Duration.Companion.seconds
 
 @SkyHanniModule
 object LowHealthAlert {
-
     private val config get() = SkyHanniMod.feature.dungeon.lowHealthAlert
     private val soundConfig get() = config.lowHealthAlertSound
     private var lastAlert: TitleContext? = null
 
+    val alertSound get() = SoundUtils.createSound(soundConfig.alertSound, soundConfig.pitch, isWarning = true)
```
hannibal002: this can stay private

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/pests/PestSpawnSound.kt
lunaynx: Yeah, well, normally replacement sounds shouldn't boost, but this is inherently a "warning" type sound that players often *want* to be loud. But maybe it's better to add a separate toggle here so that "Default" can also benefit from it.

hannibal002 (approved):

PR #3017, opened 2024-12-03
Improvement: Coop Support for Garden Plot Spray Detection

Chissl:
## What
Adds detection for when another coop member sprays or clears garden plots by pulling spray status from tablist

## Changelog Improvements
+ Added co-op support for Plot Spray detection in Garden. - Chissl

PR #6491, opened 2026-09-01
Fix: Pests widget blanking after plot change

lunaynx:
## What
`TabWidget.PESTS` was being blanked on every plot change, which made Pest Cooldown appear to "randomly" switch to Unknown while farming, especially when using multiple plots. This PR gets rid of that blanking and instead adds a guard to ensure we don't save stale spray data from the old plot to the new plot by delaying the update until Hypixel sends another tab list update.

## Changelog Fixes
+ Fixed Pest Cooldown saying Unknown for a few seconds when moving between plots. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/model/TabWidget.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/model/TabWidget.kt
@@ -408,6 +408,9 @@ enum class TabWidget(
     /** Internal value for the checking to set [isActive] */
     private var gotChecked = false
 
+    // Makes postNewEvent post even if the lines are unchanged, see forceUpdateWidget
+    private var forceNextUpdate = false
+
     private var sendOnThisIsland = false
 
     /** A [matchMatcher] for the first line using the pattern from the widget*/
@@ -418,7 +421,8 @@ enum class TabWidget(
 
     private fun postNewEvent(lines: List<Component>) {
         // Prevent Post if lines are equal
-        if (lines == this.lines) return
+        if (lines == this.lines && !forceNextUpdate) return
+        forceNextUpdate = false
         this.lines = lines
         isActive = true
         WidgetUpdateEvent(this, lines).post()
@@ -440,7 +444,6 @@ enum class TabWidget(
 
     @SkyHanniModule
     companion object {
-
         /** The index for the start of each Widget (inclusive) */
         private val separatorIndexes = mutableListOf<Pair<Int, TabWidget?>>()
 
@@ -457,7 +460,7 @@ enum class TabWidget(
         private var lastTabComponents: List<Component> = emptyList()
 
         @HandleEvent(onlyOnSkyblock = true)
-        fun onSecondPassed(event: SecondPassedEvent) {
+        private fun onSecondPassed(event: SecondPassedEvent) {
             if (sentSinceWorldChange) return
             if (SkyBlockUtils.lastWorldSwitch.passedSince() < FORCE_UPDATE_DELAY) return
             sentSinceWorldChange = true
@@ -466,7 +469,7 @@ enum class TabWidget(
         }
 
         @HandleEvent(priority = HandleEvent.HIGH)
-        fun onTabListUpdate(event: TabListUpdateEvent) {
+        private fun onTabListUpdate(event: TabListUpdateEvent) {
             if (!SkyBlockUtils.inSkyBlock) {
                 if (separatorIndexes.isNotEmpty()) {
                     separatorIndexes.forEach { it.second?.updateIsActive() }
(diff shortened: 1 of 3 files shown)


src/main/java/at/hannibal2/skyhanni/features/garden/plot/GardenPlotApi.kt
```diff
@@ -348,15 +346,14 @@ object GardenPlotApi {
     }
 
     @HandleEvent
-    fun onPlotChange(event: PlotChangeEvent) {
-        DelayedRun.runDelayed(3.seconds) {
-            TabWidget.forceUpdateWidget(TabWidget.PESTS)
-        }
+    private fun onPlotChange() {
```
hannibal002, line 349: What makes the next tab list update guaranteed to be for the new plot? If Hypixel is late, the old plot's Spray line gets applied to the new one.

hannibal002: Why was the 3 second delay dropped?

src/main/java/at/hannibal2/skyhanni/data/model/TabWidget.kt
```diff
@@ -563,9 +566,13 @@ enum class TabWidget(
             }
         }
 
+        /**
+         * Makes the widget post a [WidgetUpdateEvent] on the next tab list update even if its lines are unchanged.
+         * Listeners are only fed fresh tab list data; the currently cached lines are never re-posted.
+         */
         fun forceUpdateWidget(widget: TabWidget) {
```
hannibal002, line 572: `reSendEvents` posts `postNewEvent(lines)` with the cached list, so with the flag set the cached lines do get re-posted.

src/main/java/at/hannibal2/skyhanni/data/model/TabWidget.kt
```diff
@@ -408,6 +408,9 @@ enum class TabWidget(
     /** Internal value for the checking to set [isActive] */
     private var gotChecked = false
 
+    // Makes postNewEvent post even if the lines are unchanged, see forceUpdateWidget
+    private var forceNextUpdate = false
```
hannibal002, line 412: Which case does this actually cover? PESTS already posts every second while a cooldown runs.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/garden/plot/GardenPlotApi.kt
lunaynx: Well, I'd be damned if Hypixel sent stale data in a tab list update packet. That seems unlikely and contradicts my experience. It only updates once every 3 seconds, but I've never seen it be already stale by the time it updates.

The delay presumably existed previously as a way to try to not kill the data entirely because it was clearing the widget immediately and Hypixel only updates the tab list once every 3 seconds. I don't think it's necessary anymore.

src/main/java/at/hannibal2/skyhanni/data/model/TabWidget.kt
lunaynx: You're right, though not caused by `forceUpdateWidgets`. I updated the comment.

src/main/java/at/hannibal2/skyhanni/data/model/TabWidget.kt
lunaynx: Honestly, I'm not sure why the force update exists in the first place. There was probably a good reason for it, but you'd have to ask @Chissl – it was added in #3017. I'd rather not change too much here and risk breaking things, nor have a highly timely bugfix PR (Grand Feast active and a 4 perk Finnegan is expected after Derpy) held up over you questioning a poorly documented mechanic that predates my PR.

src/main/java/at/hannibal2/skyhanni/features/garden/plot/GardenPlotApi.kt
hannibal002: Then drop "guaranteed" from the comment and write it as the assumption it is.

src/main/java/at/hannibal2/skyhanni/data/model/TabWidget.kt
hannibal002: Still not true: with the flag set, postNewEvent in reSendEvents no longer returns early.

src/main/java/at/hannibal2/skyhanni/data/model/TabWidget.kt
hannibal002: Fair, out of scope here.

hannibal002 (changes requested):
bump

src/main/java/at/hannibal2/skyhanni/data/model/TabWidget.kt
lunaynx: Ah, you're right. I removed that part of the comment. The logic may still be somewhat dubious but I don't think we're making it worse here at least?

hannibal002 (approved):

PR #6499, opened 2026-09-02
Fix: Crash on launch with certain mods due to non-sequential VFE IDs

lunaynx:
## What
SkyHanni's VertexFormatElement registration logic incorrectly assumed VFE IDs must be sequential – taking the *last* registered ID and attempting to register ours after that. This caused a crash on launch if any other mod registered a VFE with ID ≥ 27 before us.

Also replaced the existing TODO comment with an explanation of why the "user will have enough mods to register 27 more vertex format elements" scenario is in fact *not* something we should account for. No other mod does either: Sodium and Iris both hard-crash in this scenario. The central VertexFormatElement registry is also gone on 26.2, so it won't be a problem at all in a couple months.

## Changelog Fixes
+ Fixed the game crashing on launch with certain mods (e.g. MusicDisplay) installed on Minecraft 26.1. - Luna
  * This fixes the following crash: "Caused by: java.util.NoSuchElementException: Collection contains no element matching the predicate".

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/render/SkyHanniVertexFormats.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/render/SkyHanniVertexFormats.kt
@@ -8,35 +8,26 @@ import org.lwjgl.system.MemoryUtil
 import com.mojang.blaze3d.GpuFormat
 import com.mojang.blaze3d.vertex.DefaultVertexFormat
 //?} else {
-/*import at.hannibal2.skyhanni.test.command.ErrorManager
-import at.hannibal2.skyhanni.utils.system.PlatformUtils
-import com.mojang.blaze3d.vertex.VertexFormatElement
+/*import com.mojang.blaze3d.vertex.VertexFormatElement
 *///?}
 
-//? if < 26.2
-//private typealias VFEType = VertexFormatElement.Type
-
 object SkyHanniVertexFormats {
-
     //? if < 26.2 {
-    /*// Different versions of MC use differing counts, so load the last registered ID dynamically.
-    val lastRegisteredId by lazy {
-        (0 until VertexFormatElement.MAX_COUNT).filter { VertexFormatElement.byId(it) != null }.max()
-    }
+    /*// Vanilla registers IDs 0-6.
+    private var startingId = 7
     *///?}
 
-    @Suppress("EmptyDefaultConstructor", "UnusedPrivateProperty")
+    @Suppress("EmptyDefaultConstructor")
     internal enum class VertexElement(
         //? if < 26.2 {
         /*private val index: Int = 0,
-        private val type: VFEType = VFEType.FLOAT,
+        private val type: VertexFormatElement.Type = FLOAT,
         private val normalized: Boolean = false,
         private val count: Int = 4,
         *///?}
     ) {
         // {radius, smoothness/borderThickness, adjustedHalfSizeX, adjustedHalfSizeY}
         ROUNDED_PARAMS_0,
-
         // {adjustedCenterPosX, adjustedCenterPosY, borderBlur/angle1/0, angle2/0}
         ROUNDED_PARAMS_1,
         // {angle, progress, phaseOffset, reverse(float)}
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/utils/render/SkyHanniVertexFormats.kt
```diff
@@ -13,30 +13,23 @@ import at.hannibal2.skyhanni.utils.system.PlatformUtils
 import com.mojang.blaze3d.vertex.VertexFormatElement
 *///?}
```
hannibal002, line 12: ErrorManager and PlatformUtils imports are unused

src/main/java/at/hannibal2/skyhanni/utils/render/SkyHanniVertexFormats.kt
```diff
@@ -51,41 +44,36 @@ object SkyHanniVertexFormats {
             name.lowercase().split("_").joinToString("") { it.replaceFirstChar(Char::uppercaseChar) }
 
         //? if < 26.2 {
-        /*// The ID we use to register the format element with Minecraft.
-        // see safeRegister() for details on how this is used and determined at runtime.
-        private val registrationId: Int by lazy { lastRegisteredId + (ordinal + 1) }
-        val element by lazy {
-            safeRegister(registrationId, index, type, false, count)
+        /*val element by lazy {
+            safeRegister(index, type, false, count)
```
hannibal002: ▎ pass normalized here instead of a hardcoded false, then UnusedPrivateProperty can go from the suppression

src/main/java/at/hannibal2/skyhanni/utils/render/SkyHanniVertexFormats.kt
```diff
@@ -51,41 +44,36 @@ object SkyHanniVertexFormats {
...
-        // Todo, it is exceptionally unlikely that a user will have enough mods to register 27 more vertex format elements,
-        //  but, technically possible, and something we should account for eventually.
-        val id = (desiredId until VertexFormatElement.MAX_COUNT).first { VertexFormatElement.byId(it) == null }
-        if (id != desiredId && PlatformUtils.isDevEnvironment) ErrorManager.logErrorStateWithData(
-            "VertexFormatElement ID $desiredId was already taken, using $id instead",
-            "SkyHanni vertex format element ID conflict. Desired ID $desiredId was already registered",
-        )
+        // It is technically possible, but exceptionally unlikely, that a user will have enough
+        // mods to register 27 more vertex format elements. We deliberately let the game crash
+        // here rather rather than attempting to recover, because a failed registration would
```
hannibal002: ▎ typo: "rather rather"

src/main/java/at/hannibal2/skyhanni/utils/render/SkyHanniVertexFormats.kt
```diff
@@ -51,41 +44,36 @@ object SkyHanniVertexFormats {
...
     /*/**
-     * Registers a VertexFormatElement with the given parameters, automatically finding an available ID if the desired one is taken.
-     * Logs an error if the desired ID was already taken, but still registers the element with a valid ID.
-     * @param desiredId The preferred ID for the VertexFormatElement.
+     * Registers a VertexFormatElement with the given parameters,
+     * automatically finding an available ID.
+     *
      * @param index The index of the element in the vertex format (default is 0).
      * @param type The data type of the element (default is FLOAT).
      * @param usage The intended usage of the element (default is GENERIC).
```
hannibal002: `@param usage` does not exist, and `@param normalized` is missing

src/main/java/at/hannibal2/skyhanni/utils/render/SkyHanniVertexFormats.kt
```diff
@@ -51,41 +44,36 @@ object SkyHanniVertexFormats {
...
-        if (id != desiredId && PlatformUtils.isDevEnvironment) ErrorManager.logErrorStateWithData(
-            "VertexFormatElement ID $desiredId was already taken, using $id instead",
-            "SkyHanni vertex format element ID conflict. Desired ID $desiredId was already registered",
-        )
+        // It is technically possible, but exceptionally unlikely, that a user will have enough
+        // mods to register 27 more vertex format elements. We deliberately let the game crash
+        // here rather rather than attempting to recover, because a failed registration would
+        // leave a lot of mod features broken, and forcibly extending the array would be likely
+        // unsafe and require mixins into mods such as Sodium that hardcode the size as 32.
+        val id = (startingId++ until VertexFormatElement.MAX_COUNT).firstOrNull { VertexFormatElement.byId(it) == null }
```
hannibal002, line 74: ▎ startingId is never needed, the scan skips taken ids anyway. A const val gives the same ids.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/render/SkyHanniVertexFormats.kt
lunaynx: Yes, but without it, wouldn't we waste a few extra cycles going over IDs that were already registered by us before? Probably not significant enough to matter, but I think it doesn't hurt to have this.

hannibal002: fair enough. keep it

src/main/java/at/hannibal2/skyhanni/utils/render/SkyHanniVertexFormats.kt
```diff
@@ -8,35 +8,26 @@ import org.lwjgl.system.MemoryUtil
...
     //? if < 26.2 {
-    /*// Different versions of MC use differing counts, so load the last registered ID dynamically.
-    val lastRegisteredId by lazy {
-        (0 until VertexFormatElement.MAX_COUNT).filter { VertexFormatElement.byId(it) != null }.max()
-    }
+    /*// Vanilla registers IDs 0-6.
+    private var startingId = 7
     *///?}
 
     @Suppress("EmptyDefaultConstructor", "UnusedPrivateProperty")
```
hannibal002: UnusedPrivateProperty is not needed

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/render/SkyHanniVertexFormats.kt
lunaynx: Ah, you're right, I missed this. I thought you were referencing the one I had already removed from the commented out 26.1 code.

hannibal002 (approved):

PR #6500, opened 2026-09-03
Backend: Support for Faction Rabbits in Resolution Query

jani270:
+ Added support for Faction Rabbits to the Item Resolution Query. - jani


src/main/java/at/hannibal2/skyhanni/api/enoughupdates/ItemResolutionQuery.kt
```diff
@@ -309,6 +310,12 @@ class ItemResolutionQuery {
         return rawInternalName.toInternalName()
     }
 
+    private fun resolveFactionRabbitName(): NeuInternalName {
+        val name = getExtraAttributes().getStringOrDefault("faction_rabbit_id")
```
hannibal002, line 314: `name` is the id, not a name. `id` fits better.

hannibal002 (changes requested):
Any user visible effect here? Faction Rabbits have AH prices, so this might be an Improvement rather than Technical Details.

PR #6503, opened 2026-09-03
Backend: Fix some RepositoryReloadEvent handlers not being disableable

lunaynx:
## What
Changes `RepositoryReloadEvent` handlers outside `SkyHanniEvents` that were using `HIGHEST` priority to `HIGH` to ensure they can be disabled. There was no reason for them to use `HIGHEST`, as there were no `HIGH` priority listeners for them to take precedence over.

+ Lowered the priority of IslandType.onRepoReload, PowderMiningChatFilter.onRepoReload, and RemainingSlayerKills.onRepoReload from HIGHEST to HIGH to ensure SkyHanniEvents can disable them. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/data/IslandType.kt
+++ b/src/main/java/at/hannibal2/skyhanni/data/IslandType.kt
@@ -2,7 +2,6 @@ package at.hannibal2.skyhanni.data
 
 import at.hannibal2.skyhanni.SkyHanniMod.launch
 import at.hannibal2.skyhanni.api.event.HandleEvent
-import at.hannibal2.skyhanni.api.event.HandleEvent.Companion.HIGHEST
 import at.hannibal2.skyhanni.data.jsonobjects.repo.IslandTypeJson
 import at.hannibal2.skyhanni.events.RepositoryReloadEvent
 import at.hannibal2.skyhanni.skyhannimodule.SkyHanniModule
@@ -93,7 +92,6 @@ enum class IslandType(private val nameFallback: String, private val apiNameFallb
 
     @SkyHanniModule
     companion object {
-
         fun Collection<IslandType>.isInAnyIsland(): Boolean = any { it.isInIsland() }
         private val repoReloadCoroutine = CoroutineSettings("island type repo reload")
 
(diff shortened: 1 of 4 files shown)


src/main/java/at/hannibal2/skyhanni/data/IslandType.kt
```diff
@@ -116,9 +111,11 @@ enum class IslandType(private val nameFallback: String, private val apiNameFallb
         fun getByIdOrNull(id: String): IslandType? = entries.find { it.apiName == id }
         fun getByIdOrUnknown(id: String): IslandType = getByIdOrNull(id) ?: UNKNOWN
 
-        @HandleEvent(priority = HIGHEST)
-        fun onRepoReload(event: RepositoryReloadEvent) = repoReloadCoroutine.launch {
-            val data = event.getConstantAsync<IslandTypeJson>("misc/IslandType")
+        // This is intentionally not async right now because using a fire-and-forget
+        // coroutine means priority won't be fully respected. PR #6064 will fix this.
+        @HandleEvent(priority = HandleEvent.HIGH)
```
hannibal002, line 117: Why does this need raised priority at all, which handler reads islandData during the same reload?

src/main/java/at/hannibal2/skyhanni/data/IslandType.kt
```diff
@@ -116,9 +111,11 @@ enum class IslandType(private val nameFallback: String, private val apiNameFallb
...
+        private fun onRepoReload(event: RepositoryReloadEvent) {
+            val data = event.getConstant<IslandTypeJson>("misc/IslandType")
```
hannibal002: `getConstant` is deprecated, worth naming that in the comment above so nobody reverts it.

hannibal002 (changes requested):
Worth noting in `SkyHanniEvents.onRepoReload` that no other `RepositoryReloadEvent` handler may use `HIGHEST`?

src/main/java/at/hannibal2/skyhanni/data/IslandType.kt
lunaynx: `getConstantAsync` is a `suspend fun` so that would not compile, and the reason it's synchronous is already explained above, so I don't think an additional comment is necessary.

src/main/java/at/hannibal2/skyhanni/data/IslandType.kt
lunaynx: Actually, good point, none of them do right now. I'll revert making it synchronous.

Can also drop it down to `NORMAL` priority if you want, though there might have been an intention behind giving it a higher priority originally that could be useful in the future.

hannibal002 (approved):

PR #6507, opened 2026-09-04
Fix: Sea Creature catch messages not using proper articles

lunaynx:
## What
`StringUtils.optionalAn` was only returning `"an"` if the string starts with a *lowercase* vowel; otherwise it would return `"a"`. The only practical user-facing effect of this today is that Sea Creature catch messages are wrong: "I caught a Oasis Sheep!" rather than "I caught an Oasis Sheep!".

## Changelog Fixes
+ Fixed Sea Creature catch messages incorrectly saying "a" instead of "an" (e.g. "a Oasis Sheep"). - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
@@ -525,9 +525,10 @@ object StringUtils {
 
     fun String.isValidUuid(): Boolean = runCatching(UUID::fromString).isSuccess
 
-    fun optionalAn(string: String): String {
-        if (string.isEmpty()) return ""
-        return if (string[0] in "aeiou") "an" else "a"
+    fun optionalAn(string: String): String = when {
+        string.isEmpty() -> ""
+        string.first().isVowel() -> "an"
+        else -> "a"
     }
 
     fun String.hasWhitespace(): Boolean = any { it.isWhitespace() }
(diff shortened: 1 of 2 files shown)


src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt
```diff
@@ -527,7 +527,7 @@ object StringUtils {
 
     fun optionalAn(string: String): String {
         if (string.isEmpty()) return ""
-        return if (string[0] in "aeiou") "an" else "a"
+        return if (string[0].lowercase() in "aeiou") "an" else "a"
```
hannibal002: `Char.isVowel` exists already, couple lines above

hannibal002: `EventHandler.post` uses this too, so error messages change as well.

src/test/java/at/hannibal2/skyhanni/test/StringUtilsTest.kt
```diff
@@ -0,0 +1,17 @@
...
+
+class StringUtilsTest {
+    @Test
+    fun `optionalAn returns the correct article`() {
+        assertEquals("an", StringUtils.optionalAn("apple"))
+        assertEquals("an", StringUtils.optionalAn("Apple"))
+        assertEquals("a", StringUtils.optionalAn("pear"))
+        assertEquals("a", StringUtils.optionalAn("Pear"))
+        assertEquals("a", StringUtils.optionalAn("yeti"))
+        assertEquals("a", StringUtils.optionalAn("Yeti"))
```
hannibal002, line 15: Also test the empty string.

hannibal002 (changes requested):

hannibal002 (approved):

PR #6084, opened 2026-07-13
Backend: Remove unnecessary usages of GuiKeyPressEvent.GuiMouseButtonPressEvent

Fazfoxy:
## Dependencies
- https://github.com/hannibal002/SkyHanni/pull/6305
## What
This functionality was entirely pointless because mouse events were firing the main GuiKeyPressEvent prior to 6305 anyway, so this PR just moves the very few usages of mouse event directly where it was already using GuiKeyPressEvent to the unified event (and removes a singular magic number from the loadout code)

exclude_from_changelog

PR #6510, opened 2026-09-05
Feature: Add 26.3 support

lunaynx:
## Dependencies
- #6323
- #6084
- #6538

## Other Dependencies
- MoulConfig 26.3 release
- Render Chest 26.3 release (can probably ship initial betas without if needed)

## What
Highly WIP. It builds but game doesn't boot due to shaders. Also a bit of the rendering code is just commented out for the time being.

## Changelog New Features
+ Added support for Minecraft 26.3. - Luna


src/main/java/at/hannibal2/skyhanni/events/ParticleEvent.kt
```diff
@@ -31,16 +33,22 @@ class ParticleEvent(
...
     val longDistance: Boolean,
     val particleArgs: IntArray? = null,
 ) : CancellableWorldEvent() {
-
     val distanceToPlayer by lazy { location.distanceToPlayer() }
 
+    /**
+     * Checks whether the particle's [xSpeed], [ySpeed], and [zSpeed] all equal [speed].
+     */
+    fun isSpeed(speed: Float): Boolean = xSpeed == speed && ySpeed == speed && zSpeed == speed
```
AverageUser125, line 48: I think just `val speed: Float? get() { … }`
Where it returns null if not all are equal would be better.
(Also less refactoring changes).

Though tbh, this is fine too.

lunaynx: Speed being nullable sounds weird, and it would likely require refactoring regardless because the result of `!=` would be true when it's `null`.

src/main/java/at/hannibal2/skyhanni/utils/compat/MouseCompat.kt
```diff
@@ -39,13 +39,14 @@ object MouseCompat {
 
     private val mouse by lazy { Minecraft.getInstance().mouseHandler }
 
+    fun isMouseButton(button: Int) = button in 0..5
```
AverageUser125, line 42:
```suggestion
    fun isMouseButton(button: Int) = button in 0..3
```
Yeah, this is kinda problematic.
Will just have to give up on the side mouse keys.

AverageUser125: Maybe can make negative values mean mouse button.
Or smth similiar.

but I can’t think of an elegant solution.

src/main/java/at/hannibal2/skyhanni/config/UpdateKeybinds.kt
```diff
@@ -1,62 +1,30 @@
...
+            event.transform(147, keybindPath) { element ->
+                val oldCode = element.asInt
+                val type = if (oldCode in 0 until MouseCompat.NUMBER_OF_MOUSE_BUTTONS) {
+                    InputConstants.Type.MOUSE
+                } else {
+                    //~ if < 26.3 'KEYBOARD' -> 'KEYSYM'
+                    InputConstants.Type.KEYBOARD
                 }
-                ChatUtils.chat("§aSuccessfully reset all SkyHanni Keybinds")
+                val newStringName = type.getOrCreate(oldCode).name
```
AverageUser125, line 26: Since the key constants have changed in 26.3.
You can’t do this, you will have to manually create a legacy integer to string list conversation.

AverageUser125 (changes requested):

PR #6511, opened 2026-09-05
end island features (End Stone Protector and Dragons)

hordinii:
## What

Adds a set of features for the End island, covering both the dragons at the Dragon's Nest
and the End Stone Protector.

**Dragons**
- A weight HUD and an end-of-fight chat message, calculated from placed summoning eyes,
  placement and own damage share.
- A damage list HUD reading the `DRAGON` tab list widget, so every damager and their
  damage is visible during the fight.
- Rare drop alerts split in two tiers: special drops (pet, Dragon Horn, Pearlescent Dye)
  interrupt the screen with a title, everything else on the whitelist is reported in chat
  only. Both carry the item's auction value.

**End Stone Protector**
- A framed HUD showing the awakening stage, the spawn point and a countdown until the
  protector becomes attackable. The countdown runs on server ticks, so it holds still while
  the server lags instead of skipping numbers.
- The six fixed spawn points are detected from the head block in the ground, marked with a
  beacon beam and a highlighted block.
- A weight message and rare drop alerts, mirroring the dragon side.

**Rare Drop Tracker**
- Counts epic and legendary Ender Dragon pets and Tier Boost Cores, and shows how many
  eligible fights have passed since the last one, together with the accumulated chance.
- Only fights whose weight was high enough to roll the drop count towards the streak, so a
  leeched kill does not make the streak look worse than it is. Magic Find and Pet Luck at
  the moment of the kill are factored in; Pet Luck applies to pets only, not to the
  Tier Boost Core.
- Lines can be copied individually by right clicking them while the chat is open. Used to show other people your own stats.

Drops are found by a shared scanner that reads the floating loot labels and the items the
loot armor stands carry, so nothing has to be picked up first. Items collected automatically
after the fight are reported as a fallback.

All features are off by default.

## Screenshot
<details>
<summary> Images </summary>

Screenshot_2026-09-06_014513
Screenshot_2026-09-06_014737
image2
image3
image4

</details>

## Changelog New Features
+ Added a Dragon Damage List HUD showing the damage leaderboard from the tab list. - hordinii
+ Added rare drop alerts for dragons, with a screen title for the rarest drops and the auction value in chat. - hordinii
+ Added an End Stone Protector HUD showing stage, spawn point and spawn countdown. - hordinii
    * The countdown runs on server ticks, so it pauses during lag instead of jumping.
+ Added spawn point waypoints for the End Stone Protector. - hordinii
+ Added rare drop alerts for the End Stone Protector. - hordinii
+ Added an End Rare Drop Tracker counting Ender Dragon pets and Tier Boost Cores. - hordinii
    * Shows the current dry streak and how likely a drop would have been over those fights, including Magic Find and Pet Luck.

## Changelog Improvements
+ The Dragon Profit Tracker now records kills no matter which weight options are enabled. - hordinii

## Changelog Removed Features
+ Removed the Dragon Weight HUD. The weight is still shown in chat once the fight ends. - hordinii


src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonDropAlert.kt
```diff
@@ -0,0 +1,118 @@
...
+ * their floating item labels, so nothing has to be picked up first.
+ */
+@SkyHanniModule
+object DragonDropAlert {
+
+    private val config get() = SkyHanniMod.feature.combat.endIsland.dragon
+
+    /** Special drops: rare enough to be worth covering the screen for. */
+    private val specialDrops = mapOf(
+        // NEU rarity suffixes: 3 is epic, 4 is legendary.
```
lunaynx, line 29: Unnecessary comment.

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonDropAlert.kt
```diff
@@ -0,0 +1,118 @@
...
+        "DRAGON_SCALE".toInternalName() to Drop(RareDropAlert.RARE, "Dragon Scale", withTitle = false),
+        "DRAGON_NEST_TRAVEL_SCROLL".toInternalName() to
+            Drop(RareDropAlert.EPIC, "Travel Scroll to Dragon's Nest", withTitle = false),
+    )
+
+    /**
+     * Listed by hand rather than derived from [DragonType]: a dragon added in the future will
+     * not be tracked until its four pieces are added here.
+     */
+    private val dragonArmor = mapOf(
```
lunaynx: Pull this data from the NEU repo rather than duplicating it, e.g.: https://github.com/NotEnoughUpdates/NotEnoughUpdates-REPO/blob/c1ff6290fe7d3a2c47cf2e462386c3e5d9104c69/items/PROTECTOR_ENDER_DRAGON_BOSS.json

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
```diff
@@ -0,0 +1,260 @@
...
+        .filter { it != DragonType.UNKNOWN }
+        .joinToString("|") { it.name.firstLetterUppercase() }
+
+    private val group = RepoPattern.group("combat.end-dragon-fight")
+    private val chatGroup = group.group("chat")
+    private val endGroup = group.group("chat.end")
+    private val scoreboardGroup = group.group("scoreboard")
+
+    // <editor-fold desc="Patterns">
+
```
lunaynx: Unnecessary blank line.

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
```diff
@@ -0,0 +1,260 @@
...
+    /**
+     * REGEX-TEST: §5☬ §r§d§lThe §r§5§c§lProtector Dragon§r§d§l has spawned!
+     * REGEX-TEST: §5☬ §r§d§lThe §r§5§c§lYoung Dragon§r§d§l has spawned!
+     */
+    private val dragonSpawnPattern by chatGroup.pattern(
+        "spawn",
+        "§5☬ §r§d§lThe §r§5§c§l(?<dragon>$dragonNames) Dragon§r§d§l has spawned!",
```
lunaynx: Please make all patterns you touch colorless (remove the `§*` color codes from the pattern and match against `event.cleanMessage` or similar).

If any of the pattern keys match what already exists in `beta`, don't forget to add a `.colorless` suffix.

This applies to other edited files too.

src/main/java/at/hannibal2/skyhanni/features/combat/end/golem/GolemLocation.kt
```diff
@@ -0,0 +1,127 @@
...
+ * Knows the six fixed spawn points of the End Stone Protector, names the one it rises at and
+ * marks them in the world: a beacon beam on the head structure and a highlighted block where
+ * the golem actually appears.
+ */
+@SkyHanniModule
+object GolemLocation {
+
+    private val config get() = SkyHanniMod.feature.combat.endIsland.golem
+
+    internal const val GOLEM_NAME = "Endstone Protector"
```
lunaynx, line 36: Double check this, is it still "Endstone Protector" rather than "End Stone Protector"? If so, worth noting in a comment.

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
```diff
@@ -0,0 +1,260 @@
...
+        "\\s+§r§6§l(?:PROTECTOR|OLD|UNSTABLE|YOUNG|STRONG|WISE|SUPERIOR) DRAGON DOWN!",
+    )
+
+    /**
+     * WRAPPED-REGEX-TEST: "                    §r§6§lEND STONE PROTECTOR DOWN!"
+     */
+    private val protectorDownPattern by endGroup.pattern(
+        "down.protector",
+        "\\s+§r§6§lEND ?STONE PROTECTOR DOWN!",
+    )
```
lunaynx, line 88: I doubt Hypixel is going to rename it back from End Stone Protector to Endstone Protector. (The colorless note from above still applies.)

```suggestion
    /**
     * WRAPPED-REGEX-TEST: "                    §r§6§lEND STONE PROTECTOR DOWN!"
     */
    private val protectorDownPattern by endGroup.pattern(
        "down.protector",
        "\\s+§r§6§lEND STONE PROTECTOR DOWN!",
    )
```

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndLootScanner.kt
```diff
@@ -0,0 +1,170 @@
...
+     * in colour, so the generic name resolver cannot tell them apart. The colour is therefore
+     * turned into the rarity suffix here.
+     *
+     * REGEX-TEST: §7[Lvl 1] §6Golem
+     * REGEX-TEST: §7[Lvl 1] §5Ender Dragon
+     * REGEX-TEST: §7[Lvl 100] §6Ender Dragon
+     */
+    private val petPattern by repoGroup.pattern(
+        "pet",
+        "§7\\[Lvl \\d+] §(?<rarity>[56])(?<name>.+)",
```
lunaynx, line 52: This one is an exception to the colorless note, it probably needs to keep the color check unless you can get it working with component matchers.

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndRareDropTracker.kt
```diff
@@ -0,0 +1,267 @@
...
+        val petLuck = SkyblockStat.PET_LUCK.lastKnownValue ?: 0.0
+        bothBonus = (magicFind + petLuck) / 100.0
+        magicFindBonus = magicFind / 100.0
+    }
+
+    /** Chance = BaseChance x (1 + (MagicFind + PetLuck) / 100). */
+    private fun withPetBonus(baseChance: Double) = baseChance * (1 + bothBonus)
+
+    /** Chance = BaseChance x (1 + MagicFind / 100), for drops that are not pets themselves. */
+    private fun withMagicFindBonus(baseChance: Double) = baseChance * (1 + magicFindBonus)
```
lunaynx: I'd rather do something like this:
```suggestion
    private var magicFind = 0.0
    private var petLuck = 0.0

    @HandleEvent
    private fun onEndBossDeath(event: EndBossDeathEvent) {
        magicFind = SkyblockStat.MAGIC_FIND.lastKnownValue ?: 0.0
        petLuck = SkyblockStat.PET_LUCK.lastKnownValue ?: 0.0
    }

    private fun withPetBonus(baseChance: Double) = baseChance * (1 + (magicFind + petLuck) / 100)

    private fun withMagicFindBonus(baseChance: Double) = baseChance * (1 + magicFind / 100)
```

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
lunaynx, line 43: Unnecessary blank line.

lunaynx (changes requested):

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndRareDropTracker.kt
```diff
@@ -0,0 +1,264 @@
...
+        petLuck = SkyblockStat.PET_LUCK.lastKnownValue ?: 0.0
+    }
+
+    private fun withPetBonus(baseChance: Double) = baseChance * (1 + (magicFind + petLuck) / 100)
+
+    /** Pet Luck only applies to pets, and the Tier Boost Core is a pet item rather than a pet. */
+    private fun withMagicFindBonus(baseChance: Double) = baseChance * (1 + magicFind / 100)
+
+    @HandleEvent
+    private fun onEndLootFound(event: EndLootFoundEvent) {
```
hannibal002, line 134: Doesn't this count the same drop twice? EndLootScanner posts once from the label and again from ItemAddEvent when it is picked up, and unlike RareDropAlert there is no dedup here.

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
```diff
@@ -0,0 +1,263 @@
...
+    private val endGroup = group.group("chat.end")
+    private val scoreboardGroup = group.group("scoreboard")
+
+
+
+    /**
+     * REGEX-TEST: ☬ The Protector Dragon has spawned!
+     * REGEX-TEST: ☬ The Young Dragon has spawned!
+     */
+    private val dragonSpawnPattern by chatGroup.pattern(
```
hannibal002, line 46: This resolves to the same key as beta (combat.end-dragon-fight.chat.spawn), but the capture group is now dragon instead of type. This needs a new key, otherwise it would break older versions.

src/main/java/at/hannibal2/skyhanni/config/features/combat/end/EndIslandConfig.kt
```diff
@@ -18,10 +16,14 @@ class EndIslandConfig {
     val dragon: DragonConfig = DragonConfig()
 
     @Expose
-    @ConfigOption(name = "Weight Endstone Protector", desc = "Shows your Endstone Protector weight in chat after it died.")
-    @ConfigEditorBoolean
-    @FeatureToggle
-    var endstoneProtectorChat: Boolean = true
```
hannibal002, line 24: Is there a migration for the old value anywhere? Without one everybody loses the setting and gets the new false default.

src/main/java/at/hannibal2/skyhanni/config/features/combat/end/DragonConfig.kt
```diff
@@ -14,12 +14,6 @@ class DragonConfig {
     @Accordion
     val dragonProfitTracker: DragonProfitTrackerConfig = DragonProfitTrackerConfig()
 
-    @Expose
-    @ConfigOption(name = "Superior Notification", desc = "Show a title when a Superior Dragon spawns.")
-    @ConfigEditorBoolean
-    @FeatureToggle
-    var superiorNotify: Boolean = true
```
hannibal002, line 21: Superior Notification and Skyhanni Prefix are gone entirely, intended?

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
```diff
@@ -0,0 +1,263 @@
...
+    private val tabDamagePattern by group.pattern(
+        "tablist.player",
+        "\\s(?<name>.+): (?<damage>[\\d.]+[kMB]?)[❤\\uE010].*",
+    )
+
+    /**
+     * REGEX-TEST: Dragon HP: 14,659,354 ❤
+     * REGEX-TEST: Protector HP: 2,317,156 ❤
+     */
+    private val scoreboardHpPattern by scoreboardGroup.pattern(
```
hannibal002, line 125: Since this now matches Protector HP: as well, doesn't dragonSpawned become true during a protector fight and pull the Dragon Weight HUD up there?

src/main/java/at/hannibal2/skyhanni/features/combat/end/golem/GolemDisplay.kt
```diff
@@ -0,0 +1,102 @@
...
+
+
+    private const val TITLE_COLOR = 0xE3D5A0
+
+    private const val TITLE_ICON_SPACING = 4
+
+    private val PROTECTOR_HEAD = "ENDSTONE_PROTECTOR_BOSS".toInternalName()
+
+    @HandleEvent(onlyOnIsland = IslandType.THE_END)
+    private fun onRender(event: GuiRenderEvent) {
```
hannibal002, line 78: Is this meant to stay on screen everywhere on the island? Outside a fight it just shows "Stage: ?" and "not revealed". Also, the old code cached the display behind a dirty flag, was dropping that intentional?

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
```diff
@@ -0,0 +1,263 @@
...
+        scoreboardHpPattern.matchMatcher(lines[index]) {
+            currentHp = group("hp").formatIntOrNull()
+        }
+        scoreboardDamagePattern.matchMatcher(lines[index + 1]) {
+            DragonFightState.yourDamage = group("damage").formatDouble()
+        }
+    }
+
+    @HandleEvent
+    private fun onIslandChange(event: IslandChangeEvent) {
```
hannibal002: IslandChangeEvent is deprecated, would WorldChangeEvent work here? Same in EndLootScanner, GolemLocation, GolemSpawnTimer, GolemStage and GolemWeight.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/config/features/combat/end/DragonConfig.kt
hordinii: The removed superiorNotify and skyhanniMessagePrefix options were dropped on purpose
In my eyes a Superior Dragon is nothing special anymore. Especially because of the new recombed summoning eyes the chance for a superior dragon got pretty high. 
If u want to keep this feature, i would add it back otherwise leave it like this,

src/main/java/at/hannibal2/skyhanni/config/features/combat/end/EndIslandConfig.kt
hordinii: Right, there was no migration. I'll add a move from endstoneProtectorChat to golem.weightChat and bump the config version

src/main/java/at/hannibal2/skyhanni/features/combat/end/golem/GolemDisplay.kt
hordinii: Good point. 
This HUD is developed for People that are doing End Stone Protectors as their mmm. Since these players lobby-hop to find Stage 4 Protectors, they need to see the current stage and spawn location immediately upon spawning into the End. For this reason, I prefer keeping the overlay visible everywhere on the island and for every Stage.

hordinii: the dirty flag gets added again.

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndRareDropTracker.kt
hordinii: Good point. It's now getting capped. Every item just getting reported 1 time per kill.

src/main/java/at/hannibal2/skyhanni/features/combat/end/golem/GolemDropAlert.kt
```diff
@@ -0,0 +1,47 @@
...
+ */
+@SkyHanniModule
+object GolemDropAlert {
+
+    private val config get() = SkyHanniMod.feature.combat.endIsland.golem
+
+    /** Special drops: rare enough to be worth covering the screen for. */
+    private val specialDrops = mapOf(
+        "PET_ITEM_TIER_BOOST_DROP".toInternalName() to Drop(RareDropAlert.LEGENDARY, "TIER BOOST CORE"),
+        "DYE_PEARLESCENT".toInternalName() to Drop(RareDropAlert.DYE, "PEARLESCENT DYE"),
```
hannibal002: DYE_PEARLESCENT is in both specialDrops maps and neither handler checks the boss, so doesn't the dye now fire two identical titles and two chat lines? The announced set in RareDropAlert used to catch that.

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndRareDropTracker.kt
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    @HandleEvent(onlyOnIsland = IslandType.THE_END)
+    private fun onRender(event: GuiRenderEvent) {
+        if (!config.enabled) return
+        val data = storage ?: return
+        val lines = buildLines(data)
+        if (lines.isEmpty()) return
+        config.position.renderRenderable(
+            Renderable.vertical(lines, spacing = LINE_SPACING)
```
hannibal002: The other three HUDs moved their build into a tick handler, this one still rebuilds every frame. Same treatment here?

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndLootScanner.kt
```diff
@@ -0,0 +1,190 @@
...
+        }
+    }
+
+    /**
+     * Second, slower path: drops that were too far away to ever be rendered still end up in the
+     * inventory when they are collected automatically. Reporting them late beats not at all, and a
+     * drop already reported from its stand is filtered out by [report].
+     */
+    @HandleEvent(onlyOnIsland = IslandType.THE_END)
+    private fun onItemAdd(event: ItemAddEvent) {
```
hannibal002, line 105: Should this be limited to Source.ITEM_ADD? Sacks, Hunting Box and the command source all post ItemAddEvent too.

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
```diff
@@ -0,0 +1,272 @@
...
+            if (isDragon) currentHp = group("hp").formatIntOrNull()
+        }
+        val damageLine = lines.getOrNull(index + 1) ?: return
+        scoreboardDamagePattern.matchMatcher(damageLine) {
+            DragonFightState.yourDamage = group("damage").formatDouble()
+        }
+    }
+
+    @HandleEvent
+    private fun onWorldChange(event: WorldChangeEvent) {
```
hannibal002: unused parameter

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
```diff
@@ -0,0 +1,272 @@
...
+            endingBoss = null
+            endTopDamage = 0.0
+            currentType = null
+            currentHp = null
+            if (boss == EndBoss.DRAGON) DragonFightState.reset()
+        }
+    }
+
+    @HandleEvent(onlyOnIsland = IslandType.THE_END)
+    private fun onTabList(event: WidgetUpdateEvent) {
```
hannibal002: rename the function name to onWidgetUpdate

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
```diff
@@ -0,0 +1,272 @@
...
+    /**
+     * REGEX-TEST: Your Damage: 2,003.2
+     */
+    private val scoreboardDamagePattern by scoreboardGroup.pattern(
+        "your-damage.colorless",
+        "Your Damage: (?<damage>[\\w,.]+)",
+    )
+
+    private val nestAreaPattern by group.pattern("area.nest", "Dragon's Nest")
+    // </editor-fold>
```
hannibal002: no opening tag anywhere in the file

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
```diff
@@ -0,0 +1,272 @@
...
+        .joinToString("|") { it.name.firstLetterUppercase() }
+
+    // Keys that beta already uses for the older coloured patterns carry a .colorless suffix. Their
+    // regex changed, and an older client picking up the new one under the old key would break.
+    private val group = RepoPattern.group("combat.end-dragon-fight")
+    private val chatGroup = group.group("chat")
+    private val endGroup = group.group("chat.end")
+    private val scoreboardGroup = group.group("scoreboard")
+
+
```
hannibal002: three empty lines

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndLootScanner.kt
```diff
@@ -0,0 +1,190 @@
...
+     * Boss loot exists only as an armor stand with a floating label - measured in game, it has
+     * no item entity behind it. Scanning item entities was tried and only ever turned up
+     * unrelated ground loot from other players, so the label is the one reliable source.
+     *
+     * The stand's equipment is still preferred when present, because a stack names its item
+     * exactly, including a pet's rarity, without any text parsing.
+     */
+    @OptIn(AllEntitiesGetter::class)
+    private fun scanNameplates(boss: EndBoss) {
+        for (entity in EntityUtils.getEntities<ArmorStand>()) {
```
hannibal002: this will check all armor stand entries in the world on each tick. do the check less frequent, and only in a specific radius/box around the player.

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndLootScanner.kt
```diff
@@ -0,0 +1,190 @@
...
+
+            if (fromStack == null && label == null) continue
+            seenDrops.add(entity.uuid)
+
+            val internalName = fromStack
+                ?: label?.let { resolvePet(it) ?: NeuInternalName.fromItemNameOrNull(it) }
+            if (internalName == null) continue
+
+            val amount = carried?.count?.takeIf { it > 1 }
+                ?: label?.split("§8x")?.last()?.toIntOrNull()
```
hannibal002: instead of split, use regex patterns

src/main/java/at/hannibal2/skyhanni/features/combat/end/golem/GolemWeight.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    @HandleEvent(onlyOnIsland = IslandType.THE_END)
+    private fun onChat(event: SkyHanniChatEvent.Allow) {
+        val result = pendingResult ?: return
+        if (pendingSince.passedSince() > PENDING_TIMEOUT) {
+            pendingResult = null
+            return
+        }
+        zealotsPattern.matchMatcher(event.cleanMessage) {
+            val zealots = group("amount").toInt()
```
hannibal002: use `formatInt` instead

src/main/java/at/hannibal2/skyhanni/features/combat/end/golem/GolemWeight.kt
```diff
@@ -0,0 +1,117 @@
...
+     * When that result arrived. The zealot line follows within a moment, so anything later belongs
+     * to a different fight - without this, a summary whose zealot line never came would be paired
+     * with the next fight's line and produce a weight from two different fights.
+     */
+    private var pendingSince = SimpleTimeMark.farPast()
+
+    private val PENDING_TIMEOUT = 10.seconds
+
+    /** Weight of the last finished protector fight, read by the dry streak tracker. */
+    var weight = 0.0
```
hannibal002: Why is this a member of golem weight? Can be moved inside onChat.

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
```diff
@@ -0,0 +1,272 @@
...
+    private val dragonSpawnPattern by chatGroup.pattern(
+        "spawn.colorless",
+        "☬ The (?<dragon>$dragonNames) Dragon has spawned!",
+    )
+
+    /**
+     * REGEX-TEST: ☬ You placed a Summoning Eye! (2/8)
+     * REGEX-TEST: ☬ You placed a Summoning Eye! Brace yourselves! (8/8)
+     */
+    @Suppress("MaxLineLength")
```
hannibal002: unnecessary suppress

src/main/java/at/hannibal2/skyhanni/features/combat/end/golem/GolemSpawnTimer.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    /** A server tick is 50ms by definition - the count itself already carries any lag. */
+    private const val MILLIS_PER_TICK = 50.0
+
+    /**
+     * REGEX-TEST: ☬ The ground begins to shake as an End Stone Protector rises from below!
+     */
+    private val risingPattern by repoGroup.pattern(
+        "chat.rising",
+        ".*The ground begins to shake as an End Stone Protector rises from below!.*",
```
hannibal002: this pattern will match too much. e.g. when a user writes in chat `The ground begins to shake as an End Stone Protector rises from below!`. please use the exact same pattern as the chat message hypixel sends, to avoid false positives.

src/main/java/at/hannibal2/skyhanni/features/combat/end/golem/GolemLocation.kt
```diff
@@ -0,0 +1,131 @@
...
+    /** Name of the spawn point the golem is currently at, null while none is known. */
+    fun currentLocationText(): String? = activePoint?.label
+
+    /**
+     * The golem's head sits in the ground as a player head block long before it rises, so the
+     * spawn point can be named immediately instead of only once the entity exists.
+     */
+    private fun detectFromHeadBlock(): SpawnPoint? = spawnPoints.firstOrNull { point ->
+        (0..HEAD_SCAN_HEIGHT).any { dy ->
+            val position = point.head.up(dy.toDouble())
```
hannibal002: toDouble is unnecessary

src/main/java/at/hannibal2/skyhanni/features/combat/end/golem/GolemLocation.kt
```diff
@@ -0,0 +1,131 @@
...
+
+    /** Name of the spawn point the golem is currently at, null while none is known. */
+    fun currentLocationText(): String? = activePoint?.label
+
+    /**
+     * The golem's head sits in the ground as a player head block long before it rises, so the
+     * spawn point can be named immediately instead of only once the entity exists.
+     */
+    private fun detectFromHeadBlock(): SpawnPoint? = spawnPoints.firstOrNull { point ->
+        (0..HEAD_SCAN_HEIGHT).any { dy ->
```
hannibal002: what does dy stand for? rename please

src/main/java/at/hannibal2/skyhanni/features/combat/end/golem/GolemLocation.kt
```diff
@@ -0,0 +1,131 @@
...
+@SkyHanniModule
+object GolemLocation {
+
+    private val config get() = SkyHanniMod.feature.combat.endIsland.golem
+
+    /**
+     * Deliberately one word: the entity is named "Endstone Protector", while the chat messages
+     * around the fight spell it "End Stone Protector". Both spellings are current.
+     */
+    internal const val GOLEM_NAME = "Endstone Protector"
```
hannibal002, line 34: MobFinder already has this exact literal for the same mob, worth sharing one constant?

src/main/java/at/hannibal2/skyhanni/features/combat/end/RareDropAlert.kt
```diff
@@ -0,0 +1,67 @@
...
+import at.hannibal2.skyhanni.utils.PlayerUtils
+import kotlin.time.Duration.Companion.seconds
+
+/**
+ * Shared presentation for rare End island drops: a red "DROP!" headline with the item underneath
+ * in its rarity colour, plus a chat line carrying its market value. Used by both the dragon and
+ * the protector alerts so the two cannot drift apart in wording or styling.
+ */
+@SkyHanniModule
+object RareDropAlert {
```
hannibal002: With announced gone this object has no event handlers and no repo patterns left, is the annotation still needed?

hannibal002: Also the class name is too generic

src/main/java/at/hannibal2/skyhanni/utils/renderables/decorators/TitledFrameRenderable.kt
```diff
@@ -0,0 +1,95 @@
...
+        private const val DEFAULT_BACKGROUND = 0xB0101014.toInt()
+
+        fun Renderable.withTitledFrame(
+            title: Renderable,
+            accentColor: Int = DEFAULT_ACCENT,
+            backgroundColor: Int = DEFAULT_BACKGROUND,
+            padding: Int = DEFAULT_PADDING,
+        ) = TitledFrameRenderable(this, title, accentColor, backgroundColor, padding)
+
+        fun Renderable.withTitledFrame(
```
hannibal002: overload without callers. delete?

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightState.kt
```diff
@@ -0,0 +1,49 @@
...
+    fun reset() {
+        eyesPlaced = 0
+        dragonSpawned = false
+        dragonType = null
+        yourDamage = 0.0
+    }
+
+    @HandleEvent
+    private fun onConfigFix(event: ConfigUpdaterMigrator.ConfigFixEvent) {
+        event.move(78, "combat.dragon", "combat.endIsland.dragon")
```
hannibal002: The End island migrations now sit in two unrelated feature classes. Would both be better off in EndIslandConfig?

src/main/java/at/hannibal2/skyhanni/config/features/combat/end/GolemConfig.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    @Expose
+    @ConfigOption(name = "Show Location", desc = "Show at which fixed point the protector spawns.")
+    @ConfigEditorBoolean
+    var showLocation: Boolean = true
+
+    @Expose
+    @ConfigOption(
+        name = "Show Spawn Countdown",
+        desc = "Count down the 20 seconds until the protector can be attacked."
```
hannibal002: GolemSpawnTimer.SPAWN_DELAY_TICKS talks about 21 seconds.

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndRareDropTracker.kt
```diff
@@ -0,0 +1,264 @@
...
+        data.dryDragonChance = 0.0
+    }
+
+    /**
+     * Counted on the fight result rather than on the kill, because only then is the weight known.
+     */
+    @HandleEvent
+    private fun onEndBossFightEnd(event: EndBossFightEndEvent) {
+        val data = storage ?: return
+        when (event.boss) {
```
hannibal002: an `if` is probably better suited here

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndRareDropTracker.kt
```diff
@@ -0,0 +1,264 @@
...
+                val weight = DragonWeight.calculateWeight(
+                    DragonFightState.eyesPlaced,
+                    event.place,
+                    event.topDamage,
+                    event.yourDamage,
+                )
+                if (weight < DRAGON_WEIGHT_REQUIREMENT) return
+                data.dryDragons++
+                data.dryDragonChance += withPetBonus(
+                    DragonFightState.eyesPlaced * (LEGENDARY_PET_CHANCE_PER_EYE + EPIC_PET_CHANCE_PER_EYE),
```
hannibal002: this reads a bit confusing. please use an explicit name for the patial calculation step. or even better, remove the parameter and do the line inside withPetBonus. All objects are accessible there too

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndRareDropTracker.kt
```diff
@@ -0,0 +1,264 @@
...
+            EndBoss.DRAGON -> {
+                val weight = DragonWeight.calculateWeight(
+                    DragonFightState.eyesPlaced,
+                    event.place,
+                    event.topDamage,
+                    event.yourDamage,
+                )
+                if (weight < DRAGON_WEIGHT_REQUIREMENT) return
+                data.dryDragons++
+                data.dryDragonChance += withPetBonus(
```
hannibal002: does this chance go above 100%?

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
```diff
@@ -0,0 +1,272 @@
...
+
+    /** Live damage leaderboard of the running fight, highest first. */
+    var damageEntries: List<DamageEntry> = emptyList()
+        private set
+
+    /** Damage of the current first place, 0.0 while unknown. */
+    val topDamage: Double get() = damageEntries.firstOrNull()?.damage ?: 0.0
+
+    /** Own placement (1-based) among the listed entries, null while not listed. */
+    val ownPlace: Int?
```
hannibal002: Before this pr, we had this capped with `if (i > 3) null`, which reads like the widget can list more entries than it has real placements for. Was the cap wrong, or does this now report a wrong place?

hannibal002 (changes requested):
many super small things

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndRareDropTracker.kt
hordinii: Yes

src/main/java/at/hannibal2/skyhanni/features/combat/end/golem/GolemLocation.kt
hordinii: dy stood for difference in y. 
i will rename it to heightOffset

src/main/java/at/hannibal2/skyhanni/features/combat/end/golem/GolemSpawnTimer.kt
hordinii: Good Point !

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndLootScanner.kt
hordinii: I would advise against specifying a dedicated radius for this. The radius would need to be large enough for the drop to scan from one corner to the other where the player is standing, which ultimately ends up covering almost the entire island anyway. That's why I preferred reducing the scan rate to every 5 ticks instead.

lunaynx: You should be able to use `EntityCustomNameUpdateEvent`, `EntityEquipmentChangeEvent`, and/or `DataWatcherUpdatedEvent` (the last one being the broadest encompassing all `SynchedEntityData` updates in one event). Getting all entities unbounded is pretty much never the right solution except for one-off config reloads and such.

hordinii: Fair point. I've been running this code in the End since I opened the PR, and the nameplate scan
has been reliable the whole time: not a single drop it missed, and no armor stand it
failed to pick up.
1 important thing about the scan itself: 
It never runs permanently. It only runs while a boss died within the last 60 seconds, and in the update I'm about to push it is throttled to every 5th tick inside that window. Outside of it, nothing is scanned at all.

On the event: Both strategies read the same packets, so they wouldn't widen what we can see - the
label only arrives once you are within ~25 blocks either way. The difference is that
they only report changes from the moment you subscribe, while the sweep also sees the
state that accumulated before the death message arrived. So the initial scan would have
to stay regardless.

I do think the packet-driven variant is a good alternative and would work well. If I
implement it, I'd add it as the second path and keep the scan as the first one, running
once when the boss dies. Happy to do that if you'd rather have it in this PR - otherwise
I'd prefer to keep the current implementation, since it has proven itself in practice.

lunaynx: "It has proven itself in practice" is not a good reason to keep running an expensive call. If you *really* need to, a `getAllEntities` only on boss death rather than periodically is still better than the current state, but I'd rather look into purely using the event and then early-return if not within the "X seconds after boss death" time period.

hannibal002: Please use `EntityCustomNameUpdateEvent<ArmorStand>` and `EntityEquipmentChangeEvent<ArmorStand>` instead of tick based entity reading. for this kind of loot detection, we can even abstract it an reuse the utils for other cases, e.g. dungeons or canyon (behemooth)

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndLootScanner.kt
```diff
@@ -0,0 +1,142 @@
...
+
+    @HandleEvent(onlyOnIsland = IslandType.THE_END)
+    private fun onEntityNameUpdate(event: EntityCustomNameUpdateEvent<ArmorStand>) {
+        inspect(event.entity)
+    }
+
+    /** What a stand carries is read by [LootStandUtils]; which boss it belongs to is decided here. */
+    private fun inspect(stand: ArmorStand) {
+        // Every stand in the End reports here, so anything outside a loot window leaves at once.
+        val boss = mostRecentBoss() ?: return
```
hannibal002, line 89: A stand whose packet arrives before the DOWN message is dropped here, which the sweep would still have caught. Did you re-test a few kills after the switch, especially the armor and the pet?

src/main/java/at/hannibal2/skyhanni/features/combat/end/golem/GolemStage.kt
```diff
@@ -0,0 +1,102 @@
...
+     * too, in which case the elapsed time is measured from that moment rather than from the
+     * stage actually beginning - the server does not tell us when that was.
+     */
+    private var awakeningSince: SimpleTimeMark? = null
+
+    /** How long the protector has been awakening, or null outside that stage. */
+    fun timeInAwakening(): Duration? = awakeningSince?.passedSince()
+
+    @HandleEvent(onlyOnIsland = IslandType.THE_END)
+    private fun onTabList(event: WidgetUpdateEvent) {
```
hannibal002: ▎ This one still has the old name, onWidgetUpdate here too.

src/main/java/at/hannibal2/skyhanni/events/EndLootFoundEvent.kt
```diff
@@ -0,0 +1,16 @@
+package at.hannibal2.skyhanni.events
+
+import at.hannibal2.skyhanni.api.event.SkyHanniEvent
+import at.hannibal2.skyhanni.skyhannimodule.PrimaryFunction
+import at.hannibal2.skyhanni.utils.NeuInternalName
+
+/**
+ * Sent for every drop that appears after an End boss died, read from its floating item label
```
hannibal002: "read from its floating item label rather than from picking it up" no longer holds now that onItemAdd posts this as well.

src/main/java/at/hannibal2/skyhanni/features/combat/end/golem/GolemDisplay.kt
```diff
@@ -0,0 +1,137 @@
...
+            )
+        }
+        if (state.showLocation) {
+            // Before the protector starts rising there is no head block to read the spot from.
+            val location = state.location?.let { "§f$it" } ?: "§8not revealed"
+            add(Renderable.text("§7Location: $location"))
+        }
+        if (state.showSpawnTimer) {
+            // The line keeps its place while the protector is still awakening, so it does not
+            // pop into existence only for the last 20 seconds.
```
hannibal002: ▎ Says 20 seconds, the config and the timer now say 21.

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightState.kt
```diff
@@ -0,0 +1,36 @@
+package at.hannibal2.skyhanni.features.combat.end.dragon
+
+
```
hannibal002, line 3: two blank lines

src/main/java/at/hannibal2/skyhanni/config/features/combat/end/DragonConfig.kt
```diff
@@ -14,35 +14,32 @@ class DragonConfig {
...
-        name = "Weight HUD",
-        desc = "Shows your current dragon weight on the HUD and if hovered shows the breakdown." +
+        name = "Damage List HUD",
+        desc = "Shows a list of all damage sources and their damage during the dragon fight." +
             " The dragon widget needs to be enabled for this to work."
     )
     @ConfigEditorBoolean
     @FeatureToggle
-    var display: Boolean = false
+    var damageList: Boolean = false
```
hannibal002, line 25: The display / "Weight HUD" option that stood here is not moved, it is replaced: damageList takes its place and displayPosition becomes damageListPosition. DragonWeight lost its onTick and onRender too, so only the chat message is left. This is the live class behind combat.endIsland.dragon, the Weight HUD in config/features/combat/DragonConfig.kt is the orphaned one from the v78 move and does not count.

Why is the Weight HUD going away? If it stays removed, display and displayPosition need a migration or an explicit drop, plus a ## Changelog Removed Features entry, and the "Added a Dragon Weight HUD" line in the changelog has to go.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndLootScanner.kt
hordinii: I tried it out for a few kills, and it worked quite well. Unfortunately getting the pet as a drop is pretty difficult. I’ve invested 1b so far without getting the pet, which is well below average
But the armor, aotd, dragon claw and dragon scale work quiet well.

src/main/java/at/hannibal2/skyhanni/config/features/combat/end/DragonConfig.kt
hordinii: Good point and this one comes straight from testing. I'm in the End more or less daily
while working on this.

A live weight display sounds useful at first, but it doesn't hold up in practice. A dragon
today isn't a 20 second fight, it's a 0.2 to 1 second one, and the HUD never gets to show a
meaningful number in that window. And even when a dragon does live longer, knowing your
weight mid-fight buys you nothing - there is no decision you can make with it while the
fight is still running. So I dropped the live display and kept the weight where it is
actually readable, in chat, once the fight is over.

In the meantime, I am developing additional features for the end. Especially a DPS Tracker, Zealot counter and Dragons summoned display will be next.

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndLootScanner.kt
hordinii: I'll test some specific max range kills later on to make sure everything works 100%

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightState.kt
hannibal002, line 3: two empty lines, please remove one

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndLootScanner.kt
hannibal002, line 89: Reopening until the max range kills you mentioned are done. This is the one spot where the switch from the sweep to the events can lose a drop.

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
```diff
@@ -0,0 +1,260 @@
...
+    /**
+     * REGEX-TEST: ☬ The Dragon Egg has spawned!
+     */
+    private val eggSpawnedPattern by chatGroup.pattern("egg.spawn", "☬ The Dragon Egg has spawned!")
+
+    /**
+     * WRAPPED-REGEX-TEST: "                          PROTECTOR DRAGON DOWN!"
+     * WRAPPED-REGEX-TEST: "                          YOUNG DRAGON DOWN!"
+     */
+    private val dragonDownPattern by endGroup.pattern(
```
hannibal002, line 74: dragonSpawnPattern builds its names from DragonType, could this do the same? Beta had an uppercase variant of that list for exactly this pattern.

src/main/java/at/hannibal2/skyhanni/features/combat/end/dragon/DragonFightAPI.kt
```diff
@@ -0,0 +1,260 @@
...
+import at.hannibal2.skyhanni.utils.StringUtils.removeColor
+import at.hannibal2.skyhanni.utils.repopatterns.RepoPattern
+
+/**
+ * Single data layer for the End boss fights: owns every regex, reads chat, scoreboard and tab
+ * list, keeps [DragonFightState] up to date and publishes the result as [EndBossFightEndEvent].
+ * Features only consume this and never parse anything themselves.
+ */
+@SkyHanniModule
+object DragonFightAPI {
```
hannibal002, line 29: please change the spelling of `API` to `Api`.

src/main/java/at/hannibal2/skyhanni/config/features/combat/end/DragonConfig.kt
hannibal002: Let's keep both for now. Removing options that are on by default for everyone is its own decision and does not belong in a PR about End features, so please add them back and open a separate PR if you still want them gone. For the prefix there is also a practical reason: with prefix = false hardcoded the weight message is indistinguishable from Hypixel's own summary, so nobody can tell which mod to disable.

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndRareDropTracker.kt
```diff
@@ -0,0 +1,303 @@
...
+        @Expose var legendaryPets: Int = 0
+        @Expose var epicPets: Int = 0
+        @Expose var tierBoostCores: Int = 0
+
+        /** Eligible fights since the last drop of the matching kind. */
+        @Expose var dryDragons: Int = 0
+        @Expose var dryGolems: Int = 0
+
+        /** Summed drop chance of those fights, in percent. */
+        @Expose var dryDragonChance: Double = 0.0
```
hannibal002, line 77: Summing the per fight chances reads as a percentage but is not one, it passes 100% after a few hundred fights. Could this be the combined probability instead, chance = 1 - (1 - chance) * (1 - newChance)? Storing the complement keeps 0.0 as the neutral value, which matters because Gson skips the field initialisers. Nothing is stored yet, so it is free to change.

hannibal002 (changes requested):

hannibal002:
Also please note i have updated the changelog. Feel free to edit something back if you dont like the changes.

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndLootScanner.kt
hordinii: ive tested the max range kills a lot in the last day and everything is working as planned.
Luckily ive dropped a pet to notice another bug. 
Every rarity of a pet uses the same skull, so resolving the item could only guess between
epic and legendary and ended up matching nothing. The rarity is now read from the colour
in the item's own name, which still arrives with the equipment packet, so the range stays.
Im about to test it rn but as i mentioned before, the chance for a pet drop is pretty low.

src/main/java/at/hannibal2/skyhanni/features/combat/end/EndRareDropTracker.kt
hordinii: Honestly i'd rather keep this one as it is. It is cumulative on purpose: it sums the chance of every
fight in the streak, so passing 100% is the point. That is what makes a long dry streak
visible at a glance. Standing right behind the kill count it reads unambiguously, whereas a
value that saturates near 100% no longer tells a 300 kill streak apart from a 900 one.
I'm a big fan of showing dry streaks as a percentage. It is quicker to read what the number
means, especially on sharing in the chat.
What do u think about that. Maybe luna could share her thoughts about that too.

PR #6518, opened 2026-09-07
Fix: Clip 3D lines to the camera near plane

lunaynx:
## What
SkyHanni line rendering on 26.2 has a weird bug where certain lines we render are randomly wider and jump around even when the camera and the player remains still. This may be somewhat dependent on the GPU and rendering backend, as I was able to reproduce this on my MacBook Air M1 with OpenGL but not with Vulkan. After this change, OpenGL is fixed and Vulkan continues to work fine.

<details>
<summary>Video</summary>

https://github.com/user-attachments/assets/a399bb4f-ad4d-4f5f-9f88-a257771beac2
</details>

## Changelog Fixes
+ Fixed lines drawn by SkyHanni sometimes being wider than intended and jumping around even though they're not supposed to move. - Luna

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/render/LineDrawer.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/render/LineDrawer.kt
@@ -5,7 +5,9 @@ import at.hannibal2.skyhanni.utils.LocationUtils.calculateEdges
 import at.hannibal2.skyhanni.utils.LorenzVec
 import at.hannibal2.skyhanni.utils.collection.CollectionUtils.zipWithNext3
 import at.hannibal2.skyhanni.utils.render.WorldRenderUtils.submitCustomGeometry
+import at.hannibal2.skyhanni.utils.toLorenzVec
 import net.minecraft.world.phys.AABB
+import org.joml.Vector3f
 import java.awt.Color
 
 class LineDrawer @PublishedApi internal constructor(
@@ -14,13 +16,17 @@ class LineDrawer @PublishedApi internal constructor(
     val depth: Boolean,
 ) {
     private val queuedLines = mutableListOf<QueuedLine>()
+    private val cameraPos = event.camera.pos.toLorenzVec()
+    private val cameraLook = Vector3f(0f, 0f, -1f).rotate(event.camera.orientation).let {
+        LorenzVec(it.x.toDouble(), it.y.toDouble(), it.z.toDouble())
+    }
 
     @PublishedApi
     internal fun drawQueuedLines() {
         if (queuedLines.isEmpty()) return
 
         val layer = SkyHanniRenderLayers.getLines(!depth)
-        val lines = queuedLines.toList()
+        val lines = queuedLines.mapNotNull { it.clipToCamera() }
         event.submitCustomGeometry(layer) { pose, buf ->
             fun QueuedLine.addVertexForPoint(point: LorenzVec) {
                 buf.addVertex(pose.pose(), point.x.toFloat(), point.y.toFloat(), point.z.toFloat())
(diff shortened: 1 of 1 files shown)


src/main/java/at/hannibal2/skyhanni/utils/render/LineDrawer.kt
```diff
@@ -14,13 +16,17 @@ class LineDrawer @PublishedApi internal constructor(
...
 ) {
     private val queuedLines = mutableListOf<QueuedLine>()
+    private val cameraPos = event.camera.pos.toLorenzVec()
+    private val cameraLook = Vector3f(0f, 0f, -1f).rotate(event.camera.orientation).let {
+        LorenzVec(it.x.toDouble(), it.y.toDouble(), it.z.toDouble())
+    }
 
     @PublishedApi
     internal fun drawQueuedLines() {
         if (queuedLines.isEmpty()) return
```
hannibal002, line 26: ▎ isEmpty check should move after the clipping, otherwise an all clipped batch still submits.

src/main/java/at/hannibal2/skyhanni/utils/render/LineDrawer.kt
```diff
@@ -38,6 +44,18 @@ class LineDrawer @PublishedApi internal constructor(
         queuedLines.clear()
     }
 
+    // The line shader expands each vertex into a quad edge that folds over itself once the vertex is behind the
+    // camera, and which of the two folded shapes gets rasterized depends on gl_VertexID parity. Since 26.2 all
+    // world geometry shares one vertex buffer with a per-frame base vertex, so that parity flickers between frames.
+    private fun QueuedLine.clipToCamera(): QueuedLine? {
```
hannibal002: ▎ Use camera.position and camera.rotation() from CameraRenderStateCompat, that is what it exists for.

src/main/java/at/hannibal2/skyhanni/utils/render/LineDrawer.kt
```diff
@@ -14,13 +16,17 @@ class LineDrawer @PublishedApi internal constructor(
     val depth: Boolean,
 ) {
     private val queuedLines = mutableListOf<QueuedLine>()
+    private val cameraPos = event.camera.pos.toLorenzVec()
+    private val cameraLook = Vector3f(0f, 0f, -1f).rotate(event.camera.orientation).let {
```
hannibal002, line 20: compare with `SkyHanniRenderWorldEvent.exactPlayerCrosshairLocation` in WorldRenderUtils. the logic looks identical. can we reuse existing code?

src/main/java/at/hannibal2/skyhanni/utils/render/LineDrawer.kt
```diff
@@ -14,13 +16,17 @@ class LineDrawer @PublishedApi internal constructor(
     val depth: Boolean,
 ) {
     private val queuedLines = mutableListOf<QueuedLine>()
+    private val cameraPos = event.camera.pos.toLorenzVec()
```
hannibal002, line 19: Use camera.position and camera.rotation() from CameraRenderStateCompat, that is what it exists for.

src/main/java/at/hannibal2/skyhanni/utils/render/LineDrawer.kt
hannibal002, line 50: Did all the jumping lines in the video cross the camera plane, or were some fully in front of it?

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/render/LineDrawer.kt
lunaynx: I considered that, but that class is essentially a leftover from back when we still supported 1.21.11. It basically only exists because `SkyHanniRenderWorldEvent` carried `Camera` on 1.21.11 but `CameraRenderState` on 26.1+, and the two objects have different ways to access the same information.

I'd say it has little to no reason to exist today. It's only used in a single file (`WorldRenderUtils`) and it's cleaner to just delete the imports and use the intended fields directly. It also doesn't follow the "no top-level functions" rule. The only reason it wasn't removed with the 26.2 port in the end is because I was trying to aggressively minimize the diff to get the PR merged ASAP.

If you *really* want a `CameraUtils` or `CameraCompat` class to stay, I'd rewrite it slightly so that we have a consistent interface (`.position`/`.rotation` fields, not methods) on both `Camera` and `CameraRenderState`, but this feels like an unnecessary abstraction at this time – IDE autocomplete is free.

hannibal002: no thats fine as is. the 1.21.11 leftover code should prob get a comment or something

PR #6525, opened 2026-09-08
  Backend: Split up StringUtils and TextHelper

hannibal002:
## What
Trying to split up large utils files.

+ Split up StringUtils and TextHelper. - hannibal2
    * Moved chat component helpers into the new ChatComponentUtils.
    * Moved the paginated chat list from TextHelper into the new PaginatedListHelper.
    * Moved font-dependent text helpers from StringUtils into TextHelper.
    * Kept the old names as deprecated aliases until November 2026.

### DIFF
--- a/src/main/java/at/hannibal2/skyhanni/utils/chat/ChatComponentUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/chat/ChatComponentUtils.kt
@@ -48,23 +48,23 @@ object ChatComponentUtils {
                         val formattingChar = component.unformattedTextForChatCompat()[idx++]
                         val formatting = enumChatFormattingByCode(formattingChar) ?: continue
                         when (formatting) {
-                            ChatFormatting.OBFUSCATED -> {
+                            OBFUSCATED -> {
                                 colorOverride.withObfuscated(true)
                             }
 
-                            ChatFormatting.BOLD -> {
+                            BOLD -> {
                                 colorOverride.withBold(true)
                             }
 
-                            ChatFormatting.STRIKETHROUGH -> {
+                            STRIKETHROUGH -> {
                                 colorOverride.withStrikethrough(true)
                             }
 
-                            ChatFormatting.UNDERLINE -> {
+                            UNDERLINE -> {
                                 colorOverride.withUnderlined(true)
                             }
 
-                            ChatFormatting.ITALIC -> {
+                            ITALIC -> {
                                 colorOverride.withItalic(true)
                             }
 
--- a/src/main/java/at/hannibal2/skyhanni/utils/chat/ChatComponentUtils.kt
+++ b/src/main/java/at/hannibal2/skyhanni/utils/chat/ChatComponentUtils.kt
@@ -1,0 +1,137 @@
+package at.hannibal2.skyhanni.utils.chat
+
+import at.hannibal2.skyhanni.data.hypixel.chat.event.SystemMessageEvent
+import at.hannibal2.skyhanni.utils.chat.TextHelper.asComponent
+import at.hannibal2.skyhanni.utils.compat.command
+import at.hannibal2.skyhanni.utils.compat.defaultStyleConstructor
+import at.hannibal2.skyhanni.utils.compat.formattedTextCompat
+import at.hannibal2.skyhanni.utils.compat.hover
+import at.hannibal2.skyhanni.utils.compat.unformattedTextForChatCompat
+import at.hannibal2.skyhanni.utils.compat.value
+import net.minecraft.ChatFormatting
+import net.minecraft.network.chat.ClickEvent
+import net.minecraft.network.chat.Component
+import net.minecraft.network.chat.HoverEvent
+import net.minecraft.network.chat.Style
+
+object ChatComponentUtils {
+
+    private val colorMap = ChatFormatting.entries.associateBy { it.toString()[1] }
+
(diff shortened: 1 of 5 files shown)


src/main/java/at/hannibal2/skyhanni/utils/chat/ChatComponentUtils.kt
```diff
@@ -0,0 +1,137 @@
...
+                        colorOverride = defaultStyleConstructor
+                        component = queue.removeFirstOrNull() ?: return null
+                    }
+                    val char = component.unformattedTextForChatCompat()[idx++]
+                    if (char == '§' && idx < component.unformattedTextForChatCompat().length) {
+                        val formattingChar = component.unformattedTextForChatCompat()[idx++]
+                        val formatting = enumChatFormattingByCode(formattingChar) ?: continue
+                        when (formatting) {
+                            ChatFormatting.OBFUSCATED -> {
+                                colorOverride.withObfuscated(true)
```
AverageUser125, line 52: Context sensitive thing with enums

hannibal002: Good catch, fixed.

src/main/java/at/hannibal2/skyhanni/utils/chat/ChatComponentUtils.kt
```diff
@@ -0,0 +1,137 @@
...
+    ) {
+        val original = chatComponent.formattedTextCompat()
+        val new = transform(original)
+        if (new == original) return
+
+        val clickEvents = mutableListOf<ClickEvent>()
+        val hoverEvents = mutableListOf<HoverEvent>()
+        chatComponent.findAllEvents(clickEvents, hoverEvents)
+
+        if (clickEvents.size > 1 || hoverEvents.size > 1) return
```
AverageUser125, line 107: Shouldn’t this just be an `isEmpty` check?

src/main/java/at/hannibal2/skyhanni/utils/chat/PaginatedList.kt
```diff
@@ -0,0 +1,90 @@
...
+     * @param formatter A function to format each entry into a Component.
+     */
+    fun <T> displayPaginatedList(
+        title: String,
+        list: List<T>,
+        chatLineId: Int,
+        emptyMessage: String,
+        currentPage: Int = 1,
+        maxPerPage: Int = 15,
+        dividerColor: ChatFormatting = ChatFormatting.BLUE,
```
AverageUser125:
```suggestion
        dividerColor: ChatFormatting = BLUE,
```

src/main/java/at/hannibal2/skyhanni/utils/chat/PaginatedList.kt
```diff
@@ -0,0 +1,90 @@
...
+        formatter: (T) -> Component,
+    ): Unit = DelayedRun.runOrNextTick("paginated list: $title") {
```
AverageUser125, line 42:
```suggestion
    ) = DelayedRun.runOrNextTick("paginated list: $title") {
```
I don’t think writing return type when it’s `Unit` is standard

src/main/java/at/hannibal2/skyhanni/utils/chat/TextHelper.kt
```diff
@@ -87,6 +97,59 @@ object TextHelper {
...
+        }
+    }
+
+    fun String.splitLines(width: Int): String = splitText(
+        this,
+        width,
+    ).joinToString("\n") { it.removePrefix("§r") }
+
+    private fun splitText(text: String, width: Int): List<String> {
+        val lines = ComponentRenderUtils.wrapComponents(Component.literal(text), width, Minecraft.getInstance().font)
```
AverageUser125: Use `asComponent`

AverageUser125 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/chat/TextHelper.kt
AverageUser125: Also the `wrapComponents` should probably not accept the `font` as a parameter, and just always assume `Minecraft.getInstance().font`

src/main/java/at/hannibal2/skyhanni/utils/chat/PaginatedList.kt
AverageUser125: Actually, it’s probably needed here. So nvm

hannibal002: it is not standard, but in this case its needed. otherwise you dont have any idea what the function returns. yes it might be unit because its  omitted. but it might also be something else that you dont know. thats the reason i want explicit return types on public functions all the time. so that this is always clear

src/main/java/at/hannibal2/skyhanni/utils/chat/TextHelper.kt
hannibal002: Done. `wrapComponents` is a vanilla Minecraft method though, so its parameters aren't ours to change.

src/main/java/at/hannibal2/skyhanni/utils/chat/ChatComponentUtils.kt
hannibal002: `> 1` is intentional, only a single click and hover event can be carried over, so this bails out when there are several.

PR #6528, opened 2026-09-10
Fix: Make GraphUtils.updatePlayerPosition null-safe

lunaynx:
## What
Fixes an IllegalStateException in `GraphUtils.updatePlayerPosition` (can be called through `DelayedRun.runDelayed` without any check whether the player exists, and `LocationUtils.playerEyeLocation` throws when it doesn't).

Ideally, `LocationUtils.playerEyeLocation` should be made null-safe (either nullable or returning a zero vector as fallback). But this fix is staying narrow for now to avoid a silent behavior change elsewhere or a bigger refactor.

```
SkyHanni 7.55.0 26.1.2: Delayed task crashed while executing: player is null
 
Caused by java.lang.IllegalStateException: player is null
    at SH.utils.compat.MinecraftCompat.getLocalPlayerOrThrow(MinecraftCompat.kt:71)
    at SH.utils.LocationUtils.playerEyeLocation(LocationUtils.kt:63)
    at SH.utils.GraphUtils.updatePlayerPosition(GraphUtils.kt:174)
    at SH.data.IslandGraphs.processNavigation(IslandGraphs.kt:351)
    at SH.data.IslandGraphs.setNewGraph-BNdquAs(IslandGraphs.kt:316)
    at SH.data.IslandGraphs$reloadFromJson$1.invokeSuspend$lambda$0(IslandGraphs.kt:291)
    at SH.utils.DelayedRun.withErrorHandling$lambda$0$0(DelayedRun.kt:95)
    at SH.utils.DelayedRun.runWithErrorHandling(DelayedRun.kt:79)
    at SH.utils.DelayedRun.withErrorHandling$lambda$0(DelayedRun.kt:95)
    at MC.util.thread.BlockableEventLoop.doRunTask(BlockableEventLoop.java:173)
    at MC.util.thread.ReentrantBlockableEventLoop.doRunTask(ReentrantBlockableEventLoop.java:23)
    at MC.util.thread.BlockableEventLoop.pollTask(BlockableEventLoop.java:147)
    at MC.util.thread.BlockableEventLoop.runAllTasks(BlockableEventLoop.java:127)
    at MC.client.Minecraft.runTick(Minecraft.java:1345)
    at MC.client.Minecraft.run(Minecraft.java:991)
    at java.lang.Thread.run(Thread.java:1474)

Extra data:
label: '<none>'
source: 'DelayedRun.runNextTick'
```

## Changelog Fixes
+ Fixed occasional "player is null" error in chat when switching lobbies during navigation. - Luna

+ Added LocationUtils.playerEyeLocationOrNull. - Luna
  * Safe version of LocationUtils.playerEyeLocation that returns null instead of throwing when the player is null.


hannibal002 (changes requested):
Guard in `updatePlayerPosition` instead, a silent fallback in a shared util affects every caller, not just the crashing one.

lunaynx:
Why are you suggesting to weaken the guard? A generic method like `LocationUtils.playerEyeLocation` asserting that the player will always exist is wrong. I *could* make its return value nullable instead, which would be technically more correct than a fallback to the zero vector, though I doubt it realistically matters for any existing callers.

lunaynx:
You know what, fine. Let's keep the fix focused. But we should refactor it later.

src/main/java/at/hannibal2/skyhanni/utils/GraphUtils.kt
```diff
@@ -167,11 +167,13 @@ object GraphUtils {
...
 
-    var playerPosition = LorenzVec(0, 0, 0)
+    var playerPosition = LorenzVec()
         private set
 
     fun updatePlayerPosition() {
-        playerPosition = LocationUtils.playerEyeLocation().roundToBlock()
+        playerPosition = runCatching {
+            LocationUtils.playerEyeLocation().roundToBlock()
+        }.getOrElse { LorenzVec() }
```
hannibal002: Use `MinecraftCompat.localPlayerOrNull ?: return`, `runCatching` swallows every throwable here, not just the missing player.

hannibal002: Early return instead, a fake position in `playerPosition` is read by every `distanceToPlayer` caller.

hannibal002 (changes requested):
> Why are you suggesting to weaken the guard?

Then add a `playerEyeLocationOrNull()` next to it, like `playerLocationOrNull` two functions above, instead of changing the existing one.

Does this need active navigation? `setNewGraph` calls `processNavigation` unconditionally.

src/main/java/at/hannibal2/skyhanni/utils/GraphUtils.kt
lunaynx: > Use `MinecraftCompat.localPlayerOrNull ?: return`, `runCatching` swallows every throwable here, not just the missing player.

Are you sure that would be safe? This method gets called from two different threads, which makes me worried about TOCTOU race conditions.

lunaynx:
> Then add a `playerEyeLocationOrNull()` next to it, like `playerLocationOrNull` two functions above, instead of changing the existing one.

Sure, I can do that for now. But letting footguns like that exist is not good long term. Ideally, silently unsafe functions should be phased out entirely, or at least made clear via `OrThrow` naming. So definitely worth revisiting in a follow-up PR.

src/main/java/at/hannibal2/skyhanni/utils/GraphUtils.kt
hannibal002: not sure. but either way its better to then have an error log that  proofs what we have instead of eating up the error with getOrElse

src/main/java/at/hannibal2/skyhanni/utils/GraphUtils.kt
```diff
@@ -167,11 +167,11 @@ object GraphUtils {
         return mappedNodes.zipWithNext { a, b -> findShortestDistance(a, b) }.sum()
     }
 
-    var playerPosition = LorenzVec(0, 0, 0)
+    var playerPosition = LorenzVec()
         private set
 
     fun updatePlayerPosition() {
-        playerPosition = LocationUtils.playerEyeLocation().roundToBlock()
+        playerPosition = LocationUtils.playerEyeLocationOrNull()?.roundToBlock() ?: LorenzVec()
```
hannibal002, line 174: Still writes a fake position, please early return instead.

src/main/java/at/hannibal2/skyhanni/utils/LocationUtils.kt
```diff
@@ -59,11 +58,20 @@ object LocationUtils {
...
-        val player = MinecraftCompat.localPlayerOrThrow
+    /**
+     * Returns the player's eye location, or null if there is no active player (e.g. during a world change).
+     */
+    fun playerEyeLocationOrNull(): LorenzVec? {
+        val player = MinecraftCompat.localPlayerOrNull ?: return null
         val vec = player.getLorenzVec()
         return vec.up(player.eyeHeight.toDouble())
     }
+    /**
```
hannibal002, line 69: missing empty line

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/GraphUtils.kt
lunaynx: IT LITERALLY DEFAULTS TO A GODDAMN FAKE POSITION. Fix your shitty code yourself if you're that bothered.

hannibal002: The throw at that exact spot was the design, it tells us when this gets called wrong. A fake position in `playerPosition` just hides that.

The real bug is the call site: `processNavigation` runs a tick after a lobby switch, when there is no player. Checking for null there is the right place, but on that branch it should write nothing and keep the last value instead of substituting one we know is wrong. Everything that reads `playerPosition` gets that value, so the next failure of this kind shows up as wrong navigation instead of an error report we can act on.

You are right that the default is bad too, `LorenzVec(0, 0, 0)` is not a position we ever really have. `playerPosition` should be nullable, so that "we have no position" is a state the callers have to handle instead of a zero vector pretending to be one.

The `OrThrow` accessors exist to surface issues at the call site, that is why I do not want them routed around.

PR #6530, opened 2026-09-11
Backend: Hide Far Entities Using Entity Events instead of an all entity get

Fazfoxy:
## What
Testing currently, threw this PR out while I'm in the middle of booting it up
will make Improvement/Fix if the spark profile shows a noticeable performance diff but idk

exclude_from_changelog


src/main/java/at/hannibal2/skyhanni/features/misc/HideFarEntities.kt
```diff
@@ -19,39 +22,42 @@ import at.hannibal2.skyhanni.utils.EntityUtils.isNpc
...
         if (GlobalRender.renderDisabled) return
         if (!isEnabled()) return
-        if (event.isMod(20)) {
-            updateNeverHide()
-        }
+        updateNeverHide(event.entity)
+    }
 
+    @HandleEvent
+    private fun onEntityMoveEvent(event: EntityMoveEvent<*>) {
```
hannibal002, line 51: this only works if the player never moves.

hannibal002 (changes requested):

PR #6533, opened 2026-09-12
Fix: Zealot Highlight not working during Derpy

AverageUser125:
## What
Noticed this while working on my max health update event PR (#6532)
Just makes it use the value from the event, which already accounts for derpy.

## Changelog Fixes
+ Fixed Zealot Mob Highlighting not working during Derpy. - Avrg


src/main/java/at/hannibal2/skyhanni/features/combat/mobs/MobHighlight.kt
```diff
@@ -122,6 +120,6 @@ object MobHighlight {
         )
     }
 
-    private fun LivingEntity.isZealotOrBruiser() = baseMaxHealth == 13_000 || baseMaxHealth == 65_000 ||
-        baseMaxHealth == 13_000 * 4 || baseMaxHealth == 65_000 * 4 // runic
+    private fun Int.isZealotOrBruiser() = this == 13_000 || this == 65_000 ||
```
hannibal002: An Int is not a Zealot, name it after the value (health) instead.

hannibal002 (changes requested):
> + Fixed Mob Highlighting not working during Derpy. - Avrg

Changelog entry claims all mob highlighting, this only fixes the Zealot highlight.

hannibal002 (approved):

PR #6536, opened 2026-09-12
Fix: Particle Event Hang & Broken Detection

lunaynx:
## What
Currently, SkyHanni posts `ParticleEvent` *twice* for each particle: once on the network thread, and then again on the render thread. Why? Because it injects at `HEAD`, which is *before* the `ensureRunningOnSameThread` guard that checks if it's on the render thread, and if not, schedules itself to be executed again on the render thread, and terminates execution of the method with an exception.

The fact that it's posted on the network thread at all causes issues with features that call methods that get entities, resulting in chat errors and sometimes even causing Minecraft to hang (one user claimed at end of Kuudra runs, but which specific particle hider feature was responsible for it was not revealed).

The fact that it's posted twice breaks some particle detection features such as Motes Orb detection, and possibly makes others less reliable.

Intentional trade-off made in this PR: SkyHanni no longer injects as early as possible to detect particles. The Sodium Extra crisis was averted a while ago, and other mods like Skyblocker inject late for detection as well. A mod that cancels the processing of particle packets to stop them from rendering is a broken mod and it's not our job to bend over backwards and work around every possibly broken mod. Competent mods like Particle Core (and now also Sodium Extra) don't do this, they generally cancel in `ParticleEngine`.

We continue to cancel in `ClientPacketListener` for the time being because we depend on fields that get irreversibly corrupted by being multiplied by a random number, and working around that would require more complex code and would needlessly delay a high priority bugfix. However, I tried to make it as compatible as possible with other mods. The existing cancellation method is actually a bit *less* compatible than the new one. (`ci.cancel()` cancels the whole method's execution, we only cancel the `addParticle` call and then let it carry on – it should be safe, there's nothing after that would cause corruption, this simply lets it reach `RETURN` which is where Skyblocker injects to detect particles.)

## Changelog Fixes
+ Fixed the game sometimes hanging when certain SkyHanni features that hide particles are enabled. - Luna
+ Fixed Highlight Motes Orbs and Hide Particles not working. - Luna
  * Also possibly fixed some other particle-related features working less reliably.
+ Fixed some other mods such as Skyblocker not being able to detect particles hidden by SkyHanni. - Luna


src/main/java/at/hannibal2/skyhanni/mixins/transformers/MixinClientPacketListener.java
```diff
@@ -33,13 +33,24 @@ public boolean shouldShowPacketSizeAndPingCharts(boolean original) {
...
+            value = "INVOKE",
+            target = "Lnet/minecraft/client/multiplayer/ClientLevel;addParticle(Lnet/minecraft/core/particles/ParticleOptions;ZZDDDDDD)V"
+        )
+    )
+    public ParticleOptions wrapAddParticle(
+        ParticleOptions particleOptions,
+        @Cancellable CallbackInfo ci,
+        @Local(argsOnly = true) ClientboundLevelParticlesPacket packet
+    ) {
         if (ParticleUtils.shouldSuppressParticle()) {
```
AverageUser125: voodoo magic? 🥺 
(This is mostly a joke)

lunaynx: Just a bit. https://github.com/LlamaLad7/MixinExtras/wiki/Cancellable

lunaynx: I could have simply rewritten this as a `@WrapOperation`, but this is a smaller diff.

src/main/java/at/hannibal2/skyhanni/utils/ParticleUtils.kt
```diff
@@ -32,10 +38,32 @@ object ParticleUtils {
...
+    ) {
+        if (shouldSuppressParticle()) return
+
+        val event = ParticleChangeEvent(particleOptions, packet)
+        event.post()
+
+        original.call(level, event.particleOptions, overrideLimiter, alwaysShow, x, y, z, xd, yd, zd)
+    }
+
+    private fun shouldSuppressParticle(): Boolean {
```
hannibal002: this resets on first read, but vanilla calls `addParticle` once per `packet.getCount()`, so only the first particle of a batch gets hidden now.

src/main/java/at/hannibal2/skyhanni/utils/ParticleUtils.kt
```diff
@@ -1,12 +1,18 @@
...
 import at.hannibal2.skyhanni.utils.compat.MinecraftCompat
+import com.llamalad7.mixinextras.injector.wrapoperation.Operation
+import net.minecraft.client.multiplayer.ClientLevel
+import net.minecraft.core.particles.ParticleOptions
 import net.minecraft.core.registries.BuiltInRegistries
 import net.minecraft.network.protocol.game.ClientboundLevelParticlesPacket
 import net.minecraft.resources.Identifier
 
 object ParticleUtils {
+    private val cancelled = ThreadLocal.withInitial { false }
```
hannibal002: Write and read both happen on the render thread now, a plain `var` would do.

src/main/java/at/hannibal2/skyhanni/events/ParticleEvent.kt
```diff
@@ -15,7 +15,7 @@ import net.minecraft.core.registries.BuiltInRegistries
  * suppresses the particle later in the packet handler, after other mixins have
  * had a chance to process it.
  *
- * Runs on the network thread.
+ * Runs on the render thread.
```
hannibal002, line 19: Worth adding that this fires once per packet while the packet spawns `count` particles.

hannibal002 (changes requested):
1. Did you confirm in game that the Motes Orb waypoint shows up again after this change?

2. Changelog: Motes Orb Detection is the only feature that actually counted particles, name it here and move "some other features" into a sub bullet.


3. Several ParticleEvent handlers wrap their body in DelayedRun because the event used to run off the main thread, worth a follow up PR once this lands: GriffinBurrowParticleFinder.onParticle, PreciseGuessBurrow.onParticle, ArrowGuessBurrow.onParticle, FishingHotspotRadar.onParticle, and possibly more.

src/main/java/at/hannibal2/skyhanni/utils/ParticleUtils.kt
lunaynx: Good point, time to do the interface injection approach I mentioned in the PR description then.

src/main/java/at/hannibal2/skyhanni/utils/ParticleUtils.kt
lunaynx: Will be resolved by the above.

src/main/java/at/hannibal2/skyhanni/events/ParticleEvent.kt
lunaynx: Will do.

hannibal002 (approved):

PR #6537, opened 2026-09-13
Backend: Simplify WasdInputMatrix iteration and add input sequence tracker

hannibal002:
## What

Also keeps track of the input sequence to avoid this in the future.

+ Simplified WasdInputMatrix in KeyboardManager. - hannibal2


src/main/java/at/hannibal2/skyhanni/utils/compat/GuiScreenUtils.kt
```diff
@@ -13,6 +14,19 @@ object GuiScreenUtils {
...
 
+    /**
+     * The screen that is currently open, or null if none is.
+     *
+     * This delegates to [MinecraftCompat.screen], which feature classes are not supposed
+     * to touch directly. It exists so the version differences stay in one place instead
+     * of every caller handling them, not to grant feature classes new access.
+     */
+    var currentScreen: Screen?
+        get() = MinecraftCompat.screen
```
AverageUser125: So much stuff refrences it directly already, so there is no point making this alias.

Just make it call `MinecraftCompat.screen` directly

AverageUser125 (changes requested):

src/main/java/at/hannibal2/skyhanni/utils/compat/GuiScreenUtils.kt
hannibal002: > So much stuff refrences it directly already, so there is no point making this alias.

fair for one feature, but every caller breaks the convention and they need a eventual follow up refactor pr anyway, so imo no point in delaying the beginning further/making the eventual pr bigger by repeating the wrong pattern.

lunaynx: I feel like this is kind of a pointless restriction, but whatever.

lunaynx: If you find the need to satisfy the rule by creating a wrapper class that forwards it 1:1 so you can say *technically* you're following the rule, then maybe that rule is a bad rule.

hannibal002: > I feel like this is kind of a pointless restriction, but whatever.

on one hand i agree, this is totally pointless and makes it just more complicated, on the other hand this allows a clearer separation in the three layers minecraft/skyhanni lib/skyhanni features.

not sure

lunaynx: How about just fully move this to `GuiScreenUtils` and deprecate `MinecraftCompat.screen`?

hannibal002: Removed, calls MinecraftCompat.screen directly now. The layer separation gets its own pr.

PR #6538, opened 2026-09-13
Backend: Keybind Rework

AverageUser125:
## What
But anyways, made an enum for all key constants, which just wraps `InputConstants.Key`. also means context sensitive makes life easier.
needs to make a custom MoulConfig annotation and gui element for it.

It also stores the keybinds in config as the vanila string identifier instead of the traditional enum name.

+ Reworked the Keybind system from scratch. - Avrg


src/main/java/at/hannibal2/skyhanni/data/NotificationManager.kt
```diff
@@ -30,7 +30,7 @@ object NotificationManager {
     fun onKeyPress(event: KeyPressEvent) {
         currentNotification ?: return
         if (lastNotificationClosed.passedSince() < 200.milliseconds) return
-        if (event.keyCode != GLFW.GLFW_KEY_X) return
+        if (event.key == InputCode.KEY_X) return
```
hannibal002: this got accidentally inverted.

src/main/java/at/hannibal2/skyhanni/utils/compat/SkyHanniBaseScreen.kt
```diff
@@ -67,13 +69,14 @@ abstract class SkyHanniBaseScreen(title: Component = Component.empty()) : Screen
 
     private fun postKeyTyped(typedChar: Char?, keyCode: Int?) {
         try {
-            onKeyTyped(typedChar, keyCode)
+            val key = keyCode?.let { InputCode.fromKeyCode(it) } ?: InputCode.UNKNOWN
```
hannibal002, line 72: not sure i read that correct, but doesn't firing the key typed logic with the typed key set to "unknown" activate all actions in the config that are set to "no key"? with the gui reset hotkey on none, typing anything in the position editor resets the hovered element now. keyCode is null only on the charTyped path, so that is the normal case and not an error, the parameter should probably just stay nullable.

src/main/java/at/hannibal2/skyhanni/features/inventory/loadout/CustomLoadoutKeybinds.kt
```diff
@@ -57,8 +57,8 @@ object CustomLoadoutKeybinds {
         return false
     }
 
-    fun allowMouseClick() = isEnabled() && keybinds.filter { it < 0 }.any { it.isKeyHeld() }
-    fun allowKeyboardClick() = isEnabled() && keybinds.filter { it > 0 }.any { it.isKeyHeld() }
+    fun allowMouseClick() = isEnabled() && keybinds.filter { it != InputCode.UNKNOWN }.any { it.isKeyHeld() }
+    fun allowKeyboardClick() = isEnabled() && keybinds.filter { it != InputCode.UNKNOWN }.any { it.isKeyHeld() }
```
hannibal002: those two functions are identical. one should probably call the other, or both should call the same private function. right now this is code duplication/code smell

src/main/java/at/hannibal2/skyhanni/features/event/diana/BurrowWarpHelper.kt
```diff
@@ -62,9 +61,8 @@ object BurrowWarpHelper {
         if (GriffinBurrowHelper.mobAlive) return
 
         val text = "§bWarp to " + warp.displayName
-        val keybindSuffix = if (config.keyBindWarp != GLFW.GLFW_KEY_UNKNOWN) {
-            val keyName = KeyboardManager.getKeyName(config.keyBindWarp)
-            " §7(§ePress $keyName§7)"
+        val keybindSuffix = if (config.keyBindWarp != InputCode.UNKNOWN) {
+            " §7(§ePress $${config.keyBindWarp.displayName}§7)"
```
hannibal002, line 65: there is a `$` too much

src/main/java/at/hannibal2/skyhanni/utils/json/SkyHanniTypeAdapters.kt
```diff
@@ -216,6 +217,16 @@ enum class SkyHanniTypeAdapters(
...
     ),
     SESSION_UPTIME(SessionUptime::class.java, SessionUptimeTypeAdapter()),
+    INPUT_CODE(
+        InputCode::class.java,
+        object : TypeAdapter<InputCode>() {
+            override fun write(out: JsonWriter, value: InputCode) {
+                out.value(value.toKeyIdentifier())
+            }
+
+            override fun read(reader: JsonReader): InputCode = InputCode.fromKeyIdentifier(reader.nextString())
```
hannibal002, line 227: fromKeyIdentifier can throw an error on an invalid code. this should be caught here in the read and translated into InputCode.UNKNOWN with a user facing error printed, not into null, since gson writes a null return into the non nullable kotlin field.

src/main/java/at/hannibal2/skyhanni/config/UpdateKeybinds.kt
```diff
@@ -1,62 +1,27 @@
...
-            description = "Resets all of your skyhanni keybinds"
-            aliases = listOf("shkeybindreset")
-            simpleCallback {
-                for (keybind in keybinds) {
-                    resetKeybind(keybind)
-                }
-                ChatUtils.chat("§aSuccessfully reset all SkyHanni Keybinds")
+        for (keybindPath in SkyHanniConfigSearchResetCommand.allKeybinds) {
+            event.transform(147, keybindPath) { element ->
+                val oldCode = element.asInt
```
hannibal002, line 19: we do not have a guarantee that element is an int. could throw an error. should probably be handled explicitly

src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt
```diff
@@ -57,21 +56,12 @@ abstract class ScrollInput(
...
             override fun update(isValid: Boolean) {
                 if (maxValue < minValue) return
                 if (!isValid) return
                 var changed = false
-                if (
-                    dragScrollMouseButton != null &&
-                    MouseCompat.isButtonDown(dragScrollMouseButton) &&
-                    consumeMouseMoveEvent()
-                ) {
-                    scroll += MouseCompat.getEventDY() * velocity
```
hannibal002, line 72: this was allowing "press mouse button, then move mouse" to scroll, and this behaviour is now gone. intentional? if so, please mention it in the changelog. if not intentional, please restore that functionality

src/main/java/at/hannibal2/skyhanni/features/inventory/loadout/CustomLoadoutKeybinds.kt
```diff
@@ -57,8 +57,8 @@ object CustomLoadoutKeybinds {
         return false
     }
 
-    fun allowMouseClick() = isEnabled() && keybinds.filter { it < 0 }.any { it.isKeyHeld() }
-    fun allowKeyboardClick() = isEnabled() && keybinds.filter { it > 0 }.any { it.isKeyHeld() }
+    fun allowMouseClick() = isEnabled() && keybinds.filter { it != InputCode.UNKNOWN }.any { it.isKeyHeld() }
```
hannibal002, line 60: this function now returns true for any keybind press. this is wrong, before the change it was false for every input, because the old "it < 0" filter only matches unbound keys under glfw. optionally: keeping mouse and keyboard apart properly needs key.type, but that is a behaviour change and fits better into its own pr.

src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCustomKeybinds.kt
```diff
@@ -174,36 +172,39 @@ object GardenCustomKeybinds {
 
     private fun hasGuiOpen() = MinecraftCompat.screen != null
 
+    private fun Map<KeyMapping, InputCode>.contains(value: KeyMapping): Boolean =
```
hannibal002, line 175: this function name shadows the stdlib map function. rename please. e.g. to "isUsedAsOverride"

src/main/java/at/hannibal2/skyhanni/features/garden/farming/GardenCustomKeybinds.kt
```diff
@@ -174,36 +172,39 @@ object GardenCustomKeybinds {
...
+        values.any { it.value == value.key.value }
```
hannibal002, line 176: we do not need to compare values, we can compare the key instances directly.

hannibal002 (changes requested):
`SkyHanniConfig.lastMinecraftVersion` is now unused, should get removed.

src/main/java/at/hannibal2/skyhanni/utils/renderables/ScrollInput.kt
AverageUser125: I was too lazy to try and figure out how to restore this functionality.
And it's only usage was the changelog screen anyways.

src/main/java/at/hannibal2/skyhanni/features/inventory/loadout/CustomLoadoutKeybinds.kt
AverageUser125: The GuiMouseInput PR exists for a reason.

src/main/java/at/hannibal2/skyhanni/config/UpdateKeybinds.kt
AverageUser125: I'll make it fallback on the default value in config.

src/main/java/at/hannibal2/skyhanni/utils/compat/SkyHanniBaseScreen.kt
AverageUser125: I am pretty sure I don't understand you.
since `isKeyHeld` returns false for UNKNOWN.
So I am confused.

AverageUser125:
> `SkyHanniConfig.lastMinecraftVersion` is now unused, should get removed.

Uh, It could be useful.
Minecraft changed 2 string identifier for keys between versions.
So I should probably add the migration for those 2.

src/main/java/at/hannibal2/skyhanni/utils/compat/SkyHanniBaseScreen.kt
AverageUser125: Oooh, I get it. yeah imma make it nullable.

src/main/java/at/hannibal2/skyhanni/config/UpdateKeybinds.kt
AverageUser125: I'll even show the user that some of their keybinds got reset.

src/main/java/at/hannibal2/skyhanni/features/inventory/loadout/CustomLoadoutKeybinds.kt
AverageUser125: But yeah, I got no clue what I should do here.

PR #6545, opened 2026-09-14
Improvement: Better contributor discovery message

lunaynx:
## What
Currently, if you've never discovered any SkyHanni contributors before, and you discover one, you see something like this in chat:

```
[SkyHanni] Achievement Get! EEEEKK!! [hover text: Be in the same lobby as a SkyHanni contributor]
[SkyHanni] You have discovered qtLuna for the first time!
```

While for subsequent unique contributor discoveries you get this:
```
[SkyHanni] A wild SkyHanni contributor appears! (hover) [hover text: You have discovered qtLuna for the first time!]
```

I don't know if it was ever intended to be this way, but it's very confusing for players who don't know they can hover over the first message or are simply too lazy to do so ("why did I just get an achievement for meeting this random person?"), and also massively increases unwanted attention/pings.

I've changed the former case to be like this:
```
[SkyHanni] Achievement Get! EEEEKK!! [hover text: Be in the same lobby as a SkyHanni contributor]
[SkyHanni] A wild SkyHanni contributor appears! (hover) [hover text: You have discovered qtLuna for the first time!]
```

## Changelog Improvements
+ Changed the message shown when discovering your first SkyHanni contributor. - Luna
    * The name is now behind a hover, the same as with every later discovery.


src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
```diff
@@ -148,8 +148,7 @@ object ContributorAchievement {
 
     private fun showContributorAchievement(profile: GameProfile, completed: Boolean): Boolean {
         if (!completed || !AchievementManager.shouldShowMessages) return false
-        val message = getDiscoverComponent(profile)
-        ChatUtils.chat(message)
+        showContributorDiscovered(profile)
```
hannibal002: With `discoverContributorMessage` off (the default) this drops the second message entirely, only the achievement line is left. Intended?

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
```diff
@@ -148,8 +148,7 @@ object ContributorAchievement {
 
     private fun showContributorAchievement(profile: GameProfile, completed: Boolean): Boolean {
```
hannibal002, line 149: All three paths end in `showContributorDiscovered(profile)` now, so this function and the `completed`/`shouldShowMessages` checks no longer do anything.

hannibal002 (changes requested):

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
lunaynx: I guess we could change the option to only control whether subsequent discoveries notify, or make it a three-way option, with default being first time only.

hannibal002: Prefer the first one, the option only gating later discoveries. A three way option needs a config migration for a state nobody asked for yet.

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
lunaynx: Is this still relevant? Now there's a distinction between already having the achievement or not once again.

hannibal002: its not

src/main/java/at/hannibal2/skyhanni/config/features/dev/DevConfig.kt
```diff
@@ -106,7 +106,8 @@ class DevConfig {
     @Expose
     @ConfigOption(
         name = "Discover Contributor Message",
-        desc = "Sends a message to your chat when you see a unique contributor.",
+        desc = "Sends a message to your chat when you see a unique contributor.\n" +
+            "Your first-ever contributor discovery is always shown if you have achievements enabled.",
```
hannibal002: The toggle is called §eShow Achievement Messages§7, worth naming it here.

hannibal002 (changes requested):

hannibal002 (approved):

PR #6543, opened 2026-09-13
Improvement: Reworked Compact Gift Chats into an enum and added a disable tree gift messages option

Ironfort9:
## What

<details>
<summary>Removed Compact Gifts and changed it into Display Tree Gifts</summary>
Reworked the Compact Gifts option to an enum and updated ForagingTracker.kt to use the new format.

The new enum is called TreeGiftDisplayMode and has the following elements:
FULL_GIFT: Equivalent to Compact Gifts being false.
COMPACT_GIFT: Equivalent to Compact Gifts being true.
NO_GIFT: Disables Tree Gift messages altogether.

The config attribute pertaining to Compact Gifts was changed to displayTreeGifts, with its possible values stored in TreeGiftDisplayMode.
</details>

## Changelog Improvements
+ Added Display Tree Gifts. - Ironfort9
    * This dropdown in the Foraging Tracker substitutes the old Compact Gifts.
    * You can set it to display the full tree gift message, to compact the tree gift message or to disable the tree gift messages altogether.

PR #6547, opened 2026-09-15
Improvement: Added "In Dwarven Mines" option to Fallen Star Cult Timer

Ironfort9:
## What
<details>
<summary>Added new option to Fallen Star Cult Timer for when the player is in the Dwarven Mines</summary>
Updated FallenStarCultTimer.kt to have a new option IN_DWARVEN_MINES which is checked with the private function inDwarvenMines(). I have also updated inCultRoom to check if the player is in the Dwarven Mines via the inDwarvenMines function to reduce code repetition.

Added the option for IN_DWARVEN_MINES to the enum ShowCondition like all other config options. Placed it above IN_CULT_ROOM since for the prior to be true IN_DWARVEN_MINES has to be already true.

Also bumped CONFIG_VERSION. Tell me if I need to roll that back.
</details>

## Changelog Improvements
+ Added an "In Dwarven Mines" option to the Fallen Star Cult Timer. - Ironfort9


src/main/java/at/hannibal2/skyhanni/config/ConfigUpdaterMigrator.kt
```diff
@@ -12,7 +12,7 @@ import com.google.gson.JsonPrimitive
 
 object ConfigUpdaterMigrator {
     val logger = SkyHanniLogger("ConfigMigration")
-    const val CONFIG_VERSION = 146
+    const val CONFIG_VERSION = 148
```
hannibal002: only bump by one

hannibal002 (changes requested):
overall lgtm

src/main/java/at/hannibal2/skyhanni/config/ConfigUpdaterMigrator.kt
Ironfort9: I also have #6543 which already bumped it to 147, so I used 148 to avoid conflict with it.

lunaynx: A config version bump is generally not needed if there's no migration to be done. This is an entirely new value and its addition is backwards compatible. We don't support downgrading anyway.

Ironfort9: Very well, pushed a commit to revert the CONFIG_VERSION bump.

src/main/java/at/hannibal2/skyhanni/config/features/mining/dwarves/FallenStarCultConfig.kt
```diff
@@ -25,11 +25,13 @@ class FallenStarCultConfig {
     )
     @ConfigEditorDraggableList
     val onlyShowWhen: MutableList<ShowCondition> = mutableListOf(
+        ShowCondition.IN_DWARVEN_MINES,
```
hannibal002: Please leave this out of the default list, the conditions are OR'd so this makes "In Cult Room" a no-op per default and would show this display by default always while in dwarven mines. too annoying for default value imo

hannibal002 (changes requested):

hannibal002 (approved):

PR #6549, opened 2026-09-16
Fix: Change Color Particle Fix from RBG to RGB

lunaynx:
## What
@Fazfoxy has been saying for a long time how Color Particle Fix "feels off". I dug into it and found out why: he made an incorrect assumption that 1.8.9 stores the colors as XYZ = RBG based on the misspelled `setRBGColorF` method name in the MCP mappings. A look at the actual source code reveals that it's clearly XYZ = RGB.

## Changelog Fixes
+ Fixed wrong colors in Fix Colored Particles. - Luna
    * Green and blue were swapped, so green particles rendered blue and blue ones green.


Fazfoxy:
> has been saying for a long time how Color Particle Fix "feels off". 

I mean I did at least know for a fact some colouring was off... I just assumed it was some funky rounding......

src/main/java/at/hannibal2/skyhanni/features/misc/ColorParticleFix.kt
```diff
@@ -8,28 +8,33 @@ import at.hannibal2.skyhanni.utils.SkyBlockUtils
...
             )
         }
         event.particleOptions = particleOptions
     }
-    /*
-    1.8 has, some, questionable, ways of working out color for particles where it uses the offset to calculate the Coloring,
-    this is done as RBG with XYZ, so in modern since it uses ARGB we go to XZY.
-     */
 
     fun isEnabled() = SkyBlockUtils.inSkyBlock && config.fixColorParticles
```
Fazfoxy: not changed in this PR but this is remnant from the pre-event-ified version of this feature in the original PR and nobody made a review to make this isEnabled private at the time I guess?

PR #6551, opened 2026-09-16
Fix: Not hiding visitor chat messages

CalMWolfs:
## What
It should still be made colorless or something in future but idc

## Changelog Fixes
+ Fixed visitor chat messages not being hidden. - CalMWolfs


hannibal002:
Just note that this chanhelog will only show up in 9.1.0, and the 9.0.0 tag will copy the 7.57.0 state, ignoring this fix

Issue #195, opened 2023-06-02
More Options to Chat Filter


hannibal002:
Split up the existing “others” category into separate filters.
Give each filter an ConfigEditorDraggableList element with every single message to filter manually.

PR #6554, opened 2026-09-16
Improvement: Split chat filter "Others" toggle into individual filters

badblobb:
## What
Split the `others` Chat Filter Types toggle into 12 toggles, made sure to match the existing pattern used by every other filter in `FilterTypesConfig.kt`.
`others` had a `// TODO remove` and bundled a bunch of unrelated chat messages together in one big toggle without allowing choice between them, lowkey bugged me personally so I decided to fix it myself.
A few of the splits were a bit oudated, so I made some edits:
- The old "Useless Notification" bucket got split into three separate toggles: `Useless Notifications`, `Mining Abilities`, `Deployables`. Also removed ability messages hiding "ability ready again" since that's legit useful to players.
- `Mining Abilities` now covers used/expired messages for Mining Speed Boost, Maniac Miner, Tunnel Vision, Gemstone Infusion, Sheer Force, and Pickobulus.
- `Deployables` now uses a colorless and repo pattern backed match on "was removed" rather than an enumerated name list, this should cover all power orbs, mining lanterns, flares, black holes, umberella and totem of corruption now.
- Replaced the old "Annoying Spam" catch-all with `Ability Damage`, `Blocked Actions`, and `System/Dev Noise`.
- Entirely dropped hiding the "GOOD CATCH!" fishing treasure messages since as of recently this is actual relevant gameplay info.
- Cleaned up old, obsolete code like the wardrobe in combat warning, blessing enchant doubling drops, the ad for the 12th anniversary Youtube video, and a feast chef ted line that was already covered by an earlier toggle.

`Mining Abilities` and `Deployables` default to off and don't inherit from the old `others` value on migration because this is potentially useful information and were only incidentally bundled into `others`.

This PR only handles the toggle-split half of #195, not the second part (a `ConfigEditorDraggableList` for picking individual messages within each filter). I couldn't find a clean existing precedent for a raw string message list and some of these groups are regex-based rather than fixed strings, so "every message" didn't work with existing config patterns.

<details>
<summary>Images</summary>
Before:
image

After:
image
image
image

</details>

## Changelog Improvements
+ Split the "Others" Chat Filter option into 15 individual filters, so each type of message can be hidden independently instead of all of them being toggled at once. - badblobb
    * Existing users who had "Others" enabled will automatically have 13 of the 15 toggles enabled, with the exception of Mining Abilities and Deployables toggles as they may contain useful information.


badblobb:
Actually, I'm noticing now that under the existing "Useless Messages", the mining abilities only included "Mining Speed Boost" and deployables only included Plasmaflux. Splitting these into "Mining Abilities" and "Deployables" toggles (Deployable Expiration was a bit long and description fits well). The rest I chose to stay with "Useless Notifications" despite it possibly reading as a "Useless Option" joke entry.

I'm going to expand on Mining Abilities to include all abilities as of 16 Sep 2026 and also add a Foraging Abilities toggle with the new HoTF abilities.
Deployables will now include all deployables, including black holes, dwarven lanterns, flares, totem of corruption, and umberella. I didn't include the miniature nuke because lmfao???

ALSO these three options (Mining Ability, Foraging Ability, Deployables) should *not* be enabled by default anymore since they could potentially be useful and were lumped into the old "Other" category regardless, will update PR once I figure out all the regex matching

badblobb:
Reading now that the old code would just not work if you had a skin applied, Maybe tracking "was removed." might just be easier and cleaner

lunaynx:
This is a good idea, but I really want to refactor this to all be colorless and backed by repo patterns first. I suppose your current changes aren't too difficult to reconcile, but I'd prefer if you'd hold off on changing the actual filter patterns until after that is done.

badblobb:
I've made some pretty good progress worth keeping, but I'll hold off on adding anything else for now. Stuff like deployables I used a `"§eYour previous §r.*§r§ewas removed!"` pattern for now. Fixing things up and capturing some images before finishing this PR ^^

src/main/java/at/hannibal2/skyhanni/config/features/chat/FilterTypesConfig.kt
```diff
@@ -193,10 +193,101 @@ class FilterTypesConfig {
...
+        desc = "Hide annoying warning messages, e.g. rate limit warnings, combat restrictions.",
+    )
+    @ConfigEditorBoolean
+    @FeatureToggle
+    var uselessWarning: Boolean = false
+
+    @Expose
+    @ConfigOption(
+        name = "Annoying Spam",
+        desc = "Hide miscellaneous spammy messages, e.g. ability damage numbers, fishing catches, dev warnings.",
```
zumbiepig: want to split this up too?

badblobb: Yeah, good call. Making the changes now. This whole thing might need a proper refactor soon since scrolling through 30 toggles lowkey sucks

badblobb:
As per @zumbiepig's suggestion, I'm going to edit this to split "Annoying Spam" up a bit more:

Ability Damage - Hide ability damage notifications for abilities like Wither Impact, Guided Bat, etc.
Blocked Actions - Hide messages when an action is blocked, e.g. Instant Transmission having blocks in the way, insufficient Breaking Power.
NPC Announcements - Hide NPC announcement messages, e.g. Jacob's contest starting, Booster cookie required for action.
System/dev noise - Hide internal system and debug messages that sometimes leak into the chat.

I'm also cleaning things up a bit by removing these:
- "Good" treasure catches: new updates have made this relevant to gameplay now
- "Your Blessing enchant got you double drops!": it works differently now
- 12th Anniversary Youtube premier message (lol???)
- "You can't use the wardrobe in combat!": This doesn't exist anymore

The old damage notification code hard matched stuff like implosion and spirit sceptre directly, I've cleaned it up and used the new colorless regex matching `§7Your .* hit (.*) for §r§c(.*) §r§7damage.` and wildcards to generalise it to the new standard.

zumbiepig:
> Yeah, good call. Making the changes now. This whole thing might need a proper refactor soon since scrolling through 30 toggles lowkey sucks

Make the whole thing a multiselect menu

badblobb:
> > Yeah, good call. Making the changes now. This whole thing might need a proper refactor soon since scrolling through 30 toggles lowkey sucks
> 
> Make the whole thing a multiselect menu

I disagree for now, UX wise, the rest of the options are not a multi select menu too and there's too many competing detection types, the whole thing would need to be refactored. Not to mention people accessing this menu are usually doing it through search functionality and these toggles benefit from a longer description. Happy to get other feedback on this, but as it stands IMO the best thing we can do is to chuck them into well organised groups for people to open and peruse

hannibal002:
> Make the whole thing a multiselect menu

whats the benefit of a multiselect?
imo its fine in the current apporach

zumbiepig:
You could use more nested accordions to organize them

src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
```diff
@@ -298,31 +325,36 @@ object ChatFilter {
         "§7Your radio lost signal. There's too many enjoyers on this channel.",
     )
 
-    // Annoying Spam
+    // Ability Damage
+    /**
+     * REGEX-TEST: §7Your Implosion hit Zealot for §r§c1,234 §r§7damage.
+     * REGEX-TEST: §7Your Molten Wave hit Zealot for §r§c1,234 §r§7damage.
+     * REGEX-TEST: §7Your Spirit Sceptre hit Zealot for §r§c1,234 §r§7damage.
```
lunaynx: Regex tests must be examples of real in-game messages, not strings made up to fit the pattern. The game does not say "Your Implosion hit Zealot for 1,234 damage", it says "Your Implosion hit 1 enemy for 1,234 damage" – same with the others.

src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
```diff
@@ -298,31 +325,36 @@ object ChatFilter {
...
+     */
+    private val abilityDamagePattern by abilityDamagePatternGroup.pattern(
+        "hit",
+        "§7Your .* hit (.*) for §r§c(.*) §r§7damage.",
```
lunaynx: Since you're turning this into a RepoPattern, please clean it up while you're at it:
- Make it colorless – remove the color codes (`§*`) and match against `event.cleanMessage`. Since the full colorless ChatFilter refactor is planned after this PR, you'll probably have to special case this for the time being.
- The capturing groups are unnecessary, change both `(.*)` to `.*`. 
- When something is always expected to match at least one character, use `.+` rather than `.*`. Or better yet, something more specific where appropriate.

```suggestion
        """Your [\w ]+ hit \d+ enem(?:y|ies) for [\d,.]+ damage\.""",
```

lunaynx (changes requested):

lunaynx:
Confirmed the rough plan with Hannibal on Discord:
1. This PR: split up existing "others" / "annoying spam" filters, and possibly minor fixes/extensions to existing filters, but preferably no major additions
2. My PR: full colorless + RepoPattern refactor
3. If needed, other PRs adding further filters

src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
badblobb: > Regex tests must be examples of real in-game messages, not strings made up to fit the pattern. The game does not say "Your Implosion hit Zealot for 1,234 damage", it says "Your Implosion hit 1 enemy for 1,234 damage" – same with the others.

Completely missed that from earlier, oopsies :(
Fixing now

badblobb:
Made some changes this morning:
- Fixed the regex tests i drafted originally without checking ingame, now it matches expected game behaviour (sorry!!! I'll remember this for future PRs!!)
- Ability damage pattern is now colourless, it's just a special case for now I've done ahead of your refactor, should be okay.
- The capturing groups were dropped from ability damage as suggested.
- Removed it from the generic `isPresent()` mapping system since the matching logic was different now that we're using clean message.
- I actually tightened `.*` to `.+` on two other patterns (mining ability used / deployable replaced) since those fields can never be empty either, just to clean things up a bit.

Thanks for the feedback!

src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
```diff
@@ -298,31 +325,35 @@ object ChatFilter {
...
+
+    // NPC Announcements
+    private val npcAnnouncementsMessages = listOf(
         "§e[NPC] Jacob§f: §rMy contest has started!",
         "§eObtain a §r§6Booster Cookie §r§efrom the community shop in the hub!",
+    )
+
+    // System/Dev Noise
+    private val systemNoiseMessages = listOf(
         "Unknown command. Type \"/help\" for help. ('uhfdsolguhkjdjfhgkjhdfdlgkjhldkjhlkjhsldkjfhldshkjf')",
```
Fazfoxy: This message should be unique to Soopy which isn't on modern afaik and shouldn't even be doing this solution

src/main/java/at/hannibal2/skyhanni/features/chat/ChatFilter.kt
```diff
@@ -298,31 +325,35 @@ object ChatFilter {
...
         "§3[SBE] §a§cUnable to download bin data. This may result in certain features not working!",
```
Fazfoxy: Seems weird for SH to filter another Mod's error message personally (I know this isn't from this PR but it's semi relevant), but SBE also isn't on modern and as much as this split will make this less bad it's a lot weirder when mods can generically block this at a user's writing anyway.

PR #6516, opened 2026-09-06
Improvement: Add option to defer the contributor achievement

JohnnyCena123:
## What
Contributors who add their name to ContributorList.json could choose to make it so SkyHanni users only get notified of their presence after they leave the lobby, in order to minimize unwanted attention.

## Changelog Improvements
+ Add option to defer the contributor notification per contributor

## Changelog Technical Details
+ Add optional `deferAchievement` field for each contributor in ContributorList.json in the SkyHanni repo

PR #6556, opened 2026-09-18
Improvement: Contributor Achivement Opt-out

ItsEmpa:
## What
Makes it so that contributors can opt out of counting for contributor-related achievements, and not announcing to the lobby when people see them. Some people just want to play and not constantly notify people in every lobby they go to.

Contributors can do this by setting `disabled_achievement_discovery` to true in the ContributorList.json in repo.

## Changelog Improvements
+ Allow Contributors to Opt-out of achievements. - Empa


AverageUser125:
Is #6516 to your liking instead?
Cause this is also fine tbh.

ItsEmpa:
> Is #6516 to your liking instead? Cause this is also fine tbh.

both are good i think. the problem with queuing it until the contributor leaves the lobby is that there will still be people sending you friend requests just because they got notified, and some people may also not want that

src/main/java/at/hannibal2/skyhanni/data/garden/EliteFarmersLeaderboard.kt
```diff
@@ -129,7 +129,9 @@ object EliteFarmersLeaderboard {
                             }
                             val contribUUID = ContributorManager.getUUIDFromDisplayName(name)
                             if (contribUUID != null) {
-                                AchievementManager.completeAchievement(BETTER_THAN_DEV_ACHIEVEMENT)
+                                if (!ContributorManager.hasDisabledAchievementDiscovery(contribUUID)) {
+                                    AchievementManager.completeAchievement(BETTER_THAN_DEV_ACHIEVEMENT)
```
AverageUser125: Could you add a new parameter to `AchievementManager.completeAchievement` which just silences the message instead of just not completing it?

And obviously still use it in the rest of the achievment stuff

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
```diff
@@ -101,31 +101,35 @@ object ContributorAchievement {
...
 
+    private fun isContributorAndShouldDiscover(ign: String): Boolean {
+        return ign in ContributorManager.contributorNames && !ContributorManager.hasDisabledAchievementDiscovery(ign)
+    }
+
     @HandleEvent(priority = HandleEvent.LOW)
     private fun onProfileJoin() {
         val friends = FriendApi.getAllFriends()
-        if (friends.any { it.name in ContributorManager.contributorNames }) {
+        if (friends.any { isContributorAndShouldDiscover(it.name) }) {
```
AverageUser125: this is from being a friend of a contributor.
So it's probably fine to give a chat message here.

src/main/java/at/hannibal2/skyhanni/features/achievements/ContributorAchievement.kt
```diff
@@ -101,31 +101,35 @@ object ContributorAchievement {
...
             AchievementManager.completeAchievement(CONTRIBUTOR_FRIEND_ACHIEVEMENT)
         }
     }
 
     @HandleEvent
     private fun onFriendAdd(event: FriendAddEvent) {
-        if (event.playerName in ContributorManager.contributorNames) {
```
AverageUser125: if a contributor added you as a friend. that's probably find to show the message.

AverageUser125 (changes requested):

